PR #243, opened 2020-09-22
add the rebase threshold

sparrowDom:
Adds a rebase threshold along side allocate threshold.


DanielVF (approved):
I just:

- Added actual gas numbers from mainnet transactions, with about 10-15% margin of error above.
- Added a third line for simple mints, so that we hard code this value as well. (We out-of-gassed two of these last night.)
- Added a 96% adjustment to the threshold, so that even if coin prices are a little out of line, we won't send too little gas.

The one thing I wonder about is using an env variable for these values. It would seem much better to have these in a config file that can live in version control and be updated as we change our contracts. Otherwise, when we add a new strategy, it's going to be much harder to track our gas price changes.

sparrowDom:
@DanielVF yup agreed, moving all the gas limits to a config file

PR #246, opened 2020-09-23
better handle Ledger errors

sparrowDom:
Adds error handling to Mint / Approve function calls. When Ledger user doesn't have contract data enabled an error and error is thrown the user is presented with this modal: 
Screenshot 2020-09-23 at 20 13 45

Have done additional testing with Logging in With Ledger device. For some reason the library wouldn't consistently throw an error when my Ledger is locked / not inside Ethereum app. But when the error does get thrown this is shown to the user: 
Screenshot 2020-09-23 at 19 44 52

### DIFF
--- a/dapp/src/components/LoginWidget.js
+++ b/dapp/src/components/LoginWidget.js
@@ -107,7 +107,7 @@ const LoginWidget = ({}) => {
                    * are suppressed.
                    */
                   (err) => {
-                    console.log('Setting the error: ', err)
+                    console.debug('Setting the error: ', err)
                     setError(err)
                   },
                   // do not throw the error, handle it in the onError callback above
--- a/dapp/src/components/LoginWidget.js
+++ b/dapp/src/components/LoginWidget.js
@@ -98,10 +98,20 @@ const LoginWidget = ({}) => {
                 setActivatingConnector(currentConnector)
                 await activate(
                   currentConnector,
+                  /* According to documentation: https://github.com/NoahZinsmeister/web3-react/tree/v6/docs#understanding-error-bubbling
+                   * if this onError function is specified no changes shall be done to the "useWeb3React" global context.
+                   * also with the 3rd parameter being false errors should not be thrown.
+                   *
+                   * When I test using my ledger Nano S [sparrowDom] the below function doesn't consistently throw errors
+                   * when I either lock my Ledger or exit Ethereum app. On my end it sometimes just seems that the errors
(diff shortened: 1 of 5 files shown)


mikeshultz:
When doing initial "Connect":

On first "boot" with the device still locked, this unhandled error hits the console:

    {
        id: "U2F_4"
        message: "Failed to sign with Ledger device: U2F DEVICE_INELIGIBLE"
        name: "TransportError"
        originalError: Error: Sign failed at [...]
    }

Once I hit the menu (before opening the Ethereum app), and hit "Connect", nothing happens.  After clicking a second time, I get: `MULTIPLE_OPEN_CONNECTIONS_DISALLOWED` in red in the connect modal.

If I keep hammering the Ledger button in the connect modal, I get `Failed to sign with Ledger device: U2F DEVICE_INELIGIBLE` displayed in the modal.

Haven't tried the rest yet.

mikeshultz:
Interestingly, with the timeout lock, it's properly handled and reports:

> Unlock your Ledger wallet and open Ethereum application

So I guess it's throwing different errors on boot lock versus timeout lock?

mikeshultz:
Pretty much all of my initial feedback comment comes down to confusion in how `activate()` is returning/throwing errors.  There seems to be a 15 second delay between a warning-level console error, and the error reaching our handler.  I've tried replacing the handler with just setting `activate()` to throw but the delay is still there.  It has to be internal to web3-react, I guess?

sparrowDom:
@mikeshultz here is the documentation on how the error bubbling is suppose to happen: 
https://github.com/NoahZinsmeister/web3-react/tree/v6/docs#understanding-error-bubbling

Though when I was testing I didn't see any improvement if the errors were either: thrown / handled by the error callback function / set to web3React context. I assume the way errors are reported doesn't affect the way underlying code handles different Ledger states.

After some more testing I think that the ~15 second timeout is the period the underlying library is trying to connect before it throws the error. If I enter the Etereum app inside those 15 seconds after I try to login from the dapp. The Ledger login usually succeeds. 

I think we could mitigate this by showing a Warning after a few seconds prompting users to Unlock Ledger and enter Etereum app: 

https://www.loom.com/share/8a530d495dd64804b627f6474ff47df3

sparrowDom:
@mikeshultz going to merge this is, since I think it is an improvement over the current state that is in prod. Happy to address any issues if needed later.

Issue #240, opened 2020-09-22
JSON-RPC requests


mikeshultz:
The amount JSON-RPC requests are.. intense.  Not sure where they're coming from, but requests for network ID and other trivial things are sent very often that it could actually DoS the client.  Should probably dial that back a bit.

Will fill out more detail when I have it.

mikeshultz:
Seems to be OUSD price checks(probalby for the running total) and `eth_chainId` calls happening every less-than-a-second.

    {"method":"eth_chainId","params":[],"id":48,"jsonrpc":"2.0"}


    {"method":"eth_call","params":[{"to":"0x277e80f3e14e7fb3fc40a9d6184088e0241034bd","data":"0x686b37ca000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000045553445400000000000000000000000000000000000000000000000000000000"},"latest"],"id":49,"jsonrpc":"2.0"}


    {"method":"eth_call","params":[{"to":"0x277e80f3e14e7fb3fc40a9d6184088e0241034bd","data":"0x29a903ec000000000000000000000000000000000000000000000000000000000000002000000000000000000000000000000000000000000000000000000000000000045553445400000000000000000000000000000000000000000000000000000000"},"latest"],"id":51,"jsonrpc":"2.0"}

mikeshultz:
Every 5 seconds...

https://github.com/OriginProtocol/origin-dollar/blob/ce3c0f0a4d2ad5a01053e44d162339c73901a05a/dapp/src/components/AccountListener.js#L131-L133

...it loads the balance and allowance of of every relevant token.

https://github.com/OriginProtocol/origin-dollar/blob/ce3c0f0a4d2ad5a01053e44d162339c73901a05a/dapp/src/components/AccountListener.js#L24-L103

I suspect the `eth_chainId` calls are internal to ethersjs.

mikeshultz:
@micahalcorn I think this might be a bit of a product decision.  We're getting a lot of rate limited requests from Alchemy, so this has to change, but it changes how the app operates.  Right now the app expects to see added tokens transfer in(and approvals) within 5sec.  

I don't know if we want to jack up the interval timing and add a manual refresh button?  Could change it drastically, even, and only do this update during a deposit phase and hide the deposit form when not in use (this allows the dapp to stop looking for allowances and balances as often)

[...]

PR #247, opened 2020-09-23
JSON-RPC request reduction - polling interval change

mikeshultz:
This reducing polling for balances and APR from 5s to 30s.  Slight improvement that will hopefully reduce the rate limiting we're seeing from Alchemy.

I think this is probably good for now but it does impact things like the balance ticker, and how quickly we notice an approval/balance change, so there's a UX impact here.  Perhaps there's a more intelligent way of doing this that requires a bit more of a UI change?

CC: @micahalcorn @franckc

### DIFF
--- a/dapp/src/components/AccountListener.js
+++ b/dapp/src/components/AccountListener.js
@@ -10,6 +10,9 @@ import { useStoreState } from 'pullstate'
 import { setupContracts } from 'utils/contracts'
 import { login } from 'utils/account'
 
+const BALANCE_CHECK_INTERVAL = 30000 // ms
+let balanceInterval
+
 const AccountListener = (props) => {
   const web3react = useWeb3React()
   const { account, chainId, library } = web3react
@@ -108,8 +111,6 @@ const AccountListener = (props) => {
       login(account, setCookie)
     }
 
-    let balanceInterval
-
     const setupContractsAndLoad = async () => {
       /* If we have a web3 provider present and is signed into the allowed network:
        * - NODE_ENV === production -> mainnet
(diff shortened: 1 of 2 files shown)


dapp/src/utils/contracts.js
```diff
@@ -126,7 +126,7 @@ export async function setupContracts(account, library, chainId) {
   // execute in parallel and repeat in an interval
   window.fetchInterval = setInterval(() => {
     callWithDelay()
-  }, 5000)
+  }, 30000)
```
franckc, line 129: do we really need to refresh the APR every 30sec? The APR is not going to change a lot that quickly. I would think a 1hr refresh is good enough. But maybe I'm missing something CC @micahalcorn

dapp/src/components/AccountListener.js
```diff
@@ -10,6 +10,9 @@ import { useStoreState } from 'pullstate'
 import { setupContracts } from 'utils/contracts'
 import { login } from 'utils/account'
 
+const BALANCE_CHECK_INTERVAL = 30000 // ms
```
franckc, line 13: Do we think users are going to stare at their balance that much?
If yes, as opposed to pulling that frequently from the contract, would it be worth to "simulating" the balance growing by artificially increasing it by a tiny amount on the DApp without calling the contract? CC @micahalcorn for product input

mikeshultz: Perhaps move this commentary to the issue #240?

micahalcorn:
Closing this in favor of https://github.com/OriginProtocol/origin-dollar/issues/240#issuecomment-698129245 🙂

dapp/src/components/AccountListener.js
```diff
@@ -10,6 +10,9 @@ import { useStoreState } from 'pullstate'
...
+let balanceInterval
```
shahthepro, line 14: It's better to clean up during unmount than storing it as a global variable. If you return a function inside `useEffect`, it will be called when unmounting.

PR #249, opened 2020-09-23
New storage contracts that makes the Core Vault contract about 20% sm…

crazybuster:
…aller

### DIFF
--- a/contracts/contracts/vault/VaultAdmin.sol
+++ b/contracts/contracts/vault/VaultAdmin.sol
@@ -1,0 +1,221 @@
+pragma solidity 0.5.11;
+
+/**
+ * @title OUSD Vault Admin Contract
+ * @notice The VaultAdmin contract makes configuration and admin calls on the vault.
+ * @author Origin Protocol Inc
+ */
+
+import "./VaultStorage.sol";
+
+contract VaultAdmin is VaultStorage {
+    /***************************************
+                 Configuration
+    ****************************************/
+
+    /**
+     * @dev Set address of price provider.
+     * @param _priceProvider Address of price provider
+     */
+    function setPriceProvider(address _priceProvider) external onlyGovernor {
+        priceProvider = _priceProvider;
+    }
+
+    /**
+     * @dev Set a fee in basis points to be charged for a redeem.
+     * @param _redeemFeeBps Basis point fee to be charged
+     */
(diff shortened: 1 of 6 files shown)


franckc:
C'est bien, tu utilises le Français dans tes PRs maintenant. Tu progresses.

contracts/contracts/vault/VaultAdmin.sol
```diff
@@ -0,0 +1,270 @@
...
+            Vault.
+     * @param _strategyAddr Address of the strategy to collect rewards from
+     */
+    function collectRewardTokens(address _strategyAddr) public onlyGovernor {
+        IStrategy strategy = IStrategy(_strategyAddr);
+        strategy.collectRewardToken();
+    }
+
+    /***************************************
+                    Pricing
```
franckc, line 222: Those pricing methods are not for governor only so why putting them here?

contracts/contracts/vault/VaultStorage.sol
```diff
@@ -0,0 +1,85 @@
...
+    // Mints over this amount automatically allocate funds. 18 decimals.
+    uint256 public autoAllocateThreshold;
+    // Mints over this amount automatically rebase. 18 decimals.
+    uint256 public rebaseThreshold;
+
+    OUSD oUSD;
+
+    bytes32 constant adminImplPosition = 0xa2bd3d3cf188a41358c8b401076eb59066b09dec5775650c0de4c55187d17bd9;
+
+    //keccak256("OUSD.vault.governor.admin.impl");
```
franckc, line 73: nit: is this a comment for line 71? if yes, put it before that line.

contracts/contracts/vault/VaultAdmin.sol
crazybuster: To save more space on the Core

franckc: Makes sense.

It's kinda weird to call this contract VaultAdmin then if it includes non admin related methods that are called by the DApp... But that's just a naming thing. Maybe we could call it "VaultUtils"? I'm terrible at namig so I let you guys make that very difficult decision :)

franckc (approved):
LGTM aside from potentially considering renaming VaultAdmin to VaultUtils or equivalent.

PR #251, opened 2020-09-24
Make less web3 calls

sparrowDom:
In this PR: 
- make only 1 getArp call on dapp init
- make balances & allowances calls only when user is active in the dapp

related issue: https://github.com/OriginProtocol/origin-dollar/issues/240#issuecomment-698129245

### DIFF
--- a/dapp/src/components/AccountListener.js
+++ b/dapp/src/components/AccountListener.js
@@ -14,7 +14,10 @@ const AccountListener = (props) => {
   const web3react = useWeb3React()
   const { account, chainId, library } = web3react
   const prevAccount = usePrevious(account)
+  const [contracts, setContracts] = useState(null)
+  const [balancesInterval, setBalancesInterval] = useState(null)
   const [cookies, setCookie, removeCookie] = useCookies(['loggedIn'])
+  const userActive = useStoreState(AccountStore, (s) => s.active)
 
   const displayCurrency = async (balance, contract) => {
     if (!balance) return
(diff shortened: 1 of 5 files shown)


dapp/src/components/AccountListener.js
```diff
@@ -14,7 +14,10 @@ const AccountListener = (props) => {
   const web3react = useWeb3React()
   const { account, chainId, library } = web3react
   const prevAccount = usePrevious(account)
+  const [contracts, setContracts] = useState(null)
+  const [balancesInterval, setBalancesInterval] = useState(null)
```
shahthepro, line 18: Any reason to have it as a state and not a local variable inside `useEffect`?

dapp/src/components/UserActivityListener.js
```diff
@@ -0,0 +1,55 @@
...
+      })
+      this.setState({ userActivityState: 'active' })
+    }
+  }
+
+  componentDidMount() {
+    document.addEventListener('mousemove', this.onMouseMove)
+    // check for user activity and adjust states when needed
+    setInterval(this.onStateCheck, 500)
+  }
```
shahthepro, line 48: Cleanup and remove event listeners on `componentWillUnmount`

shahthepro (commented):
A few comments inline

dapp/src/components/UserActivityListener.js
sparrowDom: yup god point thanks

dapp/src/components/AccountListener.js
sparrowDom: Yup your right cleaning that up

sparrowDom:
thanks @shahthepro both comments addressed. If nothing else we are good to merge this one I think?

shahthepro (approved):
Everything else looks good to me. LGTM

PR #255, opened 2020-09-25
Vault restructure + Uniswap syncing and liquidation of reward tokens

tomlinton:
- Splits Vault storage, admin and core code to slim down sizes
- RebaseHooks contract for actions post rebase which right now is only syncing the single Uniswap pair we have
- Collects reward tokens from all strategies and attempts to liquidate to USDT via Uniswap

Post deploy we need to do the following with multisig:

- Change VaultProxy to use new VaultCore implementation
- Change VaultCore to use new VaultAdmin implementation (this is a double delegate call)
- Set the RebaseHooks address via `setRebaseHooksAddr` on Vault
- Set the Uniswap pair via `setUniswapPairs` on RebaseHooks
- Set Uniswap address via `setUniswapAddr` on Vault

Some other notes:

- The new `harvest` function which collects reward tokens and liquidates is not called by anything, right now it is manually called
- setUniswapPairs requires an array and I'm unsure about calling this with the multisig, maybe @franckc knows?
- RebaseHooks isn't upgradeable


crazybuster (approved):
lgtm

contracts/contracts/interfaces/uniswap/IUniswapV2Pair.sol
```diff
@@ -0,0 +1,22 @@
...
+            uint112 reserve0,
+            uint112 reserve1,
+            uint32 blockTimestampLast
+        );
+
+    function price0CumulativeLast() external view returns (uint256);
+
+    function price1CumulativeLast() external view returns (uint256);
+
+  function sync() external;
```
franckc, line 21: super nit: run prettier to align

contracts/contracts/vault/VaultAdmin.sol
```diff
@@ -0,0 +1,323 @@
...
+    function harvest(address _strategyAddr) external onlyGovernor {
+        _harvest(_strategyAddr);
+    }
+
+    /**
+     * @dev Collect reward tokens from a single strategy and swap them for a
+     *      supported stablecoin via Uniswap
+     * @param _strategyAddr Address of the strategy to collect rewards from
+     */
+    function _harvest(address _strategyAddr) internal {
```
franckc, line 245: Should this method emit an event for each harvested token including the token address and amount. Though I assume Uniswap emits events as well ? in that case this could be redundant

contracts/deploy/9_vault_split.js
```diff
@@ -0,0 +1,114 @@
...
+    dVaultAdmin.receipt.transactionHash,
+    NUM_CONFIRMATIONS
+  );
+  log("Deployed VaultAdmin", dVaultAdmin);
+
+  const dRebaseHooks = await deploy("RebaseHooks", {
+    from: deployerAddr,
+    ...(await getTxOpts()),
+  });
+
```
franckc, line 55: let's wait for confirmation hash for the rebase hook deploy.

franckc (approved):
wow. great work guys. a few minor comments inline otherwise LGTM

franckc:
Regarding how to pass an array as a parameter when calling a function via the gnosis multi-sig, I tested on the SetStrategiesWeight and it is only showing a single field in the UI. So we need to research the proper format for putting an array
Screen Shot 2020-09-25 at 10 44 00 AM

franckc:
We should test it but I think the format for arrays in gnosis might be like this:
["0xD85A569F3C26f81070544451131c742283360400","0xD85A569F3C26f81070544451131c742283360400"]

contracts/contracts/vault/VaultAdmin.sol
tomlinton: The strategy emits an event like that, so that is probably enough to cover us.

PR #257, opened 2020-09-25
Governor contract

crazybuster:
This is a companion to the tomlinton/uniswap-sync PR

basically that's going to be a lot of multisig requests going through. Instead of getting that many signatures, we can instead do one call to from the multisig to the governor post deploy that will handle the upgrade, setting of implementation, and also registering the pools. This should reduce the number of multisigs drastically, and also allow us to test the timelock.


contracts/contracts/governance/Governor.sol
```diff
@@ -0,0 +1,159 @@
...
+
+
+    constructor(address timelock_, address guardian_) public {
+        timelock = ITimelock(timelock_);
+        guardian = guardian_;
+    }
+
+    function proposeAndQueue(address[] memory targets, uint[] memory values, string[] memory signatures, bytes[] memory calldatas, string memory description) public returns (uint) {
+        require(msg.sender == guardian, "Governor::proposeAndQueue: sender must be gov guardian");
+        require(targets.length == values.length && targets.length == signatures.length && targets.length == calldatas.length, "Governor::propose: proposal function information arity mismatch");
```
franckc, line 73: lol. I had to lookup what "arity" means. fancy wording cPan!

contracts/contracts/governance/Governor.sol
```diff
@@ -0,0 +1,159 @@
...
+
+    function _queueOrRevert(address target, uint value, string memory signature, bytes memory data, uint eta) internal {
+        require(!timelock.queuedTransactions(keccak256(abi.encode(target, value, signature, data, eta))), "Governor::_queueOrRevert: proposal action already queued at eta");
+        timelock.queueTransaction(target, value, signature, data, eta);
+    }
+
+    function execute(uint proposalId) public payable {
+        require(state(proposalId) == ProposalState.Queued, "Governor::execute: proposal can only be executed if it is queued");
+        Proposal storage proposal = proposals[proposalId];
+        proposal.executed = true;
```
franckc, line 121: Paranoia: should this be set after the proposal is executed (e.g. after the for loop) in case of weird revert behavior?

contracts/contracts/governance/Governor.sol
```diff
@@ -0,0 +1,159 @@
...
+        for (uint i = 0; i < proposal.targets.length; i++) {
+            timelock.executeTransaction.value(proposal.values[i])(proposal.targets[i], proposal.values[i], proposal.signatures[i], proposal.calldatas[i], proposal.eta);
```
franckc, line 123: should we check the return value of the executeTransaction function call?

contracts/contracts/governance/Governor.sol
```diff
@@ -0,0 +1,159 @@
...
+        }
+        emit ProposalExecuted(proposalId);
+    }
+
+    function getActions(uint proposalId) public view returns (address[] memory targets, uint[] memory values, string[] memory signatures, bytes[] memory calldatas) {
+        Proposal storage p = proposals[proposalId];
+        return (p.targets, p.values, p.signatures, p.calldatas);
+    }
+
+    function __acceptAdmin() public {
```
franckc, line 133: I'm a solidity noob... What's up with the __ prefix on those functions?

franckc (approved):
Mainly reviewed Governor.sol - a few questions inline otherwise LGTM

contracts/contracts/governance/Governor.sol
crazybuster: not mine this is mostly compound

contracts/contracts/governance/Governor.sol
crazybuster: I think it's better before, the Timelock will revert if anything fails. And if it does fail silent I think that's time for another proposal

contracts/contracts/governance/Governor.sol
crazybuster: basically we should almost never touch the governor on this thing. Again from compound with love.

contracts/contracts/governance/Governor.sol
crazybuster: executeTransaction in the timelock checks the return value. If the timelock is compromise there're a lot more problems. Hence you can't change the timelock.

Issue #263, opened 2020-09-27
Redeem failures


franckc:
It looks like we had a couple of redeems involving strategy liquidation that failed:
 - First instance:
    - https://etherscan.io/tx/0x32c91119f1da55741e8a6c202d2c4f9c513f5d4d77c56c607e8e2f43967f7b81
    - Debug: https://majorhammers.com/tx/debug/0x32c91119f1da55741e8a6c202d2c4f9c513f5d4d77c56c607e8e2f43967f7b81
 - Second instance:
   - https://etherscan.io/tx/0x93bd0b0550865fdf4d7b54694ddf5b42d11319248966de058fc814850c032d52
   - Debug: https://majorhammers.com/tx/debug/0x93bd0b0550865fdf4d7b54694ddf5b42d11319248966de058fc814850c032d52

Both reverts are due to this assert: https://github.com/OriginProtocol/origin-dollar/blob/master/contracts/contracts/token/OUSD.sol#L273

franckc:
CC @tomlinton @crazybuster @DanielVF

tomlinton:
As discussed in Discord. This is a bug with the UI where the input value can become greater than the users actual balance due to populating the input with the animation value.

micahalcorn:
@sparrowDom I haven't reproduced this because my local environment doesn't seem to work, but assuming @tomlinton's right about the issues, I see two potential solutions:

1) call `redeem` and pass the max of (input value, on-chain balance)
2) change the form validation to compare the input value with the on-chain balance instead of comparing it with the estimated (animated) balance

I think 1) makes sense because we already tell the user that the redeem output is estimated and 2) would be confusing if the input value is invalid while being less than or equal to the estimated balance.

sparrowDom:
Fix for the issue in this PR: https://github.com/OriginProtocol/origin-dollar/pull/265

sparrowDom:
@micahalcorn if it is all right with you went with the solution where: 
if the OUSD amount is bigger than OUSD balance but smaller or equal than animated balance we call redeemAll instead of redeem. That seemed the most elegant to me. Let me know if you have any objections.

I could reproduce it locally by:
- having a positive APR
- clicking on the sellAll toggle (that sets OUSD to sell to a higher value than OUSD balance)
- clicking sell all toggle again
- And clicking the big "Sell" button.

micahalcorn:
Sounds great, thx 👍

PR #265, opened 2020-09-28
call redeemAll instead of redeem

sparrowDom:
When user toggles the `sellAll` on and then off the value of OUSD is above user's balance and below the animated one. This caused the dapp to falsely send a too hight OUSD value to the redeem function. 

This PR fixes the issue, so that `redeemAll` is called instead of redeem when the OUSD values is between OUSD balance and OUSD animated value.

**Related issue:** https://github.com/OriginProtocol/origin-dollar/issues/263#issuecomment-699777340

### DIFF
--- a/dapp/src/components/GetOUSD.js
+++ b/dapp/src/components/GetOUSD.js
@@ -40,7 +40,11 @@ const GetOUSD = ({
             source: trackSource,
           })
           const provider = providerName()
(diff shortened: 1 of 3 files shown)


shahthepro (approved):

franckc (commented):
LGTM - thanks for improving this!

PR #266, opened 2020-09-28
Sparrow dom/mobile listener fix

sparrowDom:
User activity listener defaults to 'active' on mobile devices.

### DIFF
--- a/dapp/src/components/UserActivityListener.js
+++ b/dapp/src/components/UserActivityListener.js
@@ -5,6 +5,20 @@ import AccountStore from 'stores/AccountStore'
 
 const activeDelay = 10000 // time in miliseconds after last activity and we consider user still active
 
+/* Intentionally not using withIsMobile since that one is more interested in screen sizes. Here it is important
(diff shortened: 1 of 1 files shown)


sparrowDom:
Merging this in, happy to address stuff later if needed.

dapp/src/components/UserActivityListener.js
```diff
@@ -5,6 +5,30 @@ import AccountStore from 'stores/AccountStore'
...
+        userAgent.substr(0, 4)
+      )
+    ) {
+      check = true
```
shahthepro, line 23: why not return `true` directly?

shahthepro (commented):
LGTM.

dapp/src/components/UserActivityListener.js
sparrowDom: Because I like to complicate things for no special reason :)

Thanks, fix here: https://github.com/OriginProtocol/origin-dollar/commit/2745259265c248cc6f4bea8c189830c0867acec8

Issue #270, opened 2020-09-29
Add Token Help


micahalcorn:
After a user mints OUSD for the first time, she may need to manually add the token address to her wallet. This is currently the case with MetaMask while we wait for them to merge [our PR](https://github.com/MetaMask/eth-contract-metadata/pull/572). However, it's not necessary for most of the mobile wallets. So we should be intelligent about when to encourage the user to add the token.

I recommend that we show a modal similar to [this one](https://zpl.io/V105K5k) only under the following circumstances:
- the current provider is MetaMask
- immediately after the balance changes from 0 to non-zero
- the user has not previously seen the modal

We only want to show it after a first-time mint, but it's not necessary to show it to a first-time minter who came to the DApp with OUSD already in her wallet. And, ideally, the modal would only appear _after_ all of the other animations are done so that she can see the success transition. Only after a balance is shown in the DApp is it relevant to see the balance in the wallet.

As for the content of the modal, we can show a copyable token address and/or implement a button using [EIP-747](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-747.md). I'm told by one of the authors of the EIP that it hasn't see much adoption across various wallets. But the main one we care about is MetaMask, and we know they support it. We should probably still show a small version of the address even if we include the button. I recommend timeboxing the EIP implementation and skipping altogether unless it can be done quickly.

sparrowDom:
PR for this issue: https://github.com/OriginProtocol/origin-dollar/pull/276

PR #276, opened 2020-09-30
show "add ousd coin to wallet" modal

sparrowDom:
After the user mints OUSD for the first time, and the OUSD balance finishes animating. We show the modal that prompts users to add OUSD coin to their wallet. 

If the provider is Metamask, we just show a button that automatically pre-fills all the needed data and user only needs to approve Metamask prompt.
If the provider is any other we show the contract address + decimals information. 

**related issue:** https://github.com/OriginProtocol/origin-dollar/issues/270


micahalcorn:
Most of the mobile wallets that I've tested detect the token automatically. So it's unnecessary (and potentially confusing) to show this information in those cases. I was proposing that we only show this for MetaMask users. I suppose the more thorough approach would be to manually blacklist all the environments that we know definitively detect the token.

dapp/src/utils/web3.js
```diff
@@ -44,7 +52,36 @@ export function truncateAddress(address) {
...
+    },
+    console.log
+  )
+}
+
+/* status of token wallets and OUSD:
+ * https://docs.google.com/spreadsheets/d/1bunkxBxfkAVz9C14vAFH8CZ53rImDNHTXp94AOEjpq0/edit#gid=1608902436
+ */
+export function providersNotAutoDetectingOUSD() {
+  return ['metamask', 'trust', 'alphawallet', 'mist', 'parity']
```
sparrowDom, line 77: @micahalcorn added this bit of functionality where we whitelist the providers where we show the "add OUSD modal". We can cross them out from here when the wallets pick up OUSD, or add new ones if we see in analytics that some provider is popular and isn't auto detecting OUSD.

sparrowDom:
@shahthepro merging this in. Happy to do any changes later in case need be.

shahthepro (commented):
Sorry just saw this PR. Looks good to me. LGTM

PR #286, opened 2020-10-05
public on state var generates not that useful getter on MixOracle

fxfactorial:
the generated getter is a bit useless as it gives a method that gives addr by index, not the full `address[]` 

hence:

1. removed public on `    address[] ethUsdOracles`
2. added view method to get the full `address[]`

### DIFF
--- a/contracts/contracts/oracle/MixOracle.sol
+++ b/contracts/contracts/oracle/MixOracle.sol
@@ -14,7 +14,7 @@ import {
 } from "../governance/InitializableGovernable.sol";
 
 contract MixOracle is IMinMaxOracle, InitializableGovernable {
-    address[] public ethUsdOracles;
+    address[] ethUsdOracles;
 
     struct MixConfig {
         address[] usdOracles;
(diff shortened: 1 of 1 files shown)


franckc (commented):
Code change LGTM

Though in order to deploy this to mainnet we'll need to add a new migration. Those are there:
https://github.com/OriginProtocol/origin-dollar/tree/master/contracts/deploy

The migration will likely involve
 - deploying the new contract
 - setting Chainlink/Uniswap Oracle on that contract
 - transfering governance of the contract from deployer account to governor
 - submitting a proposal to the governor to claim the MixOracle contract and for the vault to point to the new MixOracle

So overall we may want to hold off on merging that PR until we have more changes to the Oracle since the cost/benefit seems low.

Just my 2 francs :)

contracts/contracts/oracle/MixOracle.sol
```diff
@@ -40,6 +40,10 @@ contract MixOracle is IMinMaxOracle, InitializableGovernable {
         minDrift = _minDrift;
     }
 
+    function ethUsdOrcales() external view returns (address[] memory) {
```
tomlinton, line 43: Typo :)

tomlinton:
Will merge so it gets picked up if we ever redeploy the MixOracle!

PR #292, opened 2020-10-06
Decimal places

sparrowDom:
Show 2 decimal places in Transaction details when values are under a 100. Looks like this: 
Screenshot 2020-10-06 at 23 42 50
Screenshot 2020-10-06 at 23 40 14
Screenshot 2020-10-06 at 23 40 09


sparrowDom:
merging in happy to address stuff later if needed

PR #293, opened 2020-10-06
Redo rebasing!

DanielVF:
Primarily:
- Rebasing works off stablecoin quantities (no more downwards rebases due to stablecoin oracle fluctuations)
- Rebases ratchets upwards
- Redeem never returns more than one stablecoin per OUSD.
- Mint never pays more than one OUSD per stablecoin.

And:
- Massive gas reduction related to rebasing, redeeming, and allocating.
- Moved to hardcoded APR for now (will pull from a web app in the future).

Needed:
- [ ] **Deployment scripts** @tomlinton


contracts/contracts/mocks/MockVault.sol
```diff
@@ -10,15 +10,11 @@ contract MockVault is VaultCore, VaultInitializer {
         storedTotalValue = _totalValue;
     }
 
-    function totalValue() external returns (uint256) {
+    function totalValue() external view returns (uint256) {
```
tomlinton, line 13: 💙

contracts/test/vault/redeem.js
```diff
@@ -211,33 +204,30 @@ describe("Vault Redeem", function () {
...
+    // Total Vault Coins	450
+    // USDC Percentage	100	/	450	=	0.222222222222222
+    // DAI Percentage	350	/	450	=	0.777777777777778
+    // USDC Value Percentage			0.222222222222222	*	1.3	=	0.288888888888889
+    // DAI Value Percentage			0.777777777777778	*	1.2	=	0.933333333333333
+    // Output to Dollar Ratio	1.22222222222222
+    // USDC Output	250	*	0.222222222222222	/	1.22222222222222	=	45.4545454545454
+    // DAI Output	250	*	0.777777777777778	/	1.22222222222222	=	159.090909090909
+    // Expected USDC	900	+	45.4545454545454	=	945.454545454545
+    // Expected DAI	850	+	159.090909090909	=	1009.09090909091
```
tomlinton, line 223: Pretty minor nit but the formatting looks off here in GitHub code review, haven't looked at in vim yet

contracts/contracts/vault/VaultCore.sol
```diff
@@ -53,16 +52,10 @@ contract VaultCore is VaultStorage {
...
-            (priceAdjustedDeposit > rebaseThreshold && !rebasePaused) ||
-            (priceAdjustedDeposit >= autoAllocateThreshold)
-        ) {
-            assetPrices = _getAssetPrices(false);
-        }
 
         // Rebase must happen before any transfers occur.
         if (priceAdjustedDeposit > rebaseThreshold && !rebasePaused) {
-            rebase(assetPrices, true);
+            rebase(true);
```
fxfactorial, line 58: maybe a nit - bool params are icky - perhaps better to make rebaseWithSync() and rebaseWithoutSync - and in any case - why need to to include the sync on the pairs - anyway whoever does the first interaction with the pair will cause a sync , so could avoid a cost to that call to the pairs

tomlinton: Not sure I understand, the sync needs to be called on every Uniswap pair every time the rebase is called because it changes the balances held in Uniswap which breaks the constant product requirement

fxfactorial: The pair itself will call sync anytime anyone interacts with it - like add liquidity/remove liquidity or a swap. So its isn't strictly necessary for us to call sync , since sync anyway will be called. The pair itself will check balances of its tokens

contracts/contracts/vault/VaultCore.sol
```diff
@@ -307,45 +292,31 @@ contract VaultCore is VaultStorage {
...
-    function _totalValue(uint256[] memory assetPrices)
-        internal
-        view
-        returns (uint256 value)
-    {
-        return
-            _totalValueInVault(assetPrices) +
-            _totalValueInStrategies(assetPrices);
+    function _totalValue() internal view returns (uint256 value) {
+        return _totalValueInVault() + _totalValueInStrategies();
```
fxfactorial, line 305: maybe won't happen anytime soon - but what happens when the vault is super successful and has so many assets values - then it overflows - or what if I take some massive flashloan on multiple assets and deposit - then cause overflow.

contracts/contracts/vault/VaultCore.sol
```diff
@@ -307,45 +292,31 @@ contract VaultCore is VaultStorage {
...
      * @dev Internal to calculate total value of all assets held in Vault.
      * @return uint256 Total value in ETH (1e18)
      */
-    function _totalValueInVault(uint256[] memory assetPrices)
-        internal
-        view
-        returns (uint256 value)
-    {
+    function _totalValueInVault() internal view returns (uint256 value) {
         value = 0;
```
fxfactorial, line 313: value already starts at 0

contracts/contracts/vault/VaultCore.sol
tomlinton: It'll call sync after a swap (for example), but the transaction will revert prior to that if the balances are out of sync https://github.com/Uniswap/uniswap-v2-core/blob/master/contracts/UniswapV2Pair.sol#L182

fxfactorial: what you listed was making sure that the LP fee was taken into account - for example - after the yam rebase occured (the one that caused a massive amount of new token) , people were still able to do a swap - and that initial swap caused a sync (that first person got a great deal) 

Let me try to give an example wrt this PR

contracts/contracts/vault/VaultCore.sol
```diff
@@ -109,7 +102,7 @@ contract VaultCore is VaultStorage {
         }
         // Rebase must happen before any transfers occur.
         if (priceAdjustedTotal > rebaseThreshold && !rebasePaused) {
```
fxfactorial, line 104: not fully fleshed out in my mind, but thinking - why is the rebaseThreshold at 1000e18 (seems low?) - what if I just keep doing flashloans and keep depositing and causing a rebase on redeem. Then the proportions change up on withdraw - and in this way i can bleed out the vault of more valuable token. Will think some more of pulling this attack off

contracts/contracts/vault/VaultCore.sol
tomlinton: I can't see the fee used in that swap function. Per the error `UniswapV2: K`, k is the constant product.

contracts/contracts/vault/VaultCore.sol
DanielVF: Both the mint and the redeem do a rebase if the amount is over $1,000. So I don't think flashloans aren't super helpful? Anyone can cause a rebase to happen anytime - it's not required to deposit or withdraw.

As of this morning a mint/redeem combo is around $25 USD in gas. Along with a withdrawal fee that works out to $5 , that means that in order to abuse selectively rebasing under, I think that means that there needs to be more than 3% of the value of the contract in unrebased yield in order to break even on minting in, rebasing, and redeeming out. That's a whole lot of yield. Even at 20% APY, that's about two months of going without a rebase.

contracts/contracts/vault/VaultCore.sol
DanielVF: Lets say I run an apple store during these times of coronavirus. Only one customer at time in my store, and I only go into the store after each customer to check on things. I also have a time machine that can revert. 

So a customer walks into the store, takes some apples, leaves some cash. I walk into the store, carrying my accounting books. I count up the cash in the store, the number of apples in the store, compare both to my books, and if the customer left too little money I use the time machine to revert. Otherwise, I update my books and walk back out.

This means that I don't use the time machine if a customer left too much money, or took too few apples. Importantly, this means that if the apples in the store spontaneously replicate, than a customer can take those extra apples out with a normal order, and I won't know the difference.

In our case, calling sync provides two protections. First, if a horse snuck into the store and ate some apples, then the next customer could walk in, do a perfectly fair transaction, and get reverted. Syncing stops this. Secondly, the sync prevents someone who knows when there are extra apples from gradually gaining a greater and greater share of the  at the expense of everyone else.

contracts/test/vault/redeem.js
DanielVF: Lol, it's literally a spreadsheet copy/pasted. I can take out the extra spaces.

contracts/contracts/vault/VaultCore.sol
DanielVF: I was curious about this, so I did some back of napkin math.

2^256 is kind of a big number. Conservatively it's 1e77. Let's use fixed point, 1e18, so we can represent 1e59 dollars. The current number of dollars in the world is around 1e13, so we are fairly good there. The current mass of the universe is estimated at 1e53 kilograms, which is 1e56 grams. I think we have enough bits available to store $1,000 for each gram of matter in the universe. Scaling ethereum's transaction per second might be a limiting factor here.

A bigger danger could be the compounding effects of yields. With year after year of 120% APY, (more than doubling every year) and the world using OUSD for its wealth, it could take 138 years for us run out of bits. I guess around year 136, we'd make a new contract with fewer decimals, and let people convert from one to the new OUSD.

contracts/contracts/vault/VaultCore.sol
DanielVF: It's definitely worth thinking about the proportions.

contracts/test/vault/redeem.js
tomlinton: And would you look at that, it looks fine here when I've commented on it haha

contracts/contracts/vault/VaultCore.sol
```diff
@@ -530,53 +488,70 @@ contract VaultCore is VaultStorage {
         internal
         returns (uint256[] memory outputs)
     {
-        uint256[] memory assetPrices = _getAssetPrices(true);
+        // We always give out coins in proportion to how many we have,
+        // So for every 1 DAI we give out, we'll be handing out 2 USDT
```
tomlinton, line 492: * if we had twice as many USDT as DAI

contracts/test/vault/redeem.js
DanielVF: Tabs VS spaces. :)

contracts/contracts/vault/VaultCore.sol
DanielVF: Thanks! That sentence got out of order.

contracts/contracts/vault/VaultCore.sol
DanielVF: I beat on it for a while, didn't find a bug there. Still be glad if you can find something.

DanielVF:
Merging change to rebase off stablecoin numbers, ratchet rebases, and cap mint/redeem oracle amounts.

Issue #129, opened 2020-09-02
Support MetaMask Mobile


micahalcorn:
Since [it appears that MetaMask is moving forward with their DApp browser](https://www.bloomberg.com/news/articles/2020-09-02/metamask-s-blockchain-mobile-app-opens-doors-for-next-level-web), let's make sure that our DApp works with it. Hopefully, the environment is similar to the desktop browser extension.

sparrowDom:
Done via: https://github.com/OriginProtocol/origin-dollar/pull/297

PR #297, opened 2020-10-07
Mobile Metamask

sparrowDom:
- login popup is no longer shows up on mobile metamask, but directly logs in
- automatically hides the "waiting for you to approve" modal after 5 seconds if environment is Metamask app.
- adding vConsole for easier debugging on mobile devices

**corresponding issue:** https://github.com/OriginProtocol/origin-dollar/issues/129

The second workaround is required because the Metamask app has a bug where the contract call does not throw an exception when user rejects the transaction. The javascript execution is just stuck on the `await` line calling the contract function. Interestingly if you quit the app and re-enter immediately, the correct error with "user rejected the transaction..." still gets thrown.

The other interesting error I've found is that in 60% of the cases on my android (Oneplus 8Pro) the metamask app wouldn't inject the provider to window object (window.ethereum). Not reproducible on the iPhone, but happens constantly on Android. The dapp login doesn't work of course when that happens, and there is not much we can do about it.

### DIFF
--- a/dapp/package.json
+++ b/dapp/package.json
@@ -75,11 +75,11 @@
     "react-autosize-textarea": "^7.1.0",
     "react-bootstrap": "^1.0.0-beta.16",
     "react-cookie": "^4.0.3",
+    "react-copy-to-clipboard": "^5.0.2",
     "react-dom": "16.13.1",
     "react-router-dom": "5.2.0",
     "react-styl": "^0.0.3",
     "react-toastify": "^6.0.8",
-    "react-copy-to-clipboard": "^5.0.2",
     "sass": "^1.26.10",
     "web3modal": "^1.9.0"
   },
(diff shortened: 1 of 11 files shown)


micahalcorn:
Nice work, and good to know about the issues that are out of our control, thanks 🦊

micahalcorn:
@sparrowDom it might be good to create corresponding issues at https://github.com/MetaMask/metamask-mobile or link to existing ones. I did a quick search and didn't find any that I thought would match.

shahthepro (approved):

sparrowDom:
Just for tracking purposes found a similar issue with `window.ethereum` being undefined and left a comment [here](https://github.com/MetaMask/metamask-mobile/issues/1140).

And reported [a separate issue](https://github.com/MetaMask/metamask-mobile/issues/1880) regarding the rejected transactions not throwing errors.

Issue #291, opened 2020-10-06
Alchemy calls


sparrowDom:
When a wallet is not connected the dapp still makes way too many calls. (See call frequency [here](https://dashboard.alchemyapi.io/apps/fc9ecpa5aiddqqjl))

Reported by @nick & @franckc 🙏

sparrowDom:
Pr for this prepared here: https://github.com/OriginProtocol/origin-dollar/pull/298

PR #298, opened 2020-10-07
reduce the amount of calls to Alchemy

sparrowDom:
- Slightly increase the interval time with which we fetch account balances and moderately increase interval of fetching exchange rates
- do not fetch exchange rates when wallet is not connected
- do not fetch exchange rates when user is not active 

**corresponding issue:** https://github.com/OriginProtocol/origin-dollar/issues/291


sparrowDom:
@shahthepro @franckc merging this into master, and deploying to see if the alchemy api request numbers drop. Happy to address stuff later if needed.

franckc:
Nice work @sparrowDom ! :)

PR #300, opened 2020-10-08
Script to propose and execute proposal on the governor

franckc:
- Added scripts for sending propose() and execute() transactions to the governor
 - Small refactor to move method proposeArgs() from helpers/ to utils/


franckc:
Merging this in. Happy to address any feedback in a subsequent PR...

PR #303, opened 2020-10-09
Migration scripts for upgrading VaultCore

franckc:
On Mainnet, we'll deploy the contract via the migration 12, then manually send and execute a governance proposal to do the upgrade (we'll use the scripts under scripts/governor to do that).

On other networks, the migration 12 handles everything.


DanielVF (approved):
lgtm

tomlinton (approved):

franckc:
Renamed migration files to use 3 digits so that they execute in the proper order.

Issue #259, opened 2020-09-27
Handle Dropped & Replaced Transactions


micahalcorn:
If a user decides to "speed up" her transaction by submitting a second one with the same none and a higher gas price, our DApp is unaware of the new transaction and gets stuck waiting for the old one. This is a common problem with no simple solution yet.

See the following:
- [EIP 2831](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-2831.md)
- [EIP discussion](https://ethereum-magicians.org/t/eip-2831-transaction-replacement-message-type/4448)
- https://github.com/ethereum/web3.js/issues/3585

joshfraser:
In the meantime...

Until the transaction is confirmed, we could continue watching the blockchain for additional transactions that are signed by the original wallet & see if they look like replacements. Doesn't seem too hard.

sparrowDom:
From my research when a client wants to increase gas it creates another transactions with higher gas price and the same nonce. Now the problem is that the web3 browser / Metamask do not communicate this back to the Dapp. There are some proposed solutions (that Micah outlined above) but nothing we could use today.

I see 2 ways we could get this done:
1. **slower way** - develop a backend that listens to the blockchain for all events / transactions on OUSD contract and expose that data via API. We could use @DanielVF  's `debug.ousd.com` for that, or fork it and develop a separate service. We need a similar service anyway to display all the debug information that were once part of the DApp designs.
2. **faster way** - use Etherscan API and pass the api key to the client to fetch all the transactions of an account. And see if there is a transaction that has replace the initial mint/redeem/allow one.

Personally I think the option 2 is not so messy and doable relatively quickly and would vote for that one if we want to have this functionality reasonably quickly. Actually I see that Etherscan allows 1 request every 5 seconds [to their api](https://api.etherscan.io/api?module=account&action=txlist&address=0xFD9E6005187F448957a0972a7d0C0A6dA2911236&startblock=0&endblock=99999999&sort=desc) without the need for an API key. That is more than enough to get the latest 20 transactions of an account and check for possible gas price upgrade transaction.

shahthepro:
I would suggest to subscribe to new transactions using WS provider and adding a `from` filter instead of polling though.

sparrowDom:
That might be even more elegant, thanks @shahthepro !

sparrowDom:
PR for this issue: https://github.com/OriginProtocol/origin-dollar/pull/304

micahalcorn:
[...]

PR #304, opened 2020-10-09
Sparrow dom/dropped connections

sparrowDom:
Handle the case when a wallet drops a transaction and re-issues one with higher gas price. 

**Correlating issue:** https://github.com/OriginProtocol/origin-dollar/issues/259


dapp/src/components/TransactionListener.js
```diff
@@ -57,6 +57,82 @@ class TransactionListener extends Component {
       })
       this.save(newTransactions)
     }
+
+    const nonMinedTx = this.props.transactions.filter((t) => !t.mined)
+    if (nonMinedTx.length > 0 && !this.state.wsProvider) {
+      this.startWebsocketListener()
+    } else if (nonMinedTx.length === 0 && this.state.wsProvider) {
+      this.cleanupWebSocketProvider()
```
shahthepro, line 65: Should this be called in `componentWillUnmount` too?

dapp/src/components/TransactionListener.js
```diff
@@ -57,6 +57,82 @@ class TransactionListener extends Component {
...
+              const newTx = { ...tx }
+
+              // change the necessary fields
+              newTx.hash = eventTx.hash
+              newTx.isError = false
+              newTx.mined = true
+              newTx.observed = false
+              newTx.blockNumber = eventTx.blockNumber
+
+              const newTransactions = [...otherTransactions, newTx]
```
shahthepro, line 103: Do we want to show that the Tx got replaced somewhere in the DApp?

shahthepro (approved):
LGTM

dapp/src/components/TransactionListener.js
sparrowDom: I don't think it is super necessary, since the listener only runs while the transaction is being mined. But it still improves the code, so adding it in. Thanks!

dapp/src/components/TransactionListener.js
sparrowDom: good question. It definitely is relevant info, just not sure if it is important enough.

@micahalcorn what do you think?

sparrowDom:
Thanks for the CR @shahthepro merging this in. In case we want to show the additional info for the dropped transactions happy to do it in later commits.

dapp/src/components/TransactionListener.js
micahalcorn: When the user "speeds up" the transaction or manually submits another one with the same nonce, will it produce a second transaction card in the info feed or will the data from newTx just overwrite the data from nonMinedTx and take its place? I think the simplest solution is just to replace the data in the card and ignore the nonMinedTx. It might get a little confusing if MetaMask shows both and we only show the new one, but I think it's consistent with the user's intent to replace the transaction instead of showing both.

sparrowDom: yeah that is what happens. The old transaction data gets replaced with the new tx data. 

If there was really a need to show a dropped transaction I would add some info in the expanded transaction space saying that this transaction replaced the other submitted one. And show an etherscan link of the old one. Though we can probably spend engineering resources on a lot of other stuff first.

PR #305, opened 2020-10-11
Change oracle to allow register/unregister of eth price oracle

crazybuster:
The deploy script still needs a bit more testing, but running it against ganache doesn't force it to upload new contract(because the 1_core already deploy the same one), should be ok on rinkeby?


contracts/deploy/_11_new_mix_oracle.js
```diff
@@ -0,0 +1,196 @@
...
+  );
+  log("Deployed MixOracle", d);
+
+
+  const chainlinkOracle = await ethers.getContract("ChainlinkOracle");
+
+  // Deploy the New OpenUniswap oracle.
+  d = await deploy("OpenUniswapOracle", {
+    from: deployerAddr,
+    args: [oracleAddresses.openOracle, assetAddresses.WETH],  // REMEMBER to update the oracle address
```
franckc, line 55: Is this comment relevant anymore?

contracts/deploy/_11_new_mix_oracle.js
```diff
@@ -0,0 +1,196 @@
...
+  log("Registered uniswap pair USDC/ETH");
+
+  t = await uniswapOracle
+    .connect(sDeployer)
+    .registerPair(oracleAddresses.uniswap.USDT_ETH, await getTxOpts());
+  await ethers.provider.waitForTransaction(t.hash, NUM_CONFIRMATIONS);
+  log("Registered uniswap pair USDT/ETH");
+
+  const mixOracle = await ethers.getContract("MixOracle");
+  if (isMainnetOrRinkebyOrFork) {
```
franckc, line 85: Not sure we need this if / else statement given that we skip this migration if it's not Mainnet/Rinkeby/Fork.

I would just keep only the code that's in the if statement and not make it conditional. And get rid of all the code in the else statement. Basically just relying on the skip on live 194 to determine if this whole migration runs or not.

franckc (approved):
Thanks for sending this PR so quickly! Couple minor comments otherwise LGTM

franckc:
In terms of the steps for deploying to Mainnet, does the following sound right:
1. run migration 11
2. create a proposal for the Governor to call
  - claimGovernance() on the new mixOracle
  - claimGovernance() on the new uniswapOracle
  - setPriceProvider(newMixOracleAddress) on the VaultAdmin
3. submit a transaction on the multi-sig to call enqueue(proposalId) and get it signed/executed
4. execute proposal

tomlinton (approved):

franckc:
Ran mig11 on mainnet.

PR #306, opened 2020-10-12
Curve 3pool strategies

tomlinton:
Adds two 3pool strategies, one for USDT and one for USDC. Each supports only a single asset to make it cleaner. Includes Curve gauge staking but no boosting. Includes the deployment of the strategies + proxies. Does not include the work we need to do to migrate/reallocate funds.

I'm happy for us to merge and run the deploys on this stuff but propose we hold off from actually using them until we deploy the changes from #293, #303, and #305. After that we can integrate these with the Vault. I wrote out what needs to happen for that in Discord but included here as well:

- [x] Deploy the two new 3pool strategies and proxies (`013_three_pool_strategies.js`)
- [x] Deploy a new Compound Strategy proxy (we have to do this because we don't want the storage from the old one) (`014_dai_compound_strategy.js` in `tomlinton/3pool-fund-migration`)
- [x] Deploy a new Compound Strategy that only supports DAI (`014_dai_compound_strategy.js` in `tomlinton/3pool-fund-migration`)
- [x] Call removeStrategy on the Vault with the address of the Compound strategy
- [x] Call addStrategy * 3 on the Vault with the address of the new Compound strategy and the two 3pool strategies, the weights aren't important because each strategy only supports a single asset
- [x] Call allocate


contracts/utils/addresses.js
```diff
@@ -24,17 +24,21 @@ addresses.mainnet.COMP = "0xc00e94Cb662C3520282E6f5717214004A7f26888";
...
-
+// Curve
+addresses.mainnet.CRV = "0xd533a949740bb3306d119cc777fa900ba034cd52";
+addresses.mainnet.CRVMinter = "0xd061D61a4d941c39E5453435B6345Dc261C2fcE0";
+addresses.mainnet.ThreePool = "0xbEbc44782C7dB0a1A60Cb6fe97d0b483032FF1C7";
+addresses.mainnet.ThreePoolToken = "0x6c3f90f043a72fa612cbac8115ee7e52bde6e490";
+addresses.mainnet.ThreePoolGauge = "0xbFcF63294aD7105dEa65aA58F8AE5BE2D9d0952A";
 // Open Oracle
-addresses.mainnet.openOracle = "0x922018674c12a7f0d394ebeef9b58f186cde13c1";
+addresses.mainnet.openOracle = "0x9b8eb8b3d6e2e0db36f41455185fef7049a35cae";
```
franckc: I don't think this should be updated. The current feed is 0x922018674c12a7f0d394ebeef9b58f186cde13c1

contracts/contracts/strategies/ThreePoolStrategy.sol
```diff
@@ -0,0 +1,256 @@
...
+     * @param _amount Amount of asset to deposit
+     * @return amountDeposited Amount of asset that was deposited
+     */
+    function deposit(address _asset, uint256 _amount)
+        external
+        onlyVault
+        returns (uint256 amountDeposited)
+    {
+        require(_amount > 0, "Must deposit something");
+        // 3Pool requires passing depodit amounts for all 3 assets, set to 0 for
```
franckc, line 87: super nit: typo on deposit and on Gauge on line 94 below.

contracts/contracts/utils/InitializableAbstractStrategy.sol
```diff
@@ -76,6 +72,22 @@ contract InitializableAbstractStrategy is IStrategy, Initializable, Governable {
         }
     }
 
+    /**
+     * @dev Single asset variant of the internal initialize.
+     */
+    function _initialize(
```
franckc, line 78: This is also used by the CompoundStrategy. 
When we run migration 13 I don't think buidler will try to re-run the CompoundStrategy deployment but something I'm taking a mental note to watch for.

franckc (commented):
I just did a quick scan of the contracts. Looks very clean but would be great if somebody else could do an in-depth review.

A few minor comments inline.

I'll work on this branch to add the migration file for deploying the new CompoundSrategy.

tomlinton:
Thanks @franckc, I started on a branch at `tomlinton/3pool-fund-migration`. I've got as far as deploying Compound strategy with DAI only.

contracts/contracts/utils/InitializableAbstractStrategy.sol
tomlinton: It shouldn't do, I think you made the changes to the `.migration.json` which should be the source of truth for which migrations it runs. Changes to the underlying files shouldn't matter. It's important to be careful with these initializers though, we don't want to inadvertently allow reinitialization of an already initialized strategy.

contracts/deploy/013_three_pool_strategies.js
```diff
@@ -0,0 +1,219 @@
...
+  );
+
+  transaction = await CurveUSDTStrategy.connect(sDeployer).transferGovernance(
+    strategyGovernorAddress,
+    await getTxOpts()
+  );
+  await ethers.provider.waitForTransaction(transaction.hash, NUM_CONFIRMATIONS);
+  log(`CurveUSDTStrategy transferGovernance(${strategyGovernorAddress} called`);
+
+  if (!isMainnet && !isRinkeby) {
```
franckc, line 130: Why not running those calls to initialize on Mainnet?

tomlinton: 👍

contracts/contracts/strategies/ThreePoolStrategy.sol
```diff
@@ -0,0 +1,256 @@
...
+    function collectRewardToken() external onlyVault {
+        ICRVMinter minter = ICRVMinter(crvMinterAddress);
+        minter.mint(crvGaugeAddress);
+        IERC20 crvToken = IERC20(rewardTokenAddress);
+        uint256 balance = crvToken.balanceOf(address(this));
+        require(
+            crvToken.transfer(vaultAddress, balance),
+            "Reward token transfer failed"
+        );
+        emit RewardTokenCollected(vaultAddress, balance);
```
fxfactorial, line 72: small things: 

1) wrt transfer, maybe can use safeTransfer since you're putting a require on it anyway 
2) anyway the erc20 will emit the transfer, so what's the extra event benefit

contracts/contracts/strategies/ThreePoolStrategy.sol
```diff
@@ -0,0 +1,256 @@
...
+        // Do the deposit to 3pool
+        ICurvePool(platformAddress).add_liquidity(_amounts, uint256(0));
+        // Deposit into Gauge
+        IERC20 pToken = IERC20(assetToPToken[_asset]);
+        ICurveGauge(crvGaugeAddress).deposit(
+            pToken.balanceOf(address(this)),
+            address(this)
+        );
+        amountDeposited = _amount;
+        emit Deposit(_asset, address(platformAddress), amountDeposited);
```
fxfactorial, line 101: same here for deposit - the gauge will anyway emit a deposit event - so what extra benefit we'd get here

contracts/contracts/strategies/ThreePoolStrategy.sol
```diff
@@ -0,0 +1,256 @@
...
+            poolCoinIndex
+        );
+        // Calculate how many platform tokens we need to withdraw the asset amount
+        uint256 withdrawPTokens = totalPTokens.mul(_amount).div(maxAmount);
+        if (contractPTokens < withdrawPTokens) {
+            // Not enough of pool token exists on this contract, must be staked
+            // in Gauge, unstake
+            ICurveGauge(crvGaugeAddress).withdraw(withdrawPTokens);
+        }
+        curvePool.remove_liquidity_one_coin(withdrawPTokens, poolCoinIndex, 0);
```
fxfactorial, line 138: are we sure to leave the min amount removed as 0?

contracts/contracts/strategies/ThreePoolStrategy.sol
```diff
@@ -0,0 +1,256 @@
...
+        IERC20(_asset).transfer(_recipient, _amount);
+        // Transfer any leftover dust back to the vault buffer.
+        uint256 dust = IERC20(_asset).balanceOf(address(this));
+        if (dust > 0) {
+            IERC20(_asset).safeTransfer(vaultAddress, dust);
```
fxfactorial, line 143: why safeTransfer sometimes used and plain transfer otherwise used?

contracts/test/_trace.js
```diff
@@ -0,0 +1,91 @@
+/**
+ * Allows you to trace the execution of solidity call.
+ */
+const bre = require("@nomiclabs/buidler");
+
+// let tracingEnabled = false;
```
fxfactorial, line 6: this is pretty cool - assume it only works for the buidler vm right - not on a forked mainnet?

fxfactorial:
Looks correct - small things on perhaps redundant events emitted & small risks of using min amount as 0 in deposits (perhaps at times when the 3 pool has an unfavorable distribution of stablecoin wrt what we are depositing)

contracts/test/_trace.js
DanielVF: Correct, just the buidler vm.

contracts/contracts/strategies/ThreePoolStrategy.sol
DanielVF: SafeTransfer is only needed if the coin could fail to throw on a failed transfer, such as USDT. So when working with only a specify known good coin, it's fine to just use transfer.

However, this is potentially transferring USDT here, and so should definitely be safeTransfer. Probably my mistake.

^ @tomlinton

contracts/contracts/strategies/ThreePoolStrategy.sol
```diff
@@ -0,0 +1,256 @@
...
+
+    /**
+     * @dev Get the total asset value held in the platform
+     *  This includes any interest that was generated since depositing
+     *  We calculate this by calculating a what we would get if we liquidated
+     *  the allocated percentage of this asset.
+     * @param _asset      Address of the asset
+     * @return balance    Total value of the asset in the platform
+     */
+    function checkBalance(address _asset)
```
DanielVF, line 179: This should probably return a zero if the _asset passed along does not match the strategy's asset. That's really going to confuse someone someday.

DanielVF (changes requested):

contracts/contracts/strategies/ThreePoolStrategy.sol
tomlinton: We can collect all the events from our own contracts

fxfactorial: right but that's something we can build off chain and filter on our relevant addresses. The event is not free, it cost gas and with three parameters, its pricey. I defer to you on decision keep/not :+1:

PR #308, opened 2020-10-13
Mint Gas Increase

micahalcorn:
This increases the default gas limit for `mint` transactions according to [this quick analysis](https://docs.google.com/spreadsheets/d/1A512Y0SmpI7lHKbsIi_poOWrGp_wnSnxwuze3lOlkt0/edit#gid=0). I'll go ahead and merge/deploy this. We should come back and revisit the rebase and allocate scenarios involving multiple input currencies because the current logic is a bit too conservative on those.


franckc:
Thanks for doing the analysis and upgrading those limits @micahalcorn !

PR #317, opened 2020-10-14
Mixpanel Conversion Tracking

micahalcorn:
This adds some missing events to the end of the mint and redeem funnels. It also adds the Ethereum account address as an alias on users, which probably isn't the best solution but is a start.


sparrowDom (approved):
Looks good, yeah like you mentioned the alias solution is better than nothing and should work for most cases. We will need to implement something custom if we want to properly handle multiple wallets of the same user.

Issue #275, opened 2020-09-30
OUSD animated balance confusion


sparrowDom:
In the dapp we animate the OUSD balance by reading the current OUSD balance from the contract and simulating the current (incrementing) value by using the Apy returned by the Vault. While this gives users nice impression of OUSD balance increasing it has its shortcomings.

If a user refreshes the dapp or returns in 1-2 hours and rebase has not happened in the meantime the OUSD balance "resets" to the balance returned by the contract. A user would of course expect the balance to increase while he/she was away.

One suggestion how to tackle this would be to somehow get the `blockNumber` of the latest rebase transaction (or timestamp). Using that we could guess a more accurate approximate initial OUSD balance number. 

We could probably do that using the contract events, but we currently do not have a backend client listening for those kind of events and exposing data via API. We could also store the latest rebase timestamp on the Vault contract.

Perhaps there is a more elegant solution to this problem? cc: @DanielVF @tomlinton @micahalcorn

micahalcorn:
This seems like the sort of thing that can depend on a centralized service. Its purpose is purely for user experience and not for on-chain logic, so I would think we would want to avoid adding more on-chain storage. I'm not sure what all @tomlinton has planned for his APY solution, but maybe we take that opportunity to store rebase events, take APY snapshots, and smooth it all out and give the DApp a small growth factor to apply to balances.

tomlinton:
The fundamental problem is rebases don't happen often enough yet. The animated balance will always reset if a user goes away and comes back because we are currently rebasing once a day (if that). So I don't think increasing the accuracy of the animated value (or the APY) will solve this.

I think we should revisit this when we've got a) more usage, b) more rebases and c) higher yield/more accurate yield calculation.

sparrowDom:
@micahalcorn mentioned another user had this issue.

Discussion from Discord: 
@sparrowDom : Suggesting that for a quick fix, we could pick up the animation of OUSD from when we left it. And if significant time passes in between we just simulate the mean time increase. We would store the necessary stuff into local storage, so it wouldn't work across devices.

@micahalcorn : I think that makes sense as long as we can tell if there has been a consequential event since the balance was las retrieved. It could be a rebase, mint, redeem, or transfer

[...]

PR #320, opened 2020-10-14
make ousd animation survive refreshes

sparrowDom:
make ousd animation survive refreshes

**Connected issue:** https://github.com/OriginProtocol/origin-dollar/issues/275


sparrowDom:
@shahthepro merging this in happy to address issues in case needed later.

PR #329, opened 2020-10-16
Gas Limit Defaults Experiment

micahalcorn:
Here is an opinionated adjustment to both the default gas limit amounts and also the logic used to implement them. The primary objective is to reduce the likelihood of users manually reducing their gas limits and ending up with failed transactions. This will hopefully be accomplished with a few different changes.

- Add rebase awareness to redemptions
- Move from nice round defaults (e.g. 250000) to seemingly-precise numbers (e.g. 243729)
- Treat 2-coin mints the same as 3-coin mints since there doesn't seem to be a difference in the wild
- Reduce some limits that are now far too conservative since @DanielVF's rebase changes took effect

Here is a quick analysis showing the logic used to reach these new defaults and revealing that there's still room to lower them further: [Gas Cost Analysis](https://docs.google.com/spreadsheets/d/1A512Y0SmpI7lHKbsIi_poOWrGp_wnSnxwuze3lOlkt0/edit#gid=906056811)

I've deployed this branch to the staging environment and tested scenarios involving each of the non-allocate constants. I'll wait to merge and deploy to production until someone else gets a chance to review. 😁


dapp/src/components/buySell/BuySellWidget.js
```diff
@@ -242,20 +242,30 @@ const BuySellWidget = ({
...
-      } else if (totalMintAmount.gte(rebaseThreshold.mul(96).div(100))) {
-        gasLimit = gasLimits.MINT_REBASE_GAS_LIMIT
+      const involvingMultipleCoins = mintAddresses.length > 1
+      const aboveAllocateThreshold = totalMintAmount.gte(
+        allocateThreshold.mul(96).div(100)
+      )
+      const aboveRebaseThreshold = totalMintAmount.gte(
+        rebaseThreshold.mul(96).div(100)
+      )
+
```
micahalcorn, line 252: 👇 I realize that nested conditionals aren't great, but this is much easier to reason about than the prior base + additions thing. If I'm correct that we don't need significantly more gas for the nth coin beyond 2, then this allows us to be more precise and be more declarative with the constants.

dapp/src/components/buySell/BuySellWidget.js
```diff
@@ -242,20 +242,30 @@ const BuySellWidget = ({
...
-      } else if (mintAddresses.length === 3) {
-        gasLimit += gasLimits.MINT_3_COIN_ADDITION_GAS_LIMIT
+        if (aboveAllocateThreshold) {
+          gasLimit = gasLimits.MINT_ALLOCATE_GAS_LIMIT
+        } else if (aboveRebaseThreshold) {
+          gasLimit = gasLimits.MINT_REBASE_GAS_LIMIT
+        } else {
+          gasLimit = gasLimits.MINT_GAS_LIMIT
+        }
       }
```
micahalcorn, line 269: ☝️ I realize that nested conditionals aren't great and this is more verbose, but it's much easier to reason about than the prior base + additions thing. If I'm correct that we don't need significantly more gas for the nth coin beyond 2, then this allows us to be more precise and be more declarative with the constants.

dapp/src/components/buySell/SellWidget.js
```diff
@@ -170,13 +170,26 @@ const SellWidget = ({
     )
     coinData.ousd = ousdToSell
 
+    const [rebaseThreshold] = await Promise.all([
+      vaultContract.rebaseThreshold(),
+    ])
```
sparrowDom, line 175: Could the above be simplified as: 
```
const rebaseThreshold = await vaultContract.rebaseThreshold()
```

dapp/src/components/buySell/BuySellWidget.js
sparrowDom: I think the way this is expressed is perfectly fine and readable. Sometimes (though rarely) the nested condition statements are the best approach.

sparrowDom (approved):
Aside from the minor nitpick looks good to me. 

I very much like the idea of not going with the whole round numbers. The psychological effect makes sense. It reminds me to one of the negotiating strategies, where picking a non whole number gives the other party the impression that there is some logic to it even when there is not. Did real estate background help with this one? :)

dapp/src/components/buySell/BuySellWidget.js
```diff
@@ -242,20 +242,30 @@ const BuySellWidget = ({
...
         vaultContract.rebaseThreshold(),
       ])
-
-      if (totalMintAmount.gte(allocateThreshold.mul(96).div(100))) {
-        // Define gas limit only when the amount is over threshold
-        gasLimit = gasLimits.MINT_ALLOCATE_GAS_LIMIT
-      } else if (totalMintAmount.gte(rebaseThreshold.mul(96).div(100))) {
-        gasLimit = gasLimits.MINT_REBASE_GAS_LIMIT
+      const involvingMultipleCoins = mintAddresses.length > 1
+      const aboveAllocateThreshold = totalMintAmount.gte(
```
franckc, line 246: super nit: maybe add a comment as to why we lower the threshold to 96% so that we remember why next time we read this code.

franckc (approved):
Looks great - and love the idea of using those non rounded numbers. great experiment :)

dapp/src/components/buySell/BuySellWidget.js
micahalcorn: Definitely 👍 I inherited that part but I'll add something in.

dapp/src/components/buySell/SellWidget.js
micahalcorn: Ah, yes 😇 I meant to come back and actually read that line but got in a hurry.

PR #330, opened 2020-10-16
Bug fixes

sparrowDom:
- when allowances do not load up in the dapp show a blocking error modal prompting the user to refresh
- fix false display of Apy when the backend returns 0

This is what the error modal looks like. Not super polished but should be okayish to that couple (hopefully not more) of users who will see it
Screenshot 2020-10-16 at 16 16 14


sparrowDom:
@micahalcorn added you to the PR to give you the opportunity to disagree with the solution. 

One user had a problem where for some reason the allowances have not yet loaded up when he was trying to mint. It could be that he was super fast clicking on the mint button, or there has been some other error.

sparrowDom:
@micahalcorn @shahthepro merging this in happy to address stuff if needed later.

PR #333, opened 2020-10-20
[WIP] Adding uniswap strategy - accommodate specific deposit strategies 

fxfactorial:
Current PR presents an alternative flow for deposits of specific strategies - pushed by the need of uniswap today and any in the future. 

I think current approach leaves existing code alone (keeps old APIs - while extending for the special cases as each strategy might want) 

took me a little bit of time to think about how to do it decently elegantly - expect quite a bit more commits coming in pipeline

### DIFF
--- a/contracts/contracts/strategies/UniswapStrategy.sol
+++ b/contracts/contracts/strategies/UniswapStrategy.sol
@@ -1,0 +1,98 @@
+pragma solidity 0.5.11;
+
+import "@uniswap/v2-periphery/contracts/interfaces/IUniswapV2Router02.sol";
+import "@uniswap/v2-core/contracts/interfaces/IUniswapV2Pair.sol";
+import "@uniswap/lib/contracts/libraries/Babylonian.sol";
+
+contract UniswapStrategy {
+    using SafeMath for uint256;
+
+    address
+        private constant router = 0x7a250d5630B4cF539739dF2C5dAcb4c659F2488D;
+    address private pair;
+
+    function _get_liquidity_value() external view returns (uint256, uint256) {
+        /* (uint256 reserve0, uint256 reserve1, ) = IUniswapV2Pair(pair) */
+        /*     .getReserves(); */
+        /* uint256 pool_token = IERC20(pair).balanceOf(address(this)); */
+        /* uint256 all_pool_token = IERC20(pair).totalSupply(); */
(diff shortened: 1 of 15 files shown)


contracts/contracts/strategies/UniswapStrategy.sol
```diff
@@ -0,0 +1,256 @@
+pragma solidity 0.5.11;
+pragma experimental ABIEncoderV2;
+
+// can't just import router - it wants >=0.6.2;
+/* import "@uniswap/v2-periphery/contracts/interfaces/IUniswapV2Router02.sol"; */
```
tomlinton, line 5: We have an existing interface in interfaces/uniswap/, we can just add the additional methods to that and import it

contracts/contracts/vault/VaultCore.sol
```diff
@@ -253,8 +255,42 @@ contract VaultCore is VaultStorage {
...
+                ) {
+                    // since uniswap is the only one right now then the else is
+                    // okay - later in future refactor as switch/if/else like thing
+                    // does vault need to approve on strategy for transferFrom?
+                    PairReader r = PairReader(address(strategy));
+                    (address token0, address token1) = r._tokens(
+                        address(asset)
+                    );
+
+                    // figure out allocate amount right way
```
franckc, line 275: A few questions:
 1. How do we make sure there is enough funds for token1? We know that we have at least allocateAmount of token0 in the vault but that is not guaranteed for token1
 2. We may need to adjust the way we select the strategy. See relevant discussions in #335 (I think that is more of a discussion for the long-term) and also in PR #336 (short-term fix for strategy weight).
 3. token0 and token1 could use different precision for balances. For ex I think USDT and USDC use 6 digits while DAI uses 18

fxfactorial: Right - this spot needs more thought - I went back and forth with tom/daniel about it too. 
1. a. we pick the lesser of the two possiblities - compute how much of other side is needed - then deposit the lesser of the two 
2. The strategy weighting needs to be more liberal - perhaps price not in %s but in terms of eth since each coin easily priceable in eth. 
3. that's fine - its deterministic scaling todo ( left as todo if others greenlight things this general approach of pr )

tomlinton:
Closing this as we'll take a different approach to supporting multiple deposit stables.

PR #335, opened 2020-10-21
Rebase opt out for contracts

tomlinton:
Allows for a separate credits per token value for contracts which is frozen and so rebasing doesn't change their balance. Contracts can opt out and opt back in to rebasing.


DanielVF:
With this PR we can also remove the rebase hooks functionality in the vault contract (but we should keep the storage slot, to avoid breaking upgrades).

Or we can do that as a later upgrade, to make each upgrade smaller.

DanielVF (changes requested):
Currently there is a major problem with opting in after the credit ratios have changed. This results in the contract getting way too much OUSD. See video in discord.

I'm not sure right off how this is happening, since there's an exact test for this in the test cases.

DanielVF:
Whoooops! I was running this branch, but on code from the previous time around. That problem does not show anymore!

I'm going to continue to beat on this one for a few.

DanielVF:
I'm seeing rebaseOptIn rounding the account's dollars down to the nearest whole number.

tomlinton:
One other thing I realised is this will cause problems for existing contracts where the balance is partially transferred and a fixed creditsPerToken gets set. We need to migrate the entire balance.

DanielVF:
I've pushed a fix for the rounding down issue.

tomlinton:
Thanks @DanielVF!

contracts/contracts/token/OUSD.sol
```diff
@@ -48,6 +56,7 @@ contract OUSD is Initializable, InitializableToken, Governable {
 
         _totalSupply = 0;
         totalCredits = 0;
+        nonRebasingSupply = 0;
```
franckc, line 59: Note: initialize is not going to get called when we upgrade the OUSD contract on mainnet. But that's probably ok since by default in solidity variables are set to 0 I believe?

For consistency, should we set nonRebasingSupply to zero as well here?

contracts/contracts/token/OUSD.sol
```diff
@@ -85,8 +95,12 @@ contract OUSD is Initializable, InitializableToken, Governable {
      * @return A uint256 representing the _amount of base units owned by the
```
franckc, line 95: super nit: update the comment. both balanceOf and creditsBalanceOf method say "representing the _amount of base units"

contracts/contracts/token/OUSD.sol
```diff
@@ -115,18 +130,74 @@ contract OUSD is Initializable, InitializableToken, Governable {
...
+            nonRebasingCredits += creditsCredited - creditsDeducted;
+        }
+
+        // Make sure the fixed credits per token get set for to/from accounts if
+        // they have not been
+        if (isNonRebasingTo && nonRebasingCreditsPerToken[_to] == 0) {
+            nonRebasingCreditsPerToken[_to] = creditsPerToken;
+        }
+        if (isNonRebasingFrom && nonRebasingCreditsPerToken[_from] == 0) {
+            nonRebasingCreditsPerToken[_to] = creditsPerToken;
```
franckc, line 192: Shouldn't that be ```nonRebasingCreditsPerToken[_from]```?

contracts/contracts/token/OUSD.sol
```diff
@@ -214,17 +285,29 @@ contract OUSD is Initializable, InitializableToken, Governable {
...
+            nonRebasingCreditsPerToken[_account] = creditsPerToken;
+        }
+
+        uint256 creditAmount = _amount.mulTruncate(_creditsPerToken(_account));
         _creditBalances[_account] = _creditBalances[_account].add(creditAmount);
         totalCredits = totalCredits.add(creditAmount);
+        _totalSupply = _totalSupply.add(_amount);
+
+        if (isNonRebasingAccount) {
+            nonRebasingCredits = nonRebasingCredits.add(creditAmount);
```
franckc, line 302: Noob question. Why do we use .add() here while on line 171 for example we use += ?

contracts/contracts/token/OUSD.sol
```diff
@@ -244,36 +327,92 @@ contract OUSD is Initializable, InitializableToken, Governable {
...
         require(_account != address(0), "Burn from the zero address");
 
-        _totalSupply = _totalSupply.sub(_amount);
-        uint256 creditAmount = _removeCredits(_account, _amount);
+        bool isNonRebasingAccount = _isNonRebasingAddress(_account);
+
+        uint256 creditAmount = _amount.mulTruncate(_creditsPerToken(_account));
+        uint256 currentCredits = _creditBalances[_account];
+
+        // Remove the credits
```
franckc, line 335: nit: add comment to explain the currentCredit -1 trick. I guess it's for handling round error?

contracts/deploy/017_ousd_upgrade.js
```diff
@@ -0,0 +1,60 @@
...
+    from: deployerAddr,
+    ...(await getTxOpts()),
+  });
+  await ethers.provider.waitForTransaction(
+    dOusd.receipt.transactionHash,
+    NUM_CONFIRMATIONS
+  );
+  log("Deployed OUSD", dOusd);
+
+  // Update the proxy to use the new ousd contract.
```
franckc, line 41: This won't go thru on mainnet since the governor is multi-sig. It needs to be done via governance.
Perhaps add a ```if isRinkeby``` statement.

franckc (approved):
Looks great. A few comments inline, PTAL.

Otherwise LGTM

contracts/contracts/token/OUSD.sol
tomlinton: nonRebasingCredits you mean right? Will do 👍

contracts/contracts/token/OUSD.sol
tomlinton: I should be using .add/.sub etc everywhere. That uses the SafeMath library which will revert on overflows, rather than the unchecked + version.

PR #336, opened 2020-10-21
reduce strategy weight cost and also scale by weight

crazybuster:
This should handle the issue of low weights not firing because it scales the relative weights.

### DIFF
--- a/contracts/contracts/vault/VaultCore.sol
+++ b/contracts/contracts/vault/VaultCore.sol
@@ -265,10 +265,13 @@ contract VaultCore is VaultStorage {
         // Harvest for all reward tokens above reward liquidation threshold
         for (uint256 i = 0; i < allStrategies.length; i++) {
             IStrategy strategy = IStrategy(allStrategies[i]);
-            IERC20 rewardToken = IERC20(strategy.rewardTokenAddress());
-            uint256 rewardTokenAmount = rewardToken.balanceOf(address(this));
-            if (rewardTokenAmount > strategy.rewardLiquidationThreshold()) {
-                IVault(address(this)).harvest(allStrategies[i]);
+            address rewardTokenAddress = strategy.rewardTokenAddress();
+            if (rewardTokenAddress != address(0)) {
+                IERC20 rewardToken = IERC20(rewardTokenAddress);
+                uint256 rewardTokenAmount = rewardToken.balanceOf(address(this));
+                if (rewardTokenAmount > strategy.rewardLiquidationThreshold()) {
+                    IVault(address(this)).harvest(allStrategies[i]);
+                }
             }
         }
     }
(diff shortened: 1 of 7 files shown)


crazybuster:
Still need test and deployment, also one other issue is that we might to take into consideration the amount being added/deducted, since if it's a huge amount, then putting into the lower weight strategy might not make sense, and this can be done with just another add and parameter pass.

contracts/contracts/vault/VaultCore.sol
```diff
@@ -356,36 +358,41 @@ contract VaultCore is VaultStorage {
...
-                _totalValueInStrategy(_strategyAddr).divPrecisely(_totalValue())
-            );
+        // formula is (weight - valueInStrategy / totalValue)
+        // since we are comparing relative weights, we should scale by weight so
+        // that even small weights will be triggered, ie 1% versus 20%
+        // this becomes (weight - valueInStrategy / totalValue) / weight  -> (1 - valueInStrategy / (totalValue * weight))
+        // since we are comparing weight diffs in the same asset, the one and totalValue can fallout
+        // so we have (- valueInStrategy/ weight)
+        // using MAX_UINT - value to emulate the -
+        difference = MAX_UINT - (deposit ? _totalValueInStrategy(_strategyAddr).add(_modAmount) :
```
franckc, line 376: Makes sense.

One caveat though is that if we start having strategies that support multiple assets then this formula won't apply. 

For example imagine if we have Compound for DAI and USDT. And Aave for just DAI. Compound's strategy total value will be inclusive of both DAI and USDT.

Another example is when we add the new Uniswap strategy.

So I wonder if _strategyWeightDifference and _totalValueInStrategy should take an additional parameter that is the asset address.

crazybuster: We can certainly do that, is that something we want for this? the asset address is definitely there.

franckc: Before we make more changes, I think it would be worth having a discussion with the team about where we want to go for the allocate logic...

@tomlinton wrote a proposal in #339 for doing the allocation manually which could be an alternative.

crazybuster: So are we going with that? That feels like something further off in the future, while this I think will have a good immediate impact. Should reduce gas prices too.

DanielVF:
@crazybuster The effect of this change is that super low values for the "percent" allocated to the strategy no longer "cap" the amount of funds moved to the strategy? If so, can't we work around this for now by just using big numbers for our "percentages"? Or is there some other issue?

crazybuster:
> @crazybuster The effect of this change is that super low values for the "percent" allocated to the strategy no longer "cap" the amount of funds moved to the strategy? If so, can't we work around this for now by just using big numbers for our "percentages"? Or is there some other issue?

Yes, we can work around this(and we do currently) if our relative strategies have similar weight amounts.

franckc (approved):
Thanks for adding those tests. LGTM

@tomlinton @DanielVF ok if we merge this in and deploy this week?

Issue #334, opened 2020-10-21
Path to Duplicate Transactions (one failing)


micahalcorn:
I managed to reproduce the latest failing transaction that @DanielVF was investigating.

At least one way to get there is to do the following:

1. Submit a max transaction
2. Click the max link
3. See the warning message about the funds being in-flight
4. Ignore it and click the button anyway

<img width="581" alt="Screen Shot 2020-10-20 at 5 32 06 PM" src="https://user-images.githubusercontent.com/273937/96658818-60cb6b00-12fa-11eb-8bc2-65b73ce0f1bf.png">

I think a simple solution is just to disable that button when that warning appears. I don't think there would be any harm in doing that where a user might get stuck in a disabled state. We're already disabling the button when there literally aren't enough funds.

Refreshing the browser is _not_ required to reproduce this, and we're doing a good job of maintaining that pending funds state when a refresh does occur. There may be other scenarios that lead to duplicate transactions that I haven't noticed yet.

P.S. We need some padding in that warning.

sparrowDom:
Fixed via: https://github.com/OriginProtocol/origin-dollar/pull/337

PR #337, opened 2020-10-21
Handle buy warnings better

sparrowDom:
- disable buy button when user wants to mint with funds that are still in pending transactions
- add some more padding and margin to the warning box

**Connected issue:** https://github.com/OriginProtocol/origin-dollar/issues/334


sparrowDom:
@shahthepro merging this in, happy to address anything if needed

PR #331, opened 2020-10-19
Improvements/bug fix for removeStrategy

tomlinton:
Fixes https://github.com/OriginProtocol/origin-dollar/issues/324 and https://github.com/OriginProtocol/origin-dollar/issues/322.

PR #346, opened 2020-10-28
Migration + prop16 for deploy

franckc:
- New migration 18 for deploying Vault admin and core
 - Creation of prop16 for:
  - Add USDT+USDC to compound strategy
   - Vault admin, Vault core upgrades
   - call setRewardLiquidationThreshold on the Compound and 3Pool strategies (not needed on Aave since there is no reward token on that one)

### DIFF
--- a/contracts/deploy/018_vault_upgrade.js
+++ b/contracts/deploy/018_vault_upgrade.js
@@ -1,0 +1,133 @@
+const { isMainnet, isRinkeby } = require("../test/helpers.js");
+const { getTxOpts } = require("../utils/tx");
+const { proposeArgs } = require("../utils/governor");
+
+let totalDeployGasUsed = 0;
+
+// Wait for 3 blocks confirmation on Mainnet/Rinkeby.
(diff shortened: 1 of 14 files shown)


contracts/deploy/018_vault_upgrade.js
```diff
@@ -0,0 +1,133 @@
...
+    ...(await getTxOpts()),
+  });
+  await ethers.provider.waitForTransaction(
+    dVaultCore.receipt.transactionHash,
+    NUM_CONFIRMATIONS
+  );
+  log("Deployed VaultCore", dVaultCore);
+
+  // Deploy a new vault admin contract.
+  const dVaultAdmin = await deploy("VaultAdmin", {
```
tomlinton, line 46: Did VaultAdmin change? I don't think we need this and the later setAdminImpl call

franckc: This is to pick up the changes made in #331

PR #350, opened 2020-10-30
Fix autoharvest

tomlinton:


contracts/contracts/vault/VaultCore.sol
```diff
@@ -267,12 +267,22 @@ contract VaultCore is VaultStorage {
...
+                if (liquidationThreshold == 0) {
+                    // No threshold set, always harvest from strategy
                     IVault(address(this)).harvest(allStrategies[i]);
+                } else {
+                    // Check balance against liquidation threshold
+                    // Note some strategies don't hold the reward token balance
+                    // on their contract so the liquidation threshold should be
+                    // set to 0
+                    IERC20 rewardToken = IERC20(rewardTokenAddress);
+                    uint256 rewardTokenAmount = rewardToken.balanceOf(
```
micahalcorn, line 280: Not specific to this PR, but do I understand correctly that the threshold must be met at the strategy level regardless of aggregate token balance? In other words, if we had a CompoundDAI strategy and a CompoundUSDT strategy with each holding 0.75 COMP, neither would get liquidated. And then if one of them is 1.25, it will get liquidated while the other won't if its balance is still below the threshold?

tomlinton: Yep that is correct!

tomlinton:
Verified this one on fork so should be good to go 👍

contracts/scripts/utils/harvest.js
```diff
@@ -0,0 +1,84 @@
+// Script for calling harvest on the vault.
```
franckc, line 1: nit: to avoid any possible confusion, maybe add a note that this can not be used on Mainnet (since harvest is onlyGovernor which is controlled by the multi-sig). But I assume this is useful for testing on the fork?

franckc (approved):
Thank you for those fixes. LGTM!

PR #353, opened 2020-11-02
Tomlinton/rebasestate tweak

tomlinton:
Adds more tests to rebaseOptIn/rebaseOptOut PR.

`rebaseOptInList` has been replaced with `rebsaseState` because I needed to be able to represent not set/opt in/opt out which a bool couldn't do because of no null defaults.

### DIFF
--- a/contracts/contracts/token/OUSD.sol
+++ b/contracts/contracts/token/OUSD.sol
@@ -362,7 +362,7 @@ contract OUSD is Initializable, InitializableToken, Governable {
             nonRebasingCredits = nonRebasingCredits.sub(creditAmount);
             nonRebasingSupply = nonRebasingSupply.sub(_amount);
         } else {
-            rebasingCredits.sub(creditAmount);
+            rebasingCredits = rebasingCredits.sub(creditAmount);
         }
 
         emit Transfer(_account, address(0), _amount);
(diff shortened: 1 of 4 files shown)


contracts/contracts/token/OUSD.sol
```diff
@@ -45,7 +45,11 @@ contract OUSD is Initializable, InitializableToken, Governable {
     uint256 public nonRebasingCredits;
     uint256 public nonRebasingSupply;
     mapping(address => uint256) public nonRebasingCreditsPerToken;
-    mapping(address => bool) public rebaseOptInList;
+    // Uint8 represents opt in state
+    // -1 explicit opt out
+    // 0 (default) not set (contracts by default opt on, EOAs by default opt in)
```
franckc, line 50: nit: did you mean contract by default opt out (vs "on")?

franckc (approved):
One minor comment inline otherwise LGTM

contracts/contracts/token/OUSD.sol
DanielVF: I've updated this.

DanielVF (changes requested):
Love it! This is the code I was feeling like we were missing.

One question though is why not use an ENUM for this mapping? Seems like the natural way to do it, and would be clearer in the code.

If not an enum, how about a unit8, rather than a int8?

tomlinton:
Yea an ENUM makes more sense, will convert it. I went with int8 because its easier to parse `-1` as opting out rather than having all positive values.

tomlinton:
Merging!

PR #357, opened 2020-11-07
Gas Limits with Estimates

micahalcorn:
This moves away from hardcoded default gas limits for most transactions in favor of using `estimateGas` + one of a few different buffer implementations.

**Mints**
Most gas limits for `mint` and `mintMultiple` transactions are padded by the greater of `20000` or 10%. In cases where the mint amount (or combined amounts) is within 4% of the thresholds, a hardcoded amount is added to accommodate the possibility of `rebase` or `autoAllocate` when the estimate did not account for this. These cases should be rare. And, at least currently, this addition results in a higher gas limit than we would otherwise have if the estimate had been calculated with an amount that exceeded the threshold.

**Redeems**
Since we're less concerned about gas conservation when OUSD is burned, a higher buffer of 25% is used for `redeem` and `redeemAll` transactions. Similarly to the mint scenarios, rare cases of the amount approaching the `rebase` threshold are handled by increasing the gas limit, this time using a multiplier instead of adding a specific amount.

I have gone ahead and deployed this branch to the production environment since we're seeing a lot of users run out of gas with the hardcoded amounts. I've also tested the calculations for natural gas limit, near-threshold increases, and buffers for all predictable scenarios and [recorded a subset of the most common transactions to compare actual gas usage to the resulting defaults](https://docs.google.com/spreadsheets/d/1A512Y0SmpI7lHKbsIi_poOWrGp_wnSnxwuze3lOlkt0/edit#gid=1264639928).

Areas that have not been thoroughly tested are real-world transactions above the `autoAllocate` threshold, and those that involve liquidating from strategies due to an insufficient buffer.

### DIFF
--- a/dapp/src/components/buySell/BuySellWidget.js
+++ b/dapp/src/components/buySell/BuySellWidget.js
@@ -13,7 +13,7 @@ import AddOUSDModal from 'components/buySell/AddOUSDModal'
 import ErrorModal from 'components/buySell/ErrorModal'
 import DisclaimerTooltip from 'components/buySell/DisclaimerTooltip'
 import ApproveCurrencyInProgressModal from 'components/buySell/ApproveCurrencyInProgressModal'
-import { currencies, gasLimits } from 'constants/Contract'
+import { currencies } from 'constants/Contract'
 import { formatCurrency } from 'utils/math'
 import { sleep } from 'utils/utils'
 import { providersNotAutoDetectingOUSD, providerName } from 'utils/web3'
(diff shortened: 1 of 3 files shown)


micahalcorn:
No out of gas errors in the past 48 hours 💪

dapp/src/components/buySell/BuySellWidget.js
```diff
@@ -237,47 +237,68 @@ const BuySellWidget = ({
...
+      const absoluteGasLimitBuffer = 20000
+      const percentGasLimitBuffer = 0.1
+      const rebaseGasBuffer = 600000
+      const allocateGasBuffer = 3000000
+
+      let gasEstimate, gasLimit, result
       mobileMetaMaskHack(prependStage)
       if (mintAddresses.length === 1) {
+        gasEstimate = (
+          await vaultContract.estimateGas.mint(mintAddresses[0], mintAmounts[0])
```
sparrowDom, line 262: If I am getting this correctly the `vaultContract.estimateGas.mint` already returns higher gas estimate when the `mintAmount` is over the threshold. The code here makes sure that if the mint `mintAmount` is 4% below that threshold the code manually adds additional gas buffer. Since when the `mintAmount` is just above any of the thresholds this code adds no additional gas buffer. 

If that is true the only thing I don't like is that part of the threshold logic is controlled by the contract and the other part by the JS code.

micahalcorn: @sparrowDom if I've done what I intended, there isn't really much difference in the way this works now vs before the change. Which part of the threshold logic do you not like being controlled by the contract or by the JS?

The problem that I see is that gas limit is necessarily set by the client but gas usage is determined by the contract (via the miner). And since the amounts are calculated based on oracles at different times (when the transaction is constructed vs when it is processed), there will always be the possibility of different outcomes. The code here tries to anticipate this possibility and adds the buffer if there's a reasonable risk of the gas limit being underestimated.

sparrowDom: Let me illustrate on an example the part that confuses me. For the purpose of the argument will pick whole numbers. Let say: 
```
const allocateThreshold = 1000
const rebaseThreshold = 2000
const nearAllocateThreshold = 960 < mintAmount < 1000
const nearRebaseThreshold = 1920 < mintAmount < 2000
```

Won't the allocateGasBuffer be applied only when `mintAmount` is over 960 and below 1000. Shouldn't it apply when `mintAmount` is over 960 and below 1920?

micahalcorn: I don't think so. Let me restate your illustration using real numbers to avoid confusion:

```js
const allocateThreshold = 25000
const rebaseThreshold = 1000
const nearAllocateThreshold = 24000 <= mintAmount < 25000
const nearRebaseThreshold = 960 <= mintAmount <= 1000
```

For any `mintAmount` that is > 1000 and < 24000, we will use the estimated gas, which includes the rebase in its calculation. If the amount were >= 25000, then the estimate would also include the allocate in its calculation. But for amounts >= 24000 and < 25000, they need the `allocateGasBuffer` because the estimation only included the rebase but not the allocate in the calculation.

micahalcorn: You can verify this on [ousd.com](https://ousd.com) right now by comparing the default gas limits when minting 961 vs 1001 and 24001 vs 25001. The gas limits on 961 and 24001 will be slightly higher than 1001 and 25001 respectively.

sparrowDom: Aha thanks I think I get it now. I see 2 problems with the implementation: 

1. In the previous code the 4% threshold that was only present on the left side of the threshold number (960 - 1000 -> instead of also the right side 1000-1040) functioned because the DAPP was completely in control of the gas amounts. We never relied on the contract's `eastimatedGas` calls. That is why when amount was over the threshold the dapp was the one to always set higher gas amounts. The solution here should cause 2 undesired paths: 
    - **Oracles show stable coins above peg.** Minting using 999 USDT will mint 1001 OUSD and `estimated gas` will already include `allocate threshold` gas limit. The dapp will add additional `3000000` gas limit. And the user will overpay for the transaction
    - **Oracles show stable coins below peg when calling estimate. And above peg when executing the transaction.** Minting using 1001 USDT will mint 999 OUSD in the phase of calling `estimated gas` and will not include the `allocate threshold` gas limit. Because the `nearAllocateThreshold` only looks on the left side of the threshold number the dapp will also not add the additional gas limit. When executing the transaction (and the stablecoin pegs change) the allocate gets triggered on the contract but not `estimatedGas` nor the dapp were able to detect that and increase the gas limit. The transaction should fail. 

2. Because the Dapp was previously oblivious to oracle values when setting gas limits (I think that was the reason that) @DanielVF picked the 4% buffer margin. As in that is how much a stable coin can go out of $1 peg. In this PR we ask the contract for the estimated gas, and between that estimation and transaction submission the stable coin values shouldn't change by such a huge margin as 4%.

I propose another solution that shouldn't suffer from the above problems: 
when calling the estimated gas: `vaultContract.estimateGas.mint(mintAddresses[0], mintAmounts[0])` lets just increase all the stablecoin values by 1-2%.
- This way there should be no double spending (or no spending) of the gas. 
- That estimation will already include the oracle values. So the buffer margins that are relative to current oracle values can be lower instead of absolute margin values to $1 peg
- Should be a less complex solution

Sorry for the long winded explanation. Now given my limited contract knowledge, I might be making errors in the above logic. For that reason any input from @DanielVF or @tomlinton would be very appreciated.

micahalcorn: Great points, @sparrowDom I think I glossed over the fact that `mintAmount(s)` is tokens while the actual threshold determination considers OUSD. I also agree that the 4% buffer isn't ideal.

I don't think that we can increase the stablecoin values when estimating though because the estimate transaction will fail if the increased amount is greater than the user's balance. There may be a way around that, but I'm not aware of how to do it.

Would this be a lot simpler if we modified the contract to rebase or allocate based on input values rather than output values? If the thresholds on the contract side just compared stablecoin values on mint and OUSD values on redeem, then we wouldn't need to worry about a buffer. The thresholds are arbitrary numbers anyway, so I can't think of any risks that we would need to worry about.

micahalcorn: Thinking more about that, we can already remove the threshold logic from the SellWidget because `redeemAmount` itself is used for comparisons on the contract side.

For mints, I'm not sure that there's any reason to use [`priceAdjustedDeposit`](https://github.com/OriginProtocol/origin-dollar/blob/master/contracts/contracts/vault/VaultCore.sol#L61) and [`priceAdjustedTotal`](https://github.com/OriginProtocol/origin-dollar/blob/master/contracts/contracts/vault/VaultCore.sol#L112) when we could just as easily use the unadjusted values. What am I missing there @DanielVF @tomlinton?

sparrowDom: Hmm good point about why increasing the stablecoin values wouldn't work. 

I agree if we can change the contract to consider stablecoin values pegged to $1 instead of OUSD amount minted  when minting it would simplify things greatly.

micahalcorn: @tomlinton @DanielVF I'm proposing something like this: https://github.com/OriginProtocol/origin-dollar/commit/38355662be375e942746e10681e6672f54c451a8

tomlinton: I think #361 one is a good idea and we should simplify this whole frontend piece! Nice job guys.

micahalcorn:
I think I'll close this novel and start fresh since we've arrived at #361. 😅

PR #358, opened 2020-11-09
Tomlinton/hardhat migration

tomlinton:
This was a bit painful. I may have missed things but I can at least pass all tests/run the fork/run the dapp on the fork. I say we merge it and can clean up any other little bits as they pop up.

I removed quite a few scripts I think we no longer need, and replaced the PREMIUM_GAS with GAS_MULTIPLIER which we can set for the entire network. Also merged all the migrations into one.


contracts/contracts/interfaces/IVault.sol
```diff
@@ -98,6 +98,8 @@ interface IVault {
 
     function rebase() external returns (uint256);
 
+    function totalValue() external view returns (uint256 value);
```
franckc, line 101: did you intend to make this change in the interface?

contracts/hardhat.config.js
```diff
@@ -1,12 +1,12 @@
...
-usePlugin("buidler-ethers-v5");
-usePlugin("solidity-coverage");
-usePlugin("buidler-gas-reporter");
-usePlugin("buidler-contract-sizer");
+require("@nomiclabs/hardhat-waffle");
+require("@nomiclabs/hardhat-solhint");
+require("hardhat-deploy");
+require("hardhat-contract-sizer");
+require("hardhat-deploy-ethers");
+// require("solidity-coverage");
```
franckc: should we remove them if these imports are not used?

dapp/src/utils/web3.js
```diff
@@ -17,7 +16,7 @@ export function isCorrectNetwork(chainId) {
   if (process.env.NODE_ENV === 'production') {
     return chainId === 1
   } else if (process.env.NODE_ENV === 'development') {
-    return chainId === 1337 || chainId === 31337
+    return chainId === 1 || chainId === 31337
```
franckc, line 19: is that right?

franckc (approved):
Awesome. Thanks for doing this migration. Looks great to me.

A few super minor comments inline.

If the unit tests pass then let's merge this in asap.

dapp/src/utils/web3.js
tomlinton: Yea, due to limitations in hardhat I can't specify a different chain id for the fork anymore.

contracts/contracts/interfaces/IVault.sol
tomlinton: I was using it for testing, but then figured I might as well leave it in.

PR #359, opened 2020-11-10
Add validation on values passed to setVaultBuffer

tomlinton:
- Add a require to make sure vaultBuffer is in the correct range
- Also change some subtractions to use SafeMath


franckc (approved):
LGTM

PR #363, opened 2020-11-12
Reallocate method

tomlinton:
- Implementation of the Strategist address and the methods to set it
- Add a reallocate method to move funds from one Strategy to another callable by Governor or Strategist
- Tests and deploy script
- Couple of prettier/misc cleanup fixes

### DIFF
--- a/contracts/contracts/vault/VaultAdmin.sol
+++ b/contracts/contracts/vault/VaultAdmin.sol
@@ -22,6 +22,14 @@ contract VaultAdmin is VaultStorage {
         _;
     }
 
+    modifier onlyGovernorOrStrategist() {
+        require(
+            msg.sender == strategistAddr || isGovernor(),
+            "Caller is not the Strategist or Governor"
+        );
+        _;
+    }
+
     /***************************************
                  Configuration
     ****************************************/
@@ -88,6 +96,14 @@ contract VaultAdmin is VaultStorage {
      */
     function setUniswapAddr(address _address) external onlyGovernor {
         uniswapAddr = _address;
+    }
+
+    /**
+     * @dev Set address of Strategist
+     * @param _address Address of Strategist
+     */
+    function setStrategistAddr(address _address) external onlyGovernor {
+        strategistAddr = _address;
     }
 
     /**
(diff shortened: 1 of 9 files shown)


contracts/contracts/vault/VaultAdmin.sol
```diff
@@ -186,6 +202,41 @@ contract VaultAdmin is VaultStorage {
...
+            strategies[_strategyToAddress].isSupported,
+            "Invalid to Strategy"
+        );
+        require(_assets.length == _amounts.length, "Parameter length mismatch");
+
+        IStrategy strategyFrom = IStrategy(_strategyFromAddress);
+        IStrategy strategyTo = IStrategy(_strategyToAddress);
+
+        for (uint256 i = 0; i < _assets.length; i++) {
+            require(strategyTo.supportsAsset(_assets[i]), "Asset unsupported");
```
franckc, line 232: Should there also be a require for strategyFrom.supportsAsset(_asset[i]) ?

And is it worth having a require on strategyFrom.checkBalance(_assets[i]) <= _amounts[i] ?

contracts/hardhat.config.js
```diff
@@ -573,6 +573,90 @@ task("balance", "Get OUSD balance of an account")
     console.log("OUSD credits=", formatUnits(credits.toString(), 18));
   });
 
+task("reallocate", "Allocate assets from one Strategy to another")
```
franckc, line 576: This is super useful for testing on fork! Nice :)

franckc (approved):
Looks great. I really like the new hardhat possibilities.

One minor question inline regarding the contract change otherwise LGTM

contracts/contracts/vault/VaultAdmin.sol
tomlinton: Both of those should revert so it'll save a bit of gas not to check them here. I think the to strategy check is enough 👍

franckc: Sounds good. Thanks for checking. Let's get this in, deploy and vote for who is going to be the strategist:)
Lmk if you want me to help with the deploy stuff.

PR #367, opened 2020-11-13
Simplified Gas Limits with Estimates

micahalcorn:
This moves away from hardcoded default gas limits in favor of using `estimateGas` + a buffer.

**Mints**
`mint` and `mintMultiple` transactions are padded by the greater of `20000` or 10%.

**Redeems**
`redeem` and `redeemAll` transactions are padded with a higher buffer of 25% since we're less concerned about gas conservation when tokens are going to be burned.

There is still one possible extreme edge case that will knowingly result in an "out of gas" error. If a user were to call `redeemAll` with a balance slightly under the `rebaseThreshold` (or if rebase is paused) and end up with >= 1,000 OUSD as a result of a rebase that happens before the transaction is mined, the estimated gas will be far too low. This could be avoided by removing [the conditions](https://github.com/OriginProtocol/origin-dollar/blob/master/contracts/contracts/vault/VaultCore.sol#L196) and rebasing prior to every redeem.

I've tested a handful of Mainnet transactions with this branch on staging, so I'll go ahead and merge+deploy this instead of waiting for comments. 🤞

### DIFF
--- a/dapp/src/components/buySell/BuySellWidget.js
+++ b/dapp/src/components/buySell/BuySellWidget.js
@@ -13,7 +13,7 @@ import AddOUSDModal from 'components/buySell/AddOUSDModal'
 import ErrorModal from 'components/buySell/ErrorModal'
 import DisclaimerTooltip from 'components/buySell/DisclaimerTooltip'
 import ApproveCurrencyInProgressModal from 'components/buySell/ApproveCurrencyInProgressModal'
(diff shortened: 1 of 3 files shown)


sparrowDom:
Nice awesome work. I love how this solution simplifies the code and is a lot more stable.

micahalcorn:
Thanks for helping me find my way! 😅

PR #368, opened 2020-11-15
Single default strategy per stablecoin

tomlinton:
Remove strategy weightings in favour of a single strategy for automatic allocation/withdraws per stablecoin.


contracts/contracts/vault/VaultAdmin.sol
```diff
@@ -103,6 +103,24 @@ contract VaultAdmin is VaultStorage {
...
+     * @dev Set the default Strategy for an asset, i.e. the one which the asset
+            will be automatically allocated to and withdrawn from
+     * @param _asset Address of the asset
+     * @param _strategy Address of the Strategy
+     */
+    function setAssetDefaultStrategy(address _asset, address _strategy)
+        external
+        onlyGovernorOrStrategist
+    {
+        emit AssetDefaultStrategyUpdated(_asset, _strategy);
```
shahthepro, line 116: nit: Why is the event emitted at the start of the function just in this place?

contracts/contracts/vault/VaultAdmin.sol
```diff
@@ -144,7 +153,7 @@ contract VaultAdmin is VaultStorage {
      */
 
     function removeStrategy(address _addr) external onlyGovernor {
```
shahthepro, line 155: Can we have a similar function that takes index as an argument instead of address? Could be helpful in saving gas sometimes.

shahthepro (approved):
LGTM

contracts/contracts/vault/VaultAdmin.sol
tomlinton: It removes slither warnings about reentrant events, some of the other functions required external calls before the event arguments were available so in those cases I generated a db for slither of ignored warnings in #386

PR #372, opened 2020-11-19
Hardening contracts

shahthepro:
- [x] Allows pending/queued transactions to be cancelled by governor (`TOB-OUSD-006`)
- [x] Allow only admin to invoke `Timelock.executeTransaction` (`TOB-OUSD-008`)
- ~~[x] Disallow `setPendingAdmin` tx in proposals (`TOB-OUSD-009`)~~ #385 
- ~~[ ] Switch `InitializableGovernable` to `Governable` in Oracle contracts~~ https://github.com/OriginProtocol/origin-dollar/pull/383


contracts/contracts/governance/Governor.sol
```diff
@@ -84,6 +99,13 @@ contract Governor {
             "Governor::propose: too many actions"
         );
 
+        for (uint256 i = 0; i < signatures.length; i++) {
+            require(
+                keccak256(bytes(signatures[i])) != setPendingAdminSign,
+                "Governor::propose: setPendingAdmin transaction cannot be propsed or queued"
```
tomlinton, line 105: Typo on proposed

contracts/contracts/governance/Governor.sol
```diff
@@ -189,6 +207,28 @@ contract Governor {
...
+    function cancel(uint256 proposalId) public onlyGuardian {
+        ProposalState proposalState = state(proposalId);
+
+        require(
+            proposalState == ProposalState.Queued ||
+                proposalState == ProposalState.Pending,
+            "Governor::execute: proposal can only be cancelled if it is queued or pending"
+        );
+        Proposal storage proposal = proposals[proposalId];
+        proposal.eta = 1; // To mark the proposal as `Expired`
```
tomlinton, line 219: Do you think we need a separate cancel state? I think expired is probably sufficient

tomlinton:
This looks good @shahthepro! Bit of a pain but would you mind separating them out into separate PRs (one TOB finding per PR). They requested we do that to make reviewing easier on their end.

shahthepro:
Sure. Will do

contracts/contracts/governance/Governor.sol
shahthepro: Will fix it. The `o` key in my MBP doesn't work always 😅

contracts/contracts/governance/Governor.sol
shahthepro: I don't see a need for it now. But can add it, if it's necessary

tomlinton: Nah I don't think it is, was thinking out loud 👍

shahthepro:
Merging this in, since it is a blocker for other things on the list

PR #375, opened 2020-11-20
Remove accounting for nonRebasingCredits (TOB-OUSD-003)

tomlinton:
TOB raised a question about why we need to record this early on in the audit and I don't believe we need it so I've removed it as a simplification. It doesn't make sense to record it as the nonRebasingCredits may have different values for nonRebasingCreditsPerToken for different accounts, so it can't really be used to calculate anything useful (i.e. you can't get back to nonRebasingSupply using these values).

Do you see any issues with this @DanielVF analytics wise?


DanielVF:
I don't need it in any way.

DanielVF (approved):
That's amazing how much code that removes.

PR #380, opened 2020-11-22
Reentrancy fix (TOB-OUSD-10)

tomlinton:
💀


shahthepro (approved):
Great work, looks good to me. One minor nit: Can we add the `nonReentrant` modifier to `rebase()` method too?

tomlinton:
Done 👍

Made the public versions nonReentrant and switched all contract calls to the internal functions.

PR #384, opened 2020-11-23
Remove OpenUniswapOracle (TOB-OUSD-018, TOB-OUSD-021)

tomlinton:
We aren't going to use this, and it needs fixes, so I'm removing it.


tomlinton:
Will merge this without review as its fairly low risk due to just being code removal.

PR #386, opened 2020-11-23
Add Slither GitHub action

tomlinton:


franckc (approved):
Awesome. LGTM

Looks like we'll need to re-deploy a bunch of contracts:
 - all strategies
 - vaultAdmin, vaultCore

Let's take a note of that for when we figure out our Mainnet OUSD v2 deploy strategy...

PR #387, opened 2020-11-24
[TOB-OUSD-019] Check return values of external contract calls

shahthepro:
Note: Have skipped the check in functions that are either deprecated or don't return a boolean/binary value.


contracts/contracts/utils/InitializableAbstractStrategy.sol
```diff
@@ -191,7 +191,10 @@ contract InitializableAbstractStrategy is Initializable, Governable {
         public
         onlyGovernor
     {
-        IERC20(_asset).transfer(governor(), _amount);
+        require(
+            IERC20(_asset).transfer(governor(), _amount),
```
tomlinton, line 195: Need to merge https://github.com/OriginProtocol/origin-dollar/pull/378 before this as it no longer applies

contracts/contracts/vault/VaultAdmin.sol
```diff
@@ -288,7 +288,10 @@ contract VaultAdmin is VaultStorage {
         external
         onlyGovernor
     {
-        IERC20(_asset).transfer(governor(), _amount);
+        require(
```
tomlinton, line 291: Same as above

contracts/contracts/utils/InitializableAbstractStrategy.sol
shahthepro: Have merged master with this branch and reverted these back.

PR #388, opened 2020-11-24
[TOB-OUSD-020] Add a method to remove elements from array

shahthepro:
Note: Skipped adding similar method to `MixOracle` since it already has method to override them


contracts/contracts/utils/RebaseHooks.sol
```diff
@@ -14,6 +14,15 @@ contract RebaseHooks is Governable {
         uniswapPairs = _uniswapPairs;
     }
 
+    // Removes uniswap pair at the given index
```
tomlinton, line 17: This contract is being removed in #381

contracts/contracts/utils/InitializableAbstractStrategy.sol
```diff
@@ -159,6 +160,25 @@ contract InitializableAbstractStrategy is Initializable, Governable {
...
+     */
+    function removePToken(uint256 _assetIndex) external onlyGovernor {
+        require(_assetIndex < assetsMapped.length, "Invalid index");
+        address asset = assetsMapped[_assetIndex];
+        address pToken = assetToPToken[asset];
+
+        if (_assetIndex < assetsMapped.length - 1) {
+            assetsMapped[_assetIndex] = assetsMapped[assetsMapped.length - 1];
+        }
+        delete assetsMapped[assetsMapped.length - 1];
```
tomlinton, line 176: This won't decrease the array length I believe, also see #374 and TOB-OUSD-012

shahthepro: Have changed it to use `.pop()` method

tomlinton (approved):

PR #390, opened 2020-11-25
Add a minimum stablecoin unit argument to redeem functions (TOB-OUSD-14)

tomlinton:
Add an argument to redeem and redeemAll to allow setting of a minimum amount of stablecoins (in unit terms, not price wise).

@sparrowDom I think I caught all the spots in the dapp where this needs to be added to function calls. I just set it to 0 (i.e. don't care how many) but we may want to expose this in the UI later.


sparrowDom (approved):
Yup you got them all, LGTM

PR #396, opened 2020-12-01
Add slippage on 3pool deposit/withdraw

tomlinton:
We plan not to let anyone interact with this except the strategist due to #368 but the additional protection doesn't hurt.

We may need to make this configurable?

Not sure whats up with the slither failure, I'll try figure that out.


franckc (approved):
Changes LGTM

It does seem that Slither has a new warning though:
INFO:Detectors:
StableMath.scaleBy(uint256,int8) (contracts/utils/StableMath.sol#25-36) performs a multiplication on the result of a division:
	-x = x.mul(10 ** uint256(adjustment)) (contracts/utils/StableMath.sol#31)
	-x = x.div(10 ** uint256(adjustment * - 1)) (contracts/utils/StableMath.sol#33)
Reference: https://github.com/crytic/slither/wiki/Detector-Documentation#divide-before-multiply

But that does not look to me like a multiplication on the result of a division... Did I miss something or is Slither confused?

tomlinton:
Yea it's being a pain, I also can't repro it locally, I'll try resolve it before I merge though

tomlinton:
Got a clean slither run, there are tests failing on master that are staking related so I'll merge this.

PR #402, opened 2020-12-03
Remove IViewVault and IViewMinMaxOracle interfaces

### LINKED ISSUE #398: Make oracle methods view only
Now that the Uniswap oracle is gone and we don't need to perform cumulative window updates on oracles we can change all the oracle methods (`priceUSDMint`, `priceUSDRedeem`, etc) to views, which means `totalValue` can also be a view. We can then remove IViewVault in favour of just the IVault.

shahthepro:
Closes #398 

The contracts work fine. Have to do some more testing of the DApp though.


tomlinton:
Looks good, thanks @shahthepro!

tomlinton (approved):

sparrowDom (approved):
@shahthepro if the `fetchExchangeRates` still fetches the data when the wallet is not connected to the dapp, then it is all good.

shahthepro:
>  if the `fetchExchangeRates` still fetches the data when the wallet is not connected to the dapp, then it is all good.

@sparrowDom For the DApp, I only changed the references of `viewVault` to `vault`, so if it was there already, it should be working fine.

PR #404, opened 2020-12-04
New Merkle chain method of doing preApproved drops

crazybuster:


franckc (approved):
This is really super neat. Great work.

Code looks good to me.

The only thing I'd recommend is having a test with a larger data set (hundreds of leaves?), to double check that it works on a tree that has more depth.

Also a note that on the product side, we should be aware that once the root hash is set, the payout dataset is immutable: no individual payout can be added/edited/deleted. I think it is perfectly fine but just want to call this out since the previous implementation that based on signing data enabled to add new claims.

joshfraser:
Elegant solution. I love it.

Money might not grow on trees, but it's going to grow great on this merkle tree!

franckc (approved):
Changes to add a check on the tree depth LGTM.
Thanks for adding that extra check to make it even more secure!

Unrelated: I wonder if it would be handy to add a read method "isValidClaim" that returns if a claim is valid or not? It would be handy for testing on Mainnet and for potentially fielding inquiries from users if they complain their claim does not work...

DanielVF (approved):
I've reviewed the contract changes and the airdrop script.

**Contract**

With additional changes already made, the contract looks good.

After reviewing lots of literature, I don't think SHA3/keccak256 is vulnerable to second preimage attacks. Also, because we have the index and the depth in the hash, this makes it so that the tree can't be truncated or changed. The biggest block to attacks is that the account public key is in the hashed data. Even if you could almost instantly generate a valid set of data that matched a hash, the matching private key to that random public key would be impossible to recover.

**Airdrop script**

Minor: I see that it's recalculating the entire tree for every entry. This makes it run at O(N2), rather than O(N). We only have to run it once, and with less than a thousand users, so it would probably work okay.

crazybuster:
> Changes to add a check on the tree depth LGTM.
> Thanks for adding that extra check to make it even more secure!
> 
> Unrelated: I wonder if it we should add a read method "isValidClaim" that returns if a claim is valid or not? It would be handy for testing on Mainnet and for potentially fielding inquiries from users if they complain their claim does not work...

A function to check the claim can be done on local js, I really am not a fan of putting logic onChain that can be done off chain, feels like a slight waste. But I guess it can be useful.

franckc:
@crazybuster yeah, we can verify a claim on local... I was thinking it could be handy to have a function in the contract so that we can easily verify on etherscan, but your call. not a blocker to merge this PR to master.

PR #405, opened 2020-12-04
Nkg/reduce alchemy calls2

sparrowDom:
All done by @nickgeoca: 
Instead of continuously fetching the balance updates, the dapp logic now listens for transfer / allowance events and fetches data only based on those events.


dapp/src/components/AccountListener.js
```diff
@@ -17,98 +18,136 @@ const AccountListener = (props) => {
...
+                  s.balances[name] = balance
+                })
+              )
+          )
+      )
+      // Account receives tokens
+      rpcProvider.on(
+        contract.filters.Transfer(null, account, null), // event Transfer(address indexed from, address indexed to, uint tokens);
+        (result) =>
+          pollNTimes(5, () =>
```
nickgeoca, line 124: `pollNTimes`

For balance updates, we are going off of the Transfer event. This event doesn't tell us the balance, only the amount transferred. So once the code gets the Transfer event, it then polls for the balance. There is a latency when metamask has the updated balance, so the code uses a kludge: using `pollNTimes` to poll metamask several times for an update.

nickgeoca:
@sparrowDom I did some merge work. It seems to work well.

sparrowDom (commented):
@nickgeoca great work on this PR. Sorry for taking so long to do a code review. We were just launching a feature and was occupied with that.

The biggest reason for needing to `poolNtimes` is that "normal" wallets use the Metamask provider when connected with Metamask. And the jsonRpc provider is from Alchemy. And the code was listening for Transfer events using the Alchemy provider and then querying Metamask's provider for actuall balance. Those 2 are not always in complete sync on any one of those can lag bahind.

I've added another version of contracts that uses the `jsonRpc` provider to query data. And now `jsonRpc` is used to listen for events and also query data. The pooling multiple times should not be needed anymore and you can remove it.

Left a couple more comments below

dapp/src/components/AccountListener.js
```diff
@@ -53,6 +54,125 @@ const AccountListener = (props) => {
     }
   }, [active, prevActive, account, prevAccount])
 
+  const pollOnce = (contracts) => {
+    const { usdt, dai, usdc, ousd, vault, ogn } = contracts
+
+    Promise.all([
+      // balance
+      ousd.balanceOf(account).then((b) => displayCurrency(b, ousd)),
```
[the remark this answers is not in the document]
nickgeoca: Thanks @sparrowDom. 

Sure thing, I'll clean it up here shortly.

dapp/pages/dashboard.js
```diff
@@ -98,7 +97,7 @@ const Dashboard = ({ locale, onLocale }) => {
 
   const approveUSDT = async () => {
     notSupportedOption()
-    await usdt.approve(
+    const result = await usdt.approve(
```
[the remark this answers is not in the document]
nickgeoca: I think that was for console.logging. I just removed them 👍

dapp/src/components/AccountListener.js
```diff
@@ -53,6 +54,125 @@ const AccountListener = (props) => {
...
+  }
+
+  const subscribeToEvents = (contracts) => {
+    // Polls data first time, then rely on events
+    const { usdtRpc, daiRpc, usdcRpc, ousdRpc, vault, ognRpc } = contracts
+    const rpcProvider = props.rpcProvider
+    const pollNTimes = (n, promiseFn) => {
+      if (n === 0) return
+      // Poll every 5 seconds
+      setTimeout(() => {
```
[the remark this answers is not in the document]
nickgeoca: Alchemy is used to fetch data- gotcha. I removed pollNTimes and it is working fine. 

@sparrowDom , would you like me to resolve the conversations?

dapp/src/components/AccountListener.js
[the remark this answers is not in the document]
nickgeoca: This is fixed now.

dapp/src/components/AccountListener.js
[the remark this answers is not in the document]
sparrowDom: We usually just keep the conversations, but if you want you can also resolve them.

dapp/src/components/AccountListener.js
```diff
@@ -410,15 +438,33 @@ const AccountListener = (props) => {
 
   useEffect(() => {
     let balancesInterval
-    if (contracts && userActive === 'active' && isCorrectNetwork(chainId)) {
+    if (
+      account &&
+      contracts &&
+      contracts.ogn.provider &&
```
sparrowDom, line 444: @nickgeoca just curious was there a case where the contracts have initialised and the `contracts.ogn.provider` was undefined or null?

dapp/src/components/AccountListener.js
```diff
@@ -53,6 +54,102 @@ const AccountListener = (props) => {
     }
   }, [active, prevActive, account, prevAccount])
 
+  const pollOnce = async (contracts) => {
```
sparrowDom, line 57: should the name of this function be `pullOnce` instead of `pollOnce`?

dapp/src/components/AccountListener.js
nickgeoca: I saw it in the [loadData](https://github.com/OriginProtocol/origin-dollar/blob/nkg/reduceAlchemyCalls2/dapp/src/components/AccountListener.js#L153-L157) fn way back and kept it in for consistency.

nickgeoca: They're both using the provider to grab data

dapp/src/components/AccountListener.js
nickgeoca: yeah good point. polling is a more recurrent process.. i'll update it. 

Right now we have `loadData`, `pollOnce`, `subscribeToEvents`. Maybe we can do this?
- `pollData`, `loadDataOnStart`,  `subscribeToEventData`
- 
In run time they are all treated differently. pollData is for recurrent polling. loadDataOnStart is a one time pull. subscribeToEventData is for subscribing to events.

nickgeoca: Ok for now I renamed pollOnce to pullOnce. It looks fine to me

sparrowDom:
@nickgeoca thanks for all the work on this PR. I have deployed it [to staging](https://staging.ousd.com/mint) so we can test how it works agains the mainnet. My initial test results are: 
- for popular stable coins (dai, usdt, usdc) this works incredibly well. Time between when I see Metamask tx confirmation and updated coin values is usually 2-4  seconds. 
- for OGN balance updates the performance was worse. Time between Metamask tx confirmation and updated coin in the dapp was 25 seconds 1 run and 10 seconds another.
- when sending OUSD, the times were 37s, 5s, 3s

My theory is that Alchemy has some optimisation done for contracts that are more popular and propagates those events more effectively. Would be cool if more people can test and confirm my assumptions. 

@nickgeoca if you want to do some testing let me know and DM me your wallet address and I can wire you some stablecoins + eth for testing. 

As far as I am concerned this PR is done. But we still need to test it a bit more to see what is up with the OGN and OUSD balance update times. And we probably don't want to deploy this before the holidays

nickgeoca:
That sounds great @sparrowDom. I'll msg you and test out the site :+1:

dapp/src/components/AccountListener.js
sparrowDom: ok coo makes sense

dapp/src/components/AccountListener.js
sparrowDom: hey yeah I like the renames you suggest. Go for it 👍

sparrowDom:
Copy pasting testing results from @micahalcorn from Discord to here so we have it all in the same place: 

I sent DAI from MetaMask to imToken. It took just over 60 seconds for the reduced DAI balance to be reflected in the DApp (MetaMask side) after the confirmation notification appeared. The inverse only took a few seconds and it the increased balance was reflected in the DApp before the transaction was confirmed by imToken.

sparrowDom:
These random delays that can span from 10 seconds to up to a minute have me worried a bit. @nickgeoca would you mind putting some debug messages inside (can be usual `console.logs`) so we can determine if that is happening because of Transfer events arriving late or because querying for balance takes so much time. When you are done just ping me and I can deploy to staging (agains mainnet) again.

sparrowDom:
Will leave some of my testing results here. Have sent DAI 4 times to another wallet: 
- first time the transfer event was received and balance updated within 4 seconds
- second time similar experience, everything happened quite fast
- third time after 3 minutes of waiting the transfer event was never received and the balance has not updated
- the forth time after 28 seconds 2 transfer events were received and the balance has updated. 

My conclusion based on this and the previous test is that Alchemy's implementation of contract events is kind of sketchy and we can't really rely on it heavily. Too bad because if infrastructure on their side would be functioning perfectly this PR would be a pretty nifty upgrade.

nickgeoca:
This is unfortunate. It worked fine on the local testnet and I agree with your conclusion.

I transferred USDT twice but did not receive the event either time, in one case waiting 4 minutes. Testing USDC once, receiving the event took about 40 seconds longer than metamask's confirmation time.

Maybe this is a known Alchemy issue?

sparrowDom:
Closing this PR for now. This solution sounded like it should work. But testing on mainnet against Alchemy wasn't successful. We can reopen the PR if we find a solution for it. Things we could still try out: 
- use [enhanced Alchemy API](https://docs.alchemyapi.io/documentation/alchemy-web3/enhanced-web3-api) to query balances and allowances with less calls (probably separate PR)
- try using the wallet provider instead of alchemy to listen for transfer events
- get in touch with Alchemy team maybe they can clarify why `Transfer` events are so unreliable. @franckc have you perhaps already reached out to their team?

PR #406, opened 2020-12-04
OUSD Compensation Contract

DanielVF:
**Status: Ready for Review**

A compensation claims contract with a verification period in which everyone involved can check that all claims are correct before any actual funds are moved to the contract.

### Intended use sequence
 
1. Governor unlocks the adjuster
2. Adjuster uploads claims
3. Governor locks the adjuster
4. Everyone verifies that the claim amounts and totals are correct
5. Payout funds are moved to the contract
6. The claim period starts
7. Users claim funds
8. The claim period ends
9. Governor can collect any remaining funds

### Design goals

- Allow the governor to be 100% sure which accounts are being funded before any funds are made available to the contract, and with no way to change that allocation without the governor first unlocking.
- Allow amounts upload by a much less trusted account.
- Extreme simplicity. Nothing fancy anywhere. No extra unneeded code for the task at hand.

### Security (partial list)

- [x] Owner read through
- [x] Owner tries to attack
- [x] Slither 100% pass
- [x] Any function with state changes and external calls is reentrancy guarded
- [ ] Reviewer read through
- [ ] Reviewer tries to attack


DanielVF:
Rebased this to work off current master.

contracts/contracts/compensation/CompensationClaims.sol
```diff
@@ -0,0 +1,168 @@
...
+        return claims[_account];
+    }
+
+    function decimals() external view returns (uint8) {
+        return IERC20Decimals(token).decimals();
+    }
+
+    /* -- User -- */
+
+    function claim(address _recipient) external onlyInClaimPeriod nonReentrant {
```
tomlinton, line 75: Is it OK that anyone can claim on behalf of another user? Can we just use msg.sender instead of passing the _recipient arg?

contracts/contracts/compensation/CompensationClaims.sol
```diff
@@ -0,0 +1,168 @@
...
+        require(totalClaims > 0, "No claims");
+        uint256 funding = IERC20(token).balanceOf(address(this));
+        require(funding >= totalClaims, "Insufficient funds for all claims");
+        _lockAdjuster();
+        end = block.timestamp.add(_seconds);
+        require(end.sub(block.timestamp) < 31622400, "Duration too long"); // 31622400 = 366*24*60*60
+        emit Start(end);
+    }
+
+    function collect(address _coin)
```
tomlinton, line 136: What is this for?

contracts/contracts/compensation/CompensationClaims.sol
```diff
@@ -0,0 +1,168 @@
...
+    function start(uint256 _seconds)
+        external
+        onlyGovernor
+        notInClaimPeriod
+        nonReentrant
+    {
+        require(totalClaims > 0, "No claims");
+        uint256 funding = IERC20(token).balanceOf(address(this));
+        require(funding >= totalClaims, "Insufficient funds for all claims");
+        _lockAdjuster();
```
tomlinton, line 130: This is slightly disagreeing with the intended use sequence above, the adjuster should be locked before the claim period starts, maybe require adjuster is locked in this function or modify the intended use sequence comment.

contracts/contracts/compensation/CompensationClaims.sol
DanielVF: I wondered about this. My thought was that if money was locked up in a smart contract, it might not be setup to call this contract as a msg.sender by itself.

That said, I'm open to just using msg.sender, and handling anything else outside the normal flow.

contracts/contracts/compensation/CompensationClaims.sol
DanielVF: After the claims period, the governor can take back any unclaimed funds, plus any airdrops, or random money sent by mistake. Before the claims period begins, this gives the governor the ability to adjust how much funds are actually in the contract.

contracts/contracts/compensation/CompensationClaims.sol
DanielVF: I guess it would be better to have a require. This makes sure you didn't forget to lock the contract, and no last minute changes to the amounts snuck in.

PR #407, opened 2020-12-04
Deprecate single asset

nickgeoca:
Looks like the tests are passing :+1:


contracts/test/strategies/3pool-standalone.js
```diff
@@ -97,4 +99,23 @@ describe("3Pool Strategy Standalone", function () {
         .setRewardLiquidationThreshold(utils.parseUnits("10", 18))
     ).to.be.revertedWith("Caller is not the Governor");
   });
+
+  describe("Liquidate & Depricate", function () {
+    it("Should deposit and then liquidate the asset", async () => {
+      // Deposit
+      await expect(await threePoolGauge.balanceOf(tpStandalone.address)).to.be.equal("0");
+      await deposit('1', usdt);
+      await expect(await threePoolGauge.balanceOf(tpStandalone.address)).to.be.equal("1000000000000000000");
```
nickgeoca, line 108: User collateral balance is measured in the threePoolGauge contract like this?

contracts/contracts/strategies/ThreePoolStrategy.sol
```diff
@@ -144,7 +144,19 @@ contract ThreePoolStrategy is InitializableAbstractStrategy {
     /**
      * @dev Remove all assets from platform and send them to Vault contract.
      */
-    function liquidate() external onlyVaultOrGovernor nonReentrant {
+    function liquidate() external {
```
tomlinton, line 147: What was the internal/external visibility split needed for? :)

contracts/test/strategies/3pool-standalone.js
```diff
@@ -97,4 +99,23 @@ describe("3Pool Strategy Standalone", function () {
         .setRewardLiquidationThreshold(utils.parseUnits("10", 18))
     ).to.be.revertedWith("Caller is not the Governor");
   });
+
+  describe("Liquidate & Depricate", function () {
```
tomlinton, line 103: Typo, deprecate

contracts/contracts/strategies/ThreePoolStrategy.sol
nickgeoca: `ThreePoolStrategy` inherits the `InitializableAbstractStrategy` class, which has has two `liquidate` function signagutres. 

```
1    function liquidate() external;
2    function liquidate(address _asset) public;
```
1 `liquidate(address)` is public because `InitializableAbstractStrategy.deprecateAsset` calls it.
2 `liquidate()` is external, so it is cheaper to call it than if it were public.

In ThreePool both `liquidate(address)` and `liquidate` do the same thing since there is only one asset to liquidate. The internal `_liquidate` fn is used by both liquidate fns

``` solidity
    function liquidate() external {
        _liquidate();
    }

    /**
     * @dev Remove an asset from platform and send the asset to Vault contract.
     */
    function liquidate(address _asset) public {
        require(_asset == assetsMapped[0], "Asset is not first in assetsMapped arr");
        _liquidate();      
    }
```

contracts/test/strategies/3pool-standalone.js
nickgeoca: Thanks, fixed.

contracts/contracts/strategies/ThreePoolStrategy.sol
tomlinton: Thanks for the explanation!

We've got the core withdraw bit duplicated 3 times which we can clean up. It's not anything you've done, but just the way the code developed. I think we should (for Aave and Compound):

- Make an internal _withdraw function with the logic for Aave/Compound withdraws
- The external withdraw function should just call _withdraw
- The external liquidate function should also call _withdraw and pass the vaultAddress as _recipient and the entire balance for the amount
- Deprecate asset can also just call the internal _withdraw the same way as liquidate does in a loop which means we can remove the liquidate(address _asset) version

Let me know your thoughts, happy to help out with this.

nickgeoca: I like it. It helps simplify things which is great for security 👍

nickgeoca: In Compound, getting the underlying asset balance is not a view, right? So when checkBalance is called, it is an approximation? 
[https://compound.finance/docs/ctokens#underlying-balance](https://compound.finance/docs/ctokens#underlying-balance)

They mention this in getting the exchange rate:
_Tip: note the use of call vs. send to invoke the function from off-chain without incurring gas costs._

In Compound V1 I remember it being not a view b/c updating their state was required to calculate the latest/ever-changing exchange rate.

nickgeoca: I guess worse case ^ would be that a tiny fraction of tokens aren't liquidated. I haven't tested it, but it probably is not a big deal even if it is the case.

tomlinton: Yep exactly right!

tomlinton:
Hi @nickgeoca those changes are looking good. I made a bit of a mess of your PR through some other merged PRs, mainly liquidate got renamed withdrawAll. I merged master and started cleaning up but I'll finish fixing it up tomorrow 👍

Thanks for your work on this!

franckc:
Old PR. Closing.

PR #413, opened 2020-12-08
Sparrow dom/compensation

sparrowDom:
Compensation functionality to compensate OUSD and OGN to our users who have lost funds.

Mainnet Deploy TBD:
- currently `011_single_asset_staking_update.js ` only updates the merkle tree hash and depth on non mainnet deployments. Should we alter that and use the same script for mainnet deploy? 

Local Setup: 
Note: the main testing accounts (0x17BAd8cbCDeC350958dF0Bfe01E284dd8Fec3fcD, 0x3d89e78a9Feb7Be032D5aC01A10Ee2Ca97Ab35FD) already have an entry in the `airDrop.js` file and there is no need to do anything additional. If you want to change the compensation stake amounts or add new compensations do the following:

- in contracts npm package there is a `compute-merkle-proofs` command that takes addresses specified in `scripts/staking/airDrop.js` and creates an output consumed by the dapp `dapp/src/constants/merkleProofedAccountsToBeCompensated.json`. That file contains merkle proof data that are necessary to create stakes. We expose that file via dapp side api publicly and it is ok, since the contract verifies the wallet owner, so another wallet can not claim OGN compensation for anyone else.
- the output of the above file will also generate a root hash and tree depth that needs to be fed to the contract. Open `004_single_asset_staking.js` and modify `dropRootHash` and `dropRootDepth` variables to whatever running the script in the previous step produced.
- with that redeploy the contracts: `yarn run deploy`

Visit the /compensation page and run "Claim & Stake". You can only Claim & Stake once per account per deploy. If you want it to succeed again you need to redeploy the contracts or switch to account that hasn't staked yet.

**Related issue:** https://github.com/OriginProtocol/origin-dollar/issues/472


ZaHuPro:
I have updated the components at /compensation, as below mentioned

- [ ] Initial (with or without adjustment warning) https://zpl.io/aBrv8rL

- [ ] Claimed (OUSD and/or OGN) https://zpl.io/VQZJADx

- [ ] Ineligible for compensation https://zpl.io/aRZg8nv

- [ ] Disconnected (design pending)

sparrowDom:
@franckc the `011_single_asset_staking_update.js` currently updates the merkle tree hash and depth only on non mainnet environments. Since updating those 2 properties is the only thing that it does do you see any issue with altering the code so that it updates those 2 properties on mainnet as well?

franckc:
@sparrowDom I committed a few changes to the branch to prep for mainnet:
- Deleted deploy/012_single_asset_staking_update.js and replaced it by a new action setAirDropRoot in the scripts/governance/propose.js script. That will be used for configuring the OGN airdrop on Mainnet.
- made a few tweaks to scripts/compensation/compensationSync.js

Also I added steps/details to the push plan: https://docs.google.com/spreadsheets/d/1phyzOJMmTBPIqTTa0v7HY6XJkjRmbrcdULRZPo_JEoY/edit#gid=970484292

sparrowDom:
@franckc made some progress today with tools for verification and verification of data correctness: 
- There are new hardhat tasks that: 
  - fund all compensation accounts with eth
  - claimAll compensations as those accounts
  - check account / contract balances
- The compensationSync script now checks for OUSD balance in user's wallet aside from csv file and contract

Have done the analysis the results are in the top comment in `tasks/compensation.js` file.
- 5 accounts couldn't be funded. They are all contracts and had 0 eth in them. Except for this one: 0xeae57ce9cc1984F202e15e038B964bb8bdF7229a
- out of those 5 accounts 4 couldn't claim compensation. That amounted to 73 OUSD not claimed
- 20 accounts already had OUSD in them before claiming.

contracts/scripts/staking/constants.js
```diff
@@ -0,0 +1,9 @@
+const compensationData = {
+  type: 1,
+  rate: 25,
+  duration: 31104000,
```
franckc, line 4: This is 360 days. We should use 365 days = 31536000 sec so that it is consistent with the user staking which has duration of 30,90 or 365 days.

contracts/tasks/compensation.js
```diff
@@ -0,0 +1,120 @@
...
+  const { parseCsv } = require("../utils/fileSystem");
+  const ethers = hre.ethers
+  const compensationClaims = await hre.ethers.getContract("CompensationClaims");
+  const csv = await parseCsv(reimbursementsLocation);
+
+  let claimed = 0;
+  let errored = 0;
+  for (let i = 0; i < csv.length; i++) {
+    const account = csv[i].address
+    /* Filed because accounts not funded
```
franckc, line 60: typo: Failed

nit: maybe those could get defined as constants. Then when they fail, show a special error saying it is expected.

Something like
```
const problematicAccounts = [ "0x4853c9A7CB8f42a87dF28148F3380E35d8728043", ...]

....

if problematicAccounts.includes(account) {
    console.error("Expected failure", account, e)
} else {
    console.error("Unexpected failure on account", account, e)
}
```

franckc: Same comment applies to method fundCompAccountsWithEth below.

contracts/utils/deploy.js
```diff
@@ -159,14 +159,14 @@ const executeProposalOnFork = async (proposalId) => {
 
   const governor = await ethers.getContract("Governor");
 
-  // First enqueue the proposal, then execute it.
-  await withConfirmation(governor.connect(sGuardian).queue(proposalId, await getTxOpts()));
+  //First enqueue the proposal, then execute it.
+  await withConfirmation(governor.connect(sGuardian).queue(proposalId, await getTxOpts(executeGasLimit)));
```
franckc, line 163: nit: does the executeGasLimit need to be applied to the queue as well?

contracts/utils/fileSystem.js
```diff
@@ -0,0 +1,31 @@
...
+      header: true,
+      complete: ({ data }) => {
+        console.log("Complete", data.length, "records.");
+        resolve(data);
+      },
+    });
+  });
+};
+
+const hashFileContents = async (filePath) => {
```
franckc, line 21: Why is this async? ethers.utils.keccak256 is not async.

contracts/utils/fileSystem.js
```diff
@@ -0,0 +1,31 @@
+const fs = require("fs");
+const papa = require("papaparse");
+const { ethers } = require("hardhat");
+
+
+const parseCsv = (filePath) => {
```
franckc, line 6: shouldn't this be async since it returns a Promise?

contracts/scripts/staking/constants.js
franckc: Note that after updating this, we should re-generate the file merkleProofedAccountsToBeCompensated.json for the DApp to use.

dapp/src/constants/merkleProofedAccountsToBeCompensated.json
```diff
@@ -0,0 +1 @@
...7222bce885226cf74e6f3e185b80d8f730e3cd6acd9dab9e646e5937624f","0xd7a27ea00287607927bc97141d88f9b80210cb09545ec224376fd55a422c0a7d","0x9a383c897e5b6f003aa5de52f29edd6ae6255e58d996ad7595459506ba24181d","0x4ffbad7686561ffb6fcfcd280787343b986caa7504d22b19711fcedcb3c20b89","0x1c20c78f097ea47e60cbb82fe1c33aa66bf9e90d94429f34f840f457997519c9","0xf108a0fa3b5c75f3a81c550bbfb82a0b3c2295e1e433d4f2d7713288e1df247b","0x55e59a7bce7fb3e0bb34ddb4a13ee041c9ac355843c6650becb27211afb33efd","0xa34a289c9a165c72cb5256e5fb10e9633386099cafbe4a2f4ddd6015bc31bdd2"]},"0x4bB0E9F8d5D95bC47D80A4E49fe33A1Dd2651507":{"address":"0x4bB0E9F8d5D95bC47D80A4E49fe33A1Dd2651507","eligible_ousd_value_human":"1.01","ousd_compensation_human":"1.01","ogn_compensation_w_interest_human":"0.00","ogn_compensation_human":"0.00","eligible_ousd_value":"1008222912088142336","ousd_compensation":"1008222912088142336","ogn_compensation":"0","index":230,"type":1,"duration":31104000,"rate":"250000000000000000","proof":["0x9fbf93143b354a13ea4e92583f969b4a1f586a95bd8e1582ecdb643e77295bf8","0xd71fb3c6aab816494d11cdcfb81e3a259ee4613034aaaee33f4c28fe317bca4a","0x447e7222bce885226cf74e6f3e185b80d8f730e3cd6acd9dab9e646e5937624f","0xd7a27ea00287
```
franckc, line 1: Per my earlier comment, this needs to be re-generated with duration of 365 days

franckc (approved):
Nice work!

I focused my review on the files modified under origin-dollars/contract vs the DApp.

A few comments, mostly minor except the one regarding the duration that should be set to 365 days vs 360 days currently.

Otherwise LGTM - feel free to merge to master after addressing those.

contracts/scripts/staking/constants.js
sparrowDom: Thanks for catching this one! Interestingly on the dapp we read the durations from the contract now rather than from the API serving this csv data. That's why this probably hasn't been obvious when testing the dapp.

contracts/tasks/compensation.js
sparrowDom: That is a great ideas, changing to that

contracts/utils/deploy.js
sparrowDom: Good point, removing

contracts/utils/fileSystem.js
sparrowDom: thanks changing

dapp/src/constants/merkleProofedAccountsToBeCompensated.json
sparrowDom: 👍

contracts/utils/fileSystem.js
sparrowDom: 👍

Issue #415, opened 2020-12-09
ERC20 transfer violations due to suspected rounding errors


tomlinton:
From branch `tomlinton/echidna-action`:

```
crytic_self_transferFrom_to_other_ERC20PropertiesTransferable: failed!💥
  Call sequence:
    initialize("\145s\".\245\154\US\vn\SUB\216\239\DC1\133[`\DC1\223","\SO\&HEO\157",0xf17f52151ebef6c7334fad080c5704d77216b732) from: 0x627306090abab3a6e1400e9345bc60c78a8bef57 Time delay: 113292 seconds Block delay: 22491
    mint(0xf17f52151ebef6c7334fad080c5704d77216b732,112) from: 0xf17f52151ebef6c7334fad080c5704d77216b732 Time delay: 54309 seconds Block delay: 6289
    changeSupply(157198258) from: 0xf17f52151ebef6c7334fad080c5704d77216b732 Time delay: 257855 seconds Block delay: 3948
```

```
  Call sequence:
    initialize("\145s\".\245\154\US\vn\SUB\216\239\DC1\133[`\DC1\223","\SO\&HEO\157",0xf17f52151ebef6c7334fad080c5704d77216b732) from: 0x627306090abab3a6e1400e9345bc60c78a8bef57 Time delay: 113292 seconds Block delay: 22491
    mint(0xf17f52151ebef6c7334fad080c5704d77216b732,112) from: 0xf17f52151ebef6c7334fad080c5704d77216b732 Time delay: 54309 seconds Block delay: 6289
    changeSupply(49) from: 0xf17f52151ebef6c7334fad080c5704d77216b732
```

Still investigating the root cause but it is likely due to rounding errors. Note that not every echidna run will generate these. Sometimes I get a clean run.

tomlinton:
One potential way to fix this is to recalculate _totalSupply after the rebasingCreditsPerToken has been calculated

I have to think through the impact of that, one obvious consequence is that _totalSupply will no longer be set to exactly the _newTotalSupply argument to changeSupply.

tomlinton:
Closed by #418

PR #418, opened 2020-12-09
Echidna invariants and recalculating totalSupply

tomlinton:
This includes some Echidna setup and other things but I'm putting this up for feedback on this change:

```
        _totalSupply = rebasingCredits
            .divPrecisely(rebasingCreditsPerToken)
            .add(nonRebasingSupply);
```

Which is basically recalculating the total supply based on the calculated rebasingCreditsPerToken. It has some nice properties:

- Cleans up the issue in #415, which I suspect is due to rounding errors. This generates clean Echidna runs with the given setup (if you revert that change you'll generate errors like in #415).
- Prevents arbitrary changes in totalSupply, so totalSupply will always be > nonRebasingSupply, and in general totalSupply should be very close to the sum of all users balances (there will be some rounding error)


DanielVF:
I'll try to fully review tomorrow. After thinking about it today, seems good from a systems point of view.

contracts/contracts/token/OUSD.sol
```diff
@@ -473,22 +477,19 @@ contract OUSD is Initializable, InitializableERC20Detailed, Governable {
             );
         }
```
franckc, line 478: Aren't we missing a return statement here?

tomlinton: The return is a few lines down

franckc: Sorry, I don't see it. I was expecting at the end of the if statement to exit the function. I must be missing something completely obvious?

tomlinton: Ohh yes you are right, good catch. The github code excerpt was showing me something weird.

DanielVF:
So intuitively, I would expect the following properties to hold:

1. The totalSupply() equals the sum of the balanceOf() of all accounts.
2. If we updated the supply, and the amount went up, the total OUSD supply would be less than or equal to the new backing coins value passed in to changeSupply.

Number one now looks like it holds with this change. At least as far as me spending five minutes banging on it. This is great!

Number two doesn't seem to hold. It appears to be rounding the OUSD supply *up* by whatever error comes from working through rebasingCreditsPerToken. Example: 

```
300000000000000001702 <- changeSupply to this
300000000000000002100 <- get this totalSupply()

900300000000000000000000 <- changeSupply to this
900300000000001875324900 <- get this totalSupply()

700900300000000000000000000  <- changeSupply to this
700900300001437542660351298 <- get this totalSupply()

9000000000000000000000000000000000 <- Now I'm just being silly.
9000090000900009000090000900009000 <- get this totalSupply()
```

The size of the error increases as the supply amount goes up.

If we had 90,000 trillion OUSD, we could have too much supply compared to our backing by one thousandth of one percent!

You may or may not need to do anything about this.

DanielVF (approved):

tomlinton:
Thanks for taking a look @DanielVF! I'm okay with that finding, as the other one seems to be more of an issue so its a good trade.

PR #419, opened 2020-12-10
Rename liquidate to withdrawAll

tomlinton:
This is just a function rename, I think its clearer.


franckc (approved):
Yes, I like this new naming. It is more explicit. Thanks for doing that.

PR #426, opened 2020-12-15
Update staking rates on frontend

shahthepro:


sparrowDom:
@shahthepro LGTM. Added a small fix so on local the stake widget rate is aligned with the stake action box rate. To fix this problem - rate 197b vs 1.5m: 
Screenshot 2020-12-16 at 14 04 25

PR #429, opened 2020-12-15
Update staking deploy scripts

franckc:
Prepare for deployment of the OGN staking contract:
 - Add ability to test governance proposal execution flow on the Fork
 - Do not set staking root hash on Mainnet since we won't have finalized the numbers yet. We'll set the root hash when we are ready, via a governance proposal.
 - Update the propose script with action proposeClaimOGNStaking
 -  Update hardhat config file to handle the fork


tomlinton (approved):
👍

PR #430, opened 2020-12-15
Use get_virtual_price for pricing Curve assets

tomlinton:
Previously we calculated the balance for each asset in the 3pool strategy by determining what we'd get out if we withdrew all of that asset from the Strategy. This will become less accurate as the asset size increases due to slippage and isn't realistic because we don't intend to withdraw it all at once. It was previously necessary because redeem could interface directly with the 3pool strategy.

This change uses get_virtual_price() on the to get a price for the LP tokens and divides it by the number of assets supported by the strategy.

There is a change to the mock curve pool to support this, it burns any incoming assets and mints 1/3 of each of the supported stablecoins so the pool holds the amounts that the strategy says it does.

Contract change checklist:
  - [x] Unit tests pass


tomlinton:
Also slipped in a fix for one of slithers dangerous inequality errors.

DanielVF:
Glad to move away from calc_withdraw_one_coin! I'll sleep better at night.

I'm going to have to come back for a full review on this one, after I check out the others PRs.

I don't think getting this PR merged is a launch blocker, but I do think we probably shouldn't be using 3pool strategy in production until it is..

For several reasons, we probably don't want to hardcode a 1/3 allocation to each coin, and we also shouldn't price our assets based on purely get_virtual_price. Will post more later.

tomlinton:
Part of the awkwardness here is the abstraction requires a checkBalance for each asset, where for strategies that are decoupled from mint/redeem we only care about the checkBalance value for the whole strategy. One option might be to add checkBalance with the asset arg (for Compound and Aave it just iterates and calls checkBalance for each asset, and for 3pool it just returns the figure without the 1/3 division).

And we can either price our assets according to:

- The value of what we hold
- The value of what we get when we convert what we hold back for stablecoins

For the strategies coupled to mint/redeem I think the latter is better, but for those decoupled I think the former is better.

Keen to hear any further thoughts!

DanielVF:
Finding out how much money we have from 3pool is complicated.

First, 3pool charges you a fee every time you move money into and out of the pool. get_virtual_price does not account for the exit fee, and so naively will over estimate our holdings.

Secondly the internal price of stablecoins for 3pool tokens changes all day long based off the proportions of coins in 3pool. A 3pool token gets you more common coins at withdrawl time than it does rare ones. For example, if we ignore fees, if moved 1,000 USDC into 3pool, I might get 990 3pool tokens. If I turn around in the same transaction and cash these out for USDC, I would get roughly my entire tokens back. However, if I cashed out with DAI, I might only get 980 DAI back. 

It feels like we need to somehow account for the “price” of coins coming from curve. If we were able to do this with total perfection, then how much money we apparently had would then be entirely affected by the proportions of coins that we chose.

Given a simple 1/3 split on coins for our accounting purposes, and with a mostly USDC deposit stream like we see in real life, we would see our number of backing stable coins effectively going up less than the amount put in on each deposit. Something like depositing 1,000UDSC and seeing a balance of 300 DAI, 330 USDC, and 330 USDT. I really don’t like seeing a strong negative APR effect on the day that the deposits happen. We want positive feedback loops when things are going good, not negative ones.

calc_withdraw_one_coin is the holy grail method that completely handles the at the moment reality of prices and fees, with only one  minor downside that it will get your defi project looted by an attacker.

The issues:

- Proportion
- Prices
- Amount
- Attacker protection
- Quantities that go up and down.

## Proportion strategies:

- Even proportions - Possibly the worst of all approaches, other than the computation gas is cheap, and will get some good coins after a crash.
- Match proportions - Match the current 3pool proportions. Needs testing, but I think if we do this, the stable coins will match the virtual price, minus fees. If so, it might mean that our total apparent backing number of stable coins always increases. Will fail to only collecting crashed coins.
- Biggest coin. We check the 3pool proportions and use the largest coin. This gets us the highest quantity of stablecoins for a given amount of 3pool tokens. It could breathe quite a bit, as the proportions of stablecoins changes, though this depends on how we handle the amount.
- Manual proportions. Unfortunately would let an attacker controlled strategist force the strategy to trade the wrong coins during an attack, resulting in losses. But the strategist could do this anyway just by manually transferring coins between strategies. Otherwise this could be handy to avoid crashed coins.
- Proportions match input and output. We keep track of how many coins come in, how many go out, use that number as the proportions we use. As long as the proportions of coins stays roughly in the same vicinity on 3pool, we see as much proportion based balance changes.

## Proportion timing:

 - Recheck on every time we get totals from the contract. This is likely prohibitively expensive, since it would be 9 calls each time we got totals.
 - Strategist can call method to check proportions from 3Pool and set an internal proportions storage slot, according to a rule.
 - Strategist can just set the proportions the contract uses.
	
## Pricing:

Pricing is the hardest of all. If someone controls the prices, they can steal money. And the prices that we get from 3pool can be controlled by an attacker. It would be nice if 3pool had some kind of constantly adding price feed like uniswap does, and we could use it to get a daily average, but I don’t think it does.

- Get prices directly from 3pool. Get Rekt.
- The strategist could manually set relative prices, but might allow someone controlling the strategist to steal funds.
- The strategist being able to pull prices from 3 pool with a transaction, while better than letting the general public do so, still lets the strategist attack the OUSD with a flash loan, just as an attacker would.
- Assume 1:1 with virtual price. We underestimate our holdings.
- Do some kind of long running averaging on prices from 3pool and apply it to virtual price.
- Have some kind of strategist controlled factor that is applied to the virtual price. Only has a very narrow range it can change in (perhaps only 1% off 1).
- Other ideas.

 	
## Overall approaches

(I’m going to have to think some more)

DanielVF:
So I think you are right Tom, that this can be much simpler given the constraint that this code is never running in response to user transactions. 

I've communed with the 3pool contract, comparing the results of virtual_price, calc_withdraw_one_coin, and calc_token_amount. 

It looks like as long as you withdraw coins in amounts matching the 3pool proportions, you get exactly the virtual price's worth of coins for your tokens. Not only that, but a matched withdraw (or deposit) also waives the 3pool fee. Given that we are headed for having that capability, I think just using virtual price here is awesome.

In the long term, we might want to have a method that fetches the proportions from 3pool, and then stores them for use by the balances. As it is, we'll be slightly over representing the proportion of DAI that we have, which will bias withdraws to have more DAI in them than they should. HOWEVER this effect is going to be tiny, and this PR is going to be a massive win. We can kick adding any more complexity down the road and see if we need it.

Approach looks great here. I'm going to review the code now.

contracts/contracts/vault/VaultCore.sol
```diff
@@ -300,25 +300,26 @@ contract VaultCore is VaultStorage {
         for (uint256 i = 0; i < allAssets.length; i++) {
             IERC20 asset = IERC20(allAssets[i]);
             uint256 assetBalance = asset.balanceOf(address(this));
-            // No balance, nothing to do here
-            if (assetBalance == 0) continue;
```
DanielVF, line 304: Ugh. I'm assuming this change is for slither. 

Does it get happier if we just do < 1 ? I really like the return/continue early style rather than nesting code in ifs.

DanielVF (approved):
Reviewed code, ran tests. Looks great. One minor style comment.

PR #432, opened 2020-12-16
Timelock: make queueTransaction and executeTransaction internal

franckc:
There does not seem to be a reason to keep Timelock.queueTransaction and Timelock.executeTransaction public.
These actions should never be called directly but rather should be initiated via the Governor.queue and Governor.execute methods.

Note:
  1. In order to make Timelock.executeTransaction internal, I also had to remove its payable modifier. I can't think of a use case where a proposal would require calling a method that receives ETH. But please do double-check me on this.
  1. Per suggestion from @tomlinton I also removed the ability for a proposal to send ETH

Contract change checklist:
  - [X] Unit tests pass

### DIFF
--- a/contracts/contracts/timelock/Timelock.sol
+++ b/contracts/contracts/timelock/Timelock.sol
@@ -21,7 +21,6 @@ contract Timelock {
     event CancelTransaction(
         bytes32 indexed txHash,
         address indexed target,
-        uint256 value,
         string signature,
         bytes data,
         uint256 eta
@@ -29,7 +28,6 @@ contract Timelock {
     event ExecuteTransaction(
         bytes32 indexed txHash,
         address indexed target,
-        uint256 value,
         string signature,
         bytes data,
         uint256 eta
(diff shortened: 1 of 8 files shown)


contracts/contracts/timelock/Timelock.sol
```diff
@@ -159,7 +159,7 @@ contract Timelock {
         string memory signature,
         bytes memory data,
         uint256 eta
-    ) public payable returns (bytes memory) {
```
tomlinton, line 162: We should remove the `values` from the proposal in the Governor and the payable modifier from the Governor's execute too I think?

shahthepro (approved):

contracts/contracts/timelock/Timelock.sol
franckc: Done. PTAL...

tomlinton: 💙

franckc: BTW, I think we can also remove `function() external payable {}` from Timelock.sol?
https://github.com/OriginProtocol/origin-dollar/blob/master/contracts/contracts/timelock/Timelock.sol#L70

tomlinton (approved):

shahthepro (approved):

PR #441, opened 2020-12-18
Add stats endpoint for OGN dashboard

shahthepro:
Related PR: https://github.com/OriginProtocol/origin-website/pull/565

Contract change checklist:
  - [ ] Unit tests pass


contracts/contracts/staking/SingleAssetStaking.sol
```diff
@@ -37,6 +37,7 @@ contract SingleAssetStaking is Initializable, Governable {
     uint256[] public rates; // rates that correspond with the allowed durations
 
     uint256 public totalOutstanding;
+    uint256 public totalStakers;
```
franckc, line 40: Contract is already deployed and uses a proxy. So I believe any new slot should be appended, not inserted in the middle? Move totalStakers after bool paused.

contracts/contracts/staking/SingleAssetStaking.sol
```diff
@@ -413,6 +418,9 @@ contract SingleAssetStaking is Initializable, Governable {
         require(totalWithdraw > 0, "All stakes in lock-up");
 
         totalOutstanding = totalOutstanding.sub(totalWithdraw);
+        if (stakes[stakes.length - 1].paid) {
+            totalStakers++;
```
franckc, line 422: The totalStakers counter should be decremented

contracts/contracts/staking/SingleAssetStaking.sol
franckc: Nit: add a comment explaining this is the total number of accounts that have at least one active stake (e.g. unpaid)

contracts/contracts/staking/SingleAssetStaking.sol
```diff
@@ -169,6 +170,10 @@ contract SingleAssetStaking is Initializable, Governable {
 
         require(i < MAX_STAKES, "Max stakes");
 
+        if (i == 0 || stakes[i - 1].paid) {
```
franckc, line 173: Super nit: to be explicit, add a comment explaining that if it's a new staker or a staker that had exited all their stakes, we increment the counter.

dapp/pages/api/stats.js
```diff
@@ -0,0 +1,58 @@
+import { ethers } from 'ethers'
```
franckc, line 1: Having a stats page hosted on the DApp works.

Though I wonder if it would make more sense to move this API to the [ousd analytics stack](https://github.com/OriginProtocol/ousd-analytics) since that's where we are putting all the stats related APIs.
Also an advantage of doing that is that we would be able to differentiate the user traffic hitting the dapp vs coming from the website on our Alchemy dashboard because we use different Alchemy keys for dapp vs analytics.

CC @mikeshultz @DanielVF for their opinion

franckc (commented):
A few comments inline.

Also let's add some unit tests - it is a simple change but let's strive for having excellent tests coverage for any contract code we write.

dapp/pages/api/stats.js
```diff
@@ -0,0 +1,58 @@
...
+  if (!stakingContract) {
+    return res.status(500).send({
+      error: true
+    })
+  }
+
+  const totalStaked = await stakingContract.totalOutstanding()
+  const totalStakers = await stakingContract.totalStakers()
+
+  return res.status(500).send({
```
mikeshultz, line 54: 500?

dapp/pages/api/stats.js
DanielVF: We could definitely do this on the analytics side and keep the contracts simpler. That's almost always my vote on that trade off. Also we don't have to spend extra gas to increment the counter.

mikeshultz: This here is just 2 client-side JSON-RPC calls so I don't really think this warrants a lot of debate.  This isn't currently available via analytics so maybe just roll with this for now.

I am planning to add staking data to ousd-analytics soon, though.  We could add API endpoints for this data if you all find it useful.

mikeshultz: Ah, I missed the contract changes necessary for this.  I don't feel super strongly either way then.  I can prioritize work on the analytics collection to get it ready for release, but maybe having that counter in-contract isn't a bad idea either.

franckc: Yeah, it would anyway be useful to have the staking data in the analytics DB so that the product/business team can analyze it.

If we would capture in the DB the [Staked and Withdrawn events](https://github.com/OriginProtocol/origin-dollar/blob/master/contracts/contracts/staking/SingleAssetStaking.sol#L458) emitted by the contract I think that this by itself should be enough to calculate the number of stakers.

If we want to go beyond, upon seeing these events, we could call the contract's getAllStakes() method and store/update in the DB the details of all the stakes for the account that triggered the event.

shahthepro:
Closing this PR, since we can do this on the analytics repo

PR #442, opened 2020-12-18
Sparrow dom/staking approve call

sparrowDom:
- replace the mockOgn dev contract with our actual live contract
- adds `stakeWithSender` function to singleAssetStaking contract
- **to be confirmed this is ok** adds a `web3-eth-abi` npm package
- make the dapp use the `approveAndCallWithSender` also in dev mode

Issues: 
- the testing account `0x17BAd8cbCDeC350958dF0Bfe01E284dd8Fec3fcD` receives total supply of Ogn. Don't yet know why

Contract change checklist:
  - [ ] Unit tests pass


contracts/contracts/staking/SingleAssetStaking.sol
```diff
@@ -191,6 +191,17 @@ contract SingleAssetStaking is Initializable, Governable {
         emit Staked(staker, amount);
     }
 
+    function _stakeWithChecks(address staker, uint256 amount, uint256 duration) internal {
+        require(amount > 0, "Cannot stake 0");
+
+        uint240 rewardRate = _findDurationRate(duration);
+        require(rewardRate > 0, "Invalid duration"); // we couldn't find the rate that correspond to the passed duration
+
+        _stake(msg.sender, USER_STAKE_TYPE, duration, rewardRate, amount);
```
franckc, line 200: this should be staker rather than msg.sender

contracts/contracts/staking/SingleAssetStaking.sol
```diff
@@ -377,14 +388,21 @@ contract SingleAssetStaking is Initializable, Governable {
...
-        stakingToken.safeTransferFrom(msg.sender, address(this), amount);
+    /**
+     * @dev Stake an approved amount of staking token into the contract. This function
+     *      can only be called by OGN token contract.
+     * @param staker Address of the account that is creating the stake
+     * @param amount Number of tokens to stake in 1e18
+     * @param duration Number of seconds this stake will be held for
+     */
+    function stakeWithSender(address staker, uint256 amount, uint256 duration) external returns (bool) {
+        // no checks are performed in this function since those are already present in _stakeWithChecks
```
franckc, line 403: Important: we need to add a require checking this is being called from the OGN contract.
```
require(msg.sender == address(tokenAddr), "Token must call");
```

contracts/contracts/mocks/MockOGN.sol
```diff
@@ -1,9 +1,107 @@
...
+import "@openzeppelin/contracts/token/ERC20/ERC20Burnable.sol";
+import "@openzeppelin/contracts/token/ERC20/ERC20Mintable.sol";
+import "@openzeppelin/contracts/token/ERC20/ERC20Detailed.sol";
 
-contract MockOGN is MintableERC20 {
-    uint256 public constant decimals = 18;
-    string public constant symbol = "OGN";
-    string public constant name = "OriginToken";
-}
+import "./WhitelistedPausableToken.sol";
```
franckc, line 7: It does not hurt but I don't think this is necessary. This is used to pause the token. Which shouldn't be related to the spender whitelist.

DanielVF: I see the merit here to use the actual OGN methods. Ensures our contract's behavior is correct. The one thing I might do (open for discussion) is add back the `mint(amount)` function that all our mocks have. This way we don't have to change our tests, and the OGNMock has the same minimal subset of methods as our other token mocks.

DanielVF (commented):
Did we need to commit a changed package.lock file for this commit? Was that an error, or did something change?

DanielVF:
I'm thinking we need to add tests for using the OGN whitelist with this.

dapp/pages/stake.js
```diff
@@ -200,7 +202,20 @@ const Stake = ({ locale, onLocale, rpcProvider, isMobile }) => {
         }
         tokenToStakeDecimalsCall={ognContract.decimals}
         stakeFunctionCall={async (stakeAmount) => {
-          return ognStaking.stake(stakeAmount, selectedDuration)
+          const stakeAmountString = formatBn(stakeAmount, 18)
```
franckc, line 205: I think we should be able to get rid of the new Web3 dependency and use ethers.js. See https://docs.ethers.io/v5/single-page/#/v5/api/utils/abi/interface/
I haven't tested but something like this should work:
```
const interface = ognStaking.interface
const fragment = ognStaking.interface.getFunction("stakeWithSender(uint256, uint256)")
const fnSig = interface.getSighash(fragment)
const params = interface.encodeFunctionData("stakeWithSender(uint256, uint256)", [stakeAmountString, selectedDuration])
```

franckc (commented):
A few more thoughts inline.

contracts/deploy/006_liquidity_reward.js
```diff
@@ -153,7 +153,7 @@ const liquidityReward = async ({ getNamedAccounts, deployments }) => {
       const ogn = await ethers.getContract("MockOGN");
       const loadAmount = utils.parseUnits("7200000", 18);
       const rate = utils.parseUnits("6.1538461538", 18);
-      await ogn.connect(sGovernor).mint(loadAmount);
+      await ogn.connect(sGovernor).mint(await sGovernor.getAddress(), loadAmount);
```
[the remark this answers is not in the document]
sparrowDom: ah yes easier thanks

contracts/contracts/staking/SingleAssetStaking.sol
sparrowDom: great catch, thanks!

contracts/contracts/staking/SingleAssetStaking.sol
sparrowDom: Thanks good catch added this

contracts/contracts/mocks/MockOGN.sol
sparrowDom: Yeah I wanted to keep the MockOGN contract as close as possible to the live version.  You make a valid point, I don't feel strongly either way. Maybe @DanielVF, @crazybuster or @tomlinton can wight in?

contracts/deploy/006_liquidity_reward.js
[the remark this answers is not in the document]
sparrowDom: yup thanks, fixed

dapp/pages/stake.js
sparrowDom: great thanks using that

sparrowDom:
> Did we need to commit a changed package.lock file for this commit? Was that an error, or did something change?

Yeah I added a npm package in there. But per @franckc 's suggestion that npm package can be scrapped.

sparrowDom:
Slither tests now pass but with disabling the following detectors: `controlled-array-length,weak-prng,divide-before-multiply`

- the `controlled-array-length` in `contracts/contracts/utils/InitializableAbstractStrategy.sol [line 193]` seems to only be called at initialise and by the governor from what I can tell
- the `weak-prng` in `contracts/contracts/oracle/UniswapLib.sol[line 49]` looks like a false positive to me
- the `divide-before-multiply` in `contracts/contracts/utils/StableMath.sol [line 31 & 33]` also looks like a false positive

Would love some feedback & code review from @tomlinton @DanielVF @crazybuster

contracts/contracts/staking/SingleAssetStaking.sol
```diff
@@ -377,14 +388,22 @@ contract SingleAssetStaking is Initializable, Governable {
...
+     * @param duration Number of seconds this stake will be held for
+     */
+    function stakeWithSender(address staker, uint256 amount, uint256 duration) external returns (bool) {
+        require(msg.sender == address(stakingToken), "Only token contract can make this call");
 
-        _stake(msg.sender, USER_STAKE_TYPE, duration, rewardRate, amount);
-        // transfer in the token so that we can stake the correct amount
-        stakingToken.safeTransferFrom(msg.sender, address(this), amount);
+        _stakeWithChecks(staker, amount, duration);
+        return true;
```
franckc, line 406: What is the benefit of returning a boolean for this method since it always returns true?

contracts/package.json
```diff
@@ -19,7 +19,7 @@
...
     "copy-interface-artifacts": "mkdir -p ../dapp/abis && cp artifacts/contracts/interfaces/IVault.sol/IVault.json ../dapp/abis/IVault.json && cp artifacts/contracts/liquidity/LiquidityReward.sol/LiquidityReward.json ../dapp/abis/LiquidityReward.json && cp artifacts/contracts/interfaces/uniswap/IUniswapV2Pair.sol/IUniswapV2Pair.json ../dapp/abis/IUniswapV2Pair.json && cp artifacts/contracts/staking/SingleAssetStaking.sol/SingleAssetStaking.json ../dapp/abis/SingleAssetStaking.json",
     "echidna": "yarn run clean && echidna-test . --contract PropertiesOUSDTransferable --config contracts/crytic/TestOUSDTransferable.yaml",
-    "slither": "yarn run clean && slither . --filter-paths \"crytic|mocks|@openzeppelin\" --exclude-low --exclude-informational --exclude conformance-to-solidity-naming-conventions,different-pragma-directives-are-used,external-function,assembly",
+    "slither": "yarn run clean && slither . --filter-paths \"crytic|mocks|@openzeppelin\" --exclude-low --exclude-informational --exclude conformance-to-solidity-naming-conventions,different-pragma-directives-are-used,external-function,assembly,controlled-array-length,weak-prng,divide-before-multiply",
```
franckc, line 22: These rules may uncover issues in the future on other contracts.
So I wonder if, as opposed to excluding them globally, we should add the known false positives they return to [slither.db.json](https://github.com/OriginProtocol/origin-dollar/blob/master/contracts/slither.db.json)?
I'm actually not very familiar with the best way to update slither.db.json - would be good to get input on that from @shahthepro or @tomlinton (and we should also document it in our security repo so that everyone knows).

contracts/test/staking/single-asset-staking.js
```diff
@@ -121,6 +121,46 @@ describe("Single Asset Staking", function () {
...
+
+    const numStakeAmount = 1;
+    const stakeAmount = ognUnits(numStakeAmount.toString());
+    // 0.085 is the default reward for three months
+    const expectedReward = ognUnits((numStakeAmount * 0.085).toString());
+
+    await expect(ognStaking.connect(anna).stakeWithSender(anna.address, stakeAmount, threeMonth))
+      .to.be.revertedWith("Only token contract can make this call");
+
+    // from franck
```
franckc, line 138: nit: cleanup since not really relevant to leave my name, especially since crazybuster found a better way:)

contracts/test/staking/single-asset-staking.js
```diff
@@ -121,6 +121,46 @@ describe("Single Asset Staking", function () {
...
+    // This generate the data needed for calling stakeWithSender
+    const interface = ognStaking.interface
+    const fragment = ognStaking.interface.getFunction("stakeWithSender(address,uint256,uint256)")
+    const fnSig = interface.getSighash(fragment)
+    // calling regular stake here because stakeWithSender inserts the caller's address here
```
franckc, line 143: nit: cleanup those comments below. not relevant anymore.

dapp/pages/stake.js
```diff
@@ -200,7 +201,19 @@ const Stake = ({ locale, onLocale, rpcProvider, isMobile }) => {
         }
         tokenToStakeDecimalsCall={ognContract.decimals}
         stakeFunctionCall={async (stakeAmount) => {
-          return ognStaking.stake(stakeAmount, selectedDuration)
+          //const stakeAmountString = formatBn(stakeAmount, 18)
+          const interfce = ognStaking.interface
```
franckc, line 205: nit: typo "interface"

dapp/pages/stake.js
```diff
@@ -200,7 +201,19 @@ const Stake = ({ locale, onLocale, rpcProvider, isMobile }) => {
...
+          const fragment = interfce.getFunction("stakeWithSender(address,uint256,uint256)")
+          const fnSig = interfce.getSighash(fragment)
+          // calling regular stake here because stakeWithSender inserts the caller's address here
```
franckc, line 208: clean up. comment not relevant anymore

dapp/yarn.lock
```diff
@@ -1403,6 +1403,21 @@
   resolved "https://registry.yarnpkg.com/@emotion/unitless/-/unitless-0.7.5.tgz#77211291c1900a700b8a78cfafda3160d76949ed"
   integrity sha512-OWORNpfjMsSSUBVrRBVGECkhWcULOAJz9ZW8uK9qgxD+87M7jHRcvh/A96XXNhXTLmKcoYSQtBEX7lHMO7YRwg==
 
+"@ethersproject/abi@5.0.7":
```
franckc, line 1406: We don't need this new dependency, do we?

franckc (commented):
A few more minor comments inline. PTAL. Getting very close. Thanks for all your work on this.

franckc:
@sparrow regarding the slither warnings you mentioned in discord:
 - controlled-array-length in contracts/contracts/utils/InitializableAbstractStrategy.sol [line 193]
Since the method _setPTokenAddress is only called at initialization or by the governor, I think we can safely ignore this warning.
 - weak-prng in contracts/contracts/oracle/UniswapLib.sol[line 49]
This file was removed. Pull from latest master.
 - divide-before-multiply in contracts/contracts/utils/StableMath.sol [line 31 & 33]
Looks like a false positive to me.

contracts/contracts/staking/SingleAssetStaking.sol
sparrowDom: The `MockOGN` that calls this function expects it to return true:
https://github.com/OriginProtocol/origin-dollar/blob/sparrowDom/stakingApproveCall/contracts/contracts/mocks/MockOGN.sol#L111

contracts/test/staking/single-asset-staking.js
sparrowDom: cleaned up

contracts/test/staking/single-asset-staking.js
sparrowDom: yup thanks removed

dapp/pages/stake.js
sparrowDom: Renaming to `iface` so it looks like less of a typo, couldn't use `interface` since it is a reserved keyword

dapp/pages/stake.js
sparrowDom: yup thanks removed

dapp/yarn.lock
sparrowDom: Removed this from `package.json` but haven't rerun the `yarn install`. Thanks for catching it!

contracts/package.json
sparrowDom: From what I can tell the `slither.db.json` is not something that was intended to be user editable.

The slither command does support the `--filter-paths` parameter. We could use that to run slither multiple times on different files with different `--exclude` parameters. This way we could single out the source files that have false positives and run slither with exclusion of specific detectors only on those. 

Though I am not completely thrilled by that idea, since it is going to take more time to perform the runs and can become really messy maintenance wise.

shahthepro: You can run `yarn run slither --triage` to regenerate the slither DB

crazybuster (approved):
Would like to see more tests, but it's alright for now.

contracts/package.json
sparrowDom: Nice thanks @shahthepro will look into that tomorrow

sparrowDom:
thanks @crazybuster will add more tests tomorrow after we merge this in

PR #443, opened 2020-12-18
Remove unused internal _min function from AAVE strategy contract

DanielVF:
The AAVE contract has a `_min` function which is not currently being used. Looking at the commit logs, it looks like it was present from the contract start, but was not used at that time either.

I've removed it.

Contract change checklist:
  - [x] Unit tests pass


franckc (approved):
Good catch. I don't see any reference to it in the code.

LGTM

tomlinton (approved):

PR #446, opened 2020-12-18
Deploy script for upgrading the SingleAssetStaking contract

franckc:
For deploying to Mainnet the latest changes to the contract.

Also includes a fix to have the ability to apply a multiplier to the gas price (this is handy to speed up deployments on mainnet).
We thought hardhat supports it via the gasMultiplier option in the config but that actually applies to the gas limit, not the gas price. So I removed that option in the config and replace it with customer code that is enabled via the GAS_PRICE_MULTIPLER env variable.

### DIFF
--- a/contracts/deploy/007_upgrade_single_asset_staking.js
+++ b/contracts/deploy/007_upgrade_single_asset_staking.js
@@ -12,6 +12,7 @@ const {
   executeProposal,
 } = require("../utils/deploy");
 const { proposeArgs } = require("../utils/governor");
+const { getTxOpts } = require("../utils/tx");
 
 const deployName = "007_upgrade_single_asset_staking";
 
@@ -54,9 +55,9 @@ const upgradeSingleAssetStaking = async ({ getNamedAccounts }) => {
     await executeProposal(propArgs, propDescription);
   } else {
     // Local testing environment. Upgrade via the governor account directly.
-    await cOGNStakingProxy
+    const tx = await cOGNStakingProxy
       .connect(sGovernor)
-      .upgradeTo(dSingleAssetStaking.address);
+      .upgradeTo(dSingleAssetStaking.address, await getTxOpts());
     log(`Upgraded OGNStaking to ${dSingleAssetStaking.address}`);
   }
 
--- a/contracts/deploy/007_upgrade_single_asset_staking.js
+++ b/contracts/deploy/007_upgrade_single_asset_staking.js
@@ -2,14 +2,10 @@
 // Script to upgrade the Single Asset Staking contract.
 //
 const {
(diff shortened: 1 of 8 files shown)


contracts/deploy/007_upgrade_single_asset_staking.js
```diff
@@ -0,0 +1,82 @@
...
+const { proposeArgs } = require("../utils/governor");
+
+const deployName = "007_upgrade_single_asset_staking";
+
+const upgradeSingleAssetStaking = async ({ getNamedAccounts, deployments }) => {
+  console.log(`Running ${deployName} deployment...`);
+
+  const { governorAddr, deployerAddr } = await getNamedAccounts();
+
+  const assetAddresses = await getAssetAddresses(deployments);
```
sparrowDom, line 27: @franckc this doesn't seem to be used

contracts/deploy/007_upgrade_single_asset_staking.js
```diff
@@ -0,0 +1,82 @@
...
+
+  const sDeployer = ethers.provider.getSigner(deployerAddr);
```
sparrowDom, line 29: this also

contracts/deploy/007_upgrade_single_asset_staking.js
```diff
@@ -0,0 +1,82 @@
...
+  // Deploy the new implementation.
+  //
+  const dSingleAssetStaking = await deployWithConfirmation(
+    "SingleAssetStaking"
+  );
+
+  //
+  // Upgrade.
+  //
+  const cOGNStakingProxy = await ethers.getContract("OGNStakingProxy");
```
sparrowDom, line 42: @franckc not sure do we need to call initialise like here or it is not needed?
https://github.com/OriginProtocol/origin-dollar/blob/master/contracts/deploy/004_single_asset_staking.js#L96-L98

sparrowDom (commented):
LGTM, left a few comments inline. Would feel better another one more experiences with contract deploy scripts also takes a look.

contracts/deploy/007_upgrade_single_asset_staking.js
franckc: ah yes good catch. cleaned up.

contracts/deploy/007_upgrade_single_asset_staking.js
franckc: When using proxies, the proxy is used for storing all the data. Since we initialized this data already during the 004 deploy, we don't need to do it again. we are in effect just upgrading the logic.

franckc:
Merging this now. Will address any additional feedback as a separate PR.

PR #452, opened 2020-12-21
Add test, checking just under the max possible stake

DanielVF:
Add a test to single asset staking, checking that we allow staking of under the staking amount possible given the OGN placed into the contract..

No contract changes.


franckc (approved):
Thanks for adding that test.
LGTM

PR #459, opened 2020-12-23
Update events

shahthepro:
To be deployed before https://github.com/OriginProtocol/ousd-analytics/pull/12

- Adds the duration and rate to `Staked` event
- Adds a `stakedAmount` arg to `Withdrawn` event (to be used by ousd-analytics)

Contract change checklist:
  - [x] Unit tests pass


sparrowDom (dismissed):
LGTM.

@shahthepro how did `divide-before-multiply` not fail on CI without changing the `slither.db.json`?

shahthepro:
> how did `divide-before-multiply` not fail on CI without changing the `slither.db.json`?

@sparrowDom TBH, Idk. I was just trying to get this to fail so that I can debug it. But if it works, I guess we can just leave it as it is for now

Also, I need to make some changes on `ousd-analytics` before we deploy this.

sparrowDom:
@shahthepro thanks, hmm so weird. I hope we won't have future problems with  undetermination with of this tool

contracts/contracts/staking/SingleAssetStaking.sol
```diff
@@ -188,7 +188,7 @@ contract SingleAssetStaking is Initializable, Governable {
 
         totalOutstanding = totalOutstanding.add(_totalExpected(newStake));
 
-        emit Staked(staker, amount);
+        emit Staked(staker, amount, duration, rate);
```
franckc, line 191: rate is a uint240. But the Staked event uses a uin256. Does solidity automatically converts from uint240 to uint256 when emitting the event or do we need to cast?

franckc (commented):
Minor comment inline.
Also please add a migration for the deployment of this upgrade (it can just be a copy of the existing [007_single_asset_staking_upgrade.js](https://github.com/OriginProtocol/origin-dollar/blob/master/contracts/deploy/007_upgrade_single_asset_staking.js))

franckc:
CC @mikeshultz to keep him in the loop since that is analytics related

contracts/contracts/staking/SingleAssetStaking.sol
shahthepro: Yeah, solidity does and can uint240 to uint256 but the other way isn't possible. 

Let me know if this has to be changed to keep it consistent

franckc: Thanks for checking. No need I think, to change as long as we make sure the event data is emitted correctly.

DanielVF:
### Requirements

We'll need to adjust the analytics once this change goes live. Totally doable.

I'm assuming analytics will need to handle both old and new events types. If we are handling the old event types anyway, and able to extract the information from other sources to fill in the gaps, then adding information to these events may just result in more code on the analytics side for the same functionally.

### Logic

_No changes to logic_

### Tests

#### Overflow

_No changes to math._

#### Black magic

_No magic._

#### Authentication

_No changes to authentication._

#### Cryptographic code

_No changes to cryptographic._

#### Gas problems

This will slightly increase gas usage. This is fine from a user point of view. @sparrowDom, we are dynamically calculating gas amounts from the staking dapp, right?

_No changes to loops._

#### External calls

_No changes to/from external calls_

#### Ethereum

_Does not send or receive Ethereum._

DanielVF (approved):
### Requirements

We'll need to adjust the analytics once this change goes live. Totally doable.

I'm assuming analytics will need to handle both old and new events types. If we are handling the old event types anyway, and able to extract the information from other sources to fill in the gaps, then adding information to these events may just result in more code on the analytics side for the same functionally.

### Logic

_No changes to logic_

### Tests

#### Overflow

_No changes to math._

#### Black magic

_No magic._

#### Authentication

_No changes to authentication._

#### Cryptographic code

_No changes to cryptographic._

#### Gas problems

This will slightly increase gas usage. This is fine from a user point of view. @sparrowDom, we are dynamically calculating gas amounts from the staking dapp, right?

_No changes to loops._

#### External calls

_No changes to/from external calls_

#### Ethereum

_Does not send or receive Ethereum._

tomlinton (approved):

contracts/deploy/010_upgrade_single_asset_staking.js
```diff
@@ -0,0 +1,75 @@
...
+} = require("../test/helpers.js");
+const {
+  log,
+  deployWithConfirmation,
+  executeProposal,
+} = require("../utils/deploy");
+const { proposeArgs } = require("../utils/governor");
+const { getTxOpts } = require("../utils/tx");
+
+const deployName = "007_upgrade_single_asset_staking";
```
franckc, line 17: nit: update the name

contracts/deploy/010_upgrade_single_asset_staking.js
```diff
@@ -0,0 +1,75 @@
...
+    {
+      contract: cOGNStakingProxy,
+      signature: "upgradeTo(address)",
+      args: [dSingleAssetStaking.address],
+    },
+  ]);
+
+  if (isMainnet) {
+    // On Mainnet upgrade has to be handled manually via a multi-sig tx.
+    log(
```
franckc, line 47: I recently changed our pattern here so that we don't need to run this propose step separately.
Do you mind calling `await sendProposal(propArgs, propDescription);` here?
See <https://github.com/OriginProtocol/origin-dollar/blob/master/contracts/deploy/009_ousd_fix.js#L71>

franckc (approved):
Thanks for adding the deploy script. I can help run it tomorrow.
A couple of minor nits inline otherwise LGTM

contracts/deploy/010_upgrade_single_asset_staking.js
shahthepro: Have done this change

franckc:
LGTM - feel free to merge to master

PR #469, opened 2021-01-04
OUSD deploy - migration 9

franckc:
Files from running 009_ousd_fix.js on Mainnet.


contracts/deploy/009_ousd_fix.js
```diff
@@ -75,23 +76,27 @@ const fixOUSD = async () => {
     await executeProposal(propResetArgs, propResetDescription);
     log("Proposal executed.");
   } else {
+    // Hardcoding gas estimate on Rinkeby since it fails for an undetermined reason...
+    let opts = isRinkeby ? { gasLimit: 1000000 } : null
```
DanielVF, line 80: This changes stops `getTxOpts` from being used in the rest of the file, which means that we can't adjust gas pricing from the environment variable.

Perhaps something like: 

```
let opts = await getTxOpts();
if (isRinkeby){
  opts.gasLimit = 1000000;
}
```

Otherwise, we should remove the import for getTxOpts.

DanielVF (changes requested):

contracts/deploy/009_ousd_fix.js
franckc: Cleaned up. PTAL...

DanielVF (approved):

tomlinton (approved):

contracts/deploy/009_ousd_fix.js
tomlinton: Looks good! Did that hardhat global gas premium never work? Wish it did

franckc: It turned out that hardhat config is for gas limit and not for gas price... :)

PR #478, opened 2021-01-05
Add state of MixOracle to the hardhat debug task

franckc:
Example of output: https://gist.github.com/franckc/e43975d81c9a8715d6b1454a0e734ae2

Note: I could not dump any state for ChainlinkOracle since all the variables are private :(
But I did double check the 008_ousd_reset migration used for deploying and configuring ChainlinkOracle and that looks good to me.


DanielVF (approved):
Good to have this.

PR #481, opened 2021-01-06
Update 3pool strategy checkBalance to return assets as we can get them without a fee

tomlinton:

Contract change checklist:
  - [x] Unit tests pass


contracts/contracts/strategies/ThreePoolStrategy.sol
```diff
@@ -200,12 +200,15 @@ contract ThreePoolStrategy is InitializableAbstractStrategy {
...
         (, , uint256 totalPTokens) = _getTotalPTokens();
-        balance = 0;
-        if (totalPTokens > 0) {
-            balance += ICurvePool(platformAddress).calc_withdraw_one_coin(
-                totalPTokens,
-                poolCoinIndex
-            );
+        ICurvePool curvePool = ICurvePool(platformAddress);
+
+        uint256 pTokenTotalSupply = IERC20(assetToPToken[assetsMapped[0]])
```
franckc, line 205: Since we are using assetsMapped[0] rather than _asset, do we care that checkBalance would return an invalid result if it would be called with the address of an asset that this strategy does not support (but that Curve supports so that line 208 below does return a non zero value).

franckc (commented):
One question inline.

contracts/contracts/strategies/ThreePoolStrategy.sol
tomlinton: The Vault requires the asset to be supported before it calls checkBalance, but you could cause a revert by calling this method directly on the strategy. I cleaned it up a bit with a nicer error message, please take a look :)

franckc (approved):
Thanks for addressing my feedback. LGTM

DanielVF (approved):
#### Requirements

:white_check_mark: Yay! Moving away from calc_withdraw_one_coin is super important since it blocks 3Pool price curve bending attacks.

:white_check_mark: This method shown here should also mean that if we withdraw based on the balances shown, then we should get no withdrawal fees from 3pool.

:white_check_mark: This change should mean that our balance with 3Pool will only increase. This is a very good thing, and removes the "breathing" problems we were having before with 3Pool.

:memo: The usual recommendation for getting Curve totals is "use virtual_price". This method does not use virtual_price, and instead only looks at totals from 3pool. Given that we are not just trying to find our generic 3pool balance, but balance in a specific stablecoin, this makes sense. It will be important to verify that the rounding works out correctly, and the balances returned here can be actually be withdrawn.

:small_orange_diamond: I think withdrawing the reported amounts will fail if you:

- calc three checkBalance amounts()
- serially withdraw each stablecoin.

...since 3Pool will add fees for for each unbalanced transaction, making the last withdrawal fail because of insufficient tokens to meet the requested amount. 

:small_orange_diamond: An alternative approach, doing  (check balance, then withdraw) for one asset at a time will succeed, but withdraw much less than full total, because each check balance will still only report a proportion of the total balance.

:x: This allows a user to manipulate which stablecoins they receive when withdrawing, by flash loaning to the 3Pool contract, before redeeming. If a stable coin had lost it's peg, this could allow an attacker to receive less of that coin, and more of other, more valuable coins.

:x: The vault core only checks the balances for asset on a strategy if the strategy reports that it supports the asset. Currently the 3Pool strategy is written such that it only functions correctly if it supports one asset (the coin index is set for the entire contract, and bad things would happen if more assets were marked as supported).

If the vault is only checking one asset on this strategy, then it will only find out about a portion of the assets on the contract, and wrongly think that this contract has a massively lower balance of stablecoins than it actually does.

**THIS PRICING CHANGE PR IS NOT STANDALONE, AND WILL NOT WORK UNTIL THE OTHER CHANGES ARE IN.**

#### Internal State

This method does not change any internal state, and is correctly marked as view.

It depends only on the asset addresses, pool token addresses, and gauge address state. As long as these are correct, it should run.

#### Attack

If an attacker could execute a their own code in between checkBalance calls from the vault on multiple assets, they could falsely inflate or deflate our reported assets by shifting the proportions of coins on threepool between calls. This would not require reentrancy, and reentrancy protections would not help against it.

I think that this would require one of a malicious stablecoin, malicious 3Pool contract, malicious gauge, or somehow an attacker contract marked as a supported asset in both the vault and this contract. It's definitely good that we added the require for a supported asset. The vault currently checks totals for one strategy at a time. This means that the window of being able to manipulate the 3pool price differences is restricted to the code running in this strategy.

#### Logic

Logic is minimal, just some code to avoid divides by zero.

#### Tests

It doesn't seem like there are test for the new major difference in functionality between this method and the old? It might not make sense to have them now, given that this method is currently broken when standing alone, but the next multi-deposit PR needs to verify that this method works correctly. 

#### Flavor

I don't like the way we nest the success cases inside ever nesting if statements. I know we are doing this to avoid slither errors, but this is a case of making slither happy resulting in making our code harder to read and friendlier to bugs.

I'm going to go file a PR to see if they will consider excluding errors on checks against zero.

At first I didn't like the unused values from the multi-return but all those values are needed to compute the final value, and are used in other places in the code. Seems okay then.

#### Overflow

#### Black magic

_No black magic_

#### Authentication

_No auth needed since it's a view method._

#### Cryptographic code

_No Crypto_

#### Gas problems

-_No For loops_

When the vault wants to find out how much money it has, this function will make duplicate calls for every asset in the pool to check its own ptoken balance, the gauge balance, and the total supply of ptokens . This PR is fine to merge in, but there's definitely a bit of duplicate work being done.

#### External calls

Vulnerable to a malicious stablecoin, malicious 3Pool contract, or malicious gauge.

#### Ethereum

_Contract does not send or receive Ethereum._

PR #482, opened 2021-01-06
3pool multiple asset support and depositAll methods for strategies

tomlinton:
- Implements depositAll for strategies (deposits all assets held in the strategy contract into the platform)
- Updates 3pool so it can handle depositing/withdrawing multiple assets at the same time (each 3pool strategy previously only handled one asset and used the singular add/remove liquidity methods of Curve).
- Removes old 3pool strategies
- Updated the reallocate task to be more useful for testing

Example test of reallocation:

```
FORK=true npx hardhat reallocate --from 0xD5433168Ed0B1F7714819646606DB509D9d8EC1f --to 0xaB5E7B701B605f74AaC1b749Fd50715f0DEd7Bc5 --assets 0xdAC17F958D2ee523a2206206994597C13D831ec7 --amounts 1000000000000 --network localhost
```
This reallocates 1M USDT from Compound to 3pool on the fork (need fork node running as well of course). 

I need to get better at making smaller PRs. It was a bit of a tough one due to how the old 3pool strategies needed to be removed.

Contract change checklist:
  - [x] Unit tests pass


DanielVF:
I should finish reviewing this one on Monday.

contracts/contracts/strategies/ThreePoolStrategy.sol
```diff
@@ -271,4 +303,15 @@ contract ThreePoolStrategy is InitializableAbstractStrategy {
...
         pToken.safeApprove(crvGaugeAddress, uint256(-1));
     }
+
+    /**
+     * @dev Get the index of the coin in 3pool
+     */
+    function _getPoolCoinIndex(address _asset) internal view returns (int128) {
+        ICurvePool threePool = ICurvePool(platformAddress);
+        for (int128 i = 0; i < 3; i++) {
+            if (threePool.coins(uint256(i)) == _asset) return i;
```
DanielVF, line 313: Wonder how much extra gas this is going to cost us. We'll be making an external function call six times every time we go through all three coins to do something. I guess with just USDT/USDC this could either be three calls or five.

contracts/contracts/strategies/ThreePoolStrategy.sol
```diff
@@ -163,31 +203,23 @@ contract ThreePoolStrategy is InitializableAbstractStrategy {
...
         );
-        ICurvePool(platformAddress).remove_liquidity_one_coin(
-            pTokenBalance,
-            poolCoinIndex,
-            minWithdrawAmount
-        );
-        // Transfer the asset out to Vault
-        asset.safeTransfer(vaultAddress, asset.balanceOf(address(this)));
+        ICurvePool threePool = ICurvePool(platformAddress);
+        threePool.remove_liquidity(totalPTokens, minWithdrawAmount);
```
DanielVF, line 212: @tomlinton, I think `remove_liquidity` takes an array as its second argument. The function definition is:

`def remove_liquidity(_amount: uint256, min_amounts: uint256[N_COINS]):`

..see <https://github.com/curvefi/curve-contract/blob/master/contracts/pools/3pool/StableSwap3Pool.vy#L498>

contracts/contracts/strategies/ThreePoolStrategy.sol
```diff
@@ -163,31 +203,23 @@ contract ThreePoolStrategy is InitializableAbstractStrategy {
...
+        // Note that Curve will provide all 3 of the assets in 3pool even if
```
DanielVF, line 213: Good thinking. Good comment explaining why we are doing something out of our ordinary.

contracts/contracts/strategies/ICurvePool.sol
```diff
@@ -15,6 +15,8 @@ interface ICurvePool {
         uint256 _minAmount
     ) external;
 
+    function remove_liquidity(uint256 _amount, uint256 _minAmount) external;
```
DanielVF, line 18: Actual ABI is `def remove_liquidity(_amount: uint256, min_amounts: uint256[N_COINS])`, an array.

DanielVF (changes requested):
Two changes needed:

- We have no tests covering the deposit all functionality nor the 3pool withdraw all - we at a minimum need this for three pool since its a very different set of code than depositing one token.
- The function signature for 3pool remove_liquidity is wrong and will fail. Needs an array, not just a single uint256.

contracts/contracts/strategies/ThreePoolStrategy.sol
```diff
@@ -110,6 +107,48 @@ contract ThreePoolStrategy is InitializableAbstractStrategy {
...
+    /**
+     * @dev Deposit the entire balance of any supported asset into the Curve 3pool
+     */
+    function depositAll() external onlyVault nonReentrant {
+        uint256[3] memory _amounts = [uint256(0), uint256(0), uint256(0)];
+        uint256 depositValue = 0;
+        ICurvePool curvePool = ICurvePool(platformAddress);
+
+        for (uint256 i = 0; i < assetsMapped.length; i++) {
+            uint256 balance = IERC20(assetsMapped[i]).balanceOf(address(this));
```
franckc, line 119: assetsMapped[i] is referenced 4 times in the code below. My understanding is that lookups are pretty expensive in solidity.
Does the solidity compiler automatically optimize this and create temporary memory storage for these lookups or does it execute the lookup instructions 4 times? If the latter it would be worth creating a temporary variable to store assetsMapped[i] to minimize gas costs.

franckc (commented):
Took a super quick look and added minor comments. Will take another more in-depth look.

tomlinton:
This is still a draft PR. No objection to early reviews but I don't want you to get reviewer fatigue and not look as hard at the final version.

franckc:
> This is still a draft PR. No objection to early reviews but I don't want you to get reviewer fatigue and not look as hard at the final version.

ah sorry. I did not notice that it was sneaky @micahalcorn that added me as reviewer... :) I'll check with you next time.

micahalcorn:
Sorry, I didn't notice the draft status until after I assigned it and thought we were preparing to move forwward.

contracts/contracts/strategies/ThreePoolStrategy.sol
tomlinton: Good question, I created a temporary variable. Would be great to have an understanding of what optimisations the compiler might make here (if any).

contracts/contracts/strategies/ICurvePool.sol
tomlinton: Thanks dvf 👍

tomlinton:
@franckc @DanielVF this is ready now :)

tomlinton:
Some example reallocations:

1m USDC and 1m USDT from Compound to 3pool

```
contracts ❯ FORK=true npx hardhat reallocate --from 0xD5433168Ed0B1F7714819646606DB509D9d8EC1f --to 0xaB5E7B701B605f74AaC1b749Fd50715f0DEd7Bc5 --assets 0xdAC17F958D2ee523a2206206994597C13D831ec7,0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48 --amounts 1000000000000,1000000000000 --network localhost
failed to get chainId, falling back on net_version...
Vault totalValue():	 7226864.838926228432229787
From Strategy USDC:	 balance=2475411.562751
From Strategy USDT:	 balance=4299620.053038
To Strategy DAI:	 balance=0.0
To Strategy USDC:	 balance=0.0
To Strategy USDT:	 balance=0.0
Reallocating assets...
Vault totalValue():	 7226573.431599074659351926
From Strategy USDC:	 balance=1475413.50484
From Strategy USDT:	 balance=3299638.699043
To Strategy DAI:	 balance=664236.491281132639413398
To Strategy USDC:	 balance=704730.981714
To Strategy USDT:	 balance=630720.474045
```

1m USDC and 1m USDT from 3pool to Compound

```
Vault totalValue():	 7226573.431599074659351926
From Strategy DAI:	 balance=664236.491281132639413398
From Strategy USDC:	 balance=704730.981714
From Strategy USDT:	 balance=630720.474045
To Strategy USDC:	 balance=1475413.50484
To Strategy USDT:	 balance=3299638.699043
Reallocating assets...
Vault totalValue():	 7226527.511964712139421386
From Strategy DAI:	 balance=598010.374237398533800354
From Strategy USDC:	 balance=634128.901353
From Strategy USDT:	 balance=567497.434048
To Strategy USDC:	 balance=1575413.606494
To Strategy USDT:	 balance=3399638.973429
```

10k DAI from 3pool to Aave

```
contracts ❯ FORK=true npx hardhat reallocate --from 0xaB5E7B701B605f74AaC1b749Fd50715f0DEd7Bc5 --to 0x9f2b18751376cF6a3432eb158Ba5F9b1AbD2F7ce --assets 0x6b175474e89094c44da98b954eedeac495271d0f --amounts 10000000000000000000000 --network localhost
Vault totalValue():	 7226527.596451937154284853
From Strategy DAI:	 balance=598010.374237398533800354
From Strategy USDC:	 balance=634128.901353
From Strategy USDT:	 balance=567497.434048
To Strategy DAI:	 balance=411509.907651245742999413
Reallocating assets...
Vault totalValue():	 7226525.17383738941140461
From Strategy DAI:	 balance=594664.064149391617511209
From Strategy USDC:	 balance=630616.183039
From Strategy USDT:	 balance=564353.816677
To Strategy DAI:	 balance=421509.960819930450629008
```

franckc:
> @franckc @DanielVF this is ready now :)
ack - added to my TODO list...

DanielVF:
Guess we should bump up the deploy number to 14, now that 13 is used.

tomlinton:
Merged master and updated deploy name 👍

franckc (commented):
Took a pass at reviewing the code. Looks great.
Next I'm going to test on the fork...

contracts/contracts/strategies/AaveStrategy.sol
```diff
@@ -25,13 +25,34 @@ contract AaveStrategy is InitializableAbstractStrategy {
...
+        _deposit(_asset, _amount);
+    }
 
+    /**
+     * @dev Deposit asset into Aave
+     * @param _asset Address of asset to deposit
+     * @param _amount Amount of asset to deposit
+     * @return amountDeposited Amount of asset that was deposited
+     */
+    function _deposit(address _asset, uint256 _amount) internal {
```
franckc, line 37: Since the Aave and Compound strategies are being modified, we should add them to the deploy script?

contracts/contracts/vault/VaultAdmin.sol
```diff
@@ -221,9 +221,9 @@ contract VaultAdmin is VaultStorage {
             require(strategyTo.supportsAsset(_assets[i]), "Asset unsupported");
             // Withdraw from Strategy and pass other Strategy as recipient
             strategyFrom.withdraw(address(strategyTo), _assets[i], _amounts[i]);
-            // Tell new Strategy to deposit into protocol
-            strategyTo.deposit(_assets[i], _amounts[i]);
         }
+        // Tell new Strategy to deposit into protocol
+        strategyTo.depositAll();
```
franckc, line 226: We should include VaultAdmin in deploy 14

tomlinton: Oops yes I will need to update. I think when I wrote the deploy script I was thinking we could sneak it in with that other upgrade which had all these things. Will update 👍

franckc:
@tomlinton @DanielVF FYI deploy 14 should now work on the fork so you can use that for testing.

I did some basic testing and things look good but I'm going to do some more validation this morning.

franckc (approved):
Did a bunch of reallocate tests on the fork.

LGTM

DanielVF (commented):
#### Internal State

- [x] withdrawAll and depositAll use no internal state other deposit addresses. 

Our maxSlippage constant has me wondering about it. This constant is not used to check if total slippage is less than 1% of face value of the stablecoin, nor is it checking that prices have not changed since the user sent the transaction into the mempool. Instead it is confirming the prices/ratios that we just checked, inside this atomic transaction, still match. If I'm reading the code correctly, if the prices are even 0.001% off what we just calculated, then either our math is wrong or something is badly broken in the universe. We might be able to tighten this constant. Or maybe I'm totally off here.

#### Attack

1. This entire strategy is unsafe for the general public to call, since it's vulnerable to the usual 3pool attacks. We are planning on only having the strategist role move funds in and out of 3pool. It should never be a default strategy.

2. This code can also loose money if depositing in proportions that do not match 3Pool. On the plus side, this also gives us the ability to deposit correct proportions in a way that does not lose money, which we've not had before.

3. Because we don't have any slippage protection via the standard method of using expected numbers from before the transaction, we are vulnerable to front running / back running attacks. An attacker could see that we were about to exchange a lot of money, force 3pool out of line, let our transaction execute, then follow it up with another transaction to restore the balance and profit. 

This is a bit theoretical at the moment, with our current volume, and 3pool's current volume, but when OUSD takes over the world, it's a definite avenue to steal from this strategy, even with only a trusted address is moving funds in a trusted transaction.

(yearnDAI vault V1 lost 11 million dollars today because a 1% slippage protection to face value is still enough to get you looted. However, because of that 1% limit that they had, the attack was slowed down so that they were able pause during the attack and save 20 million.)

#### Logic

[x] Logic seems good

#### Tests

#### Flavor

_getPoolCoinIndex goes out of it's way to return an int128, however in all but one use,  it is  cast back to a uint256. Perhaps it could just return a normal uint256, and the one place that needs a int128 could cast. This would make the other rest of the code cleaner. I hate seeing ints around where it should never be negative.

#### Overflow

-  ❌Check that all for loops use uint256 (see comment under "flavor" about coin index)

#### Black magic

_No magic._

#### Authentication

Is there a reason why withdrawAll() uses onlyVaultOrGovernor ? Everything else uses onlyVault?

- [x] Never use tx.origin
- [x] Check that every external/public function should actualy be external

#### Cryptographic code

_No crypto._

#### Gas problems

We do make repeated external calls to coins() to find which 3pool address to use. This isn't very effiecent, but at least it's only in code that we will be calling manually.

#### External calls

- [x] No user supplied addresses are called
- [x] Rentrancy gaurds on all state changing functions

#### Ethereum

mikeshultz:
> This code can also loose money if depositing in proportions that do not match 3Pool. On the plus side, this also gives us the ability to deposit correct proportions in a way that does not lose money, which we've not had before.

Could this same mechanism be used in the opposite way?  For example, 3pool's "lowest" asset right now is USDT.  If we favor USDT when depositing into 3pool, it would give us a bit of an edge.

DanielVF:
@mikeshultz, yes indeed, it can!

mikeshultz:
Been using this implementation branch as a reference for the upcoming monitoring changes and came across this.  What's the deal with minting to the Gauge?  Unless the "minter" is something other than the CRV token(it's the only thing I found with `mint()` in their GH), doesn't this just drop "free" tokens into the Gauge contract?

https://github.com/OriginProtocol/origin-dollar/pull/482/files#diff-0a11ceb193ed3d9b44b477cb5a1ea9ab91723c0561ba1e7f852c83ecd6bcc1f0R70

Looks like `deposit()` would need to be called.  Though maybe I'm mixing up rewards tokens or the minter is something I haven't found yet?

And as a clarifying question, are we depositing into the Gauge for "boost"?

micahalcorn:
> And as a clarifying question, are we depositing into the Gauge for "boost"?

Not yet @mikeshultz, but see Matt's [Google Doc](URL). We've discussed it in Discord and our meeting also.

tomlinton:
> Been using this implementation branch as a reference for the upcoming monitoring changes and came across this. What's the deal with minting to the Gauge? Unless the "minter" is something other than the CRV token(it's the only thing I found with `mint()` in their GH), doesn't this just drop "free" tokens into the Gauge contract?
> 
> https://github.com/OriginProtocol/origin-dollar/pull/482/files#diff-0a11ceb193ed3d9b44b477cb5a1ea9ab91723c0561ba1e7f852c83ecd6bcc1f0R70
> 
> Looks like `deposit()` would need to be called. Though maybe I'm mixing up rewards tokens or the minter is something I haven't found yet?
> 
> And as a clarifying question, are we depositing into the Gauge for "boost"?

To receive your rewards token you need to call the `mint` function on the CRV minter with the address of the Gauge you are staking 3CRV (or other LP token) in FROM the contract/address that is staking. The minter:

- Updates checkpoints on the Gauge to finalise the amount of CRV the account is eligible for
- Mints the appropriate amount of CRV
- Transfers the CRV to the contract/address

The CRV minter is here https://etherscan.io/address/0xd061D61a4d941c39E5453435B6345Dc261C2fcE0#code with the guts being in the `mint` and `_mint_for` functions.

And per @micahalcorn we aren't yet doing any boosting.

Hope that helps, ping me on Discord if you got any questions.

tomlinton:
>  Each logical branch has a test
 Edge conditions are tested

This a call for my tests? :) I'll try hunt down the untested branches.

> _getPoolCoinIndex goes out of it's way to return an int128, however in all but one use, it is cast back to a uint256. Perhaps it could just return a normal uint256, and the one place that needs a int128 could cast. This would make the other rest of the code cleaner. I hate seeing ints around where it should never be negative.

Will see what I can do to clean this up!

> Is there a reason why withdrawAll() uses onlyVaultOrGovernor ? Everything else uses onlyVault?

There is no bare way to call `withdrawAll` from the Vault on a Strategy. The intent was that if we needed to call `withdrawAll` it would be called directly on the Strategy rather than via the Vault. The Vault has permission to call it as part of `removeStrategy` though, where the Strategy is deprecated, which requires withdrawal of everything before its removed from the list of approved strategies.

> We do make repeated external calls to coins() to find which 3pool address to use. This isn't very effiecent, but at least it's only in code that we will be calling manually. :(

We could potentially hardcode these address/indices, or pass them in the constructor. The dynamic lookup always makes testing easier, but I'll see if I can clean it up.

tomlinton:
@mikeshultz This is helpful for understanding Curve https://medium.com/better-programming/how-to-integrate-the-curve-fi-protocol-into-your-defi-protocol-e1d4c43f716d.

tomlinton:
Updates:

- Since we can support all 3 coins, just pass them in the correct order for 3pool and use that index to replace the `coins(i)` lookup from Curve
- Made the casting on _getPoolCoinIndex better

DanielVF (approved):
Looks good.

franckc (approved):
still LGTM - @tomlinton feel free to merge in master and I can start on the deploy today...

Issue #445, opened 2020-12-18
Echidna failure


DanielVF:
Got an echidna failure somehow. Will need to dig into it later.

Analyzing contract: contracts/contracts/crytic/PropertiesOUSDTransferable.sol:PropertiesOUSDTransferable
crytic_approve_overwrites: passed! 🎉
crytic_self_transferFrom_to_other_ERC20PropertiesTransferable: failed!💥  
  Call sequence:
    initialize("&\214\&4#\216)?P1\170\141\254\162.\227\142\153BiZ\209\DC3\vqC9)\NAK\ETX\212$","\251\f\b\251\139x\n]$!\249\207_______\205\194\239\128p\SYN\255",0xf17f52151ebef6c7334fad080c5704d77216b732) from: 0x627306090abab3a6e1400e9345bc60c78a8bef57
    mint(0xf17f52151ebef6c7334fad080c5704d77216b732,6) from: 0xf17f52151ebef6c7334fad080c5704d77216b732 Time delay: 363058 seconds Block delay: 6382
    changeSupply(599290589) from: 0xf17f52151ebef6c7334fad080c5704d77216b732 Time delay: 40014 seconds Block delay: 39046

crytic_zero_always_empty_ERC20Properties: passed! 🎉
crytic_self_transfer_ERC20PropertiesTransferable: passed! 🎉
crytic_self_transferFrom_ERC20PropertiesTransferable: passed! 🎉
crytic_revert_transferFrom_to_zero_ERC20PropertiesTransferable: passed! 🎉
crytic_revert_transfer_to_user_ERC20PropertiesTransferable: passed! 🎉
crytic_revert_transfer_to_zero_ERC20PropertiesTransferable: passed! 🎉
crytic_transfer_to_other_ERC20PropertiesTransferable: failed!💥  
  Call sequence:
    initialize("&\214\&4#\216)?P1\170\141\254\162.\227\142\153BiZ\209\DC3\vqC9)\NAK\ETX\212$","\251\f\b\251\139x\n]$!\249\207_______\205\194\239\128p\SYN\255",0xf17f52151ebef6c7334fad080c5704d77216b732) from: 0x627306090abab3a6e1400e9345bc60c78a8bef57 Time delay: 411308 seconds Block delay: 14451
[...]

PR #489, opened 2021-01-08
Fix for rebasing when the contract state of an address changes

crazybuster:

Contract change checklist:
  - [x] Unit tests pass


nick (approved):

DanielVF (approved):
#### Requirements

We have added a check to disallow addresses that appear to be EOAs, but have non-rebasing accounting setup without having explicitly opted into it.

This should only happend if a contract was created (or recreated) via CREATE2 after that address had already been used with OUSD AND the call is coming during a constructor.

#### Logic

Logicly we first check for if the address has chosen to be treated a certain way. If so, we use it, and the accounting is handled elsewhere.

If the address has not chosen to be treated a certain way, we choose based on the type of address. If it's a contract, we also check and handle a needed migration. If it does not appear to be a contract, but was previously holding money as a contract then we throw.

#### Tests

#### Overflow

_No change to math._

#### Black magic

- [x] Address.isContract - Does use it. Code handles the cases that the the same address could be marked as a contract or not and that could switch between calls.

#### Authentication

_No authentication._

#### Cryptographic code

_No crypto._

#### Gas problems

This change should slightly decease gas costs for contracts that have not opted into rebasing. It should slightly (900?) increase the cost for contracts that are not rebasing.

_No for loops._

#### External calls

_No external calls._

#### Ethereum

_Does not send or receive Ethereum._

franckc:
I ran the echidna tests and they passed except one test case. Which is a known issue documented in #445

PR #467, opened 2020-12-30
Fix OUSD storage

tomlinton:
Use arrays to deprecate old storage. Includes upgrade script. Had to flick some variables from private to public to allow the reset to work, but we can clean that up after.

Contract change checklist:
  - [ ] Code reviewed by 2 reviewers
  - [x] Unit tests pass
  - [ ] Slither tests pass with no warning
  - [x] Echidna tests pass (not automated, run manually on local)

PR #493, opened 2021-01-09
🏰  Simplification of _isNonRebasingAccount()

DanielVF:
After some thought, there's an **epic simplification** of `_isNonRebasingAccount()`. The reason functions call `_isNonRebasingAccount()` is to figure out what kind of accounting the account is using so that the calling function can do the correct accounting operations.

Getting the returned accounting style wrong nukes OUSD. It's absolutely important that the value returned from this function matches the actual accounting style used by the address.

In the past, `_isNonRebasingAccount()` has used two bools and one three-value, for twelve total possible states, and with those attempts to predict what `_isNonRebasingAccount()` has done in the past, and what the rest of the contract has done in the past, in order to return the current accounting style to for the account. This is complicated.

It is also very brittle - if logic elsewhere chances, or if something unexpected happens _*cough, CREATE2*_, then this function can  be disastrously wrong.

In reality, this method is super simple. To always correctly return the current accounting style.... we just need to return the current accounting style!

(Plus a little extra if statement to migrate newly interacting contracts to be non-rebasing.)

Contract change checklist:
  - [x] Unit tests pass

### DIFF
--- a/contracts/contracts/token/OUSD.sol
+++ b/contracts/contracts/token/OUSD.sol
@@ -361,44 +361,18 @@ contract OUSD is Initializable, InitializableERC20Detailed, Governable {
     }
 
     /**
-     * @dev Is an account's balance non-rebasing, i.e. does not alter with rebases.
-     *      By default contracts are non-rebasing and EOA's are rebasing.
-     *      Or the address could have chosen to override this.
+     * @dev Is an account using rebasing accounting or non-rebasing accounting?
+     *      Also, ensure contracts are non-rebasing if they have not opted in.
      * @param _account Address of the account.
      */
     function _isNonRebasingAccount(address _account) internal returns (bool) {
-        RebaseOptions accountRebaseState = rebaseState[_account];
-        if (accountRebaseState == RebaseOptions.OptIn) {
-            // The address has chosen to be rebasing.
-            return false;
-        } else if (accountRebaseState == RebaseOptions.OptOut) {
-            // The address has chosen to be non-rebasing.
-            return true;
-        } else {
-            // The address has not chosen explicitly, so use the default
-            // for its type.
-            if (Address.isContract(_account)) {
-                // Contracts default to be non-rebasing.
-
-                // Migrate if needed, making sure the rebasing/non-rebasing
-                // supply is updated and fixed credits per token is set
-                // for this account.
-                _ensureRebasingMigration(_account);
-                return true;
-            } else {
-                // User accounts default to be rebasing.
-
-                // Disallow contracts from interacting with OUSD if the following
-                // sequence has occurred: The contract has self-destructed, been
-                // recreated at the same address with CREATE2, and then is interacting
-                // with OUSD again from the contract's constructor.
-                require(
-                    nonRebasingCreditsPerToken[_account] == 0,
-                    "Previous Contract"
-                );
-                return false;
-            }
+        if (
+            Address.isContract(_account) &&
+            rebaseState[_account] == RebaseOptions.NotSet
+        ) {
+            _ensureRebasingMigration(_account);
         }
+        return nonRebasingCreditsPerToken[_account] > 0;
     }
 
     /**
--- a/contracts/contracts/token/OUSD.sol
+++ b/contracts/contracts/token/OUSD.sol
@@ -366,10 +366,8 @@ contract OUSD is Initializable, InitializableERC20Detailed, Governable {
      * @param _account Address of the account.
      */
     function _isNonRebasingAccount(address _account) internal returns (bool) {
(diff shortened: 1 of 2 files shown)


tomlinton:
This seems good, but I won't be able to give it a proper review until I get back. Sorry 😬 I think this whole thing is a lot more complex than it needs, because when it was written it was for migrating existing balances. We no longer should have that problem since we reset everything.

If you guys are comfortable with it and do merge/deploy it before I get back could I ask you to do a minor cleanup on https://github.com/OriginProtocol/origin-dollar/blob/803bcdc45e548753fa158fc215eb32e9d5d9c06f/contracts/contracts/utils/InitializableERC20Detailed.sol

The variables were made public for the reset, so its generating duplicate getters for everything. We should switch them back to private. Also `_creditBalances` in OUSD.sol, more info here https://github.com/OriginProtocol/origin-dollar/pull/467/files.

You can see the duplicate getters https://etherscan.io/address/0x2A8e1E676Ec238d8A992307B495b45B3fEAa5e86#readProxyContract.

DanielVF:
@tomlinton Would love if you can think through if this makes sense.

tomlinton (approved):

tomlinton:
Makes sense to me!

DanielVF:
Pinging the mighty @franckc to have a look.

mikeshultz (commented):
Appreciate the simplification here.  Still wonder if a migration operation should be outside of a boolean-sounding function but that's more style than substance.

One thing I think might be an issue is if a contract tries to opt in before they have a balance.  I think [this require statement](https://github.com/OriginProtocol/origin-dollar/blob/54ccc5919531a62f62a7ce4c4220db6d7ec07949/contracts/contracts/token/OUSD.sol#L428) would fail in that case.  Perhaps that's not behavior we care about, and in that case this LGTM but figure I'd bring it up.

DanielVF:
Good noticing. This shouldn't care if the contract has not received a payment yet, since the migration will convert it to non-rebasing first, before it is converted back. What is prevented, I think, is a contract setting an opt-in during its constructor. I don't think this is a change to the current behavior though?

mikeshultz:
Doens't really matter if it's within a constructor or if it's done sequentially.  But if they try opt in -> transfer in, it would still catch this.

I think this is a change on what's in `master` right now.  Before, a contract when trying to optIn would fall within this branch:

https://github.com/OriginProtocol/origin-dollar/blob/54ccc5919531a62f62a7ce4c4220db6d7ec07949/contracts/contracts/token/OUSD.sol#L383-L390

Since it would be identified as a contract, though no actual balance would be migrated(since it's `0`).  With the proposed implementation, it would return `false`.

It's probably an edge case, though.

mikeshultz (approved):
That was my misunderstanding of balance storage versus CPT storage.  LGTM!

DanielVF:
Awesome! Great to push back when something doesn't make sense!

franckc:
Sorry I haven't reviewed yet but will do before we deploy.

@crazybuster do you mind taking a look as well - this is a sensitive area of the code and it would not hurt to have another pair of eyes scrutinizing this change.

franckc:
- reviewed the change
 - reviewed and re-ran the unit test

Overall LGTM.

The one comment I have is that the method _ensureRebasingMigration is only called from _isNonRebasingAccount.
So we could probably get rid of _ensureRebasingMigration and just put the code inline, within the _isNonRebasingAccount method.

CC @DanielVF @tomlinton Should we do this before the deploy this week?

contracts/contracts/token/OUSD.sol
```diff
@@ -361,44 +361,16 @@ contract OUSD is Initializable, InitializableERC20Detailed, Governable {
...
-                _ensureRebasingMigration(_account);
-                return true;
-            } else {
-                // User accounts default to be rebasing.
-
-                // Disallow contracts from interacting with OUSD if the following
-                // sequence has occurred: The contract has self-destructed, been
-                // recreated at the same address with CREATE2, and then is interacting
-                // with OUSD again from the contract's constructor.
-                require(
```
franckc, line 395: Hey @DanielVF @tomlinton @mikeshultz - I was chatting with @crazybuster and he rightly pointed out that the simplified code won't allow us to abort on "questionable" transactions as we used to. Just want to make sure we are all ok with that...

DanielVF: Yeah. I'm very happy about not reverting. Transfer should just transfer money.

mikeshultz: I thought the revert was more of a short-term stop-gap and wasn't intended to be left permanently.

PR #497, opened 2021-01-12
Add setMaxSupplyDiff action to the propose script

franckc:
I'll use this script to generate the gov proposal for setting setMaxSupplyDiff to 5% (5e16 at precision 18) on the Vault.


franckc:
The governance proposal was executed and verified on Mainnet.

Merging this in. Happy to address any feedback in a subsequent PR.

PR #499, opened 2021-01-13
🏰  Certora-02 Don’t allow transfer out of non-supported assets.

DanielVF:
This prevents the governance/timelock from transferring out supported tokens from the vault. While there certainly are other ways governance could be malicious, we might was well have this one closed.

Flagged by Certora human auditing.

Contract change checklist:
  - [x] Unit tests pass

### DIFF
--- a/contracts/contracts/vault/VaultAdmin.sol
+++ b/contracts/contracts/vault/VaultAdmin.sol
@@ -283,6 +283,7 @@ contract VaultAdmin is VaultStorage {
         external
         onlyGovernor
     {
+        require(!assets[_asset].isSupported, "Only unsupported assets");
         IERC20(_asset).safeTransfer(governor(), _amount);
     }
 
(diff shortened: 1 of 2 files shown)


DanielVF:
I'm really confused by what the tests are doing here. When I run:
  
  - only the new test - PASS
  - the entire file - PASS
  - all tests - FAIL
  
  Our tests are supposed to be separate from each other. But here it's loading from the main fixture and yet getting different behaviors presumably based on other code having run.

tomlinton:
Yep the hardhat snapshot revert doesn't seem to work properly. It has always been like that which is why there are `z_` prefixes on some of the files. It's an issue somewhere in `ethereum-waffle`. I dug into it but couldn't resolve it. It looks like it did the snapshotting correctly so I wonder if it goes deeper to the evm implementation.

tomlinton:
Fix for this here https://github.com/OriginProtocol/origin-dollar/pull/520 :)

DanielVF:
Thanks @tomlinton, that fixed the tests.

franckc (approved):
Good safety measure.

LGTM

tomlinton (approved):

PR #501, opened 2021-01-13
🏰 Certora-03 Zero amount burn should not change state

DanielVF:
A zero amount burn combined the anti-dust code in burn can result in the user's balance decreasing. While the numbers involved would need to be crazy tiny, it makes sense to keep this clean and make sure a zero burn does nothing.

Found by Certora's automated prover.

Contract change checklist:
  - [x] Unit tests pass


tomlinton (approved):

franckc (approved):
LGTM. Thanks for fixing. Let's merge this in and then separately we'll figure our when we deploy all those small contract refinements.

PR #505, opened 2021-01-14
Update accounts tasks

franckc:
Added a new mint task to hardhat and cleaned up the fund task.

Example that funds the first test account with $1M in each stable coin, then mints OUSD:
```
FORK=true npx hardhat fund --num 1 --amount 1000000 --network localhost
FORK=true npx hardhat mint --num 1 --amount 1000000 --network localhost
```

### DIFF
--- a/contracts/tasks/account.js
+++ b/contracts/tasks/account.js
@@ -4,6 +4,18 @@ const daiAbi = require("../test/abi/erc20.json");
 const tusdAbi = require("../test/abi/erc20.json");
 const usdcAbi = require("../test/abi/erc20.json");
 
+// By default we use 10 test accounts.
+const defaultNumAccounts = 10
+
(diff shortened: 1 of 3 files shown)


sparrowDom (approved):
This is super convenient thanks!

contracts/tasks/account.js
```diff
@@ -4,6 +4,18 @@ const daiAbi = require("../test/abi/erc20.json");
...
 const usdcAbi = require("../test/abi/erc20.json");
 
+// By default we use 10 test accounts.
+const defaultNumAccounts = 10
+
+// The first 4 hardhat accounts are reserved for use as the deployer, governor, etc...
+const defaultAccountIndex = 4
+
+// By default, fund each test account with 10k worth of each stable coin.
+const defaultFundAmount = 10000
```
tomlinton, line 14: I think these need to be strings otherwise it fails to parse

PR #511, opened 2021-01-20
Add custom collectRewardToken for Compound strategy

### LINKED ISSUE #506: CompoundStrategy does not claim COMP tokens automatically
Due to a recent governance change to the Compound protocol (see https://compound.finance/governance/proposals/33), COMP tokens are not automatically claimed anymore. The [claimComp](https://github.com/compound-finance/compound-protocol/blob/master/contracts/Comptroller.sol#L1202-L1248) method needs to be called. Note that claimComp is a public method and can be called by anyone.

We should decide if we want to automatically call claimComp during a vault's allocate (drawback is extra gas), or if we just rely on calling it manually from times to time.

tomlinton:
Update to support change in Compound governance requiring us to manually claimComp.

Note the `liquidationThreshold` parameter will no longer work due to the COMP no longer accruing. I haven't investigated if there is another way to get the COMP balance of our contract. The Compound strategy was the only one that the parameter was effective for so we may need to remove the `liquidationThreshold` logic altogether if it is no longer useful.

Closes #506.

Contract change checklist:
  - [x] Unit tests pass


contracts/contracts/strategies/CompoundStrategy.sol
```diff
@@ -14,6 +15,21 @@ import {
...
+     */
+    function collectRewardToken() external onlyVault nonReentrant {
+        // Claim COMP from Comptroller
+        ICERC20 cToken = _getCTokenFor(assetsMapped[0]);
+        IComptroller comptroller = IComptroller(cToken.comptroller());
+        comptroller.claimComp(address(this));
+        // Transfer COMP to Vault
+        IERC20 rewardToken = IERC20(rewardTokenAddress);
+        uint256 balance = rewardToken.balanceOf(address(this));
+        emit RewardTokenCollected(vaultAddress, balance);
```
franckc, line 29: Should we have a conditional on balance > 0 so that we don't emit the event / do the transfer for zero amount?

franckc: Separately, it seems it would be useful interpreting those events to include the address of the reward token in the event. It can be figured out by looking up strategy.rewardToken, but still.
Anyway, not a blocker for this PR. More of a possible tweak for future.

franckc (approved):
One comment inline otherwise LGTM.

I'd be fine with leaving the liquidationThreshold code for now since it may be useful for future strategies.

Are we concerned at all by the additional gas cost this will cause on all the allocates?

Also, after we deploy this change to mainnet, we should remember to issue a governance proposal to set the liquidationThreshold to zero on the Compound strategy.

contracts/contracts/strategies/CompoundStrategy.sol
tomlinton: #519

DanielVF (approved):
#### Requirements

Compound no longer automatically gives you a rewards token balance automatically, instead you must claim COMP to see your balance increase. 

:small_orange_diamond: This change will make every large allocate claim COMP, transfer COMP, and sell COMP. That's certainly better than not automatically claiming COMP at all, but does increase the gas cost quite bit.

#### Internal State

_Does not alter any internal state. Uses stored addresses._

#### Attack

Baring an address being set wrongly, or malicious behavior from a trusted contract, I don't think this is attackable.

#### Logic

Yay! No if statements, executes straight on through.

#### Tests

🔸 No tests for verifying this code's actual functionality. I guess this code is entirely dependent on external contracts, and there's not much we could actually test here.

#### Flavor

Awesomely simple!

#### Overflow

_No math!_

#### Black magic

_No black magic._


#### Authentication

_No authentication._

#### Cryptographic code

_No cryptographic code._

#### Gas problems

See notes above in "Requirements".

#### External calls

_All calls are to trusted contracts._

#### Ethereum

_Contract does not send or receive Ethereum._

PR #516, opened 2021-01-20
Update PR template with Echidna note

tomlinton:
Added some wording to the PR template. We only have Echidna properties for OUSD right now, so any changes that aren't to that contract won't change the Echidna test results.


franckc (approved):
LGTM

PR #520, opened 2021-01-22
Fix load fixture issue

### LINKED ISSUE #117: Waffle fixtures don't revert properly
I had to commend out these tests https://github.com/oplabs/origin-dollar/blob/fee4ba2557589cb87a777de207769b02827d83f1/contracts/test/vault.js#L384-L402

Waffle doesn't seem to be reverting these [fixtures](https://ethereum-waffle.readthedocs.io/en/latest/fixtures.html) as the state is hanging around. It might be something to do with the proxy setup and the vault implementation being replaced by the mock vault.

tomlinton:
This should fix #117. Not using loadFixture when calling loading defaultFixture from the other fixtures was a bad idea. This mostly resolves the problem except mockvault still seems to cause issues if it runs before the others (something to do with it being a proxy upgrade?). 

Contract change checklist:
  - [x] Unit tests pass


DanielVF (approved):
Makes sense!

franckc (approved):
Thanks for fixing this. LGTM

PR #528, opened 2021-01-26
Allow strategist to call harvest

DanielVF:
The harvest() method on vaultCore should also be able to be called by the strategist role. This a method that should not be able to cause the contract to lose current funds, and it something that occasionally needs to be called and does not need to go through the entire governance process.

Contract change checklist:
  - [x] Unit tests pass

### DIFF
--- a/contracts/contracts/vault/VaultAdmin.sol
+++ b/contracts/contracts/vault/VaultAdmin.sol
@@ -12,12 +12,12 @@ import { IUniswapV2Router } from "../interfaces/uniswap/IUniswapV2Router02.sol";
 
 contract VaultAdmin is VaultStorage {
     /**
-     * @dev Verifies that the caller is the Vault or Governor.
+     * @dev Verifies that the caller is the Vault, Governor, or Strategist.
      */
-    modifier onlyVaultOrGovernor() {
+    modifier onlyVaultOrGovernorOrStrategist() {
         require(
-            msg.sender == address(this) || isGovernor(),
-            "Caller is not the Vault or Governor"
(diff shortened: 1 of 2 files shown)


DanielVF:
Note sure why this test is failing (at least failing locally).

contracts/contracts/vault/VaultAdmin.sol
```diff
@@ -304,7 +304,7 @@ contract VaultAdmin is VaultStorage {
      */
     function harvest(address _strategyAddr)
         external
-        onlyVaultOrGovernor
+        onlyVaultOrGovernorOrStrategist
```
franckc, line 307: We should apply the same modifier to the other harvest method above (the one without a strategy address).

BTW, I'm not sure why currently those 2 methods have different modifiers? onlyGovernor for one, onlyVaultOrGovernor for the second. I think it was probably an oversight and we forgot to update both?

DanielVF: The VaultCore makes an external call into the VaultAdmin to call the harvest(address) method, which is why harvest(address) has the additional vault access permission.

DanielVF:
Thanks @franckc for pointing out the two harvests! Updated so I'm now changing the permissions on both, and that fixed the failing test.

contracts/contracts/vault/VaultAdmin.sol
franckc: Ah yes makes sense.

franckc (approved):
Thanks for making this change. LGTM.

tomlinton (approved):

PR #529, opened 2021-01-26
Sparrow dom/smoke tests

sparrowDom:
This PR creates a framework for executing smoke tests. To achieve this: 
- a node:fork is ran and the script waits for it to be ready 
- then all tests are ran from the `scripts/test/smokeTest` folder. Each test has before & after deploy functions
- once all before deploy tests execute the contract deploy script is ran. Specified by the `deployid` parameter or if left blank, the last deploy script is used.
  - A small change is needed in deployments file, where hardhat RE is passed to the main function. See `011_ousd_fix.js` for an example
- Once the contract deploy is done all the after deploy functions execute and perform required checks

Example how to run smoke tests: 
`scripts/test/smokeTest.sh --deployid 11`

Ideas for improvement: 
- Make it integratable with CI: 
  -  fetch one of the latest block numbers and run hardhat node with that env variable
  - use `deployments/mainnet/.migrations` (thanks for the idea @franckc) to figure out which contract deployments haven't yet been ran. And run smoke tests against those.

Contract change checklist:
  - [ ] Unit tests pass


contracts/deploy/011_ousd_fix.js
```diff
@@ -62,19 +60,21 @@ const fixOUSD = async () => {
...
+  }
+  await fixOUSD(hre);
   console.log(`${deployName} deploy done.`);
   return true;
 };
 
 main.id = deployName;
 main.dependencies = ["002_upgrade_vault", "003_governor", "008_ousd_reset"];
-main.skip = () => !isMainnetOrRinkebyOrFork;
+//main.skip = () => !isMainnetOrRinkebyOrFork;
```
sparrowDom, line 78: Will remove this line. It is here just so that this thing runs on fork

tomlinton:
Looks cool!

What is going on in the beforeDeploy afterDeploy here? Is this just an example of how you might validate minting/redeeming?

contracts/tasks/smokeTest.js
```diff
@@ -0,0 +1,71 @@
...
+
+async function smokeTest(taskArguments, hre) {
+  const deployId = taskArguments.deployid
+  let deployScript
+
+  if (deployId) {
+    const scripts = await getDeployScripts()
+    deployScript = scripts[parseInt(deployId)]
+  } else {
+    deployScript = await getLastDeployScript()
```
franckc, line 41: It would be useful to be able to run smoke tests without doing a deploy.

For example, when I send a governance proposal to change a setting on the vault or a strategy, I'd like to be able to verify that it does not break anything on the fork.

So I would suggest that if no deployId is passed as an option on the command line, then we don't run any deploy. This way, I could:
 - start the fork
 - execute a governance proposal on the form manually (by running side script)
 - run the smoke test without passing a deployId

contracts/scripts/test/smokeTest/mintRedeemTest.js
```diff
@@ -0,0 +1,94 @@
+const { fund, mint } = require('../../../tasks/account');
```
franckc, line 1: As opposed to putting those tests under contracts/scripts/test/smokeTest, I wonder if we should less deep structure to make them more easily discoverable. And put them under contracts/smoke. I don't feel strongly. Just providing a possible alternative.

contracts/scripts/test/smokeTest.sh
```diff
@@ -0,0 +1,32 @@
...
+        printf "\n"
+        echo "$newLine Node failed to initialize in $nodeWaitTimeout seconds"
+        exit 1
+      fi
+    done
+    printf "\n"
+    echo "🟢 Node initialized running smoke tests"
+
+    FORK=true npx hardhat smokeTest --network localhost "$@"
+}
```
franckc, line 30: Shouldn't the script kill the NODE_PID otherwise the process will keep running in the background?

As a side node, I don't necessarily see myself needing this script. I'm perfectly fine running the fork node on a terminal (after having set BLOCK_NUMBER env var) and then running the smokeTest task in a separate terminal.

contracts/scripts/test/smokeTest.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+trap "exit" INT TERM ERR
+trap "kill 0" EXIT
+nodeWaitTimeout=60
+
+main()  
+{
+    nodeOutput=$(mktemp "${TMPDIR:-/tmp/}$(basename 0).XXX")
+    yarn run node:fork &> $nodeOutput &
```
franckc, line 9: Don't we need to set BLOCK_NUMBER here? Though see my comment below regarding not necessarily needing this script.

contracts/deploy/011_ousd_fix.js
```diff
@@ -62,19 +60,21 @@ const fixOUSD = async () => {
     log("Upgraded OUSD to new implementation");
   }
 
-  console.log(`${deployName} deploy done.`);
```
franckc, line 65: nit: Why removing this line? Not a big deal but I found it useful to know when a deployment is finished running.

franckc (approved):
Nice! Thanks for getting this done. A few comments/ideas inline.

contracts/hardhat.config.js
```diff
@@ -120,6 +123,11 @@ task(
   "Claims the OGN part of the compensation plan for all eligible users"
 ).setAction(claimOGN);
 
+// Smoke tests
+task("smokeTest", "Execute smoke test after applying the deployment script on the mainnet:fork network")
```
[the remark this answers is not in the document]
sparrowDom: yup thanks correcting 👍

contracts/scripts/test/smokeTest/mintRedeemTest.js
```diff
@@ -0,0 +1,94 @@
...
+  }, hre)
+
+  const usdtAfterMint = await getUsdtBalance()
+  const ousdAfterMint = await getOusdBalance()
+
+  const expectedUsdt = usdtBeforeMint.sub(usdtUnits(usdtToMint))
+  if (!usdtAfterMint.eq(expectedUsdt)) {
+    throw new Error(`Incorrect usdt value. Got ${usdtAfterMint.toString()} expected: ${expectedUsdt.toString()}`);
+  }
+
```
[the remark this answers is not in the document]
sparrowDom: yup had that listed as todo in the code. Since because of the fluctuating value of the stablecoins there needs to be some tolerance applied to the OUSD check. Adding that now.

contracts/scripts/test/smokeTest/mintRedeemTest.js
```diff
@@ -0,0 +1,94 @@
...
+    num: 1,
+    amount: '500'
+  }, hre)
+
+  const ousdAfterMint = await getOusdBalance()
+
+  if (!beforeDeployData.ousdAfterMint.eq(ousdBeforeMint)) {
+    throw new Error(`Deploy changed the amount of ousd in user's account from ${utils.formatUnits(beforeDeployData.ousdAfterMint, 18)} to ${utils.formatUnits(ousdBeforeMint, 18)}`)
+  }
+
```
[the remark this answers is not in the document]
sparrowDom: done 👍

contracts/scripts/test/smokeTest/mintRedeemTest.js
```diff
@@ -0,0 +1,94 @@
...
+  const expectedOusd = ousdAfterMint.sub(ousdToRedeem)
+  if (!expectedOusd.eq(ousdAfterRedeem)) {
+    throw new Error(`Incorrect OUSD amount after redeem. Expected: ${expectedOusd.toString()} got: ${ousdAfterRedeem.toString()}`)
+  }
+
+  /* TODO: 
+   * - check that ousdAfterMint is withing the allowed tolerance (1%-2%)
+   * - check that stablecoin balance is withing allowed tolerance after redeem
+   * - check that OUSD transfer works
+   */
```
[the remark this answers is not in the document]
sparrowDom: I've increased the amount to mint so that rebase is triggered (current threshold is 1000 units). But if I increase it to the level so that allocate is triggered (25k threshold) the mint call times out on the forked network. 

Added a mint multiple test, where all 3 stable coins are used.

contracts/scripts/test/smokeTest.sh
sparrowDom: The top lines take care of child processes being killed when the main process is terminated: 
```
trap "exit" INT TERM ERR
trap "kill 0" EXIT
```
added that as a comment to this script. 


I created this as a single process script so that we can in future more easily include it into the CI. With the logic that smoke tests are one process running that either exits with 0 (all fine) or an error (something wrong). And that results can be reproducible and consistent between runs.

contracts/scripts/test/smokeTest.sh
sparrowDom: Yeah from my tests it seems to run fine without it.

contracts/deploy/011_ousd_fix.js
sparrowDom: This line was duplicated. Inside this function and also inside the main function.

contracts/scripts/test/smokeTest/mintRedeemTest.js
sparrowDom: Good point moving it to `smoke`

contracts/tasks/smokeTest.js
sparrowDom: If we are changing settings on the vault on the mainnet, wouldn't it make sense that we include all those governance / contract proposals into the deploy scripts? This way dev environment should more closely match the production one.

Besides the above point, would it be ok for you if when there is no `deployId` the node just runs in the background as it is running now. And the shell is waiting for user input to run smoke tests. So the usage should be: 
- run the smokeTest script without `deployid` param
- once node is ready use `hardhat console` command to connect to the node and do necessary changes
- press enter in the smokeTest terminal and all smoke tests are run. Once finished the process waits for user input to run them again. Exiting process (ctrl + C) kills the node and smoke tests.

This way we should have the versatility we require and also have the tests in a single script to make them CI ready in the future.

sparrowDom:
@tomlinton

Yeah each smoke test has a before / after deploy functions. The logic is that: 
- all the before deploy test functions execute
- the deployment of contract changes from the `deploy` folder specified by the `deployid` is ran on the mainnet:fork
- all the after deploy test functions execute

This way we can test that a specific deploy that changes the contracts doesn't break whatever functionality tests are verifying. Mint / redeem / transfer in this case.

contracts/scripts/test/smokeTest.sh
franckc: Sounds good!

contracts/scripts/test/smokeTest.sh
franckc: What about PROVIDER_URL, that's another env var that should be required.
Maybe add a quick check in the script and exit or show a warning in the console if either BLOCK_NUMBER or PROVIDER_URL are missing.

contracts/deploy/011_ousd_fix.js
franckc: Got it. Thanks.

contracts/tasks/smokeTest.js
franckc: Agreed that ideally there would be a migration for every governance proposal. But I think practically speaking it won't happen for a couple reasons: a) sometime it is urgent to flip a setting and also b) we would end up with dozens of migrations which would make unit tests setup really heavy. Practically speaking, what we have been doing is we try to keep the initial deploy script 001_core.js fairly up to date to match the settings on prod.

Your proposal sounds good. Let's try that. Thanks! :)

sparrowDom: Then again maybe there should be an additional stopping when all the tests execute the `before` step in the interactive mode.

contracts/scripts/test/smokeTest.sh
sparrowDom: Good point thanks, adding those checks

sparrowDom:
Addressed all the comments. Needs one more CR to be mergeable.

@franckc @DanielVF @tomlinton would be really cool to use these tests when doing any contract changes. To figure out if they do all we want them to. 

And at one point move them to be a part of the CI. So if a new deploy is detected that smoke tests are automatically ran against it.

tomlinton:
Did you look at setting this up as part of the mocha framework using before/after?

sparrowDom:
@tomlinton I haven't. 

As far as I understand the before/after in mocha framework is related to hooks that happen before & after each of the tests runs. In this case the before & after are the main parts of the test that happen when the contracts are either deployed or upgraded. For that reason I don't see the utility of using mocha.

Also doing it manually we can have a good control over the node running in forked mode and can start one or more contract deploys that get smoke tested.

PR #531, opened 2021-01-27
Deploy script 012 for contract upgrades

franckc:
Upgrades the following contracts that have had changes since the last mainnet deploy (GH tag [deploy-011](https://github.com/OriginProtocol/origin-dollar/releases/tag/deploy-011)):
 - OUSD
 - VaultAdmin
 -  CompoundStrategy

Here is the diff for all contract changes since deploy-011: https://gist.github.com/franckc/6fcec3a50386bbd38f9aaf1296ea0ea0

Here is the [deploy plan](https://docs.google.com/spreadsheets/d/1phyzOJMmTBPIqTTa0v7HY6XJkjRmbrcdULRZPo_JEoY/edit#gid=1480380618).

### DIFF
--- a/contracts/deploy/012_upgrades.js
+++ b/contracts/deploy/012_upgrades.js
@@ -35,7 +35,13 @@ const upgrades = async () => {
 
   const cOUSDProxy = await ethers.getContract("OUSDProxy");
   const cVaultProxy = await ethers.getContract("VaultProxy");
-  const cCompoundStrategyProxy = await ethers.getContract("CompoundStrategyProxy");
+  const cVaultCoreProxy = await ethers.getContractAt(
+    "VaultCore",
+    cVaultProxy.address
+  );
+  const cCompoundStrategyProxy = await ethers.getContract(
+    "CompoundStrategyProxy"
+  );
 
   // Deploy a new OUSD contract.
   const dOUSD = await deployWithConfirmation("OUSD");
@@ -46,7 +52,6 @@ const upgrades = async () => {
   // Deploy a new CompooundStrategy contract.
   const dCompoundStrategy = await deployWithConfirmation("CompoundStrategy");
 
-
   // Proposal for the governor to upgrade OUSD.
   const propDescription = "OUSD, VaultAdmin, CompoundStrategy upgrades";
   const propArgs = await proposeArgs([
@@ -56,7 +61,7 @@ const upgrades = async () => {
       args: [dOUSD.address],
     },
     {
-      contract: cVaultProxy,
+      contract: cVaultCoreProxy,
       signature: "setAdminImpl(address)",
       args: [dVaultAdmin.address],
     },
--- a/contracts/deploy/012_upgrades.js
+++ b/contracts/deploy/012_upgrades.js
@@ -1,0 +1,121 @@
+const hre = require("hardhat");
+
+const {
+  isMainnet,
+  isFork,
+  isRinkeby,
+  isMainnetOrRinkebyOrFork,
+} = require("../test/helpers.js");
+const {
+  log,
+  deployWithConfirmation,
+  withConfirmation,
+  executeProposal,
+  sendProposal,
+} = require("../utils/deploy");
+const { proposeArgs } = require("../utils/governor");
+const { getTxOpts } = require("../utils/tx");
+
+const deployName = "012_upgrades";
+
+/**
+ * Deploys an upgrade for the following contracts:
+ *  - OUSD
+ *  - VaultAdmin
+ *  - Compound Strategy
+ * @returns {Promise<boolean>}
+ */
+const upgrades = async () => {
+  console.log(`Running ${deployName} deployment...`);
+
+  const { governorAddr } = await hre.getNamedAccounts();
+
+  // Signers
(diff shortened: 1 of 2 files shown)


DanielVF (approved):

tomlinton:
Looks good to me, reckon we can slip the cleanup from https://github.com/OriginProtocol/origin-dollar/pull/493#issuecomment-757365461 in here?

I just doubt we'll be upgrading OUSD for a while after this, so would be nice to get it done.

franckc:
> Looks good to me, reckon we can slip the cleanup from [#493 (comment)](https://github.com/OriginProtocol/origin-dollar/pull/493#issuecomment-757365461) in here?
> 
> I just doubt we'll be upgrading OUSD for a while after this, so would be nice to get it done.

@tomlinton Yes, that's the plan. #493 is already merged into master so this deploy will pick it up. Unless I missed something?

tomlinton:
The work from the comment in #493 wasn't done as part of #493. Summarising:

- Remove duplicate getters caused by `public` variables in `InitializableERC20Detailed` (they were switched to public to help with the reset).
- Remove OUSDReset.sol

Happy to add this to this PR or make a separate one that we merge first.

franckc:
@tomlinton Got it. Thanks for pointing this out - I missed it. Do you mind creating a separate PR for it? I find small PRs easier to review :)

sparrowDom (approved):
Looks good. Made a small commit: https://github.com/OriginProtocol/origin-dollar/pull/531/commits/4ddd8c261cc6e6d8a2e9947fc16850d67e23b402 so that upgrade file is smoke testable.

For some reason when the mint amounts are as in the master (1100 stablecoin units) mint contract calls timeout for me on forked network. So the smoke tests passed but with smaller mint / transfer / redeem amounts for me.

contracts/deploy/012_upgrades.js
```diff
@@ -108,19 +106,21 @@ const upgrades = async () => {
...
+  }
+  await upgrades(hre);
   console.log(`${deployName} deploy done.`);
   return true;
 };
 
 main.id = deployName;
 main.dependencies = ["011_ousd_fix"];
-main.skip = () => !isMainnetOrRinkebyOrFork;
+main.skip = () => !(isMainnet || isRinkeby);
```
franckc, line 124: @sparrowDom Remind me, why do we need to skip this deploy on the fork?

franckc:
> Looks good. Made a small commit: [4ddd8c2](https://github.com/OriginProtocol/origin-dollar/commit/4ddd8c261cc6e6d8a2e9947fc16850d67e23b402) so that upgrade file is smoke testable.

Great. Thanks! :)
> 
> For some reason when the mint amounts are as in the master (1100 stablecoin units) mint contract calls timeout for me on forked network. So the smoke tests passed but with smaller mint / transfer / redeem amounts for me.

That's odd... wonder if the timeout was actually caused by something else than the amount. I've seen cases on the fork where the initial run fails, but subsequent run succeed due to caching. Though in the case of a smoke test, the fork gets started/terminated every run so that probably does not apply... I'll try to reproduce.

contracts/deploy/012_upgrades.js
sparrowDom: If it is set to not skip on fork then the deploy is applied immediately when the node in `fork` mode is started. And we are not able to execute the first (before) step of the tests before the contract deploy.

sparrowDom:
@franckc yeah I've seen similar cases where if mint over 1k happened to succeed it would then always subsequently succeed. And all calls that succeed once, execute a lot faster and never fail on the next runs. Implying that like you say some caching seems to happen. 

My hunch is that when rebase/allocate happen, so much code gets executed that the node isn't able to do it before it times out. Though that would make sense only if many external contract calls are made.

contracts/deploy/012_upgrades.js
franckc: ah yeah... that's not ideal since sometime developers will want to run the fork for other reasons than running smoke tests? I can't think of a good solution though. I guess the developer could then update that line... 

I'd suggest we update the contracts/README.md to document all this in the section that covers how to use the fork? (and we should anyway document how to use the smoketests).

tomlinton:
Yea @sparrowDom your hunch is right. For the rebase/allocate to work the fork has to do a lot more work because it'll have to understand the state of all the external contracts (Curve, Aave, Compound, etc), so it'll go out and pull state for all those for the block of the fork. If you pin the block number with `export BLOCK_NUMBER=xxx` it'll cache that data so its near instant the second time around.

tomlinton:
Cleanup here https://github.com/OriginProtocol/origin-dollar/pull/537

tomlinton (approved):

sparrowDom:
aah makes sense @tomlinton thanks. Any ideas how to make allocate / rebase calls succeed? I think I've seen you comment somewhere that hardhat timeout isn't configurable. Which would be the first thing to try.

tomlinton:
I just keep running it until it works, the fork gets more and more cached data each time you run it so eventually it'll go through. Haven't successfully been able to set a timeout to fix this. It'll require a PR to hardhat. Would be great if we could solve it.

Issue #468, opened 2020-12-30
Tool for checking storage slots before a deploy


franckc:
We had an incident where, due to a mistake, storage slots were shifted during the deployment of a new contract. Writes were then issued and caused a data corruption.

To prevent this from happening in the future, we should have a tool we run prior to a deployment to check the storage slots of the current contracts on the master branch vs the ones of the contracts on Mainnet.

franckc:
CC @DanielVF @tomlinton

DanielVF:
Hardhat does some kind of upgrade checking here in this module.

https://hardhat.org/plugins/hardhat-upgrades.html

franckc:
A relevant article: https://blog.coinbase.com/usdc-v2-upgrading-a-multi-billion-dollar-erc-20-token-b57cd9437096 with pointers to USDT storage slots unit tests.

mikeshultz:
If the hardhat stuff is good enough, great.  An alternative might be a simple tool using [the antlr4 parser](https://github.com/federicobond/solidity-parser-antlr) and looking for storage def changes.

sparrowDom:
Solution for this here: https://github.com/OriginProtocol/origin-dollar/pull/536

PR #536, opened 2021-01-29
Contract storage slots checker

sparrowDom:
This PR checks before a contract is upgraded that it doesn't corrupt the storage slot layout of the currently live contract. To achieve this: 
 - When a contract gets deployed its storage slot layout is stored to `storageLayout/[network]/[contractName].json`. And when a user deploys contracts to mainnet those layout changes are reflected in repo and should be pushed to master. 
 - On a contract upgrade the currently live storage layout (stored in repo) and a new layout derived from contract source code are compared and if there is a conflict an exception is thrown.

**Connected issue:** https://github.com/OriginProtocol/origin-dollar/issues/468

**A few helpful hardhat tasks are added:**
 - saveStorageSlotLayout -> saves storage layout for all contracts
 - checkUpgradability -> checks upgradability of a specific contract
 - checkUpgradabilityAll -> checks upgradability for all contracts 

The current storage slot json files have been generated from contract code commit done on Jan 8th, when the 011_ousd_fix.js has been created. 

**Example of storage slot corruption detection:**

Foo variable is added to OUSD contract
Screenshot 2021-01-29 at 21 40 10

Upon launching the upgrade an error is thrown: 
Screenshot 2021-01-29 at 21 39 58

For solidity gurus: 
 - is excluding [Proxy and the following contracts](https://github.com/OriginProtocol/origin-dollar/pull/536/files#diff-92f367937ac4637af4322bf6c4e30970fc0f6ffea938f08d1173c60a101571dcR77) from storage slot checker ok?
 
Contract change checklist:
  - [ ] Unit tests pass


sparrowDom:
Is excluding these contract names from slot check ok?
https://github.com/OriginProtocol/origin-dollar/pull/536/files#diff-92f367937ac4637af4322bf6c4e30970fc0f6ffea938f08d1173c60a101571dcR77

Also we could be using `hardhat-upgrades` module that does these storage slot checks for us, but the initial contract proxy needs to be created using the `hardhat-upgrades` code. There is an [open issue on their repo](https://github.com/OpenZeppelin/openzeppelin-upgrades/issues/175), to allow registering a proxy even if it was not created using their code, but it has not been implemented yet.

contracts/tasks/storageSlots.js
```diff
@@ -0,0 +1,156 @@
...
+
+  // pretty print storage layout for the contract
+  await promises.writeFile(storageLayoutFile, JSON.stringify(layout, null, 2));
+  console.log(`Storage slots layout for ${contractName} saved to ${storageLayoutFile} `);
+}
+
+const getAllEligibleContractNames = async (hre) => {
+  const contractNames = Object.keys(await hre.deployments.all());
+
+  const excludeContracts = [
```
franckc, line 69: I would add a comment explaining those contracts have been deprecated. They remain present in the deployment files but the contract code has been deleted and therefore they need to be excluded.

contracts/tasks/storageSlots.js
```diff
@@ -0,0 +1,156 @@
...
+    'CurveUSDCStrategy',
+    'CurveUSDTStrategy',
+    'MinuteTimelock',
+    'OpenUniswapOracle',
+    'RebaseHooks'
+  ];
+
+  // TODO: check that excluding contracts below + proxy ones is OK
```
franckc, line 77: Proxies are not upgradeable so make sense to exclude them.

contracts/utils/deploy.js
```diff
@@ -32,6 +35,14 @@ function sleep(ms) {
 }
 
 const deployWithConfirmation = async (contractName, args, contract) => {
+  // check that upgrade doesn't corrupt the storage slots
+  await assertUpgradeIsSafe(hre, contractName);
+
+  // if upgrade happened on the mainnet save the new storage slot layout to the repo
```
franckc, line 41: Rather than here. should this happen once the deploy is done and confirmed. Like after the call to withConfirmation below? So that in case of deploy error, the storage layout is not persisted.

contracts/tasks/storageSlots.js
```diff
@@ -0,0 +1,156 @@
...
+
+const {
+  getStorageLayout,
+  getVersion,
+  getUnlinkedBytecode,
+  Manifest,
+  getStorageLayoutForAddress,
+  isCurrentValidationData,
+  assertStorageUpgradeSafe
+} = require('@openzeppelin/upgrades-core');
```
franckc, line 12: Don't we need to add this package to package.json? We currently have @openzeppelin/upgrades in there but not @openzeppelin/upgrades-core

franckc (commented):
I'm super excited we are adding this tool! Thank you for doing that.

A few minor comments inline.

contracts/tasks/storageSlots.js
sparrowDom: Aha that makes sense thanks.

contracts/tasks/storageSlots.js
sparrowDom: 👍

contracts/tasks/storageSlots.js
sparrowDom: I think it worked for me because another package already includes it. Thanks for catching this, added to `contracts/package.json`

contracts/utils/deploy.js
sparrowDom: Yup makes a lot more sense, moving to there

sparrowDom:
All comments addressed. Will stick to our soft defined deal and wait for 2 CR confirmations before merging this one in.

franckc (approved):
Thanks for addressing my comments. Nice work on this! :) LGTM

Lets merge this in and use it when running deploy 12

franckc:
@sparrowDom Changes look good. Though can you take a look at those tests failures before merging this in?

sparrowDom:
Disabling this functionality on tests for now. Reading hardhat cache fails in CI and not in local test environment for some reason. Do object if you would like to see the storage slot checker working on test env anyway.

Issue #111, opened 2020-08-28
Guaranteed Amounts vs Approximation


micahalcorn:
Our buying/minting form includes a disclaimer about the resulting OUSD being an approximate amount. Arguably, our selling/redeeming form should also include this.

<img width="562" alt="Screen Shot 2020-08-28 at 11 02 36 AM" src="https://user-images.githubusercontent.com/273937/91601251-28249c00-e91e-11ea-9c72-c1c29d12c8d4.png">

This is an approximation due to possible movements in the exchange rate(s) for the stablecoin(s) involved in the transaction between the time that the form is rendered and the time that the transaction is constructed in the contract. If https://github.com/oplabs/origin-dollar/issues/90 were to account for depth of liquidity, then there could also be slippage. With the introduction of https://github.com/oplabs/origin-dollar/issues/110, there would be actual slippage.

My hypothesis is that, as we move toward a more mainstream userbase, the approximation of prices and lack of clarity might be a deterrent that causes people to abandon the transaction. Less sophisticated users might actually prefer a worse price if they don't have to think about price fluctuations. Uniswap addresses the slippage problem by promising a minimum output below and estimated output. We should consider doing something similar.

Alternatively, I'm wondering if it's possible for us to give the user the specific amount that they ask for, account for slippage in the rate, and then collect any bonus/excess resulting OUSD (or unconverted stablecoin) as revenue for either the pool itself or for governance token. This would create a small revenue opportunity and also give users a cleaner transaction that is easier to mentally process. I suppose this could be considered "OUSD as a market maker".

@matthewliu points out that this might involve significant amounts of money for users transacting with large amounts or people for whom a few dollars is very meaningful. Users who understand what's happening might be frustrated that they're losing money on their transactions. It would probably also depend on the APY that they're likely to enjoy and the gas costs associated with their transactions.

micahalcorn:
With the advent of https://github.com/OriginProtocol/origin-dollar/pull/390 and https://github.com/OriginProtocol/origin-dollar/pull/410, our contracts now have the ability to abandon mints and redeems if the output is less than expected. We'll implement a frontend config feature for the user to set a threshold for oracle price movement much like slippage parameters on AMMs. My alternate solution proposed above might make sense with more mainstream users who are less familiar with AMMs, but those aren't likely our target users with this DApp. We probably want them to just buy OUSD on centralized exchanges or with fiat onramps when the time comes.

micahalcorn:
[...]

PR #544, opened 2021-02-03
Price tolerance

sparrowDom:
Add price tolerance functionality to mint & redeem. 

Zeplin screens: [Redeem](https://app.zeplin.io/project/5f28ed76ad0b5f20d672cde8/screen/6010c459e94291673307e4e7), [Mint](https://app.zeplin.io/project/5f28ed76ad0b5f20d672cde8/screen/60108c242dba311271b28c95) 
**connected issue:** https://github.com/OriginProtocol/origin-dollar/issues/111

Contract change checklist:
  - [ ] Unit tests pass


micahalcorn (changes requested):
I'll defer to @shahthepro on the code review, but there are a few of issues with the math.

Screen Shot 2021-02-03 at 3 15 46 PM

1) The sum of the 3 in "Stablecoin Mix" should be equal to "Estimated stablecoins".

> 5,840.56 + 628.03 + 3,465.59 == 9,934.18 != 9,984.20

2) Rounding should be consistent so that the numbers make sense to a human reader. The most important thing is for the "Min. stablecoins received" to be rounded down to properly set expectations. All of the estimates can be rounded whichever way adds up.

> 9,984.20 * (1 - .005) == 9,934.279 != 9,934.20

3) The "Exchange rate loss" should correspond to the prices shown in the stablecoin mix. If they all show "1.00/", then I would expect the loss to be 0. If this is attributable to rounding, we should round down on the rates so that they show 0.99 when the actual rate is 0.9987.

sparrowDom:
@micahalcorn thanks for verifying the math and finding issues.
- regarding the 1. issue with the sum that one is fixed
- regarding the 2nd issue, there was a bug and it is now fixed. Though still since we take the non rounded numbers when doing the calculation the end result might be off for 1 cent. Because the: `9,984.20 * (1 - .005)` can actually be `9,984.209 * (1 - .005)`. The difference though will no longer be 7 (almost 8) cents like in your example above
- the thing with redeem exchange rates that number is never below 1. Since that represents a dollar value of a stablecoin. And we don't want to redeem more than 1 stablecoin per OUSD. We could display 4 decimals here to show the non 1.00 values: 
Screenshot 2021-02-04 at 16 56 34
But that will require some design changes since it gets a bit cramped in there. On 2 decimals it still looks ok:
Screenshot 2021-02-04 at 16 57 00

Maybe we could switch to the same logic used for mobile layout: 
Screenshot 2021-02-04 at 17 01 51

the fixed are going to be live on staging in ~15min

micahalcorn:
> * the thing with redeem exchange rates that number is never below 1. Since that represents a dollar value of a stablecoin. And we don't want to redeem more than 1 stablecoin per OUSD. We could display 4 decimals here to show the non 1.00 values:

@sparrowDom I gave a bad example here. I was suggesting that we would show 1.01 when the actual rate is 1.0023.

dapp/src/components/buySell/BuySellWidget.js
```diff
@@ -112,6 +118,22 @@ const BuySellWidget = ({
...
+    localStorage.setItem(priceToleranceLocalStorageKey, value)
+  }
+
+  // store price tolerance value for future default setting
+  useEffect(() => {
+    // default price tolerance value
+    let priceTolerance = 0.5
+    let localStorageValue = localStorage.getItem(priceToleranceLocalStorageKey)
+    if (localStorageValue) {
+      priceTolerance = parseFloat(localStorageValue)
```
shahthepro, line 132: nit: A `NaN`, `null` or `undefined` check would be great if you want to be super-defensive.

dapp/src/components/buySell/SellWidget.js
```diff
@@ -73,6 +76,35 @@ const SellWidget = ({
...
+  // store price tolerance value for future default setting
+  useEffect(() => {
+    // default price tolerance value
+    let priceTolerance = 0.5
+    let localStorageValue = localStorage.getItem(priceToleranceLocalStorageKey)
+    if (localStorageValue) {
+      priceTolerance = parseFloat(localStorageValue)
+    }
+    setPriceToleranceValue(priceTolerance)
+  })
```
shahthepro, line 106: Same concern with `useEffect` as above without a second param.

shahthepro, line 106: Thinking about it, this could very well be moved into its own hook, something like `usePriceTolerance`

dapp/src/components/buySell/SellWidget.js
```diff
@@ -183,14 +215,23 @@ const SellWidget = ({
 
     setSellWidgetState('waiting-user')
 
+    const minStableCoinsReceivedBN = ethers.utils.parseUnits(
+      (Math.floor(minStableCoinsReceived * 10000) / 10000).toString(),
+      18
```
shahthepro, line 220: nit: Alternatively, could do like this: 
```ethers.utils.parseUnits(Math.floor(minStableCoinsReceived * 10000).toString(), 22)```

But I guess it'd make things complex

**Review by shahthepro — APPROVED** [review]
Looks good to me, a few comments inline, PTAL.

**Review by micahalcorn — APPROVED** [review]
There are still some minor rounding issues, but this should not block us from merging.

> 100.00 - 0.50 - 0.17 == 99.33 != 99.32

> 55.50 + 13.03 + 30.78 == 99.31 != 99.32

Screen Shot 2021-02-04 at 11 04 20 PM

Otherwise, I've tested `mint` with a price tolerance of 0.25% and received the expected result.

File: dapp/src/components/buySell/BuySellWidget.js [hunk 4]
```diff
@@ -112,6 +118,22 @@ const BuySellWidget = ({
...
+    }
+    setPriceToleranceValue(priceTolerance)
+  })
```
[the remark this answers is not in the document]
sparrowDom: Missed the second parameter, nice catch thanks!

dapp/src/components/buySell/BuySellWidget.js
```diff
@@ -606,28 +626,126 @@ const BuySellWidget = ({
...
+                            'Much like slippage, exchange rate movement can cause you to receive less OUSD than expected. The price tolerance is the maximum reduction in OUSD that you are willing to accept. If exchange rates fall below this threshold, your transaction will revert.',
+                            'Much like slippage, exchange rate movement can cause you to receive less OUSD than expected. The price tolerance is the maximum reduction in OUSD that you are willing to accept. If exchange rates fall below this threshold, your transaction will revert.'
+                          )}
+                        />
+                      </div>
+                      <Dropdown
+                        className="d-flex align-items-center min-h-42"
+                        content={
+                          <div className="d-flex flex-column dropdown-menu show">
+                            <div
```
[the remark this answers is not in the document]
sparrowDom: good point 👍

dapp/src/components/buySell/SellWidget.js
sparrowDom: yup thanks

sparrowDom: I like that! moving to a hook

dapp/src/components/buySell/SellWidget.js
sparrowDom: Yup good point. Thought I would rather keep it as it is, since I think it makes the code more readable. I would quickly overlook that 22.

dapp/src/components/buySell/BuySellWidget.js
sparrowDom: Good thinking thanks

sparrowDom (commented):
All comments addressed. @shahthepro thanks for the thorough review, really appreciate it 🙏

sparrowDom:
@micahalcorn & @shahthepro thanks for all the help with this PR 🙏 

Merging to master and deploying to prod

PR #548, opened 2021-02-05
Staging env vars

sparrowDom:
Makes staging deployment read from the staging encoded variables file.

An interesting issue was encountered here. The environmental variables used by the browser ran JS code are picked up at the next-build stage rather on the development machine. Rather than any environment on the production machine. That is why it is important to feed the right environmental var file to the build process at build time.

Contract change checklist:
  - [ ] Unit tests pass


sparrowDom:
Merging this since it doesn't impact anything but staging. Happy to address stuff later if needed.

PR #558, opened 2021-02-11
Flipper for zero slippage trading of OUSD for stablecoins

DanielVF:
## Concept

The flipper contract allows 1-1 trading between stablecoins and OUSD. This allows easy in and and out during normal as users "flip" the available capital back and forth between OUSD and stablecoins.

## Ownership

Funds in the flipper are company funds and not user funds. The owner should not be set to the usual timelock governance nor to the strategist role. Assuming OUSD growth, the owner will periodically need to withdraw stablecoins and mint more OUSD.

## Low Gas Trading

The six trading functions optimize for low gas usage. Contributing to this, they make no internal calls, nor do they read from internal storage or write to internal storage. They also have one function per trading vector, which moves complexity and gas cost out of the contract.

This low gas focus means that the trading functions are not written in our usual style and do not use SafeERC20. We'll need to test them on fork, and look carefully at each.

When we deploy the contract for real, we should change to code to having the contract addresses as constants, rather than set by the constructor.

## Availability

Unlike uniswap, there is very much not guarantee that the tokens you want to trade for are available. This is an opportunistic money saving method that may be there, or may not. UI's and user experience need to communicate this.

## Rounding

I went for pragmatic simplicity here. It is possible for request more OUSD and pay less USDC, USDT than owed. By extension is also possible to pay no USDT and get a sliver of OUSD. If ETH were $1, and gas 1 gwie, then someone would spend 8000% in gas over the OUSD received. 

Similarly, if we did every transaction that has ever gone through the Uniswap 2 router (24 million), and each one went for maximum rounding error, then our rounding losses would be $24.

Still makes me make faces. I welcome second opinions.

----

Contract change checklist:
  - [x] Unit tests pass


contracts/contracts/flipper/Flipper.sol
```diff
@@ -0,0 +1,136 @@
...
+
+    function sellOusdForDai(uint256 amount) external {
+        require(amount <= MAXIMUM_PER_TRADE, "Amount too large");
+        require(dai.transfer(msg.sender, amount));
+        require(ousd.transferFrom(msg.sender, address(this), amount));
+    }
+
+    function buyOusdWithUsdc(uint256 amount) external {
+        require(amount <= MAXIMUM_PER_TRADE, "Amount too large");
+        // Potential rounding error is an intentional tradeoff
```
franckc, line 85: super super nit: let's add this comment consistently where it applies. or just once in the header file. right now it is in some places but not others.

contracts/contracts/flipper/Flipper.sol
```diff
@@ -0,0 +1,136 @@
...
+    function sellOusdForUsdc(uint256 amount) external {
+        require(amount <= MAXIMUM_PER_TRADE, "Amount too large");
+        require(usdc.transfer(msg.sender, amount / 1e12));
+        require(ousd.transferFrom(msg.sender, address(this), amount));
+    }
+
+    function buyOusdWithUsdt(uint256 amount) external {
+        require(amount <= MAXIMUM_PER_TRADE, "Amount too large");
+        // Potential rounding error is an intentional tradeoff
+        usdt.transferFrom(msg.sender, address(this), amount / 1e12);
```
franckc, line 99: nit: add a comment mentioning USDT transferFrom reverts and does not return a boolean, therefore no need to use require

franckc (commented):
Fun! Took a first pass at reviewing. Looks very good. Just a few minor comments inline.

franckc:
Also note the failing slither test:
```
Tether (contracts/flipper/Flipper.sol#13-23) has incorrect ERC20 function interface:Tether.transferFrom(address,address,uint256) (contracts/flipper/Flipper.sol#16-20)
Reference: https://github.com/crytic/slither/wiki/Detector-Documentation#incorrect-erc20-interface
```

contracts/deploy/001_core.js
```diff
@@ -484,6 +484,24 @@ const deployCore = async () => {
   log("Initialized OUSD");
 };
 
+// Deploy the Flipper tradeing contract
```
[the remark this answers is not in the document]
DanielVF: I'd love if you would set up the deploy file.

contracts/contracts/flipper/Flipper.sol
DanielVF: There are only two places in the file that a rounding error works against us - if I'm right in my thinking. I think both of those have comments. It's only when buying OUSD with 6 precision coins (USDT and USDC).

contracts/contracts/flipper/Flipper.sol
```diff
@@ -0,0 +1,136 @@
...
+}
+
+contract Flipper is Governable {
+    using SafeERC20 for IERC20;
+
+    uint256 constant MAXIMUM_PER_TRADE = (25000 * 1e18);
+    uint256 constant MININUM_PER_TRADE = (50 * 1e12);
+
+    // ---------------------
+    // Production constructor
```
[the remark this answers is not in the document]
DanielVF: Yes. By using constants we remove two reads per trade.

contracts/contracts/flipper/Flipper.sol
```diff
@@ -0,0 +1,136 @@
...
+        require(address(dai) != address(0));
+        require(address(usdc) != address(0));
+        require(address(usdt) != address(0));
+    }
+
+    // -----------------
+    // Trading functions
+    // -----------------
+
+    function buyOusdWithDai(uint256 amount) external {
```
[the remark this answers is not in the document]
DanielVF: Done.

DanielVF:
Set slither hide the errors about Tether not being ERC20 compliant.

tomlinton:
This'll get arbed if a stablecoin goes off peg by a bit by chaining calls i.e. buy for usdt, sell for dai. Probably not an issue if we don't keep too much in it.

contracts/contracts/flipper/Flipper.sol
franckc: yes, u r right. ignore my comment.

contracts/deploy/001_core.js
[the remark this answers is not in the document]
franckc: Will do

contracts/contracts/flipper/Flipper.sol
```diff
@@ -0,0 +1,154 @@
...
+        uint256 value
+    ) external;
+
+    function balanceOf(address) external returns (uint256);
+}
+
+contract Flipper is Governable {
+    using SafeERC20 for IERC20;
+
+    uint256 constant MAXIMUM_PER_TRADE = (25000 * 1e18);
```
tomlinton, line 28: What's the reasoning behind this maximum? Trying to think through the advantage of it. If I wanted to circumvent it I'd just run multiple buys or sells in one transaction.

DanielVF:
Yeah. It's practically guaranteed to have the less valuable of the set of stablecoins.

I figure we launch it with a small amount of capital, and see what happens and how useful it is.

franckc (approved):
LGTM

franckc:
Deployed to Mainnet.

@DanielVF I let you merge in this PR to master after checking that @tomlinton does not have any objection (there is still one of his comment that is unaddressed).

DanielVF:
@tomlinton Replying to both of your comments here, since they are related.

> This'll get arbed if a stablecoin goes off peg by a bit by chaining calls i.e. buy for usdt, sell for dai. Probably not an issue if we don't keep too much in it.

> What's the reasoning behind this maximum? Trying to think through the advantage of it. If I wanted to circumvent it I'd just run multiple buys or sells in one transaction.

Yeah, doing multiple coins ensures that this will tend towards having the lowest valued coins in pile. Time will tell how it works out.

A limit does two things:
- It increases the size of the spread needed between stablecoins before it's profitable to arb them. This means that there's more likely a selection of stables available to "withdraw" to.
- It's a least a nudge while we are learning to keep this more useful for smaller players.

All that said, we may later want to raise the limit.

contracts/contracts/flipper/Flipper.sol
```diff
@@ -0,0 +1,119 @@
...
+//   - Optimized for low gas usage
+//   - No guarantee of availability
+
+
+contract Flipper is Governable {
+    using SafeERC20 for IERC20;
+
+    uint256 constant MAXIMUM_PER_TRADE = (25000 * 1e18);
+
+    // -----------
```
tomlinton, line 20: Constructor is below 😁

franckc: Fixed!

franckc:
Merged to master and created tag deploy-015

PR #568, opened 2021-02-21
Fix fund task on local

franckc:
Fixed the hardhat fund task on local.

Also set GA_ID to a fake id otherwise the DApp server was failing.


franckc:
Merging this in to unblock a contributor. Happy to address any feedback as a separate PR.

sparrowDom:
LGTM thanks for doing this

PR #575, opened 2021-02-22
Add OGN buyback contract

tomlinton:
Deploys a simple buyback contract that swaps OUSD for OGN via Uniswap. The buyback contract replaces the trustee and the swap function is called as part of allocate. Some design decisions I made:

- Non upgradeable, has a hardcoded Uniswap path and hardcoded threshold for swap amounts
- Only swap if buyback has > 1000 OUSD

Contract change checklist:
  - [x] Unit tests pass


contracts/contracts/buyback/Buyback.sol
```diff
@@ -0,0 +1,85 @@
...
+        require(vaultAddr == msg.sender, "Caller is not the Vault");
+        _;
+    }
+
+    /**
+     * @dev Set address of Uniswap for performing liquidation of strategy reward
+     * tokens. Setting to 0x0 will pause swaps.
+     * @param _address Address of Uniswap
+     */
+    function setUniswapAddr(address _address) external onlyGovernor {
```
franckc, line 42: should this be onlyStrategist so that the strategist can pause swaps?

contracts/contracts/buyback/Buyback.sol
```diff
@@ -0,0 +1,85 @@
...
+    }
+
+    /**
+     * @dev Execute a swap of OGN for OUSD via Uniswap or Uniswap compatible
+     * protocol (e.g. Sushiswap)
+     **/
+    function swap() external onlyVault {
+        if (uniswapAddr != address(0)) {
+            uint256 sourceAmount = ousd.balanceOf(address(this));
+            if (sourceAmount >= 1000 * 1e18) {
```
franckc, line 57: should this be a threshold stored as a public variable with a setter accessible by the strategist?

contracts/contracts/buyback/Buyback.sol
```diff
@@ -0,0 +1,85 @@
...
+                // Uniswap redemption path
+                address[] memory path = new address[](4);
+                path[0] = address(ousd);
+                path[1] = address(usdt); // USDT
+                path[2] = IUniswapV2Router(uniswapAddr).WETH();
+                path[3] = address(ogn);
+                IUniswapV2Router(uniswapAddr).swapExactTokensForTokens(
```
franckc, line 64: Should we emit a BuyBack event for our analytics needs?
Or is it pretty stratight forward to leverage the events this method I'm sure already emits?

contracts/deploy/015_buyback.js
```diff
@@ -0,0 +1,116 @@
...
+      args: [dVaultCore.address],
+    },
+    {
+      contract: cBuyback,
+      signature: "claimGovernance()",
+    },
+    {
+      contract: cVaultAdmin,
+      signature: "setTrusteeAddress(address)",
+      args: [cBuyback.address],
```
franckc, line 65: Should this be done separately, after the governor claimed governance?

franckc (commented):
Sorry sorry for being super later on this review.

A few minor questions inline

DanielVF (approved):
I still need to set up to run a fork test and verify that the uniswap calls work correctly.
#### Requirements

I'm wondering if we should just keeper this directly and keep the vault unmodified? I guess doing it this way adds gas cost, pushing back profitability on slippage attacks a little, but at the cost of us paying the gas all the time.

#### Internal State

_No stored state :heart:_

#### Attack

The obvious attack here is a slippage attack on the sale of the tokens. OUSD/USDT has fairly thin liquidity, with 2.3 million on each side.

If my math is right, for roughly 300 dollars in Uniswap fees, an attacker could bend Uniswap, then call this and capture 5% of the sale value. That would make this profitable, ignoring gas, at a trade of $6,000 of yield. There is currently $3,941 OUSD ready to be swapped. Now gas adds a little to this, and we should be running this more often once it's automated. But we are inside the order of magnitude where these attacks make sense.

Having this called only by the vault doesn't move running this action out of user control, just adds a little gas

#### Logic

_Logic good._

#### Tests

#### Flavor

See inline comments.

#### Overflow

No overflow, though a probably unnecessary plain addition is in the code and can be removed.

#### Black magic

_No black magic._

#### Authentication

- [x] Never use tx.origin for authentication

#### Cryptographic code

_No cryptographic code_

#### Gas problems

_No gas issues, but see inline comments on gas savings._

#### External calls

- [x] Unsafe external calls
- [-] How vulnerable is this to a slippage attack
- [-] Does this work on the fork (STILL NEED TO TEST)

#### Ethereum

contracts/contracts/buyback/Buyback.sol
DanielVF: For the flipper I just planned on watching the ERC20 Transactions out of it.

contracts/contracts/buyback/Buyback.sol
```diff
@@ -0,0 +1,85 @@
...
+     * protocol (e.g. Sushiswap)
+     **/
+    function swap() external onlyVault {
+        if (uniswapAddr != address(0)) {
+            uint256 sourceAmount = ousd.balanceOf(address(this));
+            if (sourceAmount >= 1000 * 1e18) {
+                // Uniswap redemption path
+                address[] memory path = new address[](4);
+                path[0] = address(ousd);
+                path[1] = address(usdt); // USDT
```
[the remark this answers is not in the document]
tomlinton: Yep thanks will remove, the address was hardcoded here before but I moved it up to the top and didn't remove that comment.

contracts/deploy/015_buyback.js
tomlinton: I don't see a reason to, the whole thing either works or it doesn't, any reason you had in mind?

contracts/contracts/buyback/Buyback.sol
tomlinton: I'm ok with it being hardcoded. I don't think we will need to change it and we could also just redeploy it if we did. Happy to hear any argument for making it configurable though.

contracts/contracts/buyback/Buyback.sol
franckc: Sounds good.

franckc (approved):

contracts/contracts/buyback/Buyback.sol
```diff
@@ -0,0 +1,85 @@
...
+                path[2] = IUniswapV2Router(uniswapAddr).WETH();
+                path[3] = address(ogn);
+                IUniswapV2Router(uniswapAddr).swapExactTokensForTokens(
+                    sourceAmount,
+                    uint256(0),
+                    path,
+                    address(this),
+                    now + 1800
```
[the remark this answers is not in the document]
tomlinton: Will change

contracts/contracts/buyback/Buyback.sol
```diff
@@ -0,0 +1,85 @@
...
+        ousd.safeApprove(uniswapAddr, uint256(-1));
+        emit UniswapUpdated(_address);
+    }
+
+    /**
+     * @dev Execute a swap of OGN for OUSD via Uniswap or Uniswap compatible
+     * protocol (e.g. Sushiswap)
+     **/
+    function swap() external onlyVault {
+        if (uniswapAddr != address(0)) {
```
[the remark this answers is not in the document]
tomlinton: Changed 👍

contracts/contracts/buyback/BuybackConstructor.sol
```diff
@@ -0,0 +1,25 @@
+pragma solidity 0.5.11;
+
+import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+
+import { Buyback } from "./Buyback.sol";
+
+contract BuybackConstructor is Buyback {
```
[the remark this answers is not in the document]
tomlinton: Yea I was looking for a way to decrease gas consumption of the production contract through constants but also have it configurable for tests. I didn't fully follow through here as you've spotted. I looked at the @eth-optimisim/smock package as an option.

I could duplicate the code like Flipper/FlipperDev, but surely there must be an easier way. I don't like ending up with two separate contracts, one for tests and one for the deployment.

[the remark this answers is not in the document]
tomlinton: I guess one easy solution is just to change the constants at deploy time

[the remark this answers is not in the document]
franckc: Agreed it is not the prettiest, but for now in the spirit on moving fast and focusing on NFT, should we go for the simplest solution which is to duplicate the contract? e.g. BuyBack.sol and BuyBackDev.sol
We can file an issue for figuring our a better solution later?

franckc:
FYI - I'm starting to gather the list of contracts to include when we do our next mainnet deploy so that I can iron out all the deploy scripts involved. Trying to target end of this week.

franckc:
Ah also, there is a test failure which seems related to this PR. Let's fix it before merging.

tomlinton:
I fixed the slither error. Not sure about that timeout, but seems to be unrelated to this. I'll keep pressing retry till it goes green!

contracts/contracts/buyback/BuybackConstructor.sol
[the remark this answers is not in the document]
tomlinton: Went with the duplicate and changed the prod to constant similar to Flipper.

PR #579, opened 2021-03-01
Oracle simplification

DanielVF:
Change from combining multiple oracles per stable coin to using a single oracle per stablecoin.

@franckc, would you be able to write the official deploy script for this?

----

Contract change checklist:
  - [x] Unit tests pass


contracts/contracts/vault/VaultAdmin.sol
```diff
@@ -380,63 +384,33 @@ contract VaultAdmin is VaultStorage {
...
-     * @return uint256 USD price of 1 of the asset
-     */
-    function priceUSDRedeem(string calldata symbol)
-        external
-        view
-        returns (uint256)
-    {
+    function priceUSDMint(address asset) external view returns (uint256) {
+        uint256 price = IOracle(priceProvider).price(asset);
+        if (price > 1e8) {
```
franckc, line 393: shouldn't that be price < 1e8 ?

franckc (approved):
So much simpler! :) Thanks for implementing this.
A couple comment inline, PTAL otherwise LGTM.

contracts/contracts/oracle/OracleRouter.sol
```diff
@@ -0,0 +1,74 @@
...
+    }
+}
+
+contract OracleRouter is OracleRouterBase {
+    /**
+     * @dev The price feed contract to use for a particular asset.
+     * @param asset address of the asset
+     */
+    function feed(address asset) internal view returns (address) {
+        if (asset == address(0x6B175474E89094C44Da98b954EedeAC495271d0F)) {
```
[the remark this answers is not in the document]
DanielVF: Added comments!

contracts/contracts/oracle/OracleRouter.sol
```diff
@@ -0,0 +1,74 @@
+pragma solidity 0.5.11;
+
+import "../interfaces/chainlink/AggregatorV3Interface.sol";
+import { IOracle } from "../interfaces/IOracle.sol";
+
+contract OracleRouterBase is IOracle {
+    uint256 constant MIN_DRIFT = uint256(70000000);
```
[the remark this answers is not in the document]
DanielVF: Yeah, we seem to never change it, and it saves on gas for each oracle action by being hardcoded.

contracts/contracts/vault/VaultAdmin.sol
DanielVF: Thanks so much for catching this!

DanielVF:
@tomlinton You want to take a look at this PR too?

franckc:
I'm going to be the pita who asks... should we add a unit test that would have caught the sign bug so that in the future we don't miss it in case we do any sort of refactoring, etc...?

sparrowDom:
didn't CR it but have tested it in the dapp locally and it worked nicely for me (fetches exchange rates in the expected format).

DanielVF:
Move from using merging multiple oracles per stablecoin to using a single oracle per stablecoin.

#### Requirements

Now that the vault cannot end up with fewer stablecoins as a result of oracle prices, we no longer need redundancy and the expense and complication that that brings. This code provides a direct bridge from a USD denominated Chainlink oracle for each of the stablecoins used by OUSD.

#### Internal State

_No internal state in new oracle setup._

#### Attack

There could exists short lived arbitrage opportunities if a stablecoin lost its peg , the oracles had not yet updated, and the OUSD could be bought/sold elsewhere. This is the same situation we have always had, and should not be changed by this PR, other than we now have one oracle rather than two. An attacker still should not be able to loop through minting and redeeming and take total coins from other OUSD holders.

#### Logic

_Looks correct._

#### Tests

  - [x] Each logical branch has a test (_outside picking which oracle to use in production_)

#### Flavor

The actual OracleRouter contract code is very simple. The API interface is also simple. I tried to make changes to as little of the rest of the vault as possible, outside converting to lookup oracles with addresses rather than strings. It's possible that things could be named better to go with the new system.

#### Overflow

#### Black magic

_No black magic_

#### Authentication

_No changes to authentication._

#### Cryptographic code

_No crypto code._

#### Gas problems

Big win here!

#### External calls

Only makes external calls to hardcoded Chainlink oracle addresses.

#### Ethereum

contracts/contracts/oracle/OracleRouter.sol
```diff
@@ -0,0 +1,77 @@
...
+            return address(0xAed0c38402a5d19df6E4c03F4E2DceD6e29c1ee9);
+            // USDC
+        } else if (
+            asset == address(0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48)
+        ) {
+            return address(0x8fFfFfd4AfB6115b954Bd326cbe7B4BA576818f6);
+            // USDT
+        } else if (
+            asset == address(0xdAC17F958D2ee523a2206206994597C13D831ec7)
+        ) {
```
tomlinton, line 55: Left out // DAI comment here like the others

I went through and checked them all :grin:

tomlinton: NM i see where it is

tomlinton (approved):
Looks wholesome to me

PR #580, opened 2021-03-02
DO NOT MERGE - Mathis/keeper

cipherzzz:
**Description**:
Just a WIP - I wanted to get a sign off on the basic logic before proceeding with figuring out the tests and how to mock the OUSD and VaultCore contracts. @DanielVF @tomlinton @sparrowDom @franckc 

**Test**:
`yarn run test ./test/keeper/keeper.js`

### DIFF
--- a/contracts/contracts/keeper/Keeper.sol
+++ b/contracts/contracts/keeper/Keeper.sol
@@ -2,6 +2,8 @@ pragma solidity 0.5.11;
 
 pragma experimental ABIEncoderV2;
 
+import "hardhat/console.sol";
+
 /**
  * @title OUSD Keeper Contract
  * @notice The Keeper is used to perform maintenance on vaults
@@ -16,25 +18,40 @@ contract Keeper is IKeeper, Governable {
 
   VaultCore vaultCore;
 
-  struct KeepInfo {
-        AllocateInfo allocate;
-        RebaseInfo rebase;
-    }
+  string CHECK  = 'check';
+  string EXECUTE  = 'execute';
 
-  struct AllocateInfo {
-        bool shouldRun;
-        string allocateSpecificParam;
+  string ALLOCATE  = 'allocate';
+  string REBASE  = 'rebase';
+
+  struct KeepInfo {
+        ActionInfo rebaseInfo;
+        ActionInfo allocateInfo;
+    }  
+
+  struct ActionInfo {
+        string actionType;
+        bool run;
+        string param;
     } 
 
   struct RebaseInfo {
-        bool shouldRun;
-        string rebaseSpecificParam;
-    }   
+        bool run;
+        string param;
+    }     
+
+  struct DummyInfo {
+        bool rebase;
+        bool allocate;
+        uint256 time;
+  } 
 
   event KeeperEvent(
-        KeepInfo info,
-        uint256 time
-    );  
+        string action,
+        bool success,
+        bool rebase,
+        bool allocate
+    );    
 
   // /*
   //  * @notice modifier that allows it to be simulated via eth_call by checking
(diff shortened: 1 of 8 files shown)


contracts/contracts/keeper/Keeper.sol
```diff
@@ -0,0 +1,194 @@
...
+ * @notice The Keeper is used to perform maintenance on vaults
+ * @author Origin Protocol Inc
+ */
+
+import { IKeeper } from "../interfaces/IKeeper.sol";
+import "../vault/VaultCore.sol";
+import "../token/OUSD.sol";
+import "../governance/Governable.sol";
+
+contract Keeper is IKeeper, Governable {
```
DanielVF, line 18: We might want to call this contract something like "KeeperConnector", since otherwise it could get confused with with the Keeper network itself.

contracts/contracts/keeper/Keeper.sol
```diff
@@ -0,0 +1,194 @@
...
+    }
+
+    function performUpkeep(bytes calldata dynamicData) external {
+        KeepInfo memory keepInfo = abi.decode(dynamicData, (KeepInfo));
+
+        require(
+            keepInfo.rebaseInfo.run || keepInfo.allocateInfo.run,
+            "No keeper actions are enabled"
+        );
+        require(
```
DanielVF, line 135: We could probably simplify this and just have performUpkeep check each task, and then execute if it needs to operate. This avoids having to trust user input, and avoids the overhead of passing information about what to execute out and back into the contract. As it is, an attacker could call "allocate" if a rebase was ready to execute. This wouldn't be an actual security problem since those are public anyway, but its aways worth being paranoid.

contracts/contracts/keeper/Keeper.sol
```diff
@@ -0,0 +1,194 @@
...
+        uint256 vaultValue = vaultCore.totalValue();
+        uint256 ousdSupply = oUSD.totalSupply();
+
+        if (ousdSupply == 0) {
+            return false;
+        }
+
+        bool rebase = (vaultValue + rebaseThreshold) > ousdSupply;
+
+        console.log("*** shouldRebase ***");
```
DanielVF, line 160: Logging for the win during the development!

contracts/contracts/keeper/Keeper.sol
```diff
@@ -0,0 +1,194 @@
...
+
+contract Keeper is IKeeper, Governable {
+    VaultCore vaultCore;
+    OUSD oUSD;
+
+    string CHECK = "check";
+    string EXECUTE = "execute";
+
+    string ALLOCATE = "allocate";
+    string REBASE = "rebase";
```
tomlinton, line 26: Strings can be constants I think

contracts/contracts/keeper/Keeper.sol
```diff
@@ -0,0 +1,194 @@
...
+
+    function shouldRebase() internal view returns (bool) {
+        uint256 vaultValue = vaultCore.totalValue();
+        uint256 ousdSupply = oUSD.totalSupply();
+
+        if (ousdSupply == 0) {
+            return false;
+        }
+
+        bool rebase = (vaultValue + rebaseThreshold) > ousdSupply;
```
tomlinton, line 158: We generally try to use SafeMath everywhere, even if not really required :)

contracts/contracts/keeper/Keeper.sol
cipherzzz: Done - I agree

contracts/contracts/keeper/Keeper.sol
cipherzzz: I THINK I did what you suggested. All of my abi encode/decode work out the window! 😭  I'm all for simpler though...

contracts/contracts/keeper/Keeper.sol
cipherzzz: 🍻

contracts/contracts/keeper/Keeper.sol
cipherzzz: yep that came up in the `slither` - resolved

contracts/contracts/keeper/Keeper.sol
cipherzzz: done- thanks

DanielVF (commented):
Looks like you are on the right track!

contracts/contracts/keeper/Keeper.sol
```diff
@@ -0,0 +1,152 @@
...
+    //   console.log(tx.origin);
+    //   require(tx.origin == address(0), "only for simulated backend");
+    // }
+
+    constructor(address payable _vaultCoreAddr, address payable _ousdAddr)
+        public
+    {
+        vaultCore = VaultCore(_vaultCoreAddr);
+        oUSD = OUSD(_ousdAddr);
+        nextAllocate = now + 1 days;
```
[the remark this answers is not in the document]
cipherzzz: done

franckc:
Old PR. Closing.

Issue #260, opened 2020-09-27
Sticky, Linkable Language Selection


micahalcorn:
If a user selects a language other than English and then refreshes the DApp, her selection is lost. We also should be able to link to a particular non-English version. On the [website](https://github.com/OriginProtocol/origin-website/), we handle both of these cases by supporting a locale param in the URL.

topoface:
@micahalcorn 
I am sorry, would you please give me the link to the DApp you mentioned?

topoface:
@micahalcorn 
>If a user selects a language other than English and then refreshes the DApp, her selection is lost.
This is fixed but can't make a PR because I have no write permission.
>We also should be able to link to a particular non-English version.
I am not sure why `fbt-runtime` is chosen for this, I don't see the possibility that we can perform the above unless we use next-i18n or others.

topoface:
@micahalcorn Would you please check this PR?
https://github.com/OriginProtocol/origin-dollar/pull/583
A simple fix.

micahalcorn:
Thanks, we probably inherited fbt-runtime from our pre-Next days. @sparrowDom might have some opinions on how it compares to next-i18n.

sparrowDom:
yup exactly like @micahalcorn says we've inherited that from previous projects and it is not really the ideal tool. Especially because of the lack of server side translations and as mentioned linking to a specific language.

It would be great to transition to  `next-i18n`. This probably involves: 
- combing over the project and migrating all source files with translation strings
- modifying Crowdin scripts that: 
  - push new translation strings to Crowdin
  - pull translated strings from Crowdin

topoface:
@sparrowDom 
So you want transition, okay!
I will take care of it.

sparrowDom:
@topoface you rock 🎸  ! Thank you

PR #583, opened 2021-03-08
remember locale settings

topoface:
PR for https://github.com/OriginProtocol/origin-dollar/issues/260

The reason of this issue is that it was not re-initialize the translation after fetching saved locale infos from local storage.

### DIFF
--- a/dapp/pages/_app.js
+++ b/dapp/pages/_app.js
@@ -103,9 +103,10 @@ function App({ Component, pageProps, err }) {
     }
   }, [])
 
-	useEffect(() => {
-		if (localStorage.locale) {
-			setLocale(localStorage.locale)
+  useEffect(() => {
+    if (localStorage.locale) {
+      setLocale(localStorage.locale)
+      setUtilLocale(localStorage.locale)
 		}
   }, [])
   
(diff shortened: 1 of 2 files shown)


dapp/pages/_app.js
```diff
@@ -103,9 +103,10 @@ function App({ Component, pageProps, err }) {
     }
   }, [])
 
-	useEffect(() => {
-		if (localStorage.locale) {
-			setLocale(localStorage.locale)
+  useEffect(() => {
+    if (localStorage.locale) {
+      setLocale(localStorage.locale)
+      setUtilLocale(localStorage.locale)
```
sparrowDom, line 109: The `setUtilLocale` code triggers a `analytics.track` event, since it doesn't predict that it can be called on initialise. Can you add a boolean param to the utils `setLocale` that will distinguish between initialise call and user initiated call. And not track the analytics event when it has been called during initialisation?

sparrowDom (commented):
Thanks for doing this 🙏 One comment inline

dapp/pages/_app.js
topoface: @sparrowDom 
Thank you! I have just updated it.
Can you please check the latest update?

sparrowDom: Look good thanks for fixing it

PR #587, opened 2021-03-29
Adds SingleAssetStaking liquidity check

mikeshultz:
Adds a missing liquidity check to SingleAssetStaking

Contract change checklist:
  - [ ] Unit tests pass

### DIFF
--- a/contracts/test/staking/single-asset-staking.js
+++ b/contracts/test/staking/single-asset-staking.js
@@ -157,11 +157,11 @@ describe("Single Asset Staking", function () {
     ).to.be.revertedWith("Only token contract can make this call");
 
     // This generate the data needed for calling stakeWithSender
-    const interface = ognStaking.interface;
+    const iface = ognStaking.interface;
     const fragment = ognStaking.interface.getFunction(
       "stakeWithSender(address,uint256,uint256)"
     );
-    const fnSig = interface.getSighash(fragment);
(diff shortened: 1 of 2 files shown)


tomlinton (approved):

contracts/test/staking/single-asset-staking.js
```diff
@@ -157,11 +157,11 @@ describe("Single Asset Staking", function () {
     ).to.be.revertedWith("Only token contract can make this call");
 
     // This generate the data needed for calling stakeWithSender
-    const interface = ognStaking.interface;
```
dagumak, line 160: Is this abbreviation a common convention?

mikeshultz: iface?   No idea, but `interface` is a reserved word in JS, hence the change.

dagumak: @mikeshultz Ah I see! Thank you for explaining that. How was that able to work before if `interface` is a reserved word?

mikeshultz: I could be wrong, but I think `interface` was introduced in ES6, which probably explains why we haven't see any issues.  I changed it out of an abundance of caution and to prevent any future conflicts.

DanielVF (approved):

PR #588, opened 2021-03-31
Reduce redeem gas costs by 719K/$319 by removing unnecessary totalValues()

DanielVF:
The single most expensive individual operation that the value does during a redeem is to check the total number of stablecoins deposited. This requires looping through each asset, and then for each asset, looping through all strategies and calculating the total value of each asset.

image

Unfortunately, the current code is doing this four times per redeem:

1. Pre-redeem rebase
2. Check to make sure the vault is solvent
3. Get totals of each asset to know how to proportionaly split the redeem
4. Post-redeem rebase 

The first three times are all checking the exact same underlying data - nothing in the balances changes on these checks until we actually send the user their funds between steps three and four. 

This PR reduces the number of times the total vault value is calculated per redeem to two.

First, it removes the pre-reedem rebase. This was necessary for the safe of the protocol when the rebases could rebase down. Now, however, it only server to ensure that any unrebased funds are credited to the user before the withdrawal. However, at 400K gas for this rebase, it is costing users an extra $178 per redeem to run this. Almost all of the time this is not a good tradeoff for the user.

Secondly, we reuse the totals acquired for calculating the stablecoins to give to the user, and use them for solvency check. There is no change to the underlying state during this, so the value generated is the same.

In the end, two checks, rather than four, and a **719K/$319 savings**.

image

Our tests pass unmodified, though I did need to update a mock to correctly pass through the test-set total value.

Contract change checklist:
  - [x] Unit tests pass


cipherzzz:
@DanielVF - what is this wizardry? What tool are you using that spits out that graph?

tomlinton:
Lovely!

DanielVF:
@cipherzzz It's The graph is coming from tenderly. I just pulled up a past large transaction to have it look at.

DanielVF:
@franckc This one is ready for your review. I figure we can merge to master now and deploy later with other changes.

franckc:
I have a bunch of other things and won't likely get to it in the next few days :(

@crazybuster Could you do a thorough code review? 

@DanielVF Could you add a migration script for this deployment?

Thanks to both of you :)

DanielVF:
@franckc Will make the migration script.

crazybuster (approved):
looks good to me

PR #589, opened 2021-05-06
Daniel vf/caching and static pages

DanielVF:
- Don't strip our precompiled pages from the deploy
- Add cache control headers to allow pages to cached by both GAE and Clouldfront.
- Allow deploying to staging without a Google Analytics ID


sparrowDom (approved):
LGTM

PR #594, opened 2021-05-19
Add an unpauseRebase option to the propose script

franckc:
unpauseRebase is not under the Strategist permissions so for now we still need to call it via the governor.


DanielVF (approved):
👍

Issue #496, opened 2021-01-12
Aave V2 Migration


micahalcorn:
It's getting time to migrate to AAVE v2. Interest rates are higher, liquidity is bigger and gas cost is supposed to be cheaper.

The API has changed in AAVE v2, so we will need to deploy a new strategy contract.

## API changes

_Our current [AAVE strategy code](https://github.com/OriginProtocol/origin-dollar/blob/6e2b1738fe876aa8e7e6d38c23417383ed33b3ad/contracts/contracts/strategies/AaveStrategy.sol). New v2 AAVE [lending pool code](https://github.com/aave/protocol-v2/blob/ice/mainnet-deployment-03-12-2020/contracts/protocol/lendingpool/LendingPool.sol)._

Deposit method has changed from 

```
deposit(address _reserve, uint256 _amount, uint16 _referralCode)
```

to: 

```
deposit(
    address asset,
    uint256 amount,
    address onBehalfOf,
    uint16 referralCode
  )
```

Withdraw method has changed be **on the pool contract** instead of on the aToken:

I think the method signature remains the same, but we should check that we match the new:

```
withdraw(
    address asset,
    uint256 amount,
    address to
  )
```

Balance checking remains the same, just using the new atoken.

We'll need to add the ability to set the `platformAddress` storage slot in the AaveStrategy. Unless I'm missing something obvious, I don't see a setter. Alternately, we could just hardcode it in the new contract.

## Governance Proposal

We'll need a governance proposal to:

1. Pull funds out from strategy - `VaultAdmin.withdrawAllFromStrategy(0x9f2b18751376cF6a3432eb158Ba5F9b1AbD2F7ce)`
2. Upgrade strategy
3. Set atoken AAVEStrategy.setPTokenAddress(0x6b175474e89094c44da98b954eedeac495271d0f, 0x028171bca77440897b824ca71d1c56cac55b68a3 )
4. Set platform
AAVEStrategy.setPlatformToken(0x7d2768de32b0b80b7a3454c06bdac94a69ddc7a9)
5. Reallocate Funds - VaultCore.allocate()

## Token cheatsheet

|V|Name|Address|
|-|-----|--|
|v1|aDAI|0xfC1E690f61EFd961294b3e1Ce3313fBD8aa4f85d|
|**v2**|aDAI| 0x028171bca77440897b824ca71d1c56cac55b68a3 |
|v1|LendingPool|0x398ec7346dcd622edc5ae82352f02be94c62d119|
[...]

PR #601, opened 2021-06-16
Upgrade AAVE strategy to AAVE v2

DanielVF:
Upgrade AAVE strategy to use AAVE v2.

[Discussion for upgrade](https://github.com/OriginProtocol/origin-dollar/issues/496)

We should plan on rolling this out as a new strategy address, just to guarantee a clean slate and that all addresses/state are correct. We'll remove the old strategy, then add this strategy.

### Progress:

- [x] Interface changes
- [x] Mock changes
- [x] Contract changes
- [x] Tests pass
- [x] Deployment written
- [x] Deployment tested on fork with fork funds

### Contract change checklist:
  - [x] Unit tests pass


DanielVF:
Will write deploy script once we have the new deploy library running and merged.

DanielVF:
## Notes on AAVE Liquidity rewards:

- Official [Liquidity Mining Docs](https://docs.aave.com/developers/guides/liquidity-mining#setclaimer)
- Official [Staking Docs](https://docs.aave.com/developers/protocol-governance/staking-aave)
- rewards contract: 0xd784927Ff2f95ba542BfC824c8a8a98F3495f6b5
- stkAAVE: 0x4da27a545c0c5b758a6ba100e3a049001de870f5

### 1. Get rewards amount

`rewards.getRewardsBalance(address[] calldata assets, address user)`

### 2. Convert rewards into stkAAVE

`rewards.claimRewards(address[] calldata assets, uint256 amount, address to)`

### 3. Trigger cooldown

stkAAVE cannot be turned into AAVE without a five day cool down. Sending more funds in restarts the cool down, as does claiming rewards. Additionally, two days after the cool down unlocks, it locks again automatically. To restart the cool down clock:

`stkAAVE.cooldown()`


To find out your current cool down start time:

`stkAAVE.cooldown.stakersCooldowns(address(this))`

### 4. Convert to AAVE

When the time is right:

```
stkAave.claimRewards(address(this), type(uint256).max) 
```

```
stkAave.redeem(address(this), stkAaveBalance)
```

### All together

Because moving funds or claiming rewards resets the cool down, the order matters a little bit

1. First we convert stkAave.redeem to AAVE
2. Then we claim rewards from stkAave
3. Transfer AAVE to the vault to be sold
4. start the cooldown for the next round.

franckc (approved):
LGTM

DanielVF:
## Testing out AAVE Staking


```javascript
stkAave = Contract.from_explorer("0x4da27a545c0c5b758a6ba100e3a049001de870f5")
aave = Contract.from_explorer("0x7fc66500c84a76ad7e9c93437bfc5ac33e2ddae9")
incentivesController = Contract.from_explorer("0xd784927Ff2f95ba542BfC824c8a8a98F3495f6b5")
VAULT = "0xE75D77B1865Ae93c7eaa3040B038D7aA7BC02F70"
ADAI_V2 = '0x028171bca77440897b824ca71d1c56cac55b68a3'
ME = "0x4a49985b14bd0ce42c25efde5d8c379a48ab02f3"
FROM_ME = {'from': ME}


stkAave.cooldown(FROM_ME)
chain.sleep(stkAave.COOLDOWN_SECONDS()+100)
chain.mine()

cooldown = stkAave.stakersCooldowns(ME)
windowStart = cooldown + stkAave.COOLDOWN_SECONDS()
windowEnd = windowStart + stkAave.UNSTAKE_WINDOW()
currentTimestamp = chain.time()
currentTimestamp > windowStart and currentTimestamp < windowEnd

stkAaveBalance = stkAave.balanceOf(ME)
rewardLiquidationThreshold = 50
stkAaveBalance > rewardLiquidationThreshold

stkAave.redeem(ME, stkAaveBalance, FROM_ME)
stkAave.balanceOf(ME)

aaveBalance = aave.balanceOf(ME)
aave.transfer(VAULT, aaveBalance, FROM_ME)

pendingRewards = incentivesController.getRewardsBalance([ADAI_V2], ME)
incentivesController.claimRewards([ADAI_V2], pendingRewards, ME, FROM_ME)
```

DanielVF:
Now we have tests, next up is deploy scripts.

DanielVF:
Code for testing rewards

```python
from world import *
from addresses import *
vault_core.totalValue()
OLD_AAVE_STRAT = AAVE_STRAT
NEW_AAVE_STRAT = "0xaB5E7B701B605f74AaC1b749Fd50715f0DEd7Bc5"
COMPOUND_STRAT = '0xD5433168Ed0B1F7714819646606DB509D9d8EC1f'
AAVE_INCENTIVES_CONTROLLER = "0xd784927Ff2f95ba542BfC824c8a8a98F3495f6b5"
ADAI = "0x028171bca77440897b824ca71d1c56cac55b68a3"
new_aave_strat = Contract(NEW_AAVE_STRAT)
incentives = Contract(AAVE_INCENTIVES_CONTROLLER)
stkAave = Contract('0x4da27a545c0c5b758a6ba100e3a049001de870f5')
aave = Contract('0x7Fc66500c84A76Ad7e9c93437bFc5Ac33E2DDaE9')
adai = Contract(ADAI)
brownie.network.web3.provider.make_request('hardhat_impersonateAccount', ['0x830622bdd79cc677ee6594e20bbda5b26568b781'])
brownie.network.web3.provider.make_request("evm_snapshot", [])

new_aave_strat.checkBalance(DAI) / 1e18
vault_admin.reallocate(COMPOUND_STRAT, NEW_AAVE_STRAT, [DAI], [int(1000000*1e18)], {'from': GOVERNOR})
vault_admin.reallocate( NEW_AAVE_STRAT, COMPOUND_STRAT, [DAI], [int(1000000*1e18)], {'from': GOVERNOR})
vault_admin.reallocate(COMPOUND_STRAT, NEW_AAVE_STRAT, [DAI], [int(2000000*1e18)], {'from': GOVERNOR})
new_aave_strat.checkBalance(DAI) / 1e18

for i in range(1,30):
  print("---- DAY %d" %i)
  #brownie.network.web3.provider.make_request('evm_mine', [1627490205 + 86400 * i])
  brownie.network.web3.provider.make_request('evm_increaseTime', [86400])
  print("USDT %d" % (usdt.balanceOf(vault_core) / 1e6))
  print("STK %d" % (stkAave.balanceOf(NEW_AAVE_STRAT) / 1e18))
  print("DAI %d" % (new_aave_strat.checkBalance(DAI) / 1e18))
  print("Cooldown %d" % stkAave.stakersCooldowns(NEW_AAVE_STRAT))
  print("RB %d" % incentives.getRewardsBalance([ADAI], new_aave_strat))
  vault_admin.harvest(NEW_AAVE_STRAT, {'from':GOVERNOR})

#vault_admin.harvest(NEW_AAVE_STRAT, {'from':GOVERNOR})
```

DanielVF:
@franckc, @tomlinton, @mikeshultz. This code is ready for review again, small changes to the contract and to the deploy script. I was able to run success path fork testing.

contracts/contracts/strategies/AaveStrategy.sol
```diff
@@ -11,9 +11,47 @@ import {
     InitializableAbstractStrategy
```
tomlinton, line 11: Is there any merit in leaving this strategy untouched as a V1 integration and creating a new file AaveStrategyV2 or similar?

DanielVF: Only if we thought it likely that we would use AAVE v1 again. But we can aways resurrect this from the Git history if we really need it.

contracts/deploy/020_aave_v2.js
```diff
@@ -0,0 +1,102 @@
...
+    );
+    // 3. Init the proxy to point at the implementation
+    await withConfirmation(
+      cAaveStrategyProxy
+        .connect(sDeployer)
+        ["initialize(address,address,bytes)"](
+          dAaveStrategyImpl.address,
+          deployerAddr,
+          []
+        )
```
franckc, line 51: missing the getTxOpts() which is super helpful to speed up tx by bumping up gas price
same comment for the other calls to withConfirmation that don't include it in this file.

franckc (commented):
Reviewed the deploy script - minor comment inline otherwise lgtm.
I haven't reviewed the contract piece yet.

contracts/deploy/020_aave_v2.js
DanielVF: Updated. Thanks!

contracts/contracts/strategies/AaveStrategy.sol
```diff
@@ -170,19 +213,71 @@ contract AaveStrategy is InitializableAbstractStrategy {
...
+
+        // Check staked AAVE cooldown timer
+        uint256 cooldown = stkAave.stakersCooldowns(address(this));
+        uint256 windowStart = cooldown + stkAave.COOLDOWN_SECONDS();
+        uint256 windowEnd = windowStart + stkAave.UNSTAKE_WINDOW();
+        uint256 currentTimestamp = now;
+
+        // If inside the unlock window, then we can redeem stkAave
+        // for AAVE and send it to the vault.
+        if (currentTimestamp > windowStart && currentTimestamp < windowEnd) {
```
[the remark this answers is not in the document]
DanielVF: Thanks. Corrected.

contracts/contracts/strategies/AaveStrategy.sol
```diff
@@ -170,19 +213,71 @@ contract AaveStrategy is InitializableAbstractStrategy {
...
+    function collectRewardToken() external onlyVault nonReentrant {
+        if (address(stkAave) == address(0)) {
+            return;
+        }
+
+        // Check staked AAVE cooldown timer
+        uint256 cooldown = stkAave.stakersCooldowns(address(this));
+        uint256 windowStart = cooldown + stkAave.COOLDOWN_SECONDS();
+        uint256 windowEnd = windowStart + stkAave.UNSTAKE_WINDOW();
+        uint256 currentTimestamp = now;
```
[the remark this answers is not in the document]
DanielVF: I prototyped this in python, then converted it to solidity, so I had this as a var since it was an external RPC call at the time.

Turns out "now" is only 2 gas, so I've just replaced currentTimestamp with it through the code.

Thanks!

DanielVF:
@tomlinton, could you take a look at the contract side of this when you have a chance?

contracts/contracts/strategies/AaveStrategy.sol
```diff
@@ -170,19 +213,71 @@ contract AaveStrategy is InitializableAbstractStrategy {
...
+            uint256 stkAaveBalance = stkAave.balanceOf(address(this));
+            if (stkAaveBalance > rewardLiquidationThreshold) {
+                stkAave.redeem(address(this), stkAaveBalance);
+            }
+            // Transfer AAVE to vaultAddress
+            uint256 aaveBalance = IERC20(rewardTokenAddress).balanceOf(
+                address(this)
+            );
+            if (aaveBalance > 0) {
+                IERC20(rewardTokenAddress).safeTransfer(
```
tomlinton, line 241: The Uniswap path that we use to liquidate these is hardcoded as reward token -> weth -> usdt. Did you check if that is suitable here (and for Compound for that matter). My guess is that it is, but I've always had a bit of an uneasy feeling about it.

tomlinton:
Not sure if I previously suggested this, but any thoughts on maintaining the v1 strategy as is (in the codebase) and adding a new file for the v2 strategy stuff? Might be more pain than it is worth.

contracts/contracts/strategies/AaveStrategy.sol
```diff
@@ -170,19 +213,71 @@ contract AaveStrategy is InitializableAbstractStrategy {
...
+                    vaultAddress,
+                    aaveBalance
+                );
+            }
+        }
+
+        // If we were past the start of the window,
```
tomlinton, line 248: This comment is a bit confusing. I think it needs to be updated to mention claiming of rewards. "If we have passed the start of the window, claim any outstanding rewards. If the cooldown counter is not running then start it."

tomlinton:
I think the reason I suggested the v1 vs v2 thing is that when I review the changes here I've got the blinders on with regard to how the rest of the strategy works because the stuff we originally had doesn't show up in the diff, e.g. checkBalance. Fortunately it's all reasonably simple.

contracts/contracts/strategies/AaveStrategy.sol
DanielVF: I checked and that's currently the best route on both AAVE and COMP. Looks like were are good.

DanielVF:
> Not sure if I previously suggested this, but any thoughts on maintaining the v1 strategy as is (in the codebase) and adding a new file for the v2 strategy stuff? Might be more pain than it is worth.

Yeah, maybe talked in Discord. I'd rather keep the codebase simple, and we can all ways get the old strategy back from git in the unlikely event that we need it.

contracts/contracts/strategies/AaveStrategy.sol
```diff
@@ -70,7 +108,12 @@ contract AaveStrategy is InitializableAbstractStrategy {
 
         IAaveAToken aToken = _getATokenFor(_asset);
         emit Withdrawal(_asset, address(aToken), _amount);
-        aToken.redeem(_amount);
+        uint256 actual = _getLendingPool().withdraw(
+            _asset,
+            _amount,
+            address(this)
+        );
+        require(actual >= _amount, "Did not withdraw enough");
```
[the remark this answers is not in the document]
DanielVF: I guess this was just a cheap sanity check.  Do you think it would be better to do `actual == amount`?

contracts/contracts/strategies/AaveStrategy.sol
DanielVF: Updated the comment.

contracts/contracts/strategies/AaveStrategy.sol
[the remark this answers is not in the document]
mikeshultz: IMO, I think that would be more appropriate since `actual` should never be greater than `_amount`.  I was also suggesting the revert message wasn't accurate.

tomlinton (approved):
👍

DanielVF:
#### Requirements

This PR moves from AAVE v1 to AAVE v2, and adds support for the AAVE v2 rewards.

Sheesh StkAAVE is a pain.

### Deployment Considerations

AAVE v2 rewards collection requires that the collectRewardToken() method on the strategy be called by the vault on the correct times to hit the unstaking window. We may need a keeper contract to do this.


#### Internal State

No internal state is modified during operation. Contract storage slots are configuration. All dynamic state is looked up from external sources.

💡 If we were using a more recent solidity version, we could `immutable` all these configuration addresses into the contract, and save gas on everything the contract does.

#### Attack

AAVE very difficult to misuse.

Funds are only ever transferred out to the vault, or to the address specified by the vault.

Balances are a simple 1-1 mapping of aToken balances.


#### Logic

💡 We aren't handling collecting the small amounts of StkAAVE interest that could be generated from the week worth of StkAAVE we'll build up between converting it to AAVE. If AAVE reintroduces rewards mining, we can upgrade to support it.


#### Tests

#### Flavor

_Could reorder methods for more standardization. Will do later though, if we do._

#### Overflow

#### Black magic

#### Authentication

- [x] Never use tx.origin

#### Cryptographic code

_No crypto code used._

#### Gas problems

    - [ ] A way to remove items

#### External calls

- [x] Unsafe external calls. _AAVE and trusted stables only for external calls_
- [x] Malicious behaviors. _If AAVE is malicious, problems ensue._
- [x] Could fail from stack depth problems (low level calls much require success)
- [x] Oracles? _none_
- [x] Oracles, one of:
  - [x] No oracles
  - [ ] Oracles can't be bent

#### Ethereum

- [x] Contract does not use msg.value

Issue #590, opened 2021-05-10
Increase OUSD credits resolution


DanielVF:
OUSD currently cannot mint to Uniswap v3.

### Background:

OUSD stores balances as "credits" that back each account. These value of these credit go up as OUSD rebases. When transferring funds from one account to another, the OUSD transfer method subtracts a number of credits from the sender and  adds credits to the receiver account.

Unfortunately the resolution on these credits to token ratio is larger than the 18 decimals resolution of OUSD balances. It's like trying to making dollar exact change using only bills of a $1.13 face value. This means that OUSD must either round up or down or round down when converting to credits and back up. Implementing either up or down will almost always result in either more or less being transferred than was requested. This rounding error affects both the sender account and the receiver account.

Currently, we round down for credit calculations, resulting in the sender receiving less than the amount requested to be transferred.
  
    >>> ousd.balanceOf(TARGET)
    0
    >>> ousd.transfer(TARGET, 10000, {'from': SENDER})
    ....
    >>> ousd.balanceOf(TARGET)
    9999

This is fairly unintuitive, and breaks Uniswap v3's mint checks, which insure that the exact amount requested to transfer in was transferred in.

### Options: 

Off the top of my head, some options for fixing this:

A. Add 1/1e18 credit to the receiver on each transfer for more than a credit. Extremely minimal change to code. This will make the ending receiver balance in OUSD be equal to or greater than. This will increase the actual supply on each transfer (but by insanely, unimaginable amounts). Disadvanage is that if we want to keep the accounting perfect, we also need to write this value change to the the total supply, which costs one write's worth of gas per transfer.

B. We could round up when computing transfers. While this takes makes the receiver's account be correct, the it does mean that the sender's accounting is less, and maybe?might? fail a legit transfer. A sender with 30000 raw OUSD would be unable to make 3x 10000 transfers.
 
C. We could take OUSD offline and increase the resolution of credits. This is by far the most work. If each credits was worth only a fraction of what it is now, then the increased resolution could mean that this would almost always work out to the right answer.

joshfraser:
Thanks for looking into this. Is there a magic amount of liquidity that we could provide that wouldn't have rounding issues? Uniswap v3 is really ideal for trading OUSD but I understand that we're strapped for bandwidth right now.

At first glance, option C seems like the "correct" long-term solution.

[...]

PR #603, opened 2021-06-18
Use high resolution credits on new contract accounts.

DanielVF:
We are targeting [increasing the credits resolution for OUSD](https://github.com/OriginProtocol/origin-dollar/issues/590). Our immediate motivation is Uniswap V3 support, which requires balance-exact transfers in order to work. Upgrading all accounts will require substantial work.

However, it turns out that a new, non-rebasing account has its own independent creditsPerToken ratio, and has no effect on OUSD's bookkeeping values This means that it is trivial to set the new account to start at a high resolution creditsPerToken ratio, and as a result transfers in and out of that account will be balance exact.

Contract change checklist:
  - [x] Unit tests pass


contracts/contracts/token/OUSD.sol
```diff
@@ -382,12 +382,22 @@ contract OUSD is Initializable, InitializableERC20Detailed, Governable {
...
     function _ensureRebasingMigration(address _account) internal {
         if (nonRebasingCreditsPerToken[_account] == 0) {
-            // Set fixed credits per token for this account
-            nonRebasingCreditsPerToken[_account] = rebasingCreditsPerToken;
-            // Update non rebasing supply
-            nonRebasingSupply = nonRebasingSupply.add(balanceOf(_account));
-            // Update credit tallies
-            rebasingCredits = rebasingCredits.sub(_creditBalances[_account]);
+            if (_creditBalances[_account] == 0) {
+                // Since there is no exisiting balance, we can directly set to
```
tomlinton, line 386: Typo in comment

tomlinton:
Great idea, I was a bit worried about opt ins/outs but running through those it looks like it'll be fine.

franckc (commented):
Logic looks good. Neat trick :)

Couple questions:
 1. remind me, how did we pick the 1e27 resolution? (just to double check this is what we want).
 2. is there any chance this make fixing the rest of the accounts (which I assume is the majority) more challenging since we'll have to check if an account uses high resolution or not?

Let's add a migration file. Either in this PR or in a separate PR. Up to you.

contracts/contracts/token/OUSD.sol
DanielVF: Fixed. Thanks!

DanielVF:
1e27 is a standard "RAY" in DAI world, and is used by AAVE v2 tokens as their standard internal measurement. Since our public balances is 1e18, this gives us roughly 1e9 - 1e1 = 1e8, or letting OUSD grow 100,000,000x before we run into the same resolution problems as we had before. At 20% growth per year, this would last for 101 years. 

The other side is how much total supply this would support. 

- 2^256 is roughly 1e77.
- 1e77 (underlying) - 1e27 (decimal point) is 1e50 dollars
- World circulating currency and bank accounts is roughly 37 trillion or 1e14
- If the dollar inflation averages 20% per year, for 101 years, that's 1e8
- So OUSD would be able to support 1e50 - 1e14 (world currency) - 1e8 (inflation) = 1e28 or 10 Octillions times the world's money for 100 years.

So a little bit overkill on that side, and we might honestly be able to split the difference and do our internal resolution at 1e40, I just went with 1e27 since it's common for this kind of thing.

franckc:
Thanks for the details about 1e27 resolution :) Love the back of the envelope calculation. Clearly shows we should be good for a while. Next will look at the new migration...

contracts/utils/deploy.js
```diff
@@ -236,6 +236,56 @@ const sendProposal = async (proposalArgs, description) => {
   log("Done");
 };
 
+async function deployment(deployName, fn) {
```
franckc, line 239: Maybe also add an optional dependencies (list of strings) argument.
And have something like
```
if (dependencies && dependencies.length > 0) {
  main.dependencies = dependencies.
}
```

franckc: nit: add jsdoc for the new 'deployment' method

contracts/utils/deploy.js
```diff
@@ -245,4 +295,5 @@ module.exports = {
   executeProposal,
   executeProposalOnFork,
   sendProposal,
+  // deployment
```
franckc, line 298: uncomment?

franckc (approved):
LGTM

I like the refactor of the deployment code to reduce the amount of boilerplate code we havae to write moving forward. Nice! :)

contracts/utils/deploy.js
DanielVF: Thanks, added.

contracts/utils/deploy.js
DanielVF: Fixed, thanks!

PR #604, opened 2021-06-21
Hardhat packages upgrade

franckc:
Upgrade hardhat versions.


tomlinton (approved):

franckc:
Couple of tests are failing... not sure why. Needs to be investigated before this gets merged in...

DanielVF (approved):

DanielVF:
I've debugged the failing tests. Hardhat slightly changed the format that it uses in reporting exceptions. I've updated the two tests to just match the actual revert message from the contract, rather than also looking at the hardhat added error message as well.

PR #614, opened 2021-07-07
Punk1004/tx details

jupiteruwdev:

Contract change checklist:
  - [ ] Unit tests pass


franckc:
@sparrowDom Do you mind reviewing those changes since you are the expert for this code? :)

dapp/src/components/buySell/SellWidget.js
```diff
@@ -221,10 +219,12 @@ const SellWidget = ({
         result = await vaultContract.redeemAll(minStableCoinsReceivedBN, {
           gasLimit,
         })
-        storeTransaction(result, `redeem`, returnedCoins, coinData)
+        receipt = await rpcProvider.waitForTransaction(result.hash)
+        const data = await getStableCoinLogs(receipt)
+
+        storeTransaction(result, `redeem`, returnedCoins, data)
```
sparrowDom, line 225: If I understand correctly this change would prevent transactions from showing up in the sidebar while they are not yet mined. We show transaction as soon as it is submitted, and update it once it is mined.

See [TransactionListener](https://github.com/OriginProtocol/origin-dollar/blob/6e2b1738fe876aa8e7e6d38c23417383ed33b3ad/dapp/src/components/TransactionListener.js) that is taking care of updating mined transactions.

TransactionListener would probably also be a better place where to populate the `getStableCoinLogs`

dapp/src/utils/utils.js
```diff
@@ -33,3 +36,52 @@ export function mergeDeep(target, ...sources) {
...
+  ]
+
+  await Promise.all(
+    stableCoins.map((coin) => {
+      const log = transactions.logs.find(
+        (log) => log.address.toLowerCase() === coin.address
+      )
+      if (log) {
+        const value = ethers.BigNumber.from(log.data)
+        data[coin.name] = (parseFloat(value.toString()) / 10) * coin.decimals
```
sparrowDom, line 81: there is a [formatCurrency](https://github.com/OriginProtocol/origin-dollar/blob/6e2b1738fe876aa8e7e6d38c23417383ed33b3ad/dapp/src/utils/math.js#L29) helper function that you can use to format values.

sparrowDom (commented):
Great job taking this ticket on! Left a couple of comments if you can take a look please

dapp/src/components/buySell/SellWidget.js
jupiteruwdev: Yes you're correct. in transactionListener, we can take care of every transactions whatever its type is. And that way, we can make it simple and also high performance as well. since in the transactionListener, it originally calling `waitForTransaction` method so we don't need to call this function twice.
But what I'm concerned is the transaction logs which are returned from `waitForTransaction` seems not correct on my local network. there is no stable coin addresses listed in our system at least.
Is it expected issue?

dapp/src/utils/utils.js
```diff
@@ -33,3 +37,52 @@ export function mergeDeep(target, ...sources) {
...
+export const getStableCoinLogs = async (transactions) => {
+  const data = {
+    usdt: 0,
+    dai: 0,
+    usdc: 0,
+    ousd: 0,
+  }
+  const stableCoins = [
+    {
+      address: addresses.mainnet.USDT.toLowerCase(),
```
franckc, line 54: Hardcoding those addresses to mainnet will cause this function to fail on any other network than mainnet. For example on local hardhat network we use for development.

How about leveraging the addresses already loaded in the react state by this method: https://github.com/OriginProtocol/origin-dollar/blob/dc9ec97fadb90b409be0f7749be97a9b149d1855/dapp/src/utils/contracts.js#L20

jupiteruwdev: @franckc to use this variables we need to put these variables inside store. but I don't see it's implemented yet.

sparrowDom: @jupiteruwdev for sharing a global state we are using `pullstate` instead of say Redux. Here is where the contract data gets stored into the global state: https://github.com/OriginProtocol/origin-dollar/blob/dc9ec97fadb90b409be0f7749be97a9b149d1855/dapp/src/utils/contracts.js#L351-L353

jupiteruwdev: i see

franckc:
Closing since there hasn't been any activity on this PR for a while.

PR #615, opened 2021-07-11
Sparrow dom/swap

sparrowDom:
- updates the UI as specified [in Zeplin](https://zpl.io/aw5n3LJ)
- when user inputs the value looks at 3 options (Uniswap, Vault, OUSD Swap (Flipper) and picks the best one
- renames `mint` to `swap`

Contract change checklist:
  - [ ] Unit tests pass


dapp/src/components/TransactionListener.js
```diff
@@ -152,10 +156,16 @@ const TransactionListener = ({ rpcProvider }) => {
     }
   }
 
-  const updateTransactions = (transactions) => {
-    const txHashes = transactions.map((t) => t.hash)
-    const otherTxs = transactions.filter((tx) => !txHashes.includes(tx.hash))
-    const newTransactions = [...otherTxs, ...transactions]
+  const updateTransactions = (transactionsToUpdate) => {
+    const txHashes = transactionsToUpdate.map((t) => t.hash)
+    /* With functional approach values inside a function scope can be out of date
```
sparrowDom, line 161: @tomlinton now I remember the reason I did the Tx listener in an Object oriented approach instead of functional. There were cases where multiple async functions would be executing simultaneously and sometimes values in those functions would be outdated. Lets say: 

2 routines simultaneously do some CRUD to transactions and store end result. The one finishing last will most probably override the results of the one finishing previously. Since the value of `transactions` has been copied to the function scope before the previous routine has modified it. You can see how on master (without this commit) any newer transaction overrides previous ones. 

Anyway having this component in functional approach with the use of `store.currentState` instead of OOO is still nicer IMHO.

dapp/src/components/buySell/ContractsTable.js
```diff
@@ -191,7 +221,11 @@ const ContractsTable = () => {
                     ? '-'
                     : `$${formatCurrency(estimation.effectivePrice, 2)}`}
                 </div>
-                <div className={`text-right w-18 ${redStatus ? 'red' : ''}`}>
+                <div
```
micahalcorn, line 224: @sparrowDom can we throw in a `pl-2` on these so that we don't end up with "-Amount too high"?

dapp/src/components/buySell/SwapHomepage.js
```diff
@@ -346,8 +346,8 @@ const SwapHomepage = ({
         }
       )
       setStoredCoinValuesToZero()
-      setSelectedBuyCoinAmount(0)
-      setSelectedRedeemCoinAmount(0)
+      setSelectedBuyCoinAmount('')
```
micahalcorn, line 349: @sparrowDom I was going to ask for this since it currently resets to "0" instead of the empty state. I assume that this commit just hasn't been deployed.

sparrowDom: Yup that is correct, have done this today, but not deployed it to staging

dapp/src/components/buySell/ContractsTable.js
sparrowDom: Yup ofc

sparrowDom: Never before have I worked with a project / product manager that would be able to tell me what lines of code I need to modify 👊

sparrowDom:
Merging this in since it has been extensively tested on staging. Happy to address any comments subsequently.

dapp/src/components/buySell/ContractsTable.js
micahalcorn: Probably best to ignore me most of the time 😬

sparrowDom: Oh no quite the oposite. Your helping is saving time and is welcome

PR #616, opened 2021-07-14
Snyk Vulnerability Scan - Github Action

cipherzzz:
Description:
This is a snyk github action to run upon the push of a feature branch. No contract change

Note:
There are some vulnerabilities that I'd like to be reviewed by the team before before ignoring them. Currently it is set to succeed despite the gate conditions.

### DIFF
--- a/.snyk
+++ b/.snyk
@@ -1,0 +1,10 @@
+# Snyk (https://snyk.io) policy file, patches or ignores known vulnerabilities.
+version: v1.19.0
+# ignores vulnerabilities until expiry date; change duration by modifying expiry date
+ignore:
+  SNYK-JS-UNDERSCORE-1080984:
+    - '*':
+        reason: None Given
+        expires: 2021-08-13T18:38:10.205Z
+        created: 2021-07-14T18:38:10.208Z
+patch: {}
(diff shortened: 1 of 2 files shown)


.snyk
```diff
@@ -0,0 +1,10 @@
+# Snyk (https://snyk.io) policy file, patches or ignores known vulnerabilities.
+version: v1.19.0
+# ignores vulnerabilities until expiry date; change duration by modifying expiry date
+ignore:
+  SNYK-JS-UNDERSCORE-1080984:
```
franckc, line 5: I checked and the "underscore" package is only used by the playground which is a dev tool and not used in prod. https://github.com/OriginProtocol/origin-dollar/blob/dc9ec97fadb90b409be0f7749be97a9b149d1855/playground/src/stores.js#L4

Seems fine to ignore. Or better we could upgrade without too much worries of breaking something in prod.

If you decide to ignore, it would be useful to document the reason in this .snyk file so that we don't have to look it up next time we wonder about this ignore...

franckc (approved):
One comment inline. PTAL. Otherwise LGTM

.snyk
cipherzzz: snyk ignored - `dev dependencies` so more than likely this module is in the wrong place. I ignored this one really just to generate the `.snyk` policy file, but you're right I should have put a description

PR #621, opened 2021-08-02
Gnosis Safe web3-react connector integration

tomlinton:
Integrate Gnosis Safe. If the safe is available it automatically connects to that and doesn't try anything else. It might be worth us removing the disconnect/other options if gnosis safe is in use. Also adds a rebase opt in message if they are not opted in.


dapp/src/components/layout.js
```diff
@@ -24,7 +29,39 @@ const Layout = ({
...
+  const [showRebaseOptInNotice, setShowRebaseOptInNotice] = useState(false)
+  const ousdContract = useStoreState(ContractStore, (s) =>
+    get(s, 'contracts.ousd')
+  )
+
+  useEffect(() => {
+    async function checkRebaseState() {
+      if (ousdContract && account) {
+        const isSafe = !!get(connector, 'safe.safeAddress', false)
+        const rebaseOptInState = await ousdContract.rebaseState(account)
```
sparrowDom, line 45: BalanceHeader runs very similar code to this one. Should this checking logic be moved somewhere else (like `AccountListener` for example) and the result stored in Account or Contract stores?

dapp/src/utils/hooks.js
```diff
@@ -1,57 +1,103 @@
...
+  const [safeMultisigConnector, setSafeMultisigConnector] = useState(null)
+  const [connector, setConnector] = useState(null)
+  const [connectorIcon, setConnectorIcon] = useState(null)
+
+  // Instantiate Gnosis Safe web3-react connector
+  useEffect(() => {
+    // Requires global.window to be available to must be instantiated inside
+    // useEffect for Next.js to work
+    setSafeMultisigConnector(new SafeAppConnector())
+  }, []) // Do this on mount
```
sparrowDom, line 28: I think these to `useEffects` could be joined into one to reduce complexity with smth like this: 
```
useEffect(() => {
  if (!process.browser) return 

  const SafeMultisigConnector = new SafeAppConnector()
  ...

}, [process.browser])
```

dapp/src/utils/hooks.js
```diff
@@ -1,57 +1,103 @@
...
+
+  // Attempt to use Gnosis Safe Multisig if available
+  useEffect(() => {
+    async function attemptSafeConnection() {
+      // Safe multisig connector not yet initialised, wait for it
+      if (!safeMultisigConnector) return
+
+      // OK to use Gnosis Safe?
+      const canUseGnosisSafe =
+        safeMultisigConnector && (await safeMultisigConnector.isSafeApp())
```
sparrowDom, line 38: nit: `safeMultisigConnector` should always result to a valid object here so just: 
```
const canUseGnosisSafe = await safeMultisigConnector.isSafeApp()
```

dapp/src/utils/hooks.js
```diff
@@ -1,57 +1,103 @@
...
+  const [triedSafeMultisig, setTriedSafeMultisig] = useState(false)
+  const [safeMultisigConnector, setSafeMultisigConnector] = useState(null)
+  const [connector, setConnector] = useState(null)
+  const [connectorIcon, setConnectorIcon] = useState(null)
+
+  // Instantiate Gnosis Safe web3-react connector
+  useEffect(() => {
+    // Requires global.window to be available to must be instantiated inside
+    // useEffect for Next.js to work
+    setSafeMultisigConnector(new SafeAppConnector())
```
sparrowDom, line 27: I wonder if it would be cleaner adding this connector to `utils/connectors` to `connectorsByName`. Each item there would also need a `showOnLoginPanel`, where Gnosis would have that property to `false`. 

The one challenge would be initialisation of the connector since it needs to happen in browser environment.

dapp/src/utils/hooks.js
```diff
@@ -1,57 +1,103 @@
...
-            console.error(e)
-            setTried(true)
-          })
-      } else {
-        setTried(true)
+    async function attemptInjectedConnection() {
+      // Must try Safe multisig before injected connector
+      if (!triedSafeMultisig) return
+      // Already got a connector, don't try another, use icon as indicator
+      if (connectorIcon) return
```
sparrowDom, line 64: IMHO would be better to have a separate bool indicator for this. Since it is not obvious that changing `connectorIcon ` logic prevents this function from trying a non gnosis login.

sparrowDom (commented):
Great work, a couple of comments inline.

dapp/src/utils/hooks.js
tomlinton: That was my first approach but it broke the next build for the reasons you describe (at least the static export which I was using for IPFS). I think we probably want to keep that available.

dapp/src/utils/hooks.js
tomlinton: Done

dapp/src/utils/hooks.js
tomlinton: Done

dapp/src/utils/hooks.js
tomlinton: Done

dapp/src/components/layout.js
tomlinton: Done ✅

dapp/src/utils/hooks.js
sparrowDom: Aha I see thanks

tomlinton:
@sparrowDom I did some more work here. I rewrote the TransactionListener to a functional component and tried to start integrating safe messages. It's incredible complicated to handle all the states though, I'm keen to merge this and deploy it and get the app approved. It's not perfect from the UI side (the side panel doesn't display messages when a transaction is waiting on signers) but its functional.

tomlinton:
@sparrowDom any objections to merging this?

dapp/src/components/TransactionListener.js
```diff
@@ -188,29 +184,33 @@ class TransactionListener extends Component {
...
+  /**
+   * Save transaction data to local storage.
+   */
+  const save = (transactions) => {
+    localStorage.setItem(localStorageId(account), JSON.stringify(transactions))
   }
 
-  async observeTransaction(transaction) {
+  const observeTransaction = async (transaction) => {
+    const observableHash = transaction.safeData
```
sparrowDom, line 198: would it be safer to check for `isSafe` here: 
```
const observableHash = transaction.isSafe
  ? transaction.safeData && transaction.safeData.txHash (otherwise SDK call failed I guess?)
  : transaction.hash
```

dapp/src/components/TransactionListener.js
```diff
@@ -188,29 +184,33 @@ class TransactionListener extends Component {
...
     try {
-      const receipt = await this.props.rpcProvider.waitForTransaction(
-        transaction.hash
-      )
-      const account = this.props.account
+      const receipt = await rpcProvider.waitForTransaction(observableHash)
+
+      // For a Gnosis safe transaction to the receipt to address should match
+      // the current account, otherwise the receipt from address
+      const sourceAccount = transaction.safeData
```
sparrowDom, line 207: Maybe the same thing here: 
```
const sourceAccount = transaction.isSafe ....
```

dapp/src/components/TransactionListener.js
```diff
@@ -244,57 +244,64 @@ class TransactionListener extends Component {
...
+        // If this was a Gnosis Safe transaction, we have a SafeTxHash not a
+        // real transaction hash. We need to resolve some additional data about
+        // the transaction using the Gnosis Safe SDK.
+        let safeData
+        // TODO handle a 404 here. We need to retry. A refresh is required to
+        // get this to retry at the moment.
+        try {
+          safeData = await connector.sdk.txs.getBySafeTxHash(t.hash)
+        } catch (e) {
+          console.error(e)
```
sparrowDom, line 260: nit: Don't know if this error is going to be descriptive, some context might help: 
```
console.error('Gnosis safe SDK call failed: ', e)
```

sparrowDom (approved):
LGTM (couple if inline comments)
And thanks for re-writing the listener to a functional component. Must have been a PITA, dealing with that code. I love it that it shakes of so much boilerplate code.

dapp/src/components/TransactionListener.js
```diff
@@ -10,95 +10,93 @@ import withRpcProvider from 'hoc/withRpcProvider'
...
+    } else if (nonMinedTx.length === 0 && wsProvider) {
+      cleanupWebSocketProvider()
     }
-  }
 
-  componentWillUnmount() {
-    this.cleanupWebSocketProvider()
-  }
+    return () => cleanupWebSocketProvider()
+  }, [transactions, dirtyTransactions, transactionHashesToDismiss, account])
```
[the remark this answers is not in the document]
tomlinton: True!

dapp/src/components/TransactionListener.js
tomlinton: Hmm yea if the SDK call fails observableHash will end up being false which will cause it to error out. The way it is written now transaction.hash will at least exist (it'll be wrong and the listener won't update properly but at least it is preferable to erroring out). We need to address this.

PR #624, opened 2021-08-03
Buyback with Uniswap v3

DanielVF:

Contract change checklist:
  - [x] Unit tests pass


DanielVF:
Fork testing script for swap:

```python
from addresses import *
from world import *

brownie.network.web3.provider.make_request('hardhat_impersonateAccount', [BINANCE])
brownie.network.web3.provider.make_request('hardhat_impersonateAccount', [GOVERNOR])


old_buyback = Contract("0x7294CD3C3eb4097b03E1A61EB2AD280D3dD265e6")
new_buyback = Contract.from_abi("Buyback","0xaB5E7B701B605f74AaC1b749Fd50715f0DEd7Bc5",old_buyback.abi)

# Generate a million dollars of yield
usdc.transfer(vault_core, 1e6 * 1e6, {'from': BINANCE})
vault_core.rebase({'from': GOVERNOR})

# Swap OUSD for OGN
vault_core.allocate({'from': GOVERNOR})
print(ousd.balanceOf(new_buyback) / 1e18)
print(ogn.balanceOf(new_buyback) / 1e18)
```

DanielVF:
Chainlink mockup

```python
chain_ogn_eth = Contract.from_explorer("0x2c881b6f3f6b5ff6c975813f87a4dad0b241c15b")
chain_eth_usd = Contract.from_explorer("0x5f4ec3df9cbd43714fe2740f5e3616155c5b8419")
# Expected OGN per OUSD
expected_ogn_per_ousd = 1e44 / (chain_eth_usd.latestAnswer() * chain_ogn_eth.latestAnswer())
```

DanielVF:
With the new Chainlink price protection, the previous sample code with a million dollars of yield in one shot now correctly skips the buyback since the slippage is too high.

The amount sold, and thus also slippage, can be reduced by transferring out some OUSD before doing the allocate:

```
new_buyback.transferToken(OUSD, 90*1e21, {'from': GOVERNOR})
```

contracts/contracts/buyback/Buyback.sol
```diff
@@ -52,26 +68,78 @@ contract Buyback is Governable {
...
+                usdt,
+                uint24(3000), // Pool fee, usdt -> weth9
+                weth9,
+                uint24(3000), // Pool fee, weth9 -> ogn
+                ogn
+            ),
+            recipient: address(this),
+            deadline: uint256(block.timestamp + 1000),
+            amountIn: sourceAmount,
+            amountOutMinimum: uint256(0)
```
mikeshultz, line 105: Is this safe?  If one of those three pools gets extremely lopsided somehow, we could end up with a poor `amountOut`.  Wonder if we should add a guesstimate here to prevent something like that?

contracts/contracts/buyback/Buyback.sol
```diff
@@ -52,26 +68,78 @@ contract Buyback is Governable {
...
+        });
+
+        uint256 recieved = UniswapV3Router(uniswapAddr).exactInput(params);
+
+        // Check
+        uint256 minExpected = expectedOgnPerOUSD(sourceAmount).mul(96).div(100);
```
mikeshultz, line 111: Suppose this is pretty related to the previous comment.  Why not send `minExpected` into `amountOutMinimum` rather than reverting here?

Also, does this math mean we're good with 4% slippage here?

contracts/contracts/buyback/Buyback.sol
```diff
@@ -52,26 +68,78 @@ contract Buyback is Governable {
...
+        require(_feed != address(0), "Asset not available");
+        (
+            uint80 roundID,
+            int256 _iprice,
+            uint256 startedAt,
+            uint256 timeStamp,
+            uint80 answeredInRound
+        ) = AggregatorV3Interface(_feed).latestRoundData();
+        uint256 _price = uint256(_iprice);
+        require(_price > 0, "Price must be greater than zero");
```
mikeshultz, line 139: Correct me if I'm wrong because I mix them up sometimes, but I don't think `_price` can ever be negative, being unsigned.  Might want to move this `require()` up one line and check against `_iprice`, as casting a negative signed to unsigned could lead to some unexpected weirdness.

Also, maybe rename one of those vars to prevent confusion between `_price` and `_iprice` when reading?

contracts/contracts/buyback/Buyback.sol
```diff
@@ -52,26 +68,78 @@ contract Buyback is Governable {
...
+        return uint256(_price);
```
mikeshultz, line 140: Redundant?

contracts/contracts/buyback/BuybackConstructor.sol
```diff
@@ -1,15 +1,19 @@
...
 import { Governable } from "../governance/Governable.sol";
-
-import { IUniswapV2Router } from "../interfaces/uniswap/IUniswapV2Router02.sol";
+import "../interfaces/chainlink/AggregatorV3Interface.sol";
 import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/SafeERC20.sol";
+import { SafeMath } from "@openzeppelin/contracts/math/SafeMath.sol";
+import {UniswapV3Router} from "../interfaces/UniswapV3Router.sol";
 
 contract BuybackConstructor is Governable {
```
mikeshultz, line 11: Same comments on this one from the other contract.  

I realize you aren't adding this, but why do we have both `Buyback` and `BuybackConstructor`?

mikeshultz (commented):
Took a first pass at the contracts.  Some questions inline.

contracts/contracts/buyback/Buyback.sol
```diff
@@ -52,26 +68,78 @@ contract Buyback is Governable {
      * protocol (e.g. Sushiswap)
      **/
     function swap() external onlyVault {
+        // Don't revert everything, even if the buyback fails.
+        // We want the overall transaction to continue regardless.
+        (bool success, bytes memory data) = address(this).call(
+            abi.encodeWithSignature("swapAndCheck()")
+        );
```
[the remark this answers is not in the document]
DanielVF: Buyback is currently called as part of our large mints. If something goes with the buyback price during the large mint, I don't want the overall transaction with the mint to fail. By forcing an external call here, I can catch the revert after it happens, and the sale that happened during the external call is reverted into nothingness, but execution continues cleanly from here.

That's the theory anyway.

contracts/contracts/buyback/Buyback.sol
DanielVF: Good points here.

contracts/contracts/buyback/Buyback.sol
DanielVF: Yes, I guess if I trust the router, I can just move this expected number into the param and call it a day.

contracts/contracts/buyback/BuybackConstructor.sol
DanielVF: Way more gas efficient to read a contract address from code than from a storage slot. But it makes tests so much better if you can set the addresses. Newer solidities provide an "immutable" slot attribute that writes the value to code during the constructor, giving you the best of both worlds, but we don't have it the 5.11 we are using.

contracts/contracts/buyback/Buyback.sol
DanielVF: Checked after the TX, but yes, might just change this to run here.

contracts/contracts/buyback/Buyback.sol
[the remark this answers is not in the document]
mikeshultz: Ah right, there's no way to prevent the revert from bubbling up otherwise (though I thought I heard someone mention try/catch with modern Solidity).  Thanks

contracts/contracts/buyback/Buyback.sol
mikeshultz: In this case you're trusting the router to return `received`, just the same.

contracts/contracts/buyback/Buyback.sol
DanielVF: Great, should be fixed, have a look.

contracts/contracts/buyback/Buyback.sol
DanielVF: Fixed

contracts/contracts/buyback/Buyback.sol
DanielVF: Changed.

contracts/contracts/buyback/Buyback.sol
DanielVF: Done

contracts/contracts/buyback/Buyback.sol
mikeshultz: Nice and clean :+1:

mikeshultz (approved):
Thanks for addressing my feedback.

LGTM  :+1:

mikeshultz:
Also, we should make sure the `Buyback` contract is added to analytics with relevant triggers so we know when `BuybackFailed` happens.  Probably not an event we want to miss.

DanielVF:
#### Requirements

Buys back OGN with a percentage of OUSD yield. Resistant to uniswap economic attacks.

### Deployment Considerations

_NA_

#### Internal State

No internal state is modified during operation. Contract storage slots are configuration. All dynamic state is looked up from external sources.

💡 If we were using a more recent solidity version, we could immutable all these configuration addresses into the contract, and save gas on everything the contract does.

#### Attack

If this contract gets stupid amounts of OUSD, the slippage from swapping such a large amount can go over the slippage threshold, pausing swaps. Governance can fix this by temporarily withdrawing some OUSD.

If this were a long term issue, another option would be to cap the size that can be swapped at any given moment.

#### Logic

_LGTM_


#### Tests

#### Flavor

_Made a big change to flavor. Looks better._

#### Overflow

#### Black magic

#### Authentication

- [x] Never use tx.origin

#### Cryptographic code

_No crypto_

#### Gas problems

_No for loops._

#### External calls

- [x] Contract addresses passed in are validated. _No user supplied addresses._
- [x] Unsafe external calls
- [x] Malicious behaviors. _Only talking to trusted contracts._
- [x] Oracles, one of:
  - [ ] No oracles
  - [x] Oracles can't be bent _(not bent on chain, anyway)_


#### Ethereum

DanielVF:
I've fork tested this again, and it's ready for a review, @tomlinton.

contracts/contracts/buyback/Buyback.sol
```diff
@@ -51,25 +67,70 @@ contract Buyback is Governable {
      * @dev Execute a swap of OGN for OUSD via Uniswap or Uniswap compatible
      * protocol (e.g. Sushiswap)
      **/
-    function swap() external onlyVault {
-        if (uniswapAddr == address(0)) return;
-
+    function swap() external onlyVault nonReentrant {
         uint256 sourceAmount = ousd.balanceOf(address(this));
         if (sourceAmount < 1000 * 1e18) return;
+        if (uniswapAddr == address(0)) return;
```
tomlinton, line 73: Not really related to this PR but might be nice to add: with the setUniswapAddr we should only run the safeApprove functions on uniswapAddr if address is not the zero address.

contracts/contracts/buyback/Buyback.sol
```diff
@@ -51,25 +67,70 @@ contract Buyback is Governable {
...
+        // 97% should be the limits of our oracle errors.
+        // If this swap somtimes skips when it should succeed, that’s okay,
```
tomlinton, line 75: Typo

contracts/contracts/buyback/Buyback.sol
```diff
@@ -51,25 +67,70 @@ contract Buyback is Governable {
...
+        // the amounts will get get sold the next time this runs,
+        // when presumably the oracles are more accurate.
```
tomlinton, line 77: This comment is very poetic, it feels like a haiku

tomlinton (approved):
👍

contracts/contracts/buyback/Buyback.sol
DanielVF: Hah, so it is!

contracts/contracts/buyback/Buyback.sol
DanielVF: Done

contracts/contracts/buyback/Buyback.sol
DanielVF: Fixed. Thanks

PR #630, opened 2021-08-11
Generalized Wallet Connection Text

micahalcorn:
This first commit (bf12eba) addresses an issue where the modal is launched from the OGN staking page but it refers to "stablecoins".

image0

@sparrowDom I was unable to push this without first running prettier, but I may have a different version because it modified several files in 81ea6e7d74f029e27a06b1f7057e371f1cb8b0dc that I didn't intend to change.


sparrowDom:
@micahalcorn yeah no worries regarding prettier and thanks for running it. LGTM

PR #632, opened 2021-08-12
Check for trustee/buyback before calling during allocate/mint

DanielVF:
We check that we have trustee contract when we call it during the rebase, but not when we try to call it during the allocate. This breaks large mints if no trustee contract is set. 

Since a trustee contract is set in production, this is not a problem for the live contracts. It's just good to have the code act consistently, and how you would expect.

Contract change checklist:
  - [x] Unit tests pass


franckc (approved):
LGTM

No need to deploy to Mainnet since as you stated, the trustee should always be set on Mainnet.

tomlinton (approved):
Nice fix!

PR #677, opened 2021-08-24
[L13] Lack of event emission after sensitive actions

tomlinton:
Contract change checklist:
  - [ ] Unit tests pass

### DIFF
--- a/contracts/contracts/utils/InitializableAbstractStrategy.sol
+++ b/contracts/contracts/utils/InitializableAbstractStrategy.sol
@@ -1,8 +1,6 @@
 pragma solidity 0.5.11;
 
-import {
-    Initializable
-} from "@openzeppelin/upgrades/contracts/Initializable.sol";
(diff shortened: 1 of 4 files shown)


franckc (approved):
Code changes LGTM

Should the unit tests get updated to verify the events are fired?

contracts/contracts/utils/InitializableAbstractStrategy.sol
```diff
@@ -124,6 +128,10 @@ contract InitializableAbstractStrategy is Initializable, Governable {
         external
         onlyGovernor
     {
+        emit RewardLiquidationThresholdUpdated(
```
DanielVF, line 131: If it is intended to be able to reconstruct all values that these are set to from the events, should we fire these events in the initializer as well?

DanielVF (approved):
One possible change, otherwise looks good to me.

PR #724, opened 2021-09-01
🥏 [C01] Slippage check on reward token swaps

tomlinton:
Add a swap token list to the vault and separate out the harvest/swap functions. The swap function uses chainlink oracles to determine the min expected output from Uniswap.

Also changed the approvals to use infinite approvals for vault -> uniswap when the swap token is added, and removes the approval when a swap token is removed. Previously it was doing approvals of the actual required amount every time it swapped. We save gas with the infinite approval but it is less secure, but I think it is OK since the swap tokens should never be sitting in the vault contract. Happy to discuss.

Contract change checklist:
  - [ ] Unit tests pass


tomlinton:
@DanielVF put this up for some early feedback. Are we on the same page with this? 😁

DanielVF:
@tomlinton Much Good

DanielVF (approved):

contracts/contracts/vault/VaultAdmin.sol
```diff
@@ -185,18 +193,86 @@ contract VaultAdmin is VaultStorage {
...
+
+    /**
+     * @dev Remove a swap token from the tokens that get liquidated for stablecoins.
+     * @param _addr Address of the token
+     */
+    function removeSwapToken(address _addr) external onlyGovernor {
+        bool added = false;
+        for (uint256 i = 0; i < swapTokens.length; i++) {
+            if (swapTokens[i] == _addr) {
+                added = true;
```
franckc, line 246: supernit: can add a break here to exit the loop and maybe save a tiny bit of gas.

franckc (approved):
Nice! One minor comment inline otherwise LGTM

Issue #661, opened 2021-08-21
Two Decimals on Blur


micahalcorn:
I would vote for updating this input field to "123.00" when the field is blurred. We allow up to six decimal places on the input field and we should _not_ round that down to two places. But I do think it's nice for each value to at least be that precise since we're dealing in USD terms.

<img width="611" alt="Screen Shot 2021-08-21 at 3 58 35 PM" src="https://user-images.githubusercontent.com/273937/130336829-eb64b3c2-2294-4d52-a58c-07251c54b41d.png">

**PR for issue:** https://github.com/OriginProtocol/origin-dollar/pull/737

sparrowDom:
done via: https://github.com/OriginProtocol/origin-dollar/pull/737

PR #737, opened 2021-09-10
Kurogewashu/decimal places

KurogeWashu:
PR for https://github.com/OriginProtocol/origin-dollar/issues/661
Screenshot_2021-09-09 Origin Dollar (OUSD)

### DIFF
--- a/dapp/src/components/AccountStatusDropdown.js
+++ b/dapp/src/components/AccountStatusDropdown.js
@@ -133,6 +133,7 @@ const AccountStatusDropdown = ({ className, showLogin, dapp }) => {
         .dot {
           width: 10px;
           height: 10px;
+          margin-left: 10px;
           border-radius: 5px;
           background-color: #ed2a28;
         }
--- a/dapp/src/components/buySell/SwapCurrencyPill.js
+++ b/dapp/src/components/buySell/SwapCurrencyPill.js
@@ -278,6 +278,9 @@ const SwapCurrencyPill = ({
   const showOusd =
     (swapMode === 'redeem' && topItem) || (swapMode === 'mint' && bottomItem)
 
+  const roundTo2Decimals = (value) => {
+    return formatCurrency(parseFloat(value), 2)
+  }
   const roundTo2to6Decimals = (value) => {
     return formatCurrencyMinMaxDecimals(value, {
       minDecimals: 2,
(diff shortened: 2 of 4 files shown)


joshfraser:
We should use a lighter color grey for digits 3-6.

KurogeWashu:
Good idea @joshfraser , it would be difficult to do in the input field, but I can add it into the expected output area

KurogeWashu:
Screenshot_2021-09-09 Origin Dollar (OUSD)(3)
Screenshot_2021-09-09 Origin Dollar (OUSD)(2)

joshfraser:
Thanks for the quick turnaround. We should use black instead of the dark gray on the second input field (up to the second decimal). We probably want the faded numbers to be the same hue as the background as well.

sparrowDom:
Hey, I seem to have encountered a couple of bugs: 
Screenshot 2021-09-10 at 12 05 30
(output balance is incorrect)

Screenshot 2021-09-10 at 12 05 57
(input should format to 10302.00)

let me know when you address them and I can take another look 👀

KurogeWashu:
@sparrowDom fixed the error, I may have used the provided function incorrectly.

app.hashflow.com, mentions zero slippage, tight spreads, MEV resistant. Their UI could be broken right now, but they also only use two decimal places. 
photo5035177860217809386
I would go with 2-3 decimal places in expected output for better accuracy. I do understand how it may be valuable to have more decimal places in the expected output, so if anyone requests, I can add it back (also curious at comparing between expected and actual outputs). 

Ideas for next steps:
-18 decimal places for input, or the max available decimal places for each token
-enable swapTokensForExactTokens

KurogeWashu:
@sparrowDom only the first error is fixed* Just got a chance to look into the second one: 
- It works up to three digits, which my wallet coincidentally only has. Looking at your screenshot it should still work with "insufficient balance" error.
- Currently using onBlur to add decimal places, if we use onChange, it would glitch when deleting the decimal place, (e.g. 100.00 becomes 10000 immediately as soon as the . is deleted)
- Thought we could use debounce with onChange, but it wouldn't solve the above error if someone were in the middle of updating a number
- I've isolated the problem: if it's larger than 3 digits with no decimal, then it gets caught in "checkValidInputForCoin" from "utils/math"
- Are there unit tests for this regex line somewhere? 
- `^((\\d{1,3})(?:[0-9]{3}){0,1}|(\\d{1})(?:[0-9]{3}){0,2}|(\\d{1,18}))((\\.)|(\\.\\d{1,${decimals}}))?$`

sparrowDom:
@KurogeWashu I've cleaned up the PR a bit. I think everything should be working now as expected. Please take a look if it looks ok to you

KurogeWashu:
@sparrowDom Looks good!

sparrowDom (approved):
Lgtm

PR #738, opened 2021-09-10
Upgrade to Solidity ^0.8.0

tomlinton:
I've got it to a point where tests are passing, but some of the hacks are pretty ugly and I haven't given much thought to refining it yet. I'll work on that.

**Problems**

- ~~Ugly mappings for Curve int128/uint256 conversions in both the core strategy and the mock curve pool~~
- Virtuals on a core contract (VaultCore) to allow for MockVault to override
- Probably a host of unnecessary imports that can be fixed with better inheritance
- More?

**Todo**

- ~~Improve the Buyback/Flipper constructor for test situation~~
- There is a bunch of unnecessary SafeMath that could be removed


DanielVF:
@tomlinton 

Solidity either lets you change the number of bits OR the sign in a single  cast, so you just use double casts to get what you want:

```
uint256 foo = 1;
int128 bar = int128(uint128(foo));
```

We can ditch the index mappings.

tomlinton:
Ahh I missed that in the changelog, awesome

tomlinton:
Gone 🧹

DanielVF (commented):
Looking good. Nice that the upgrade tends towards making the code cleaner looking, most of the time.

I notice that this changes to using any 0.8.x solidity in the contracts. On the plus side, we can upgrade single contracts as we deploy them to future 0.8.x solidity versions. On the minus side I sure miss being able to look at a contract, and have tools look at a contract, and know exactly what it was compiled with.

Props for doing this - a massive amount of work.

DanielVF (approved):

crazybuster (approved):
Wow, that's a lot of changes! Scary, but looks good to me.

DanielVF:
@tomlinton Would have a chance to merge in the master changes into this?

tomlinton:
Will do 👍

PR #757, opened 2021-09-23
Create method of authorized transfer of all stakes between addresses

crazybuster:

Contract change checklist:
  - [ ] Unit tests pass

### DIFF
--- a/contracts/contracts/staking/SingleAssetStaking.sol
+++ b/contracts/contracts/staking/SingleAssetStaking.sol
@@ -45,6 +45,8 @@ contract SingleAssetStaking is Initializable, Governable {
     uint8 constant USER_STAKE_TYPE = 0;
     uint256 constant MAX_STAKES = 256;
 
+    address public transferAgent;
+
     /* ========== Initialize ========== */
 
     /**
@@ -447,6 +449,28 @@ contract SingleAssetStaking is Initializable, Governable {
         stakingToken.safeTransfer(msg.sender, totalWithdraw);
     }
 
+    function transferStakes(address _dstAccount, bytes32 r, bytes32 s, uint8 v) external {
+        Stake[] storage dstStakes = userStakes[_dstAccount];
+        require(dstStakes.length == 0, "Dest stakes must be empty");
+        Stake[] storage stakes = userStakes[msg.sender];
+        require(stakes.length > 0, "Nothing to transfer");
+
+        // matches signMsg(ethers.utils.solidityKeccak256)
+        bytes32 hash = keccak256(
+          abi.encodePacked("\x19Ethereum Signed Message:\n32",
+                           keccak256(
+                             abi.encodePacked("tran", address(this), msg.sender, _dstAccount)
+                           )
+                          )
+        );
+        require(ecrecover(hash, v, r, s) == transferAgent, "Transfer not authed");
+
+        // copy the stakes into the dstAccount array and delete the old one
+        userStakes[_dstAccount] = stakes;
+        delete userStakes[msg.sender];
+        emit StakesTransfered(msg.sender, _dstAccount, stakes.length);
+    }
+
     /* ========== MODIFIERS ========== */
 
     function setPaused(bool _paused) external onlyGovernor {
@@ -464,6 +488,14 @@ contract SingleAssetStaking is Initializable, Governable {
         uint256[] calldata _rates
     ) external onlyGovernor {
         _setDurationRates(_durations, _rates);
+    }
+
+    /**
+     * @dev Set the agent that will authorize transfers
+     * @param _agent Address of agent
+     */
+    function setTransferAgent(address _agent) external onlyGovernor {
+      transferAgent = _agent;
     }
 
     /**
@@ -500,4 +532,5 @@ contract SingleAssetStaking is Initializable, Governable {
         bytes32 rootHash,
         uint256 proofDepth
     );
+    event StakesTransfered(address indexed fromUser, address toUser, uint256 numStakes);
 }
--- a/contracts/contracts/staking/SingleAssetStaking.sol
+++ b/contracts/contracts/staking/SingleAssetStaking.sol
@@ -7,7 +7,6 @@ import { Initializable } from "@openzeppelin/upgrades/contracts/Initializable.so
 import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/SafeERC20.sol";
 import { Governable } from "../governance/Governable.sol";
 import { StableMath } from "../utils/StableMath.sol";
-import "hardhat/console.sol";
 
 contract SingleAssetStaking is Initializable, Governable {
     using SafeMath for uint256;
--- a/contracts/contracts/staking/SingleAssetStaking.sol
+++ b/contracts/contracts/staking/SingleAssetStaking.sol
@@ -449,7 +449,20 @@ contract SingleAssetStaking is Initializable, Governable {
         stakingToken.safeTransfer(msg.sender, totalWithdraw);
     }
 
-    function transferStakes(address _dstAccount, bytes32 r, bytes32 s, uint8 v) external {
+    /**
+     * @dev Use to transfer all the stakes of an account in the case that the account is compromised
+     *      Requires access to both the account itself and the signature of a transfer agent
+     * @param _dstAccount the address of the address to transfer to(must be a clean address with no stakes)
+     * @param r r portion of the signature by the transfer agent(must sign packed message [string(4), address, address, address])
+     * @param s s portion of the signature
+     * @param v v portion of the signature
+     */
+    function transferStakes(
+        address _dstAccount,
+        bytes32 r,
+        bytes32 s,
+        uint8 v
+    ) external {
         Stake[] storage dstStakes = userStakes[_dstAccount];
         require(dstStakes.length == 0, "Dest stakes must be empty");
         Stake[] storage stakes = userStakes[msg.sender];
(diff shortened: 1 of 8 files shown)


franckc (commented):
Nice! Took a quick pass and makes sense to me.

Could you add a migration file that we can user deploy this change?

Also bonus point for adding a small script that given the transfer agent pk as input would output the signature to communicate to the user for recovering their funds.

DanielVF (commented):
1. So the use case for this is the case in which a user has had their ETH account seed phrase phished, and needs to recover their OGN without allowing the attacker who also has the account keys to transfer out the OGN as soon as it is unstaked.

This code requires the victim user to send the transaction, which requires ETH to be in the victim account, and this ETH amount will need to be several multiples larger than the amount of ETH that it would cost to transfer the ETH out. This means that if the account is being monitored by the attacker, when the user transfers ETH into the account to pay for the gas for recovering the staked OGN, then the attacker can immediately profitably steal that ETH, which blocks the recovery process.

It would seem better to use a signature from the victim, rather than a transaction. This would mean that an attacker cannot block it by stealing ETH, and sees no activity from the compromised account.

2. With the current implementation, the signature is good forever, which means that once a transfer has been authorized, it can be reused repeatedly. While in the normal use case that should not be a problem, it's still pretty sketchy. We should probably have an expiration timestamp passed along and in the signature.

3. Glad to see transferAgent required to be non-zero. That would be chaos otherwise.

crazybuster:
> Nice! Took a quick pass and makes sense to me.
> 
> Could you add a migration file that we can user deploy this change?
> 
> Also bonus point for adding a small script that given the transfer agent pk as input would output the signature to communicate to the user for recovering their funds.



> 1. So the use case for this is the case in which a user has had their ETH account seed phrase phished, and needs to recover their OGN without allowing the attacker who also has the account keys to transfer out the OGN as soon as it is unstaked.
> 
> This code requires the victim user to send the transaction, which requires ETH to be in the victim account, and this ETH amount will need to be several multiples larger than the amount of ETH that it would cost to transfer the ETH out. This means that if the account is being monitored by the attacker, when the user transfers ETH into the account to pay for the gas for recovering the staked OGN, then the attacker can immediately profitably steal that ETH, which blocks the recovery process.
> 
> It would seem better to use a signature from the victim, rather than a transaction. This would mean that an attacker cannot block it by stealing ETH, and sees no activity from the compromised account.
> 
> 1. With the current implementation, the signature is good forever, which means that once a transfer has been authorized, it can be reused repeatedly. While in the normal use case that should not be a problem, it's still pretty sketchy. We should probably have an expiration timestamp passed along and in the signature.
> 2. Glad to see transferAgent required to be non-zero. That would be chaos otherwise.

That is a good call, @joshfraser what do you think?

joshfraser:
Good call @DanielVF. It sounds like signature-based is the way to go.

crazybuster:
@DanielVF @franckc @joshfraser Alright added a page /signTransfer to the dapp that will allow an account to sign the transfer. Then the use must send this signature to us and we run the /scripts/staking/transferStakes.js script with the signature to do the actual transfer.

dapp/src/components/EmailForm.js
```diff
@@ -75,7 +75,7 @@ const EmailForm = ({ footer }) => {
           }}
           required
           value={email}
-          placeholder="Your email"
+          placeholder="Destination Address"
```
sparrowDom, line 78: This was probably not intended

dapp/src/components/SignTransferAuth.js
```diff
@@ -0,0 +1,148 @@
...
+  })
+
+  return (
+    <>
+      <div>
+        <div className="content-holder flex-grow d-flex flex-column shadow-div">
+          {active && (
+            <div>
+              {' '}
+              on {ognStaking.address} Transfer stakes from {account} to:
```
sparrowDom, line 30: Would be cool to use `fbt` so this text is translateable

sparrowDom (commented):
Nice, if Aure wants to make some mocks, I can help with translating those to frontend code if you want.

dapp/src/components/SignTransferAuth.js
DanielVF: This is probably going to be only used once or twice a quarter. Fine not to translate it, or do more work than making sure it works.

joshfraser: Yeah, this doesn't need translations or to look pretty at all. Let's save Aure's time for more important stuff.

sparrowDom: Makes sense, got it

dapp/src/components/EmailForm.js
crazybuster: Yeah my bad, too many tabs Open

contracts/contracts/staking/SingleAssetStaking.sol
```diff
@@ -7,6 +7,7 @@ import { Initializable } from "@openzeppelin/upgrades/contracts/Initializable.so
 import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/SafeERC20.sol";
 import { Governable } from "../governance/Governable.sol";
 import { StableMath } from "../utils/StableMath.sol";
+import "hardhat/console.sol";
```
franckc, line 10: remove

contracts/scripts/staking/signTransferAuth.js
```diff
@@ -0,0 +1,56 @@
+// Script for verifying against etherscan.
```
franckc, line 1: nit: update

contracts/scripts/staking/transferStakes.js
```diff
@@ -0,0 +1,53 @@
...
+//      node signStakingPayout.js <PrivateKey> <fromAddress> <toAddress> <r> <s> <v>
+//
+//
+//
+const { ethers } = require("hardhat");
+const { utils, BigNumber, Wallet } = require("ethers");
+const { getTxOpts } = require("../../utils/tx");
+const fs = require("fs");
+
+const TEST_AGENT_PK =
```
franckc, line 14: delete

contracts/scripts/staking/transferStakes.js
```diff
@@ -0,0 +1,53 @@
...
+  const wallet = new Wallet(pk, ethers.provider);
+  const proxyAddress = (await ethers.getContract("OGNStakingProxy")).address;
+  const contract = await ethers.getContractAt(
+    "SingleAssetStaking",
+    proxyAddress
+  );
+
+  const ops = await getTxOpts();
+  ops.gasLimit = 6000000;
+  ops.gasPrice = 9000000000;
```
franckc, line 27: With the introduction of EIP 1559, I think ehter.js by default sends type 2 transaction and setting the gas price is ignored.
I would just remove all this and only set gasLimit in the call to transferStakes below.
like 
```
.transferStakes(fromAddress, toAddress, r, s, v, { gasLimit: 9000000000);
```

contracts/scripts/staking/signTransferAuth.js
```diff
@@ -0,0 +1,56 @@
...
+//  - Run:
+//      node signStakingPayout.js <PrivateKey> <toAddress>
+//
+//
+//
+const { ethers } = require("hardhat");
+const { utils, BigNumber, Wallet } = require("ethers");
+const fs = require("fs");
+
+const TEST_AGENT_PK =
```
franckc, line 13: remove

contracts/scripts/staking/signTransferAuth.js
```diff
@@ -0,0 +1,56 @@
...
+//
+// Usage:
+//  - Run:
+//      node signStakingPayout.js <PrivateKey> <toAddress>
+//
+//
+//
+const { ethers } = require("hardhat");
+const { utils, BigNumber, Wallet } = require("ethers");
+const fs = require("fs");
```
franckc, line 11: unnecessary import? pls run linter

contracts/scripts/staking/transferStakes.js
```diff
@@ -0,0 +1,53 @@
...
+// Usage:
+//  - Run:
+//      node signStakingPayout.js <PrivateKey> <fromAddress> <toAddress> <r> <s> <v>
+//
+//
+//
+const { ethers } = require("hardhat");
+const { utils, BigNumber, Wallet } = require("ethers");
+const { getTxOpts } = require("../../utils/tx");
+const fs = require("fs");
```
franckc, line 12: unnecessary import? pls run linter

franckc (approved):
Added  few more small comments but overall code looks good.

Let's clean this up asap and deploy!

franckc:
A few remaining comments. Especially the one regarding the migration (not sure if the old migration format still works?)
@crazybuster do you mind taking a look? let's get this out! :)

contracts/scripts/staking/signTransferAuth.js
```diff
@@ -0,0 +1,52 @@
+// Script for verifying against etherscan.
+//
+// Usage:
+//  - Run:
+//      node signStakingPayout.js <PrivateKey> <toAddress>
```
sparrowDom, line 5: this probably needs updating?

contracts/scripts/staking/transferStakes.js
```diff
@@ -0,0 +1,46 @@
+// Script for verifying against etherscan.
+//
+// Usage:
+//  - Run:
+//      node signStakingPayout.js <PrivateKey> <fromAddress> <toAddress> <r> <s> <v>
```
sparrowDom, line 5: does this need updating as well?

Issue #555, opened 2021-02-10
Large mint causes revert on fork


franckc:
To reproduce:
Terminal A
```
export BLOCK_NUMBER=...
export PROVIDER_URL="https://eth-mainnet.alchemyapi.io/v2/..."
yarn run node:fork
```

Terminal B
```
FORK=true npx hardhat fund --index 4 --num 1 --amount 1000000 --network localhost
FORK=true npx hardhat mint --index 4 --num 1 --amount 50000 --network localhost

Minting 50000 OUSD for account 4 at address 0x1974f84881Af4204a21f18c43D7c4d9Dee331Bb5
An unexpected error occurred:

ProviderError: Transaction reverted without a reason
    at HttpProvider.request (/Users/franck/src/origin-dollar/contracts/node_modules/hardhat/src/internal/core/providers/http.ts:45:19)
    at GanacheGasMultiplierProvider.request (/Users/franck/src/origin-dollar/contracts/node_modules/hardhat/src/internal/core/providers/gas-providers.ts:170:34)
    at processTicksAndRejections (internal/process/task_queues.js:93:5)
```

A smaller mint (ex 1k) succeeds, so this is likely a failure in `allocate()`.

Possibly related to <https://github.com/trufflesuite/ganache-core/issues/571>

tomlinton:
Yea I think this is likely Compound. Would be interesting to try it with that strategy removed.

sparrowDom:
Fixed via: https://github.com/OriginProtocol/origin-dollar/pull/759

crypto-ticky:
I think I figured out the issue

crypto-ticky:
In VaultCore.mint function, IOracle(priceProvider).price function is called.
![image](https://user-images.githubusercontent.com/68897116/185136676-6477f396-6a96-4c5b-bb26-4282983e49e0.png)
and the priceProvider address is [https://etherscan.io/address/0x277e80f3E14E7fB3fc40A9d6184088e0241034bD#readProxyContract](url)

DanielVF:
Works in brownie: 

<img width="784" alt="image" src="https://user-images.githubusercontent.com/837/185206949-752435e8-de7e-4a4a-a3b3-a7d4a124c4de.png">

crypto-ticky:
that worked with 24999
![image](https://user-images.githubusercontent.com/68897116/185220961-fa30194f-7e05-4291-ae64-90f3fa4d1051.png)
[...]

PR #759, opened 2021-09-27
fix fork and mint cmd calls in forked environment

sparrowDom:
Fixes `fund` and `mint` cmd calls in forked environment. 

I suspect that the cause of the problem is one of the hardhat version updates. In the past we didn't require default signature accounts to call the `impersonate account` hardhat function in order to impersonate them. Now it is required. 

**Issue:** https://github.com/OriginProtocol/origin-dollar/issues/555

Contract change checklist:
  - [ ] Unit tests pass


sparrowDom:
@franckc or @DanielVF or @tomlinton mind throwing your 👀 on this one?

tomlinton:
Any description of what was wrong or what the changes are? It's not clear to me why changing the approval to max uint256 helps. I can't see anything wrong with the changes though.

sparrowDom:
@tomlinton yup you are right that change did nothing - reverted it. Is this PR ready to be merged now you think?

contracts/tasks/account.js
```diff
@@ -190,19 +193,26 @@ async function mint(taskArguments, hre) {
     throw new Error("Task can only be used on local or fork");
   }
 
-  const ousdProxy = await ethers.getContract("OUSDProxy");
-  const ousd = await ethers.getContractAt("OUSD", ousdProxy.address);
```
tomlinton, line 194: Shouldn't this correctly use the mainnet address?

sparrowDom: I might be doing something wrong, but hardhat isn't able to find the already deployed contracts for me.
Screenshot 2021-10-05 at 10 17 52

Even though the `deployments/localhost` gets copied over in `node.sh` from the `deployments/mainnet` the `localhost` folder is immediately deleted when the `FORK=true npx --no-install hardhat node --export...` is ran.

It would be super useful to make fork already source the deployed mainnet contracts from deployments folder. Is this not happening to you?

To test it these 2 commands can be ran: 
```
FORK=true npx hardhat fund --index 4 --num 1 --amount 1000000 --network localhost
FORK=true npx hardhat mint --index 4 --num 1 --amount 50000 --network localhost
```

tomlinton: Ahh yea the behaviour of hardhat changed and we upgraded. Fix was here https://github.com/OriginProtocol/origin-dollar/pull/765

sparrowDom: Thanks, was wondering why that reset was happening... Works like a charm now. 

Have undid the unnecessary code changes. Can you please take another look?

sparrowDom:
tnx @DanielVF  merging...

tomlinton:
I think this still has issues. Why the `return hre.network.provider.request({`? Doesn't that return break the code because the rest of the function now does nothing?

PR #765, opened 2021-09-29
Add no reset flag to hardhat fork

tomlinton:


franckc (approved):
LGTM

This used to work. So I guess that's something that is now needed because we upgraded to a new hardhat version?

tomlinton:
Yep it was added to `hardhat-deploy`

PR #777, opened 2021-10-06
Solidity upgrade deploy changes

tomlinton:
Separate PR being merged into the existing one to make it easier to review. While running through predeploy checks I found a few issues:

**Contracts**

- Can't set a default strategy to nothing, you can only set it to another existing strategy. Added code to allow setting it to the zero address.
- You could remove a strategy even while it was the default strategy for an asset, with the result that allocations will get sent to a removed strategy.
- Added a way to get revert messages from the timelock rather than a generic error

**Deployment**
- Added code to remove the old AAVE and 3pool strategies which were still approved on the Vault

### DIFF
--- a/contracts/contracts/timelock/Timelock.sol
+++ b/contracts/contracts/timelock/Timelock.sol
@@ -151,6 +151,17 @@ contract Timelock {
         emit CancelTransaction(txHash, target, signature, data, eta);
     }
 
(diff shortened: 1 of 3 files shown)


DanielVF (approved):

franckc (approved):
Thanks for diligently testing the deploy and identifying this issue!

LGTM

PR #783, opened 2021-10-07
Upgrade proposal script

tomlinton:


contracts/proposal_fix.js
```diff
@@ -0,0 +1,250 @@
...
+        args: ["0xD5433168Ed0B1F7714819646606DB509D9d8EC1f"],
+      },
+      {
+        // Remove old Aave strategy
+        contract: cVault,
+        signature: "removeStrategy(address)",
+        args: ["0x9f2b18751376cF6a3432eb158Ba5F9b1AbD2F7ce"],
+      },
+      {
+        // Remove old Aave strategy
```
franckc, line 200: it's 3pool

contracts/proposal_fix.js
```diff
@@ -0,0 +1,250 @@
...
+        args: [assetAddresses.AAVE],
+      },
+      {
+        // Approve Compound strategy in Vault
+        contract: cVault,
+        signature: "approveStrategy(address)",
+        args: [cCompoundStrategyProxy.address],
+      },
+      {
+        // Remove the deafult strategy for DAI
```
DanielVF, line 164: @tomlinton This removes the default strategy for DAI, but never adds it back. Should this not be setting the default DAI strategy to Compound?

tomlinton: Yea I wasn't sure about that. If I set it to Compound all the DAI will end up in there and I think we were planning to send some to AAVE. I had thought we might manually allocate some to aave, some to compound.

DanielVF: The default strategy allowed the invested funds to be withdrawn by users, so if we have most of the funds invested, then we should a default strat.

With all the funds in compound, the strategist just needs to to run a reallocate to move a portion of funds over to AAVE. So just one strategist action, afterwords, rather than two.

DanielVF: The vault has a method to reallocate funds from one strategy to another, and we can withdraw certain or all funds from a strategy to the vault, but I don't think we have a method to send funds from the vault to a non-default strategy. In OUSD's normal operation, this is not a probables. Ideally there's sort of a three layer flow going from the vault buffer into and out of a coin's default strategy, and then from the default strategy into and out of other longer term strategies, and reallocate works great for this.

But the only strategy deposit methods in both the VaultCore and VaultAdmin deposit to the default strategy for an asset. This means we will have 9.6 million DAI uninvested until we set a default strategy and allocate.

tomlinton:
Not going to merge this, it was just for review.

PR #785, opened 2021-10-11
Proposal to allocate all DAI to Compound

tomlinton:
- Add DAI support to Compound
- Set Compound as default strategy for DAI
- Set the trustee fee to 1000 BPS/10%
- Allocate


franckc (approved):
This does not really have to be a deployment since no new contract gets deployed? alternatively we could use the propose script. https://github.com/OriginProtocol/origin-dollar/blob/master/contracts/scripts/governor/propose.js

But that works and is probably less code...

LGTM

PR #799, opened 2021-11-01
Sparrow dom/disable flag

sparrowDom:
- add a "Temporarily unavailable due to maintenance" (that is also configurable) warning message to swap page and disable the Swapping
- add staging/prod env variables for the behaviour

**Important:** This should be merged / deployed right before we want to disable the dapp. The deploy usually takes ~30 minutes. If we want to time it perfectly we could also deploy this version earlier and then in GAE panel disable it and route traffic to previous version. And when we want to disable the dapp, just swap the versions. 

Screenshot 2021-11-01 at 05 09 43


sparrowDom:
@DanielVF @franckc left some thoughts on how to deploy this in the original message. Not sure how time sensitive it is.

franckc (approved):
Thanks for adding this. LGTM

Let's deploy tomorrow before we start the maintenance...

PR #812, opened 2021-11-08
Redo deploy scripts for OGN rescue.

DanielVF:
This adds a separate proposal to run from the old governor to transfer ownership to the new governor, as well as moves the deploys to the new method.

I've left a hole in the deploy numbering by moving this to the end, since it needed two deployment migrations numbers, and this seemed like the least ugly solution. I'm open to anyone changing this around who's sense of aesthetics can find less unsightly option.

@franckc Please check, and change if needed, the `TRANSFER_AGENT` constant. I left this as I found it, but I am not sure that it is correct.

I have fork tested the deploy, but not the operation of staking or of the rescue feature after the deploy.


contracts/deploy/031_staking_add_rescue.js
```diff
@@ -0,0 +1,42 @@
+const { deploymentWithProposal } = require("../utils/deploy");
+
+const TRANSFER_AGENT = "0x522731a061e896B5Db9dDff9234fB5461A533710";
```
tomlinton, line 3: Who dis 🤣

franckc (approved):
Thanks for rewriting this deployment!
LGTM

contracts/deploy/031_staking_add_rescue.js
franckc: Checked with the authority. Let's use 0x449E0B5564e0d141b3bc3829E74fFA0Ea8C08ad5

franckc:
For some reason the tests are failing and I suspect it might be because this deployment fails on local?

contracts/deploy/031_staking_add_rescue.js
DanielVF: Used it

DanielVF:
Fixed tests - hadn't pushed up my latest changes.

PR #819, opened 2021-11-13
Fix race condition

sparrowDom:
In Gnosis environment (maybe elsewhere as well) it would happen that `setupContracts` can be called 2 times very close together and the first call can take longer to finish than the second one. The first call also does not have information about user's account while the second one does. 

This can lead up to the first call overriding a more richer data of the second call. 

**note:**
this has already been deployed, so it is tested in production and it works...


shahthepro (approved):

tomlinton:
Thanks sir!

PR #826, opened 2021-11-17
Sandboxed AAVE/Compound strategies and funds reallocation

tomlinton:
Putting this up for feedback. This deploys new proxies and points them to the existing AAVE and Compound implementations, then reallocated 40M into these strategies. The idea being that this 40m is then "sandboxed" as it is out of the mint/redeem path. The only way to interact with these funds is through a governance or strategist multisig transaction and is a damage mitigation strategy.

Contract change checklist:
  - [ ] Unit tests pass


tomlinton:
Not sure on the best way to handle this in the analytics UI. We could merge the balances into the existing Compound/AAVE fields, or display them as separate strategies.

joshfraser:
I think it makes a lot of sense to segregate the funds to mitigate risk.

Playing devil's advocate, I don't like that those funds are only accessible to our team and no one else. One of our core promises is that your funds are always redeemable for the underlying assets (even if we all know that redemptions are exceedingly rare).

Perhaps this is something we turn on after implementing decentralized governance?

tomlinton:
> Playing devil's advocate, I don't like that those funds are only accessible to our team and no one else. One of our core promises is that your funds are always redeemable for the underlying assets (even if we all know that redemptions are exceedingly rare).

This holds true as long as we maintain that the non-sandboxed fund balance > largest holder at all times. A bank rush could of course break that. It's also a maintenance cost and is not decentralized.

I think in general I prefer the added security of this over the drawbacks (and it's something we've already done with Curve strategies and I assume we'll do it with Convex).

franckc:
I really like the added security.
Having to manually transfer funds out of sandbox to cover for unexpected mass withdrawals is the obvious drawback...

I wonder if the rebalancing of the funds in/out of the sandbox could somehow get automated? For example by storing in a queue the redeem amounts that could not be fulfilled, then doing a rebalance + execution of those withdrawals after a timelock has expired?

tomlinton:
Closing this for now

PR #830, opened 2021-11-23
Remove vault buffer from OUSD.

DanielVF:
Per https://vote.originprotocol.com/#/proposal/0x6d665c05cd2cc7debb64c4b66bf4a6f8d23b006e97b8ba113145783feab1b3e5

We're removing the vault buffer, which will increase the APY and simplicity of the code.

The downside is that all redeems will be gas expensive now, rather than just most redeems, however we have plenty of liquidity on swaps now to soak this up.

Contract change checklist:
  - [x] Unit tests pass


DanielVF:
Note that I left some mentions in in one-off scripts, just to keep the historical record. I'm open to changing this.

contracts/contracts/vault/VaultStorage.sol
```diff
@@ -73,7 +72,7 @@ contract VaultStorage is Initializable, Governable {
     // Redemption fee in basis points
     uint256 public redeemFeeBps;
     // Buffer of assets to keep in Vault to handle (most) withdrawals
-    uint256 public vaultBuffer;
+    uint256 public vaultBuffer; // No longer used
```
tomlinton, line 75: Elsewhere we've used `_deprecated_vaultBuffer` for these, might be nice to make it non-public too

contracts/contracts/vault/VaultCore.sol
```diff
@@ -340,11 +317,7 @@ contract VaultCore is VaultStorage {
             // No balance, nothing to do here
             if (assetBalance == 0) continue;
 
-            // Multiply the balance by the vault buffer modifier and truncate
-            // to the scale of the asset decimals
-            uint256 allocateAmount = assetBalance.mulTruncate(
-                vaultBufferModifier
-            );
+            uint256 allocateAmount = assetBalance;
```
franckc, line 320: nit: is this intermediary variable necessary? How about renaming assetBalance on line 316 to allocateAmount?

franckc (commented):
The less code the better :)

Will the deploy script be implemented in a separate PR?

contracts/contracts/vault/VaultCore.sol
DanielVF: Cleaner. Done.

contracts/contracts/vault/VaultStorage.sol
DanielVF: Done.

DanielVF:
> Will the deploy script be implemented in a separate PR?

I can make the deployment script as a part of this PR.

PR #832, opened 2021-11-24
Fix account dropdown

tomlinton:
Fixes the account change dropdown. It never opened the account screen so I couldn't disconnect an existing wallet.

Screen Shot 2021-11-24 at 1 29 00 PM


franckc (approved):
Shah and Domen are OOO. So stamping this PR on their behalf :)
Thanks for the fixin' !

sparrowDom:
Forgot about this PR. Have done something similar today, since I've experienced a similar frustration with not being able to disconnect from the incorrect network here: https://github.com/OriginProtocol/origin-dollar/pull/844

Mind if I close this one since the other PR has some other fixes?

Issue #634, opened 2021-08-13
Coinbase Wallet Connection Option


micahalcorn:
Coinbase Wallet is naturally supported in our DApp because it works with the same injected connector as MetaMask. In my experience, both of them connect automatically when "Connect" is clicked. This is different from the desktop version of MetaMask, which waits for the user to choose the MetaMask button from the listed options before connecting. As a result, there likely isn't a scenario where a user would actually click a "Coinbase Wallet" button within the mobile app. However, we should add a button basically advertising the fact that Coinbase Wallet is supported.

- [ ] Add a button to the [LoginWidget](https://github.com/OriginProtocol/origin-dollar/blob/master/dapp/src/components/LoginWidget.js) using [the icon that I recently committed](https://github.com/OriginProtocol/origin-dollar/commit/d7b5f984250393284874f873cd0dabbb06a0540f) with the text "Coinbase Wallet".
- [ ] Arrange the buttons as follows: MetaMask, Ledger, Coinbase Wallet, WalletConnect, MEW wallet.
- [ ] Wire the button up to the same injected connector as MetaMask.

micahalcorn:
Now that Coinbase Wallet has a browser extension, we can treat it just like we do MetaMask. I've updated the preceding comment accordingly.

One issue, however, is that both MetaMask and Coinbase Wallet prompts will appear if both are installed and a user clicks "MetaMask" **or** "Coinnbase Wallet". If there's a way to treat them differently, that would be a bonus. I don't think it's a big deal though because users who have both installed would likely see the same behavior on most other DApps.

joshfraser:
It might be worth figuring out how we can support Brave wallet while we're at it.

micahalcorn:
> It might be worth figuring out how we can support Brave wallet while we're at it.

Brave's built-in wallet works for me out of the box. It's the same concern that we have with Coinbase right now. You just click the MetaMask button and the same connector works. We might be able to identify that it's Brave's MetaMask fork rather than the browser extension and show a different button in that case. But it may get messy because Brave also has a config option for choosing whether to user their wallet or MetaMask when they're both enabled.

calvintoes:
https://github.com/OriginProtocol/origin-dollar/pull/838

franckc:
For the record, draft PR #838 could be useful starting point to look for the whoever works on this issue...

sparrowDom:
[...]

PR #838, opened 2021-11-28
added coinbase wallet option

calvintoes:
Added `@web3-react/walletlink-connector` package to add Coinbase Wallet option. Only Coinbase prompt pops up when you click Coinbase wallet. Metamask **and** Coinbase prompt pops up when you click Metamask. 

Walletlink connector implementation was borrowed from the way Uniswap does their connections. 

Also tested on Brave and it works out of the box.

---

Contract change checklist:
  - [ ] Unit tests pass


micahalcorn:
We shouldn't need to introduce WalletLink for this, should we? What do we gain by doing that?

calvintoes:
@micahalcorn I noticed that with WalletLink, only the Coinbase prompt pops up when you connect a Coinbase wallet. Thought that UX was better than using the same connector as MM and having both prompt pop up for connecting MM and/or Coinbase.

Another UX benefit with WalletLink, if the user tries to connect with Coinbase and doesn't have the extension installed, it will pop up a Coinbase modal asking to install the extension or scan QR code with phone. 
If we used the MM connector instead and the user doesn't have Coinbase extension, then it'll pop up a MM prompt instead (assuming user has MM extension installed), which doesn't make sense to me.

micahalcorn:
> @micahalcorn I noticed that with WalletLink, only the Coinbase prompt pops up when you connect a Coinbase wallet. Thought that UX was better than using the same connector as MM and having both prompt pop up for connecting MM and/or Coinbase.
> 
> Another UX benefit with WalletLink, if the user tries to connect with Coinbase and doesn't have the extension installed, it will pop up a Coinbase modal asking to install the extension or scan QR code with phone. If we used the MM connector instead and the user doesn't have Coinbase extension, then it'll pop up a MM prompt instead (assuming user has MM extension installed), which doesn't make sense to me.

Okay, as you point out, it still gives you redundant popups when clicking MetaMask, so I'm not sure if it's worth the additional dependency or not. I'll defer to @sparrowDom and @tomlinton. It would also be good to confirm that it plays nicely with WalletConnect.

sparrowDom:
I might be missing what was the goal with this PR. The Coinbase wallet connects nicely to what is currently in production: 
Screenshot_20211130-133745

micahalcorn:
> I might be missing what was the goal with this PR. The Coinbase wallet connects nicely to what is currently in production:

See https://github.com/OriginProtocol/origin-dollar/issues/634 🙂 Users were complaining because they thought we didn't support it. If you're in a desktop environment, you have to explicitly click a button and they didn't know to click the MetaMask option. So I was proposing that we just add a button, particularly now that Coinbase Wallet has a browser extension.

sparrowDom (commented):
Aside 2 minor comments I think it is great the way you did it. I've checked this brach-es build size and current master build size and it is almost identical. So the extra packages don't pose a thread to enlarge the website size too much. 

And we already use other connectors from `@web3-reach` repo, so aside from non problematic build size I see no problems having those extra QR-code features. 

great work 👍

sparrowDom:
> > I might be missing what was the goal with this PR. The Coinbase wallet connects nicely to what is currently in production:
> 
> See #634 🙂 Users were complaining because they thought we didn't support it. If you're in a desktop environment, you have to explicitly click a button and they didn't know to click the MetaMask option. So I was proposing that we just add a button, particularly now that Coinbase Wallet has a browser extension.

Thanks, I should have figured that out :)

package.json
```diff
@@ -0,0 +1,3 @@
+{
+  "dependencies": {}
```
sparrowDom, line 2: this can probably go out

dapp/src/utils/connectors.js
```diff
@@ -52,5 +61,37 @@ export const connectorNameIconMap = {
   WalletConnect: 'walletconnect-icon.svg',
 }
 
+export const connectorsByName = {
```
sparrowDom, line 64: This is probably a result of a bad merge?

sparrowDom (commented):
A couple of minor comments inline

dapp/src/utils/connectors.js
calvintoes: What should be removed/edited here?

tomlinton:
Thanks for the fixes @calvintoes. Looks good to me.

The dependencies of walletlink are a bit rough, preact and rxjs? I wonder why it requires that.

dapp/src/utils/connectors.js
sparrowDom: The `connectorsByName` has just been refactored to `connectorNameIconMap` while you are working on this PR: 
https://github.com/OriginProtocol/origin-dollar/blob/master/dapp/src/utils/connectors.js#L48-L53

franckc:
Closing this PR since @calvintoes mentioned to me he won't be able to bring it to the finish line...

PR #844, opened 2021-12-01
Various polish

sparrowDom:
- remove injected connector listeners since that is [already handled under the hood](https://github.com/NoahZinsmeister/web3-react/blob/d0b038c748a42ec85641a307e6c588546d86afc2/packages/injected-connector/src/index.ts#L74-L79)
- fix connector icon bug in Account dropdown

### DIFF
--- a/dapp/src/components/AccountStatusDropdown.js
+++ b/dapp/src/components/AccountStatusDropdown.js
@@ -33,7 +33,7 @@ const AccountStatusDropdown = ({ className, showLogin, dapp }) => {
           className={`account-status d-flex justify-content-center align-items-center clickable ${className} ${
             open ? 'open' : ''
           }`}
-          onClick={(e) => {
+          onClick={async (e) => {
             e.preventDefault()
             if (dapp && !active) {
               showLogin()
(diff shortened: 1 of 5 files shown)


shahthepro (approved):

dapp/src/components/AccountStatusDropdown.js
```diff
@@ -33,15 +33,17 @@ const AccountStatusDropdown = ({ className, showLogin, dapp }) => {
...
             e.preventDefault()
             if (dapp && !active) {
               showLogin()
             } else if (active && !correctNetwork) {
               analytics.track('On Change network', {
                 category: 'settings',
               })
-              switchEthereumChain()
+              // open the dropdown to allow disconnecting, while also requesting an auto switch to mainnet
+              await switchEthereumChain()
```
shahthepro, line 45: nit: should this be wrapped inside a try catch block?

sparrowDom: This in effect [calls this function](https://github.com/OriginProtocol/origin-dollar/blob/master/dapp/src/utils/web3.js#L18) on the window object and from the documentation it does not throw an error. Rather returns null on success and returns error on fail: https://eips.ethereum.org/EIPS/eip-3326

Anyway your comment made me check, because I wasn't sure before :)

sparrowDom:
tnx @shahthepro

PR #862, opened 2021-12-15
Ledger Curve fix

sparrowDom:
Ledger's provider underestimated the gas required for transaction when using Curve. This fixes the issue.


tomlinton:
Is this ledger specific? I think this is probably using the estimateGas function from [web3-provider-engine](https://github.com/MetaMask/web3-provider-engine/blob/main/subproviders/vm.js#L71), it probably should be a problem for other connectors if it is for ledger. We might need to pass some arguments to the constructor of web3-provider-engine or even the connectors if they allow it?

PR #869, opened 2021-12-21
Sparrow dom/multiple rewards strategy

sparrowDom:

Contract change checklist:
  - [ ] Unit tests pass

### DIFF
--- a/contracts/contracts/interfaces/IStrategy.sol
+++ b/contracts/contracts/interfaces/IStrategy.sol
@@ -53,11 +53,11 @@ interface IStrategy {
     /**
      * @dev The address array of the reward tokens for the Strategy.
      */
-    function rewardTokenAddresses() external pure returns (address[] memory);
+    function getRewardTokenAddresses() external pure returns (address[] memory);
 
     /**
      * @dev The threshold array (denominated in the reward token) over which the
      * vault will auto harvest on allocate calls.
      */
-    function rewardLiquidationThresholds() external pure returns (uint256[] memory);
+    function getRewardLiquidationThresholds() external pure returns (uint256[] memory);
 }
--- a/contracts/contracts/interfaces/IStrategy.sol
+++ b/contracts/contracts/interfaces/IStrategy.sol
@@ -48,16 +48,16 @@ interface IStrategy {
     /**
      * @dev Collect reward tokens from the Strategy.
      */
-    function collectRewardToken() external;
+    function collectRewardTokens() external;
 
     /**
-     * @dev The address of the reward token for the Strategy.
+     * @dev The address array of the reward tokens for the Strategy.
      */
-    function rewardTokenAddress() external pure returns (address);
+    function rewardTokenAddresses() external pure returns (address[] memory);
 
     /**
-     * @dev The threshold (denominated in the reward token) over which the
+     * @dev The threshold array (denominated in the reward token) over which the
      * vault will auto harvest on allocate calls.
      */
-    function rewardLiquidationThreshold() external pure returns (uint256);
+    function rewardLiquidationThresholds() external pure returns (uint256[] memory);
 }
--- a/contracts/contracts/strategies/ConvexStrategy.sol
+++ b/contracts/contracts/strategies/ConvexStrategy.sol
@@ -18,12 +18,6 @@ contract ConvexStrategy is BaseCurveStrategy {
     using StableMath for uint256;
     using SafeERC20 for IERC20;
 
-    event RewardTokenCollected(
-        address recipient,
-        address token,
-        uint256 amount
-    );
-
     event CvxRewardTokenAddressUpdated(
         address _oldAddress,
         address _newAddress
@@ -53,7 +47,7 @@ contract ConvexStrategy is BaseCurveStrategy {
     function initialize(
         address _platformAddress, // 3Pool address
         address _vaultAddress,
-        address _rewardTokenAddress, // CRV
+        address[] calldata _rewardTokenAddress, // CRV
         address _cvxRewardTokenAddress, // CVX
         address[] calldata _assets,
         address[] calldata _pTokens,
@@ -151,13 +145,13 @@ contract ConvexStrategy is BaseCurveStrategy {
     /**
      * @dev Collect accumulated CRV and CVX and send to Vault.
      */
-    function collectRewardToken() external override onlyVault nonReentrant {
+    function collectRewardTokens() external override onlyVault nonReentrant {
         // Collect CRV and CVX
         IRewardStaking(cvxRewardStakerAddress).getReward();
         // Send CRV
-        IERC20 crvToken = IERC20(rewardTokenAddress);
+        IERC20 crvToken = IERC20(rewardTokenAddresses[0]);
         uint256 balance = crvToken.balanceOf(address(this));
-        emit RewardTokenCollected(vaultAddress, rewardTokenAddress, balance);
+        emit RewardTokenCollected(vaultAddress, rewardTokenAddresses[0], balance);
         crvToken.safeTransfer(vaultAddress, balance);
         // Send CVX
         IERC20 cvxToken = IERC20(cvxRewardTokenAddress);
--- a/contracts/contracts/utils/InitializableAbstractStrategy.sol
+++ b/contracts/contracts/utils/InitializableAbstractStrategy.sol
@@ -80,6 +80,11 @@ abstract contract InitializableAbstractStrategy is Initializable, Governable {
         for (uint256 i = 0; i < assetCount; i++) {
             _setPTokenAddress(_assets[i], _pTokens[i]);
         }
+        uint256 rewardsCount = _rewardTokenAddresses.length;
+        rewardLiquidationThresholds = new uint256[](rewardsCount);
+        for (uint256 i = 0; i < rewardsCount; i++) {
+            rewardLiquidationThresholds[i] = 0;
+        }
     }
 
     /**
@@ -114,7 +119,7 @@ abstract contract InitializableAbstractStrategy is Initializable, Governable {
     }
 
     /**
-     * @dev Set the reward token address.
+     * @dev Set the reward token addresses.
      * @param _rewardTokenAddresses Address array of the reward token
      */
     function setRewardTokenAddresses(address[] calldata _rewardTokenAddresses)
@@ -126,7 +131,15 @@ abstract contract InitializableAbstractStrategy is Initializable, Governable {
     }
 
     /**
-     * @dev Set the reward token liquidation threshold.
+     * @dev Get the reward token addresses.
+     * @return address[] the reward token addresses.
+     */
+    function getRewardTokenAddresses() external view returns (address[] memory) {
+        return rewardTokenAddresses;
+    }
+
+    /**
+     * @dev Set the reward token liquidation thresholds.
      * @param _thresholds Threshold array in decimals of reward token that will
      * cause the Vault to claim and withdrawAll on allocate() calls.
      */
@@ -139,6 +152,14 @@ abstract contract InitializableAbstractStrategy is Initializable, Governable {
             _thresholds
         );
         rewardLiquidationThresholds = _thresholds;
+    }
+
+    /**
+     * @dev Get the reward token liquidation thresholds.
+     * @return uint256[] the reward token liquidation thresholds.
+     */
+    function getRewardLiquidationThresholds() external view returns (uint256[] memory) {
+        return rewardLiquidationThresholds;
     }
 
     /**
--- a/contracts/contracts/utils/InitializableAbstractStrategy.sol
+++ b/contracts/contracts/utils/InitializableAbstractStrategy.sol
@@ -128,6 +128,16 @@ abstract contract InitializableAbstractStrategy is Initializable, Governable {
     {
         emit RewardTokenAddressesUpdated(rewardTokenAddresses, _rewardTokenAddresses);
         rewardTokenAddresses = _rewardTokenAddresses;
+
+        uint256[] memory previousThresholds = rewardLiquidationThresholds;
+        rewardLiquidationThresholds = new uint256[](_rewardTokenAddresses.length);
+        for (uint256 i = 0; i < _rewardTokenAddresses.length; i++) {
+            if (previousThresholds.length > i) {
+                rewardLiquidationThresholds[i] = previousThresholds[i];
+            } else {
+                rewardLiquidationThresholds[i] = 0;
+            }
+        }
     }
 
     /**
@@ -147,6 +157,7 @@ abstract contract InitializableAbstractStrategy is Initializable, Governable {
         external
         onlyGovernor
     {
+        require(_thresholds.length == rewardTokenAddresses.length, "Invalid thresholds array size");
         emit RewardLiquidationThresholdsUpdated(
             rewardLiquidationThresholds,
             _thresholds
--- a/contracts/contracts/utils/InitializableAbstractStrategy.sol
+++ b/contracts/contracts/utils/InitializableAbstractStrategy.sol
@@ -16,11 +16,11 @@ abstract contract InitializableAbstractStrategy is Initializable, Governable {
     event PTokenRemoved(address indexed _asset, address _pToken);
     event Deposit(address indexed _asset, address _pToken, uint256 _amount);
     event Withdrawal(address indexed _asset, address _pToken, uint256 _amount);
-    event RewardTokenCollected(address recipient, uint256 amount);
-    event RewardTokenAddressUpdated(address _oldAddress, address _newAddress);
-    event RewardLiquidationThresholdUpdated(
-        uint256 _oldThreshold,
-        uint256 _newThreshold
+    event RewardTokenCollected(address recipient, address rewardToken, uint256 amount);
+    event RewardTokenAddressesUpdated(address[] _oldAddresses, address[] _newAddresses);
+    event RewardLiquidationThresholdsUpdated(
+        uint256[] _oldThreshold,
+        uint256[] _newThreshold
     );
 
     // Core address for the given platform
@@ -35,8 +35,8 @@ abstract contract InitializableAbstractStrategy is Initializable, Governable {
     address[] internal assetsMapped;
 
     // Reward token address
-    address public rewardTokenAddress;
-    uint256 public rewardLiquidationThreshold;
+    address[] public rewardTokenAddresses;
+    uint256[] public rewardLiquidationThresholds;
 
     // Reserved for future expansion
     int256[100] private _reserved;
@@ -45,21 +45,21 @@ abstract contract InitializableAbstractStrategy is Initializable, Governable {
      * @dev Internal initialize function, to set up initial internal state
      * @param _platformAddress Generic platform address
      * @param _vaultAddress Address of the Vault
-     * @param _rewardTokenAddress Address of reward token for platform
+     * @param _rewardTokenAddresses Address of reward token for platform
      * @param _assets Addresses of initial supported assets
      * @param _pTokens Platform Token corresponding addresses
      */
     function initialize(
         address _platformAddress,
         address _vaultAddress,
-        address _rewardTokenAddress,
+        address[] calldata _rewardTokenAddresses,
         address[] calldata _assets,
         address[] calldata _pTokens
     ) external onlyGovernor initializer {
         InitializableAbstractStrategy._initialize(
             _platformAddress,
             _vaultAddress,
-            _rewardTokenAddress,
+            _rewardTokenAddresses,
             _assets,
             _pTokens
         );
@@ -68,13 +68,13 @@ abstract contract InitializableAbstractStrategy is Initializable, Governable {
     function _initialize(
         address _platformAddress,
         address _vaultAddress,
-        address _rewardTokenAddress,
+        address[] calldata _rewardTokenAddresses,
         address[] memory _assets,
         address[] memory _pTokens
     ) internal {
         platformAddress = _platformAddress;
         vaultAddress = _vaultAddress;
-        rewardTokenAddress = _rewardTokenAddress;
+        rewardTokenAddresses = _rewardTokenAddresses;
         uint256 assetCount = _assets.length;
         require(assetCount == _pTokens.length, "Invalid input arrays");
         for (uint256 i = 0; i < assetCount; i++) {
@@ -85,11 +85,13 @@ abstract contract InitializableAbstractStrategy is Initializable, Governable {
     /**
      * @dev Collect accumulated reward token and send to Vault.
      */
-    function collectRewardToken() external virtual onlyVault nonReentrant {
-        IERC20 rewardToken = IERC20(rewardTokenAddress);
-        uint256 balance = rewardToken.balanceOf(address(this));
-        emit RewardTokenCollected(vaultAddress, balance);
-        rewardToken.safeTransfer(vaultAddress, balance);
+    function collectRewardTokens() external virtual onlyVault nonReentrant {
+        for (uint256 i = 0; i < rewardTokenAddresses.length; i++) {
+            IERC20 rewardToken = IERC20(rewardTokenAddresses[i]);
+            uint256 balance = rewardToken.balanceOf(address(this));
+            emit RewardTokenCollected(vaultAddress, rewardTokenAddresses[i], balance);
+            rewardToken.safeTransfer(vaultAddress, balance);
+        }
     }
 
     /**
@@ -113,30 +115,30 @@ abstract contract InitializableAbstractStrategy is Initializable, Governable {
 
     /**
      * @dev Set the reward token address.
-     * @param _rewardTokenAddress Address of the reward token
+     * @param _rewardTokenAddresses Address array of the reward token
      */
-    function setRewardTokenAddress(address _rewardTokenAddress)
+    function setRewardTokenAddresses(address[] calldata _rewardTokenAddresses)
         external
         onlyGovernor
     {
-        emit RewardTokenAddressUpdated(rewardTokenAddress, _rewardTokenAddress);
-        rewardTokenAddress = _rewardTokenAddress;
+        emit RewardTokenAddressesUpdated(rewardTokenAddresses, _rewardTokenAddresses);
+        rewardTokenAddresses = _rewardTokenAddresses;
     }
 
     /**
      * @dev Set the reward token liquidation threshold.
-     * @param _threshold Threshold amount in decimals of reward token that will
+     * @param _thresholds Threshold array in decimals of reward token that will
      * cause the Vault to claim and withdrawAll on allocate() calls.
      */
-    function setRewardLiquidationThreshold(uint256 _threshold)
+    function setRewardLiquidationThresholds(uint256[] calldata _thresholds)
         external
         onlyGovernor
     {
-        emit RewardLiquidationThresholdUpdated(
-            rewardLiquidationThreshold,
-            _threshold
+        emit RewardLiquidationThresholdsUpdated(
+            rewardLiquidationThresholds,
+            _thresholds
         );
-        rewardLiquidationThreshold = _threshold;
+        rewardLiquidationThresholds = _thresholds;
     }
 
     /**
--- a/contracts/contracts/vault/VaultAdmin.sol
+++ b/contracts/contracts/vault/VaultAdmin.sol
@@ -449,16 +449,13 @@ contract VaultAdmin is VaultStorage {
     {
         IStrategy strategy = IStrategy(_strategyAddr);
         _harvest(address(strategy));
-        address[] memory rewardTokens = strategy.rewardTokenAddresses();
+        address[] memory rewardTokens = strategy.getRewardTokenAddresses();
         uint256[] memory results = new uint256[](rewardTokens.length * 2);
 
         for (uint256 i = 0; i < rewardTokens.length; i++) {
-            uint256[] memory swapResult = _swap(rewardTokens[i]);
-            results[i * 2] = swapResult[0];
-            results[i * 2 + 1] = swapResult[1];
+            // TODO: is it necessary we return the swap results here?
+            _swap(rewardTokens[i]);
         }
-
-        return results;
     }
 
     /**
@@ -468,7 +465,7 @@ contract VaultAdmin is VaultStorage {
      */
     function _harvest(address _strategyAddr) internal {
         IStrategy strategy = IStrategy(_strategyAddr);
-        address[] memory rewardTokenAddresses = strategy.rewardTokenAddresses();
+        address[] memory rewardTokenAddresses = strategy.getRewardTokenAddresses();
         bool collectReward = true;
 
         for (uint256 i = 0; i < rewardTokenAddresses.length; i++) {
--- a/contracts/contracts/vault/VaultAdmin.sol
+++ b/contracts/contracts/vault/VaultAdmin.sol
@@ -450,8 +450,6 @@ contract VaultAdmin is VaultStorage {
         IStrategy strategy = IStrategy(_strategyAddr);
         _harvest(address(strategy));
         address[] memory rewardTokens = strategy.getRewardTokenAddresses();
-        uint256[] memory results = new uint256[](rewardTokens.length * 2);
-
         for (uint256 i = 0; i < rewardTokens.length; i++) {
             // TODO: is it necessary we return the swap results here?
             _swap(rewardTokens[i]);
--- a/contracts/contracts/vault/VaultAdmin.sol
+++ b/contracts/contracts/vault/VaultAdmin.sol
@@ -449,7 +449,16 @@ contract VaultAdmin is VaultStorage {
     {
         IStrategy strategy = IStrategy(_strategyAddr);
         _harvest(address(strategy));
-        return _swap(strategy.rewardTokenAddress());
+        address[] memory rewardTokens = strategy.rewardTokenAddresses();
+        uint256[] memory results = new uint256[](rewardTokens.length * 2);
+
+        for (uint256 i = 0; i < rewardTokens.length; i++) {
+            uint256[] memory swapResult = _swap(rewardTokens[i]);
+            results[i * 2] = swapResult[0];
+            results[i * 2 + 1] = swapResult[1];
+        }
+
+        return results;
     }
 
     /**
@@ -459,9 +468,17 @@ contract VaultAdmin is VaultStorage {
      */
     function _harvest(address _strategyAddr) internal {
         IStrategy strategy = IStrategy(_strategyAddr);
-        address rewardTokenAddress = strategy.rewardTokenAddress();
-        if (rewardTokenAddress != address(0)) {
-            strategy.collectRewardToken();
+        address[] memory rewardTokenAddresses = strategy.rewardTokenAddresses();
+        bool collectReward = true;
+
+        for (uint256 i = 0; i < rewardTokenAddresses.length; i++) {
+            if (rewardTokenAddresses[i] == address(0)) {
+                collectReward = false;
+            }
+        }
+
+        if (collectReward) {
+            strategy.collectRewardTokens();
         }
     }
 
--- a/contracts/contracts/vault/VaultCore.sol
+++ b/contracts/contracts/vault/VaultCore.sol
@@ -217,24 +217,27 @@ contract VaultCore is VaultStorage {
         // date picture of total assets before allocating to strategies.
         for (uint256 i = 0; i < allStrategies.length; i++) {
             IStrategy strategy = IStrategy(allStrategies[i]);
-            address rewardTokenAddress = strategy.rewardTokenAddress();
-            if (rewardTokenAddress != address(0)) {
-                uint256 liquidationThreshold = strategy
-                    .rewardLiquidationThreshold();
-                if (liquidationThreshold == 0) {
-                    // No threshold set, always harvest from strategy
-                    IVault(address(this)).harvestAndSwap(allStrategies[i]);
-                } else {
-                    // Check balance against liquidation threshold
-                    // Note some strategies don't hold the reward token balance
-                    // on their contract so the liquidation threshold should be
-                    // set to 0
-                    IERC20 rewardToken = IERC20(rewardTokenAddress);
-                    uint256 rewardTokenAmount = rewardToken.balanceOf(
-                        allStrategies[i]
-                    );
-                    if (rewardTokenAmount >= liquidationThreshold) {
+            address[] memory rewardTokenAddresses = strategy.rewardTokenAddresses();
+            uint256[] memory liquidationThresholds = strategy.rewardLiquidationThresholds();
+            require(rewardTokenAddresses.length == liquidationThresholds.length, "Reward token array and liquidation array must be of the same size");
+
+            for (uint256 j = 0; j < rewardTokenAddresses.length; j++) {
+                if (rewardTokenAddresses[j] != address(0)) {
+                    if (liquidationThresholds[j] == 0) {
+                        // No threshold set, always harvest from strategy
                         IVault(address(this)).harvestAndSwap(allStrategies[i]);
+                    } else {
+                        // Check balance against liquidation threshold
+                        // Note some strategies don't hold the reward token balance
+                        // on their contract so the liquidation threshold should be
+                        // set to 0
+                        IERC20 rewardToken = IERC20(rewardTokenAddresses[j]);
+                        uint256 rewardTokenAmount = rewardToken.balanceOf(
+                            allStrategies[i]
+                        );
+                        if (rewardTokenAmount >= liquidationThresholds[j]) {
+                            IVault(address(this)).harvestAndSwap(allStrategies[i]);
+                        }
                     }
                 }
             }
--- a/contracts/contracts/vault/VaultCore.sol
+++ b/contracts/contracts/vault/VaultCore.sol
@@ -217,8 +217,8 @@ contract VaultCore is VaultStorage {
         // date picture of total assets before allocating to strategies.
         for (uint256 i = 0; i < allStrategies.length; i++) {
             IStrategy strategy = IStrategy(allStrategies[i]);
-            address[] memory rewardTokenAddresses = strategy.rewardTokenAddresses();
-            uint256[] memory liquidationThresholds = strategy.rewardLiquidationThresholds();
+            address[] memory rewardTokenAddresses = strategy.getRewardTokenAddresses();
+            uint256[] memory liquidationThresholds = strategy.getRewardLiquidationThresholds();
             require(rewardTokenAddresses.length == liquidationThresholds.length, "Reward token array and liquidation array must be of the same size");
 
             for (uint256 j = 0; j < rewardTokenAddresses.length; j++) {
--- a/contracts/contracts/strategies/AaveStrategy.sol
+++ b/contracts/contracts/strategies/AaveStrategy.sol
@@ -28,7 +28,7 @@ contract AaveStrategy is InitializableAbstractStrategy {
      * addresses for the rewards program.
      * @param _platformAddress Address of the AAVE pool
      * @param _vaultAddress Address of the vault
-     * @param _rewardTokenAddress Address of the AAVE token
+     * @param _rewardTokenAddresses Address of the AAVE token
      * @param _assets Addresses of supported assets
      * @param _pTokens Platform Token corresponding addresses
      * @param _incentivesAddress Address of the AAVE incentives controller
@@ -37,7 +37,7 @@ contract AaveStrategy is InitializableAbstractStrategy {
     function initialize(
         address _platformAddress, // AAVE pool
         address _vaultAddress,
-        address _rewardTokenAddress, // AAVE
+        address[] calldata _rewardTokenAddresses, // AAVE
         address[] calldata _assets,
         address[] calldata _pTokens,
         address _incentivesAddress,
@@ -48,7 +48,7 @@ contract AaveStrategy is InitializableAbstractStrategy {
         InitializableAbstractStrategy._initialize(
             _platformAddress,
             _vaultAddress,
-            _rewardTokenAddress,
+            _rewardTokenAddresses,
             _assets,
             _pTokens
         );
@@ -234,7 +234,7 @@ contract AaveStrategy is InitializableAbstractStrategy {
     /**
      * @dev Collect stkAave, convert it to AAVE send to Vault.
      */
-    function collectRewardToken() external override onlyVault nonReentrant {
+    function collectRewardTokens() external override onlyVault nonReentrant {
         if (address(stkAave) == address(0)) {
             return;
         }
@@ -249,15 +249,15 @@ contract AaveStrategy is InitializableAbstractStrategy {
         if (block.timestamp > windowStart && block.timestamp <= windowEnd) {
             // Redeem to AAVE
             uint256 stkAaveBalance = stkAave.balanceOf(address(this));
-            if (stkAaveBalance > rewardLiquidationThreshold) {
+            if (stkAaveBalance > rewardLiquidationThresholds[0]) {
                 stkAave.redeem(address(this), stkAaveBalance);
             }
             // Transfer AAVE to vaultAddress
-            uint256 aaveBalance = IERC20(rewardTokenAddress).balanceOf(
+            uint256 aaveBalance = IERC20(rewardTokenAddresses[0]).balanceOf(
                 address(this)
             );
             if (aaveBalance > 0) {
-                IERC20(rewardTokenAddress).safeTransfer(
+                IERC20(rewardTokenAddresses[0]).safeTransfer(
                     vaultAddress,
                     aaveBalance
                 );
--- a/contracts/contracts/strategies/CompoundStrategy.sol
+++ b/contracts/contracts/strategies/CompoundStrategy.sol
@@ -20,15 +20,15 @@ contract CompoundStrategy is InitializableAbstractStrategy {
     /**
      * @dev Collect accumulated COMP and send to Vault.
      */
-    function collectRewardToken() external override onlyVault nonReentrant {
+    function collectRewardTokens() external override onlyVault nonReentrant {
         // Claim COMP from Comptroller
         ICERC20 cToken = _getCTokenFor(assetsMapped[0]);
         IComptroller comptroller = IComptroller(cToken.comptroller());
         comptroller.claimComp(address(this));
         // Transfer COMP to Vault
-        IERC20 rewardToken = IERC20(rewardTokenAddress);
+        IERC20 rewardToken = IERC20(rewardTokenAddresses[0]);
         uint256 balance = rewardToken.balanceOf(address(this));
-        emit RewardTokenCollected(vaultAddress, balance);
+        emit RewardTokenCollected(vaultAddress, rewardTokenAddresses[0], balance);
         rewardToken.safeTransfer(vaultAddress, balance);
     }
 
--- a/contracts/contracts/strategies/ThreePoolStrategy.sol
+++ b/contracts/contracts/strategies/ThreePoolStrategy.sol
@@ -38,7 +38,7 @@ contract ThreePoolStrategy is BaseCurveStrategy {
     function initialize(
         address _platformAddress, // 3Pool address
         address _vaultAddress,
-        address _rewardTokenAddress, // CRV
+        address[] calldata _rewardTokenAddress, // CRV
         address[] calldata _assets,
         address[] calldata _pTokens,
         address _crvGaugeAddress,
@@ -112,13 +112,13 @@ contract ThreePoolStrategy is BaseCurveStrategy {
     /**
      * @dev Collect accumulated CRV and send to Vault.
      */
-    function collectRewardToken() external override onlyVault nonReentrant {
+    function collectRewardTokens() external override onlyVault nonReentrant {
         // Collect
         ICRVMinter(crvMinterAddress).mint(crvGaugeAddress);
         // Send
-        IERC20 crvToken = IERC20(rewardTokenAddress);
+        IERC20 crvToken = IERC20(rewardTokenAddresses[0]);
         uint256 balance = crvToken.balanceOf(address(this));
-        emit RewardTokenCollected(vaultAddress, balance);
+        emit RewardTokenCollected(vaultAddress, rewardTokenAddresses[0], balance);
         crvToken.safeTransfer(vaultAddress, balance);
     }
 }
--- a/contracts/deploy/001_core.js
+++ b/contracts/deploy/001_core.js
@@ -52,14 +52,14 @@ const deployAaveStrategy = async () => {
 
   log("Initialized AaveStrategyProxy");
   const initFunctionName =
-    "initialize(address,address,address,address[],address[],address,address)";
+    "initialize(address,address,address[],address[],address[],address,address)";
   await withConfirmation(
     cAaveStrategy
       .connect(sDeployer)
       [initFunctionName](
         assetAddresses.AAVE_ADDRESS_PROVIDER,
         cVaultProxy.address,
-        assetAddresses.AAVE_TOKEN,
+        [assetAddresses.AAVE_TOKEN],
         [assetAddresses.DAI],
         [assetAddresses.aDAI],
         cAaveIncentivesController.address,
@@ -126,7 +126,7 @@ const deployCompoundStrategy = async () => {
       .initialize(
         addresses.dead,
         cVaultProxy.address,
-        assetAddresses.COMP,
+        [assetAddresses.COMP],
         [assetAddresses.DAI],
         [assetAddresses.cDAI]
       )
@@ -188,11 +188,11 @@ const deployThreePoolStrategy = async () => {
     cThreePoolStrategy
       .connect(sDeployer)
       [
-        "initialize(address,address,address,address[],address[],address,address)"
+        "initialize(address,address,address[],address[],address[],address,address)"
       ](
         assetAddresses.ThreePool,
         cVaultProxy.address,
-        assetAddresses.CRV,
+        [assetAddresses.CRV],
         [assetAddresses.DAI, assetAddresses.USDC, assetAddresses.USDT],
         [
           assetAddresses.ThreePoolToken,
@@ -260,11 +260,11 @@ const deployConvexStrategy = async () => {
     cConvexStrategy
       .connect(sDeployer)
       [
-        "initialize(address,address,address,address,address[],address[],address,address,uint256)"
+        "initialize(address,address,address[],address,address[],address[],address,address,uint256)"
       ](
         assetAddresses.ThreePool,
         cVaultProxy.address,
-        assetAddresses.CRV,
+        [assetAddresses.CRV],
         assetAddresses.CVX,
         [assetAddresses.DAI, assetAddresses.USDC, assetAddresses.USDT],
         [
--- a/contracts/deploy/032_convex_rewards.js
+++ b/contracts/deploy/032_convex_rewards.js
@@ -27,7 +27,7 @@ module.exports = deploymentWithProposal(
       "ConvexStrategy",
       undefined,
       undefined,
-      true // Disable storage slot checking. We are intentionaly renaming a slot.
+      true // Disable storage slot checking. We are intentionally renaming a slot.
     );
     const cConvexStrategyProxy = await ethers.getContract(
       "ConvexStrategyProxy"
@@ -53,8 +53,8 @@ module.exports = deploymentWithProposal(
         // 2. Use CRV as main rewards token
         {
           contract: cConvexStrategy,
-          signature: "setRewardTokenAddress(address)",
-          args: [assetAddresses.CRV],
+          signature: "setRewardTokenAddresses(address[])",
+          args: [[assetAddresses.CRV]],
         },
         // 2. Use correct CVX token addresss
         {
--- a/contracts/node.sh
+++ b/contracts/node.sh
@@ -50,6 +50,7 @@ main()
 
         FORK=true npx hardhat fund --amount 100000 --network localhost --accountsfromenv true &
         
+        # wait for subprocesses to finish
         for job in `jobs -p`
           do
             wait $job || let "FAIL+=1"
--- a/contracts/node.sh
+++ b/contracts/node.sh
@@ -32,13 +32,12 @@ main()
         # the --no-install is here so npx doesn't download some package on its own if it can not find one in the repo
         FORK=true npx --no-install hardhat node --no-reset --export '../dapp/network.json' ${params[@]} > $nodeOutput 2>&1 &
 
-        echo "Node output: $nodeOutput"
-        echo "Waiting for node to initialize:"
+        tail -f $nodeOutput &
+
         i=0
         until grep -q -i 'Started HTTP and WebSocket JSON-RPC server at' $nodeOutput
         do
           let i++
-          printf "."
           sleep 1
           if (( i > nodeWaitTimeout )); then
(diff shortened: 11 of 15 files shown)


brownie/scripts/convex_tvl_tokens.py
```diff
@@ -0,0 +1,3 @@
+crvRPool = Contract.from_explorer('0x689440f2Ff927E1f24c72F1087E1FAF471eCe1c8')
+# Convex strategy proxy
+crvRPool.balanceOf('0xEA2Ef2e2E5A749D4A66b41Db9aD85a38Aa264cb3')
```
DanielVF, line 3: This script is still incomplete?

contracts/contracts/interfaces/IStrategy.sol
```diff
@@ -48,16 +48,20 @@ interface IStrategy {
...
-     * @dev The address of the reward token for the Strategy.
+     * @dev The address array of the reward tokens for the Strategy.
      */
-    function rewardTokenAddress() external pure returns (address);
+    function getRewardTokenAddresses() external pure returns (address[] memory);
 
     /**
-     * @dev The threshold (denominated in the reward token) over which the
-     * vault will auto harvest on allocate calls.
+     * @dev The limit array (denominated in the reward token) which is the
```
DanielVF, line 59: Nice adding documentation here.

contracts/contracts/vault/VaultAdmin.sol
```diff
@@ -449,7 +451,14 @@ contract VaultAdmin is VaultStorage {
     {
         IStrategy strategy = IStrategy(_strategyAddr);
         _harvest(address(strategy));
-        return _swap(strategy.rewardTokenAddress());
+        address[] memory rewardTokens = strategy.getRewardTokenAddresses();
+        uint256[] memory liquidationLimits = strategy.getRewardLiquidationLimits();
+        require(rewardTokens.length == liquidationLimits.length, "Reward token array and liquidation limit array must be of the same size");
+
+        for (uint256 i = 0; i < rewardTokens.length; i++) {
+            // TODO: is it necessary we return the swap results here?
```
DanielVF, line 459: No, I don't think we need to return the swap results.

contracts/contracts/vault/VaultAdmin.sol
```diff
@@ -459,18 +468,49 @@ contract VaultAdmin is VaultStorage {
...
     function _harvest(address _strategyAddr) internal {
         IStrategy strategy = IStrategy(_strategyAddr);
-        address rewardTokenAddress = strategy.rewardTokenAddress();
-        if (rewardTokenAddress != address(0)) {
-            strategy.collectRewardToken();
+        address[] memory rewardTokenAddresses = strategy.getRewardTokenAddresses();
+        bool collectReward = true;
+
+        for (uint256 i = 0; i < rewardTokenAddresses.length; i++) {
+            if (rewardTokenAddresses[i] == address(0)) {
```
DanielVF, line 475: It would probably be better to swap this around so that if any reward token is set, we call collect

contracts/contracts/vault/VaultCore.sol
```diff
@@ -217,27 +217,7 @@ contract VaultCore is VaultStorage {
...
-                    IERC20 rewardToken = IERC20(rewardTokenAddress);
-                    uint256 rewardTokenAmount = rewardToken.balanceOf(
-                        allStrategies[i]
-                    );
-                    if (rewardTokenAmount >= liquidationThreshold) {
-                        IVault(address(this)).harvestAndSwap(allStrategies[i]);
-                    }
-                }
-            }
+            IVault(address(this)).harvestAndSwap(allStrategies[i]);
```
DanielVF, line 220: Well this is much nicer! Way to make the code better!

contracts/deploy/035_multiple_rewards_strategies.js
```diff
@@ -0,0 +1,179 @@
...
+      undefined,
+      true // Disable storage slot checking, we are renaming variables in InitializableAbstractStrategy.
+    );
+    const dAaveStrategyImpl = await deployWithConfirmation(
+      "AaveStrategy",
+      undefined,
+      undefined,
+      true // Disable storage slot checking, we are renaming variables in InitializableAbstractStrategy.
+    );
+    const dThreePoolStrategyImpl = await deployWithConfirmation(
```
DanielVF, line 52: We don't have three pool deployed to production right now. We can skip deploying an upgrade.

contracts/contracts/utils/InitializableAbstractStrategy.sol
```diff
@@ -35,8 +36,8 @@ abstract contract InitializableAbstractStrategy is Initializable, Governable {
     address[] internal assetsMapped;
 
     // Reward token address
-    address public rewardTokenAddress;
-    uint256 public rewardLiquidationThreshold;
+    address[] public rewardTokenAddresses;
```
DanielVF, line 39: We need to depreciate these two slots and add the new ones after. (and decrement the reserved appropriately.

DanielVF:
Have you fork tested it yet?

brownie/scripts/convex_tvl_tokens.py
sparrowDom: yup tnx will update this

contracts/contracts/utils/InitializableAbstractStrategy.sol
sparrowDom: I was thinking that since arrays always take a new & full solidity storage slot we might get away with just re-setting both of the array values for all of the strategies. (Haven't confirmed this with on-chain testing yet though - will do that today or tomorrow). 

I you think the deprecation and decrementing is a better approach I can still do that.

contracts/contracts/vault/VaultAdmin.sol
sparrowDom: Cool changing so that function doesn't have a return type.

contracts/contracts/vault/VaultAdmin.sol
sparrowDom: Hmm yeah that makes much more sense... thx changing

contracts/contracts/vault/VaultCore.sol
sparrowDom: tnx

e2870a557842b5a32324e83b2cdab540

😄

contracts/deploy/035_multiple_rewards_strategies.js
sparrowDom: Tnx removed

contracts/contracts/vault/VaultAdmin.sol
sparrowDom: Looking at this code again, we will probably never have zero address elements in the array right? So something like this might make more sense:
```
if (rewardTokenAddresses.length > 0) {
  strategy.collectRewardTokens();
}
```

contracts/test/helpers.js
```diff
@@ -250,6 +250,7 @@ const getAssetAddresses = async (deployments) => {
       aUSDC: addresses.mainnet.aUSDC,
       aUSDT: addresses.mainnet.aUSDT,
       AAVE: addresses.mainnet.Aave,
+      AAVE_TOKEN: addresses.mainnet.Aave,
```
sparrowDom, line 253: @DanielVF just wanted to run this by you. Didn't see AAVE_TOKEN yet setup for the mainnet address. I might be confused here since the mainent AAVE is ERC20 and the local AAVE (MockAave) is not ERC20. Not sure if what I've done here is the appropriate solution. 

With the above change I am setting `AAVE_TOKEN ` as a reward token in the deploy script.

contracts/contracts/vault/VaultAdmin.sol
```diff
@@ -401,116 +348,6 @@ contract VaultAdmin is VaultStorage {
         IERC20(_asset).safeTransfer(governor(), _amount);
     }
 
-    /**
```
DanielVF, line 404: this code getting deleted from the vault makes me so happpppppyyyyyyy!

contracts/contracts/vault/VaultAdmin.sol
```diff
@@ -57,6 +57,8 @@ contract VaultAdmin is VaultStorage {
      * @param _redeemFeeBps Basis point fee to be charged
      */
     function setRedeemFeeBps(uint256 _redeemFeeBps) external onlyGovernor {
+        // TODO: do we need an additional sanity check here? As in not more than 10% or something?
+        require(_redeemFeeBps <= 1e4, "Invalid value");
```
DanielVF, line 61: I like your thinking. Yes, I think a 10% sanity cap here is good.

contracts/contracts/vault/VaultAdmin.sol
```diff
@@ -169,6 +153,16 @@ contract VaultAdmin is VaultStorage {
         emit AssetSupported(_asset);
     }
 
+    /**
+     * @dev Is supported asset
+     * @param _asset Address of the asset
+     */
+    function isSupportedAsset(address _asset) external view returns (bool) {
+        // TODO: if the asset mapping does not yet exist mapping will point to
+        // non initialized (empty) space and should return false?
```
DanielVF, line 162: That would be my understanding too, but we should check it with a unit test.

contracts/contracts/oracle/OracleRouter.sol
```diff
@@ -78,6 +78,11 @@ contract OracleRouter is OracleRouterBase {
         ) {
             // Chainlink: CRV/USD
             return address(0xCd627aA160A6fA45Eb793D19Ef54f5062F20f33f);
+        } else if (
+            asset == address(0x4e3FBD56CD56c3e72c1403e103b45Db9da5B9D2B)
```
DanielVF, line 82: VERIFIED: Convex Token

contracts/contracts/oracle/OracleRouter.sol
```diff
@@ -78,6 +78,11 @@ contract OracleRouter is OracleRouterBase {
...
+        ) {
+            // Chainlink: CVX/USD
+            return address(0xd962fC30A72A84cE50161031391756Bf2876Af5D);
```
DanielVF, line 85: VERIFIED: CVX/USD oracle.
CAUTION: Need to verify that it is working before going live.

contracts/contracts/harvest/Harvester.sol
```diff
@@ -0,0 +1,337 @@
...
+     *      contract, i.e. mistaken sends.
+     * @param _asset Address for the asset
+     * @param _amount Amount of the asset to transfer
+     */
+    function transferToken(address _asset, uint256 _amount)
+        external
+        onlyGovernor
+    {
+        require(
+            !IVault(vaultAddress).isSupportedAsset(_asset),
```
DanielVF, line 156: We don't need this check here, since this is not the vault. On the vault we didn't want the governor to be able to steal principal.

contracts/contracts/harvest/Harvester.sol
```diff
@@ -0,0 +1,337 @@
...
+contract Harvester is Initializable, Governable {
+    using SafeERC20 for IERC20;
+    using SafeMath for uint256;
+    using StableMath for uint256;
+
+    // Tokens that should be swapped for stablecoins
+    address[] public swapTokens;
+
+    // Address of Uniswap
+    address public uniswapAddr = address(0);
```
DanielVF, line 27: We should have this be per token, since we can get better deals on different routers.

We should also have the fee be per token, but in this contract, not on the strategies. That's not really something the strategies need to be concerned about.

Both should be in a struct mapped to the token. Also in the same struct should be the max slippage allowed for selling that token.

contracts/contracts/harvest/Harvester.sol
```diff
@@ -0,0 +1,337 @@
...
+
+    // Address of Vault
+    address public vaultAddress = address(0);
+
+    // Reserved for future expansion
+    int256[100] private _reserved;
```
DanielVF, line 33: We only need this reserved if something else is going to inherit from it.

DanielVF (commented):
Did a prelim look through at the smart contracts only.

contracts/contracts/harvest/Harvester.sol
sparrowDom: Good point removing it

contracts/contracts/harvest/Harvester.sol
sparrowDom: Oh got it... stealing harvest reward tokens before calling the swap and sending them to the vault + caller is obviously significantly smaller issue.

contracts/contracts/interfaces/IStrategy.sol
sparrowDom: 👍

contracts/contracts/utils/InitializableAbstractStrategy.sol
```diff
@@ -206,6 +302,37 @@ abstract contract InitializableAbstractStrategy is Initializable, Governable {
...
+        harvesterAddress = _harvesterAddress;
+        emit HarvesterAddressesUpdated(harvesterAddress, _harvesterAddress);
+    }
+
+    /**
+     * @dev Set a fee in basis points to be rewarded when calling harvest
+     * @param _harvestRewardBps Basis point fee to be rewarded
+     */
+    function setHarvestRewardBps(uint32 _harvestRewardBps) external onlyGovernor {
+        require(_harvestRewardBps <= 1000, "Harvest reward fee should not be over 10%");
```
sparrowDom, line 322: @DanielVF you fine with having a 10% hard limit here as well?

contracts/contracts/vault/VaultAdmin.sol
sparrowDom: 🎉 🎉 🎉

contracts/contracts/utils/InitializableAbstractStrategy.sol
DanielVF: Yes.

contracts/contracts/harvest/Harvester.sol
sparrowDom: Good deal, I will add reward liquidation limit - a per token setting of how much token can be sold at once to this struct as well

contracts/contracts/harvest/Harvester.sol
```diff
@@ -0,0 +1,363 @@
...
+                ).swapExactTokensForTokens(
+                        maxBalanceToSwap,
+                        minExpected,
+                        path,
+                        address(this),
+                        block.timestamp
+                    );
+
+                IERC20 usdt = IERC20(allAssets[1]); // USDT
+                uint256 usdTbalance = usdt.balanceOf(address(this));
```
tomlinton, line 344: Inconsistent capitalisation on usdTbalance

contracts/contracts/oracle/OracleRouter.sol
DanielVF: This oracle is now showing data, and we have official word from chainlink that it is ready for production.

contracts/contracts/harvest/Harvester.sol
sparrowDom: Lol weird, thanks fixing

contracts/contracts/harvest/Harvester.sol
```diff
@@ -0,0 +1,367 @@
...
+        RewardTokenConfig memory tokenConfig = RewardTokenConfig({
+            harvestRewardBps: _harvestRewardBps,
+            allowedSlippageBps: _allowedSlippageBps,
+            uniswapV2CompatibleAddr: _uniswapV2CompatibleAddr,
+            liquidationLimit: _liquidationLimit
+        });
+
+        rewardTokenConfigs[_tokenAddress] = tokenConfig;
+
+        // Give Uniswap infinite approval when needed
```
DanielVF, line 161: We also need to remove approvals to the old uniswap if it's a change.

contracts/contracts/oracle/OracleRouter.sol
sparrowDom: sweet

contracts/contracts/harvest/Harvester.sol
sparrowDom: Good point adding that in 👍

contracts/contracts/harvest/Harvester.sol
```diff
@@ -0,0 +1,376 @@
...
+        uint256 _liquidationLimit
+    );
+
+    // Tokens that should be swapped for stablecoins
+    address[] public swapTokens;
+
+    // Strategies approved for use by the Vault
+    struct RewardTokenConfig {
+        // Max allowed slippage when swapping reward token for a stablecoin denominated in basis points.
+        uint32 allowedSlippageBps;
```
DanielVF, line 40: Since bps maxes out at 10,000 = 100%, we could change these to uint16s for storage, which hold up to 65,536.

contracts/contracts/harvest/Harvester.sol
```diff
@@ -0,0 +1,376 @@
...
+        returns (uint256[] memory swapResult)
+    {
+        RewardTokenConfig storage tokenConfig = rewardTokenConfigs[_swapToken];
+        require(
+            tokenConfig.uniswapV2CompatibleAddr != address(0),
+            "Swap token is missing token configuration."
+        );
+
+        address priceProvider = IVault(vaultAddress).priceProvider();
+        address[] memory allAssets = IVault(vaultAddress).getAllAssets();
```
DanielVF, line 315: This is gas expensive to load from the vault, and we are loading several addresses we don't need. Let's just have USDT be local to the contract.

contracts/contracts/harvest/Harvester.sol
```diff
@@ -0,0 +1,376 @@
...
+
+        if (tokenConfig.uniswapV2CompatibleAddr != address(0)) {
```
DanielVF, line 317: Don't we just check this in a require statement a few lines above?

Sometimes we may want to disable selling a token perhaps by setting the address to zero.

I think we should just return if the address is zero, in the place we currently have the require statement.

contracts/contracts/harvest/Harvester.sol
```diff
@@ -0,0 +1,376 @@
...
+    /**
+     * @dev Swap a record token for stablecoins for Uniswap. The token must have
+     *       a registered price feed with the price provider.
+     * @param _swapToken Address of the token to swap.
+     */
+    function _swap(address _swapToken)
+        internal
+        returns (uint256[] memory swapResult)
+    {
+        RewardTokenConfig storage tokenConfig = rewardTokenConfigs[_swapToken];
```
DanielVF, line 308: We probably want to load this to `memory` here, rather than reloading from `storage` every time we get a value from it. 


Would be worth comparing the gas usage difference between memory and storage in this line.

contracts/contracts/harvest/Harvester.sol
```diff
@@ -0,0 +1,388 @@
...
+                    tokenConfig.harvestRewardBps > 0,
+                    "Harvest rewards can not be zero"
+                );
+                require(
+                    vaultBps > tokenConfig.harvestRewardBps,
+                    "Address calling harvest is receiving more rewards than the vault"
+                );
+
+                // When governor calls the swap function send full reward amount to the Vault
+                if (isGovernor()) {
```
sparrowDom, line 373: @DanielVF @tomlinton made this change. The idea is that governor calling `harvest` and `swap` and `harvestAndSwap` doesn't get the 1% cut of the sold rewards. Let me know if any objections. 

Also would it be gas cheaper to pass down to this `_swap` function whether to send 1% to the `msg.sender` instead of calling `isGovernor ()`?

contracts/contracts/harvest/Harvester.sol
```diff
@@ -0,0 +1,388 @@
...
+     * @param _strategyAddr Address of the strategy to collect rewards from
+     */
+    function harvestAndSwap(address _strategyAddr) external nonReentrant {
+        IStrategy strategy = IStrategy(_strategyAddr);
+        _harvest(address(strategy));
+        address[] memory rewardTokens = strategy.getRewardTokenAddresses();
+        for (uint256 i = 0; i < rewardTokens.length; i++) {
+            _swap(rewardTokens[i]);
+        }
+        if (msg.sender != vaultAddress) {
```
sparrowDom, line 271: @DanielVF @tomlinton made this change and an identical one in a `harvestAndSwap()` without strategy address parameter. The issues is that Vault's `_allocate` calls this function and if we call back `rebase` the re-entrancy gets triggered. Also we probably do not want Vault's `allocate` to also trigger a rebase.

contracts/contracts/harvest/Harvester.sol
sparrowDom: good point thanks, changing to uint16

contracts/contracts/harvest/Harvester.sol
sparrowDom: Thanks you are right. Ran a gas estimate function and calling `harvestAndSwap` total gas usage reduced from 466115 to 465213 (reduced by 902) when changing from `storage` to `memory`.

contracts/contracts/harvest/Harvester.sol
sparrowDom: Thanks good thinking. This results in additional gas reduction from 465213 to 463599 (reduced by 1614).

contracts/contracts/harvest/Harvester.sol
sparrowDom: Thanks removed the double check. And added comments to `uniswapV2CompatibleAddr` property that setting to zero address disables swaps.

contracts/contracts/harvest/Harvester.sol
DanielVF: Vault allocate should no longer harvest - just move funds into strategies. Sorry, I had not gotten to reviewing the vault changes yet.

contracts/contracts/harvest/Harvester.sol
```diff
@@ -0,0 +1,384 @@
...
+                    maxBalanceToSwap,
+                    minExpected,
+                    path,
+                    address(this),
+                    block.timestamp
+                );
+
+            IERC20 usdt = IERC20(usdtAddress);
+            uint256 usdtBalance = usdt.balanceOf(address(this));
+            uint16 vaultBps = 1e4 - tokenConfig.harvestRewardBps;
```
DanielVF, line 362: This roughly does:

```
a = pct * total / decimals;
b = (1-pct) * total / decimals;
```

A "standard" way to calculate a percentage split of something has evolved in solidity, which looks like this:

```python
a = pct * total / decimals;
b = total - a;
```

This has a few advantages:

- No rounding errors. The other method will have rounding errors on both A and B, which will usualy result in a + b not quite equaling the total.
- Hard math in just one place. Getting all the decimals right is a pain. With the standard way, you only have to do this once.
- Can't screw up the amounts beyond the total when using uints. No matter what you do, `a + b` is always going to exactly equal total. If `a` is miscalculated too high, the subtraction will throw and save you. If A is too low, well, B gets more more, but no funds have gone into limbo, and B can't get more than total.

contracts/contracts/harvest/Harvester.sol
```diff
@@ -0,0 +1,384 @@
...
+            require(
```
DanielVF, line 363: We may want to have zero rewards. It's fine to remove this require.

contracts/contracts/harvest/Harvester.sol
```diff
@@ -0,0 +1,384 @@
...
+                tokenConfig.harvestRewardBps > 0,
+                "Harvest rewards can not be zero"
+            );
+            require(
+                vaultBps > tokenConfig.harvestRewardBps,
+                "Address calling harvest is receiving more rewards than the vault"
+            );
+
+            // When governor calls the swap function send full reward amount to the Vault
+            if (isGovernor()) {
```
DanielVF, line 373: If the above calculations return USDT amounts for vault and harvester, then this can become a simple check to see if there anything to send to the harvester, and aways send to the vault.

contracts/contracts/harvest/Harvester.sol
```diff
@@ -0,0 +1,384 @@
...
+        require(
+            tokenConfig.uniswapV2CompatibleAddr != address(0),
+            "Swap token is missing token configuration."
+        );
+
+        address priceProvider = IVault(vaultAddress).priceProvider();
+
+        IERC20 swapToken = IERC20(_swapToken);
+        uint256 balance = swapToken.balanceOf(address(this));
+        if (balance > 0) {
```
DanielVF, line 325: Rather than indenting this entire section, let's just fast return here if balance is zero instead. Will make things clearer.  In case you can't tell I'm a big fan of fast returns over nested ifs. :D

Slither will yell, but it's in the wrong.

contracts/contracts/harvest/Harvester.sol
```diff
@@ -0,0 +1,384 @@
...
+    }
+
+    /**
+     * @dev Swap a record token for stablecoins for Uniswap. The token must have
+     *       a registered price feed with the price provider.
+     * @param _swapToken Address of the token to swap.
+     */
+    function _swap(address _swapToken)
+        internal
+        returns (uint256[] memory swapResult)
```
DanielVF, line 313: I don't think we ever use this return from this internal function. Let's remove it.

contracts/contracts/harvest/Harvester.sol
```diff
@@ -0,0 +1,384 @@
...
+        address[] memory allStrategies = IVault(vaultAddress)
+            .getAllStrategies();
+
+        for (uint256 i = 0; i < swapTokens.length; i++) {
+            _swap(swapTokens[i]);
+        }
+    }
+
+    /**
+     * @dev Swap a record token for stablecoins for Uniswap. The token must have
```
DanielVF, line 307: "swap a **reward** token for stablecoins **on** Uniswap. The token must have"

contracts/contracts/harvest/Harvester.sol
```diff
@@ -0,0 +1,384 @@
...
+        if (rewardTokenAddresses.length > 0) {
+            strategy.collectRewardTokens();
+        }
+    }
+
+    /**
+     * @dev Swap all supported swap tokens for stablecoins via Uniswap.
+     */
+    function _swap() internal {
+        address[] memory allStrategies = IVault(vaultAddress)
```
DanielVF, line 298: I don't think we use this?

sparrowDom: Yup tnx removing

contracts/contracts/harvest/Harvester.sol
sparrowDom: 👍

contracts/contracts/harvest/Harvester.sol
sparrowDom: Yup thanks for pointing it out, I prefer early exits as well.

contracts/contracts/harvest/Harvester.sol
sparrowDom: Thanks this makes so much more sense. I was researching this and I think I read that divisions in solidity always round down. So the 2 numbers should never be above total, but there are as you say still small rounding errors. The suggested fix is so much cleaner.

contracts/contracts/harvest/Harvester.sol
sparrowDom: 👍

contracts/contracts/harvest/Harvester.sol
```diff
@@ -0,0 +1,376 @@
...
+                path,
+                address(this),
+                block.timestamp
+            );
+
+        IERC20 usdt = IERC20(usdtAddress);
+        uint256 usdtBalance = usdt.balanceOf(address(this));
+
+        // When governor calls the swap function send full reward amount to the Vault
+        if (isGovernor()) {
```
sparrowDom, line 358: @DanielVF addressing your comment that was lost by the code refactor: 
```
If the above calculations return USDT amounts for vault and harvester, then this can become a simple check to see if there anything to send to the harvester, and aways send to the vault.

The actual isGoverner logic can live up in the math/logic above, rather than here in the doing.
```

I've refactored this part a bit, since the share of the proceeds between vault & harvester didn't really belong outside of the if clause scope. Does this version seem ok to you or would you still rather see it refactored without the if statement?

I like how it is right now since it seems readable to me.

contracts/contracts/harvest/Harvester.sol
sparrowDom: Just pasting it here for easier reference. This was what Vault used to do before allocating: https://github.com/OriginProtocol/origin-dollar/blob/master/contracts/contracts/vault/VaultCore.sol#L218-L241

And that has been shortened to this: https://github.com/OriginProtocol/origin-dollar/pull/869/files#diff-113977321a6b18ff2702753444615646663eef373deadf81107bf316f1300448R217-R219

So to double confirm we are ok with removing the harvesting bit from `_allocate` call? It makes sense to me, and it should make allocates a lot more gas efficient.

DanielVF: Yes. I am good removing the harvest from the vault.

contracts/contracts/harvest/Harvester.sol
DanielVF: Yes, this works structurally.

contracts/contracts/harvest/Harvester.sol
```diff
@@ -0,0 +1,376 @@
...
+            usdt.safeTransfer(vaultAddress, usdtBalance);
+        } else {
+            uint16 vaultBps = 1e4 - tokenConfig.harvestRewardBps;
```
DanielVF, line 361: It would be interesting to test and see if using uin256 here is actually cheaper on gas.

contracts/contracts/harvest/Harvester.sol
```diff
@@ -0,0 +1,376 @@
...
+     * @dev Swap all supported swap tokens for stablecoins via Uniswap.
+     */
+    function _swap() internal {
+        for (uint256 i = 0; i < swapTokens.length; i++) {
+            _swap(swapTokens[i]);
+        }
+    }
+
+    /**
+     * @dev Swap a record token for stablecoins on Uniswap. The token must have
```
DanielVF, line 304: "record" should be "reward", I think.

contracts/contracts/harvest/Harvester.sol
```diff
@@ -0,0 +1,376 @@
...
+            _harvest(allStrategies[i]);
+        }
+    }
+
+    /**
+     * @dev Collect reward tokens for a specific strategy. Called from the vault.
+     * @param _strategyAddr Address of the strategy to collect rewards from
+     */
+    function harvest(address _strategyAddr) external onlyGovernor nonReentrant {
+        // TODO: add protection so that harvestAndSwap isn't called twice too closely together
```
DanielVF, line 258: I think for simplicity, we can leave this too closely check out of the contract. It's really only useful for callers, and I think there are other ways for them to handle it.

contracts/contracts/harvest/Harvester.sol
```diff
@@ -0,0 +1,376 @@
...
+        if (
+            oldUniswapAddress != address(0) &&
+            oldUniswapAddress != _uniswapV2CompatibleAddr
+        ) {
+            token.safeApprove(oldUniswapAddress, 0);
+        }
+
+        // Give Uniswap infinite approval when needed
+        if (_uniswapV2CompatibleAddr != address(0)) {
+            if (
```
DanielVF, line 180: I think it's more efficient/cleaner to only set allowances if we are changing to a new uniswap address.

2^256 an an allowance is stupidly big number. We shouldn't run it out.

contracts/contracts/harvest/Harvester.sol
```diff
@@ -0,0 +1,376 @@
...
+        external
+        onlyGovernor
+        initializer
+    {
+        _initialize(_vaultAddress, _usdtAddress);
+    }
+
+    function _initialize(address _vaultAddress, address _usdtAddress) internal {
+        vaultAddress = _vaultAddress;
+        usdtAddress = _usdtAddress;
```
DanielVF, line 74: If this was an immutable value set in the constructor, it would save 1600 ish gas per harvest. @tomlinton, do you think it's worth adding a constructor for?

contracts/contracts/interfaces/IStrategy.sol
```diff
@@ -48,16 +48,10 @@ interface IStrategy {
...
+     * @dev The address array of the reward tokens for the Strategy.
      */
-    function rewardTokenAddress() external pure returns (address);
-
-    /**
-     * @dev The threshold (denominated in the reward token) over which the
-     * vault will auto harvest on allocate calls.
-     */
-    function rewardLiquidationThreshold() external pure returns (uint256);
+    function getRewardTokenAddresses() external pure returns (address[] memory);
```
DanielVF, line 56: I don't think is is actually `pure`. It should be `view` since it reads from state.

contracts/contracts/mocks/MockUniswapRouter.sol
```diff
@@ -48,6 +58,14 @@ contract MockUniswapRouter is IUniswapV2Router {
         payable
         returns (uint256 amountOut)
     {
+        bytes memory tok0Bytes = new bytes(20);
+        for (uint256 i = 0; i < 20; i++) {
+            tok0Bytes[i] = params.path[i];
```
DanielVF, line 63: There are way faster / cleaner ways to do this, but they take assembly language. This is good enough for a mock.

contracts/contracts/vault/VaultStorage.sol
```diff
@@ -88,8 +86,8 @@ contract VaultStorage is Initializable, Governable {
     // Address of the contract responsible for post rebase syncs with AMMs
     address private _deprecated_rebaseHooksAddr = address(0);
 
-    // Address of Uniswap
-    address public uniswapAddr = address(0);
+    // Deprecated: Address of Uniswap
+    address public _deprecated_uniswapAddr = address(0);
```
DanielVF, line 90: We can change this to private if there is nothing use it. (We'll probably need to update frank's debugging info script at deploy time)

contracts/contracts/vault/VaultCore.sol
```diff
@@ -212,33 +213,9 @@ contract VaultCore is VaultStorage {
...
-                    uint256 rewardTokenAmount = rewardToken.balanceOf(
-                        allStrategies[i]
-                    );
-                    if (rewardTokenAmount >= liquidationThreshold) {
-                        IVault(address(this)).harvestAndSwap(allStrategies[i]);
-                    }
-                }
-            }
-        }
+        IHarvester(harvesterAddress).harvestAndSwap();
```
DanielVF, line 218: As discussed in Harvester.sol, we can drop the harvestAndSwap, as well as the import above.

contracts/contracts/harvest/Harvester.sol
```diff
@@ -0,0 +1,376 @@
...
+        for (uint256 i = 0; i < allStrategies.length; i++) {
+            _harvest(allStrategies[i]);
+        }
+    }
+
+    /**
+     * @dev Collect reward tokens for a specific strategy. Called from the vault.
+     * @param _strategyAddr Address of the strategy to collect rewards from
+     */
+    function harvest(address _strategyAddr) external onlyGovernor nonReentrant {
```
DanielVF, line 257: We need to make sure that we do only allow the _strategyAddr to be one of our valid strategies.

We do the correctly check internally when doing _swap(address), which protects swap(address), and harvestAndSwap(address), but we also need to check it here on harvest(address). Check all my assumptions here. :)

In general we don't want to allow our contract to make calls to arbitrary external addresses unless we absolutely have to do so.

contracts/contracts/vault/VaultAdmin.sol
```diff
@@ -221,67 +204,12 @@ contract VaultAdmin is VaultStorage {
...
-        }
-        swapTokens.pop();
-
-        if (uniswapAddr != address(0)) {
-            IERC20 token = IERC20(_addr);
-            // Remove Uniswap approval
-            token.safeApprove(uniswapAddr, 0);
+            // TODO: calling `_harvest` is replaced by `harvestAndSwap`. Good for an extra pair
+            // of eyes to check this, though it shouldn't be a problem IMHO [domeng]
+            IHarvester(harvesterAddress).harvestAndSwap(_addr);
```
DanielVF, line 210: @tomlinton, I'm still thinking we don't need this. We can always drive the harvestAndSwap later from the harvester contract, even if the strategy is no longer an active vault strategy. 

If we remove this auto harvest, we gain two things:
- VaultAdmin contract no longer depends on the harvester contract either by code, nor has to know its address.
- If a strategy fails in such a way that harvesting is broken, we can still remove it without first upgrading either the vault or the strategy code.

contracts/utils/addresses.js
```diff
@@ -75,6 +75,8 @@ addresses.mainnet.chainlinkAAVE_USD =
   "0x547a514d5e3769680Ce22B2361c10Ea13619e8a9";
 addresses.mainnet.chainlinkCRV_USD =
   "0xcd627aa160a6fa45eb793d19ef54f5062f20f33f";
+addresses.mainnet.chainlinkCVX_USD =
+  "0xd962fC30A72A84cE50161031391756Bf2876Af5D";
```
DanielVF, line 79: Verified

contracts/contracts/utils/InitializableAbstractStrategy.sol
```diff
@@ -68,28 +77,40 @@ abstract contract InitializableAbstractStrategy is Initializable, Governable {
...
         vaultAddress = _vaultAddress;
-        rewardTokenAddress = _rewardTokenAddress;
+        rewardTokenAddresses = _rewardTokenAddresses;
+
         uint256 assetCount = _assets.length;
         require(assetCount == _pTokens.length, "Invalid input arrays");
         for (uint256 i = 0; i < assetCount; i++) {
             _setPTokenAddress(_assets[i], _pTokens[i]);
         }
+        uint256 rewardsCount = _rewardTokenAddresses.length;
```
DanielVF, line 93: This looks like a local variable. Is it used anywhere?

contracts/contracts/utils/InitializableAbstractStrategy.sol
DanielVF: I still would rather depreceate.

contracts/contracts/strategies/ConvexStrategy.sol
```diff
@@ -18,12 +18,6 @@ contract ConvexStrategy is BaseCurveStrategy {
     using StableMath for uint256;
     using SafeERC20 for IERC20;
 
-    event RewardTokenCollected(
```
DanielVF, line 21: I like that everything uses the same event now.

contracts/contracts/strategies/ConvexStrategy.sol
```diff
@@ -88,10 +80,10 @@ contract ConvexStrategy is BaseCurveStrategy {
         onlyGovernor
```
DanielVF, line 80: We can ditch this method, and the event for it.

contracts/contracts/strategies/ConvexStrategy.sol
```diff
@@ -151,18 +143,29 @@ contract ConvexStrategy is BaseCurveStrategy {
...
-        emit RewardTokenCollected(vaultAddress, rewardTokenAddress, balance);
-        crvToken.safeTransfer(vaultAddress, balance);
-        // Send CVX
-        IERC20 cvxToken = IERC20(cvxRewardTokenAddress);
-        balance = cvxToken.balanceOf(address(this));
-        emit RewardTokenCollected(vaultAddress, cvxRewardTokenAddress, balance);
-        cvxToken.safeTransfer(vaultAddress, balance);
+
+        /*
+         * This code is direct copy from the same name function from InitializableAbstractStrategy
```
DanielVF, line 156: Maybe we could just make the repeated section of code an internal function, and then we can call it from here.

contracts/contracts/vault/VaultAdmin.sol
tomlinton: Yep I'm fine with that

contracts/contracts/harvest/Harvester.sol
```diff
@@ -0,0 +1,376 @@
...
+        uint16 _allowedSlippageBps,
+        uint16 _harvestRewardBps,
+        address _uniswapV2CompatibleAddr,
+        uint256 _liquidationLimit
+    );
+
+    // Tokens that should be swapped for stablecoins
+    address[] public swapTokens;
+
+    // Strategies approved for use by the Vault
```
tomlinton, line 37: Struct isn't about strategies

contracts/contracts/harvest/Harvester.sol
tomlinton: Yea I think it is, it's the most gas efficient approach right? For reference Flipper.sol does this @sparrowDom

contracts/contracts/harvest/Harvester.sol
```diff
@@ -0,0 +1,376 @@
...
+    /*
+     * @dev Collect reward tokens from all strategies and swap for supported
+     *      stablecoin via Uniswap
+     */
+    function harvestAndSwap() external nonReentrant {
+        // TODO: add protection so that harvestAndSwap isn't called twice too closely together
+        _harvest();
+        _swap();
+        if (msg.sender != vaultAddress) {
+            IVault(vaultAddress).rebase();
```
tomlinton, line 238: I may have missed some discussion on this, but do we need this rebase here? They seem to be two distinct actions. If the caller wants both of them they can call each separately, but it doesn't seem like we should mandate a rebase call here.

contracts/contracts/harvest/Harvester.sol
tomlinton: 👍

contracts/contracts/harvest/Harvester.sol
sparrowDom: Thanks fixed

contracts/contracts/harvest/Harvester.sol
sparrowDom: thanks changing to immutable property set inside constructor

contracts/contracts/harvest/Harvester.sol
DanielVF: I agree. I think we should keep the rebases out of the harvester.

contracts/contracts/vault/VaultAdmin.sol
DanielVF: image

contracts/contracts/harvest/Harvester.sol
sparrowDom: cool, agreed 👍

contracts/contracts/harvest/Harvester.sol
sparrowDom: OK removing for now. I guess we can address this issue if we hear a lot of complaints from the ones using it. And not leave out premature optimisation 👍

contracts/contracts/harvest/Harvester.sol
sparrowDom: Felt a bit of cold sweat down mu back, not thinking about strategy address potentially being malicious contract. Will add tests for this

contracts/contracts/harvest/Harvester.sol
sparrowDom: Done

contracts/contracts/interfaces/IStrategy.sol
sparrowDom: Yup true, thanks

contracts/contracts/mocks/MockUniswapRouter.sol
sparrowDom: Yeah it is messy, glued together with duct tape :)

contracts/contracts/strategies/ConvexStrategy.sol
sparrowDom: 😊

contracts/contracts/vault/VaultCore.sol
sparrowDom: Yup removed 👍

contracts/contracts/vault/VaultAdmin.sol
sparrowDom: Just one thing, I will be adding a check in harvest & swap that the `_strategyAddress` parameter is one of the approved strategies. So in theory if `harvestAndSwap` fails and we remove a strategy. Harvester will no longer be able to harvest funds.

contracts/contracts/vault/VaultStorage.sol
sparrowDom: Cool changing to private

contracts/contracts/strategies/ConvexStrategy.sol
sparrowDom: good point thanks, removing

contracts/contracts/utils/InitializableAbstractStrategy.sol
sparrowDom: tnx removing

contracts/contracts/harvest/Harvester.sol
sparrowDom: Haa there is a gas reduction. But a tiny one when calling `harvestAndSwap`. From 464709 to 464702 (7 gas points). Changing to that. 

istSomething

contracts/contracts/harvest/Harvester.sol
sparrowDom: Added the check to `_harvest(address)` function that is called by all the public & governorOnly functions. Will also add tests for this one.

contracts/contracts/vault/VaultAdmin.sol
sparrowDom: Removed the call to harvestAndSwap and added a comment, to be careful and call `harvestAndSwap` manually before removing the strategy. Since it is no longer possible after removal.

contracts/contracts/strategies/ConvexStrategy.sol
sparrowDom: That makes much more sense tnx

contracts/contracts/harvest/Harvester.sol
DanielVF: We don't need to check that it's an active strategy on the vault, just that it's a strategy that has been configured locally and is setup on the harvester. This is cleaner and cheaper, as well as meaning that we can harvest even if the strategy is removed from the vault.

sparrowDom: Currently the harvester doesn't hold any info regarding the strategy addresses. How do you propose this could be achieved? Maybe Harvester having a mapping something along the lines of: 

```
// mapping of all (active or non active) strategies that ever existed
mapping(address => bool) public strategyRegistry;
```
That mapping could be set by the Vault when `approveStrategy` is called - and not removed / unset when `removeStrategy` is called. 

Or you had something else in mind?

contracts/contracts/vault/VaultInitializer.sol
```diff
@@ -33,5 +33,7 @@ contract VaultInitializer is VaultStorage {
         autoAllocateThreshold = 25000e18;
         // Threshold for rebasing
         rebaseThreshold = 1000e18;
+        // Initialize all strategies
+        allStrategies = new address[](0);
```
sparrowDom, line 37: @DanielVF added this to make slither happy (uninitialised `allStrategies` variable). This is probably OK since the vault has already been initialised right?

DanielVF:
#### Requirements

Have users/bots compete to harvest and sell reward tokens. Callers get 1% of harvested USDT. Rest of USDT to goes to vault.

#### Failure Consequences

This contract handles yield from strategies. It does not handle principal, which reduces the size of risk. It should normally be handling around a day's worth of yield which would make it be about 20% / 365 = 0.05% of the vault's TLV at a time.

Going to be a big deployment!

- [x] Analytics
    - [x] No changes to analytics
    - [ ] Analytics update scheduled

#### Internal State

✅ The only stored internal state is config variables.

✅ No internal state is changed during user called actions.

- swapTokens - List of reward tokens
- RewardTokenConfig - per reward token

What can be always said about relationships between stored state:

- ⛔️ There should aways be a RewardTokenConfig if there is a swaptoken.
- ⛔️ Calling harvest(), then swap(), could act differently than calling harvestAndSwap(), and may swap different tokens. This would probably deeply surprise the person who called it. 

I think we could completely remove the swapTokens list? It can get out of sync with the reward config. We don't use it the normal harvestAndSwap() case. It will drop a couple gnarly functions from the code.

There's nothing that swap() does that can't be done by the caller just picking which reward coins to swap, and calling the code. I would be good dropping all of the looping shortcut functions (harvest(), swap(), and harvestAndSwap(), and requiring that the caller specify what gets swapped. If we did keep them, they should all use the same method for choosing what gets called.

- ✅ Only an admin can set uniswapV2CompatibleAddr
- ⛔️ DON'T NEED a rebase in harvest and swap? Did we settle this?
- 🟠 We probably don't need to check if the strat has reward tokens addresses set before harvesting. The extra gas is on the person calling, if they call harvest on a strategy that doesn't need it. If we don't want a strat harvested, we should just remove the config from the harvester.
- ✅ We should always have a valid token config before swapping a coin.

#### Attack

The most likely vuln would be theft of yield from a sandwhich attack.

Mev Attacks?

Is there a way for a contract failure to result in reward tokens or stablecoins being stolen. Where do reward tokens leave the contract?

- ✅ The only place reward contracts can leave the contract is via uniswap or an admin token recovery.
- ✅ Stables only leave to vault address or to the user that called the swap.
- 🟠 Harvest and swap could take a rewardTo address that chooses where to send USDT. This would be a gas savings for arbers who would be turning around and doing other things. We could also get rid of the isgovernor complication, since a governor could just choose to their share to the vault. 
- NOTE: Perhaps we need to only do send once at the end of running all swaps.
- 🟠 maxBalanceToSwap is probably a bad name, since it's not used as a max, but rather it is the actual balanceToSwap.
- TEST: VERIFY MAX SLIPPAGE COMPUTATION, VERIFY unit tests on edge cases.
- TEST vaultShare Calculations
- 🟠If we make it so that liquidationLimit has to be set a high value in order for the swap to work, then we could just do a return from the swap() if it is set to zero. This would remove the logic about sometimes calculating a max value and sometimes not.
- 🟠As we saw with curve, it's possible that we may support one of a strategy reward coin's without supporting another. Perhaps we should have the check on tokenConfig.uniswapV2CompatibleAddr just do a silent return rather than throwing. The current harvester contract would not have worked with our current setup of selling CRV and holding CVX.

#### Logic


- [ ] Verify tests for removeSwapToken and removeSwapToken, that test edge condtions, and removing from middle and start, or maybe we can remove these methods.
- 🟠 Supernit: Require checks in setRewardTokenConfig don't match order of function parameters


#### Tests

#### Flavor

- 🟠Since it will almost never be profitable to be calling this on more than one strategy at a time, we cold remove the global harvest(), swap(), and harvestAndSwap() methods and only keep the strategy and coin specific methods.
- 🟠Need an external method to swap(rewardToken) a single coin, at least from governance.

- Strategy
  - ✅ Could not return us money. Fine
  - ✅ Could crash. Fine
- Uniswap?
  - ✅ Beter behave
- Reward token?
  - ✅ Reentrant on us? We check for it.
  - ✅ Reentrant on other contracts? We don't call other OUSD contracts outside of getting configs from the vault.
  - ✅Crash, fine
- USDT
  - ✅ Beter behave

#### Overflow

- [x] _We are using Sol V8_

#### Black magic

#### Authentication

- [x] Never use tx.origin
- [X] Check that every external/public function has the correct authentication

#### Cryptographic code

_No crypto in this contract._

#### Gas problems

- [X] Contracts with for loops must have either:
    - [x] A way to remove items

#### External calls

- [x] No Unsafe external calls
- [X] Malicious behaviors
- [X] No slippage attacks (we need to validate expected tokens received)
- [x] Oracles, one of:
  - [ ] No oracles
  - [x] Oracles can't be bent onchain


#### Ethereum

contracts/contracts/harvest/Harvester.sol
DanielVF: Yes, exactly this.

joshfraser:
Wow, I love the thoroughness on this PR! Great work @sparrowDom and appreciate your detailed security review @DanielVF.

A couple questions.

I trust that Uniswap contracts will behave, but what happens if liquidity providers pull all of their capital? This could easily happen if another AMM starts offering superior rewards. I assume the Chainlink oracles keep us safe and would just hard-fail until we fix it?

Do we have a method for other smart contracts to be able to tell how many rewards they should expect if they call this? It may end up being completely unnecessary, but I'm still fond of the idea of having self-funded Chainlink Keepers running as a backup. I'm picturing a separate contract that Chainlink could query to check the expected rewards and determine whether it's worthwhile for the Keepers to claim the rewards, swap the funds into LINK, and then deposit the funds back into the keeper account to fund future maintenance.

sparrowDom:
@joshfraser we use chainlink prices as a protective measure of measuring how much reward we should receive when swapping & a slippage threshold - that becomes configurable per reward token with this PR. So a swap fails if minimum amount of expected USDT after swapping is not received.

sparrowDom:
Regarding who is eligible to call `harvest`, `swap` & `harvestAndSwap`. Good point and I agree with removal of `swapTokens`. Now that each reward tokens has configurable swap params like `uniswapAddress` and `slippage` we loose the flexibility of swapping reward tokens that are not part of reward token strategies anyway. The way we could potentially currently swap CVX for example. 

Removing swapTokens unifies & simplifies how `harvest()` and `swap()` loop through tokens. We could leave them as governor only convenience functions -> see this commit: https://github.com/OriginProtocol/origin-dollar/pull/869/commits/cf4f6782c7be971421c39360efe47495f72e1954

I am also fine with removing them in case we are sure they won't come in handy.

DanielVF:
@joshfraser The amount of rewards can be simulated or looked up off chain, by whoever is going to call it. We have to have  onchain contract support for looking it up.

contracts/contracts/harvest/Harvester.sol
sparrowDom: Ran a quick test calling `harvestAndSwap(cvxAddress)` with the existing call to Vault to fetch all strategies and check if it is a valid one and without the Vault call. The gas costs were 569637 & 562516  (7.1k saving without the check). Things affecting decision of having strategies configured locally in Harvester: 
- 🟢 7.1k saved in gas costs when calling `harvestAndSwap(address)`
- 🟢 harvester is able to harvest from strategies that were once and are no longer active on the Vault
- 🔴 Vault needs to be aware of Harvester again to be able to update its state on `approveStrategy` calls 

I would rather keep a cleaner code separation and have Vault not being Harvester aware. Open to being persuaded to the contrary.

sparrowDom:
- ✅  rebase removed from both version of `harvestAndSwap`
- ✅  rebase removed from both version of `harvestAndSwap`
- ✅  swapTokens removed. Handled exclusively by `rewardTokenConfig` and per strategy scoped `rewardTokenAddresses`

DanielVF:
> I assume the Chainlink oracles keep us safe and would just hard-fail until we fix it?

@joshfraser Yes. Fortunately because in the new setup each strategy is harvested separately, only the one strategy would be failing, and the other strategies would continue to collect normaly.

contracts/contracts/harvest/Harvester.sol
```diff
@@ -0,0 +1,371 @@
...
+     * @param _liquidationLimit uint256 Maximum amount of token to be sold per one swap function call.
+     *        When value is 0 there is no limit.
+     */
+    function setRewardTokenConfig(
+        address _tokenAddress,
+        uint16 _allowedSlippageBps,
+        uint16 _harvestRewardBps,
+        address _uniswapV2CompatibleAddr,
+        uint256 _liquidationLimit,
+        bool _doSwapRewardToken
```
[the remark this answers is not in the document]
sparrowDom: tnx fixed

contracts/contracts/harvest/Harvester.sol
DanielVF: That's a great gas savings.

I agree that it is awesome for the vault not to know about the harvester. I think it's okay if the vault doesn't configure the harvester, since that is an operation that can be done externally. 

In general my thinking is that contracts enforce properties and logic that is security critical, but outside that as much thinking that can live outside the chain lives out there. This makes the contracts have much less code in them, thus be simpler, and thus both easier to be correct, secure, and less needing of change. 

The question here is really if the vault and the harvester being out of sync is a security concern.

I don't think it is one if the new strategy is being added. We can just harvest later, nothing is running that should not be. Same non-problem if a strategy is in the vault and removed from harvesting.

The question in my mind is around risks if we remove a strategy from the vault and forget to remove a strategy from the harvester.

The harvester couldn't cause the vault a problem in such a case - it can only send the vault funds. The harvester doesn't call any state updating actions on the vault.

If a strategy has been removed from the vault, the strategy should be empty of principal funds (though it may have reward tokens laying around). We only currently allow one strategy at a time to be harvested, which means that calling harvest on one strategy should not interfere with the harvest on another strategy.

I lean towards being fine with the possible disconnect between the two. This is more a gut feeling, value call. I really like things that don't know about other things, less code, and big gas savings in our core function. I know that smart people might disagree with me. @tomlinton Got any thoughts here?

sparrowDom: I will let you make the call here. My primary instinct was to have better separation of concerns between different contracts, then again I have no objections on the points you are making above. Agree that one way or another there shouldn't be any negative security implications even in cases of disconnects between vault & harvester. 

Will add a mapping of strategyAddress -> bool indicating if a strategy is a valid one, that will be updated by the vault when governor approves the strategy on the vault. Will also let that property be updatable by the governor just in case we would ever need that control. But I agree with your points, that currently I don't see why that would be required.

sparrowDom: done via: https://github.com/OriginProtocol/origin-dollar/pull/869/commits/69f4803c88303edf0073c5789328b1d77db5e188

sparrowDom:
Tests remaining to be added: 
- [] should skip reward swap but still harvest when doSwapRewardToken is set to false
- [] set max slippage to 4% should pass with 3.9% slippage.
- [] set max slippage to 4% should fail with 4.1% slippage.
- [] set reward token share to 1% and test it works
- [] set reward token share to 99% and test it works
- [] set reward token share to 101% should fail (setting it should fail)
- [] set reward token share to -1% should fail (setting it should fail)
- [] try calling `setSupportedStrategy` from random address should fail
- [] try calling `setSupportedStrategy` as governor should succeed
- [] try calling `setSupportedStrategy` as vault should succeed
- [] try calling `harvestAndSwap(address)` with the invalid strategy address
- [] try calling `harvestAndSwap(address, rewardTo)` with the invalid strategy address
- [] test `harvestAndSwap(address)` still works even when the vault removes a certain strategy
- [] test `harvestAndSwap(address)` fails if governor sets a strategy as not supported
- [] test `harvestAndSwap(address, rewardTo)` with a 0 liquidationLimit
- [] test `harvestAndSwap(address, rewardTo)` with a small liquidationLimit
- [] test `harvestAndSwap(address, rewardTo)` where rewardTo address is not the caller

sparrowDom:
added all the missing tests. I know it is super tedious with such a long PR but would really appreciate a second set of eyes checking that I've not overlooked and missed testing something.

//cc: @tomlinton @DanielVF

contracts/test/helpers.js
DanielVF: Why use AAVE_TOKEN vs using AAVE?

sparrowDom: In case of local environment we are marking a distinction between them:
https://github.com/OriginProtocol/origin-dollar/blob/master/contracts/test/helpers.js#L281-L282

sparrowDom:
@tomlinton when you have time could you make another thorough pass please? This still needs you master approval stamp

contracts/contracts/strategies/ConvexStrategy.sol
```diff
@@ -18,20 +18,8 @@ contract ConvexStrategy is BaseCurveStrategy {
...
-    );
-
-    event CvxRewardTokenAddressUpdated(
-        address _oldAddress,
-        address _newAddress
-    );
-
     address internal cvxDepositorAddress;
     address internal cvxRewardStakerAddress;
-    address public cvxRewardTokenAddress;
```
DanielVF, line 34: Let's just mark this as deprecated, rather than dropping this storage slot.

We are setting cvxDepositorPTokenId in the initializer, which is called during the deploy, so the all storage slots we use will be correct, and this will work as is for now. However we leave a the next unused storage slot now containing a value. This has a tiny chance of tripping us up at some future time, if we expected it be just like any any other empty storage slot.

DanielVF (approved):
Other than the last two comments I made, I think this is okay to go.

contracts/contracts/strategies/ConvexStrategy.sol
sparrowDom: That's a good point... fixed. I've seen this when testing mainnet fork without deprecating slots and replacing previously used slot of type `address`, with an `address[]` type. Can't remember if the node hanged or has thrown a weird exception. 

I think slot was initialised with an `address` type and then when replaced with `address[]` it was expecting list length in that slot but got something completely different.

DanielVF: Haha, yeah, getting an address when it was expecting a length could make it thing it was an insanely long list.

contracts/deploy/036_multiple_rewards_strategies.js
```diff
@@ -0,0 +1,334 @@
...
+      cAaveStrategyProxy.address
+    );
+    log(
+      "AaveStrategyProxy proxyAddress:",
+      cAaveStrategyProxy.address,
+      " governor:",
+      await cAaveStrategyProxy.governor()
+    );
+
+    // 3Pool
```
tomlinton, line 111: I think these are unused

tomlinton:
The abi in `brownie/abi` look like they need to be updated.

tomlinton:
Tested this out on a fork, I'm happy with it, nice one @sparrowDom

tomlinton (approved):
Some minor things that don't effect the functionality remaining, but happy with the contract work. Nice one!

contracts/contracts/vault/VaultAdmin.sol
```diff
@@ -176,13 +168,15 @@ contract VaultAdmin is VaultStorage {
     function approveStrategy(address _addr) external onlyGovernor {
         require(!strategies[_addr].isSupported, "Strategy already approved");
         strategies[_addr] = Strategy({ isSupported: true, _deprecated: 0 });
+        IHarvester(harvesterAddress).setSupportedStrategy(_addr, true);
```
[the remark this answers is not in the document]
sparrowDom: Removed it all. We will now need to configure supported strategies on the Harvester. Have added a sanity check where configuring a reward token fails if it does not belong to a supported strategy. Sort of a sanity check so we don't forget the supported strategies configuration. All in this commit: https://github.com/OriginProtocol/origin-dollar/pull/869/commits/3225f2580162df44471b11fe8a7c343b68daeee2

contracts/deploy/036_multiple_rewards_strategies.js
sparrowDom: good point removing 3 pool

sparrowDom:
@DanielVF resolved the remaining comments. @tomlinton thanks good thinking about updating the brownie files, that is also done. 

Nice thanks @tomlinton and @DanielVF for a lot of help with this PR. I think we should be ready to get this deployed now.

DanielVF:
@sparrowDom I don't think we need that last set of changes to harvester. We aren't really using that strategy address, and there might be more than one strategy using a single token. (For example, if we also put funds into a different pool on curve, we might be getting CRV and CVX from two different strategies.

sparrowDom:
@DanielVF tnx good point, removed.

PR #888, opened 2022-01-24
from world import *

DanielVF:
My current brownie tooling.

Provides preconfigured objects for working with our contracts, and the contracts we interact with.


sparrowDom (approved):
LGTM, thanks these will be super useful

PR #900, opened 2022-02-11
make it possible to configure proceeds address

sparrowDom:
make it possible to configure proceeds address

Contract change checklist:
  - [x] Unit tests pass


DanielVF:
Looks good, @sparrowDom. The only change I see that we need is that it does not need a separate deploy. We can just add the setRewardsProceeds to the previous deploy, since it has not run yet.

contracts/contracts/harvest/Harvester.sol
```diff
@@ -391,17 +413,17 @@ contract Harvester is Governable {
...
+        uint256 rewardsProceedsShare = (usdtBalance * vaultBps) / 1e4;
 
         require(
             vaultBps > tokenConfig.harvestRewardBps,
-            "Address receiving harvest incentive is receiving more rewards than the vault"
+            "Address receiving harvest incentive is receiving more rewards than the rewards proceeds address"
         );
 
-        usdt.safeTransfer(vaultAddress, vaultShare);
+        usdt.safeTransfer(rewardProceedsAddress, rewardsProceedsShare);
```
tomlinton, line 423: Not really related to this PR, but we should make it possible to pass in `address(0)` here for _rewardTo and skip the incentive logic and just transfer all the rewards to `rewardProceedsAddress`.

tomlinton (approved):
Looks good to me!

contracts/contracts/harvest/Harvester.sol
tomlinton: Actually I mean if `_rewardTo == rewardProceedsAddress`

PR #903, opened 2022-02-14
Dripper Draft

### LINKED ISSUE #879: Stream rewards from harvests
We should consider [copying Convex](https://docs.convexfinance.com/convexfinance/faq/platform-rewards-current-vs-projected-apy) and stream the yield from harvests over 7 days. Basically, we would total up the reward token sales for the week, then evenly split them over the next week with our daily rebases. This would help smooth out the daily APY and make it harder for people to front-run any big harvests. We could always adjust the time period. but 7 days feels about right as a starting point.

DanielVF:
Dripper takes reward token sales process, and smooths it out into a per block reward stream.

Closes #879

--

Contract change checklist:
  - [ ] Unit tests pass

### DIFF
--- a/contracts/contracts/harvest/Dripper.sol
+++ b/contracts/contracts/harvest/Dripper.sol
@@ -1,0 +1,111 @@
+// SPDX-License-Identifier: agpl-3.0
+pragma solidity ^0.8.0;
+
+import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import { Governable } from "../governance/Governable.sol";
+import { IVault } from "../interfaces/IVault.sol";
+
+/**
+ * @title OUSD Dripper
+ *
+ * The dripper contract smooths out the yield from point-in-time yield events
+ * and spreads the yield out over a configurable time period. This ensures a
+ * continuous per block yield to makes users happy as their next rebase
+ * amount is always moving up. Also, this makes historical day to day yields
+ * smooth, rather than going from a near zero day, to a large APY day, then
+ * back to a near zero day again.
+ *
+ *
+ * Design notes
+ * - USDT has a smaller resolution than the number of seconds
+ * in a week, which can make per block payouts have a rounding error. However
+ * the total effect is not large - cents per day, and this money is
+ * not lost, just distributed in the future. While we could use a higher
+ * decimal precision for the drip perBlock, we chose simpler code.
+ * - By calculting the changing drip rates on collects only, harvests and yield
+ * events don't have to call anything on this contract or pay any extra gas.
+ * Collect() is already be paying for a single write, since it has to reset
+ * the lastCollect time.
+ * - By having a collectAndRebase method, and having our external systems call
+ * that, the OUSD vault does not need any changes, or even to know the address
+ * of the dripper.
+ * - A rejected design was to retro-calculate the drip rate on each collect,
+ * based on the balance at the time of the collect. While this would have
+ * required less state, and would also have made the contract respond more quickly
+ * to new income, it would break the predictability that is this contract's entire
+ * purpose. If we did this, the amount of fundsAvailable() would make sharp increases
+ * when funds were deposited.
+ *
+ */
+
+contract Dripper is Governable {
+    using SafeERC20 for IERC20;
+
+    struct Drip {
+        uint64 lastCollect; // overflows 262 billion years after the sun dies
+        uint192 perBlock; // drip rate per block
+    }
+
+    address immutable vault; // OUSD vault
+    address immutable token; // token to drip out
+    uint256 public dripDuration; // in seconds
+    Drip public drip; // active drip parameters
+
+    constructor(address _vault, address _token) {
+        vault = _vault;
+        token = _token;
+    }
+
+    function availableFunds() external view returns (uint256) {
+        uint256 balance = IERC20(token).balanceOf(address(this));
+        return _availableFunds(balance, drip);
+    }
+
+    function collect() external {
+        _collect();
+    }
+
+    function collectAndRebase() external {
+        _collect();
+        IVault(vault).rebase();
+    }
+
+    function setDripDuration(uint256 _durationSeconds) external onlyGovernor {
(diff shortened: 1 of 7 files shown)


contracts/contracts/harvest/Dripper.sol
```diff
@@ -0,0 +1,111 @@
...
+ * purpose. If we did this, the amount of fundsAvailable() would make sharp increases
+ * when funds were deposited.
+ *
+ */
+
+contract Dripper is Governable {
+    using SafeERC20 for IERC20;
+
+    struct Drip {
+        uint64 lastCollect; // overflows 262 billion years after the sun dies
```
sparrowDom, line 46: AKA W262BK :)

DanielVF:
Self review:

#### Requirements

The dripper contract smooths out the reward from yield tokens into OUSD. It is built as a seperate contract so that the vault code remains untouched and uncomplicatd.

The harvester contract forwards all yield to the dripper, and the dripper sends yield to the vault.


These changes will need to be made after the deploy is done. They are not critical, but are needed to see the full benefits.

- ❗️We should move the parked CVX to the harvester contract
- ❗️We will need add support for checking the dripper balance to analytics.
- ❗️The keeper contract will need to change to call dripper.collectAndRebase() instead of vault.rebase()
- ❗️The OUSD dapp will need to change to call dripper.collectAndRebase() instead of vault.rebase()

#### Internal State

Three internal state variables:

1. lastCollect. Only updated on collects. No problems if zero.
2. dripRatePer block. Only updated on collects. No problems if zero.
3. dripDuration. Only set by governor. Only used in collects. Would crash collect if not set, but will be set in deploy.

- [x] lastCollect must never be in the future

#### Funds flow

Other contracts transfer ERC20's to this contract. This contract does not pull funds

ERC20's only leave to the vault or to the governor.

#### Consequences

If there were bugs in the code:
- approximatly a week's worth of yield could be lost. 

or

- the yield to be slowly dripped out could suddenly go to OUSD

or

- yield could stop flowing to OUSD and be backed up here.

#### Attack

Calling collect more often has the effect of slightly reducing the drip rate vs calling it daily. This is because it moves the drip rate to a point on a proportional curve. No funds are lost, just defered. There is no profit to an attacker.

#### Logic

The only conditional code is a ternary operator to choose the lessor of the balance or the accrued drippage to send to the vault. This is unit tested.


#### Tests

#### Flavor

_Lol, can't really judge my own code. :) I've squezed it down as far as I see so far._


#### Overflow

- [x] Uses v8 or, Never use "+" or "-", always use safe math
- _no loops_

#### Black magic

#### Authentication

- [x] Never use tx.origin

#### Cryptographic code

_No crypto._

#### Gas problems

_No loops_

#### External calls

- _only calls trusted contracts_
- [x] Oracles, one of:
  - [x] No oracles
  - [ ] Oracles can't be bent


#### Ethereum

sparrowDom:
The OUSD dapp doesn't call rebase anymore so we can strike this on off the list: 
`The OUSD dapp will need to change to call dripper.collectAndRebase() instead of vault.rebase()`

contracts/contracts/harvest/Dripper.sol
```diff
@@ -0,0 +1,116 @@
...
+        view
+        returns (uint256)
+    {
+        uint256 elapsed = block.timestamp - _drip.lastCollect;
+        uint256 allowed = (elapsed * _drip.perBlock);
+        return (allowed > _balance) ? _balance : allowed;
+    }
+
+    function _collect() internal {
+        // Calculate send
```
sparrowDom, line 103: I am slightly worried an attacker (or competing product) can spend relatively small amount of gas (by frequently calling collect) to artificially keep our APY low. Stalling the funds in the dripper.

Seems like we can defend agains that by adding a check where collect fails if there is not a minimum difference between the last collect and the current one.

sparrowDom (approved):
Very lean and clean code. Aside from 1 inline comment LGTM

tomlinton:
No NatSpec commenting on the dripper functions. It's super simple and clear to understand without them but I think this is something that'll definitely end up in an audit report.

contracts/test/strategies/dripper.js
```diff
@@ -0,0 +1,128 @@
+const { expect } = require("chai");
+const { defaultFixture } = require("../_fixture");
+
+const { loadFixture, usdtUnits, advanceTime } = require("../helpers");
+
+describe.only("Dripper", async () => {
```
tomlinton, line 6: .only

DanielVF: Thanks!

DanielVF: One day we need a GitHub action for that. :D

sparrowDom: Yeah, and also one that checks that no deploy after nr.6 has a `forceDeploy` set to `true`

DanielVF:
> No NatSpec commenting on the dripper functions. It's super simple and clear to understand without them but I think this is something that'll definitely end up in an audit report.

Added NatSpec comments.

contracts/contracts/proxies/Proxies.sol
```diff
@@ -51,3 +51,10 @@ contract ConvexStrategyProxy is InitializeGovernedUpgradeabilityProxy {
 contract HarvesterProxy is InitializeGovernedUpgradeabilityProxy {
 
 }
+
+/**
+ * @notice HarvesterProxy delegates calls to a Harvester implementation
```
franckc, line 56: nit: update comment. DripperProxy

contracts/deploy/037_dripper.js
```diff
@@ -0,0 +1,125 @@
+const { deploymentWithProposal, log } = require("../utils/deploy");
+const { MAX_UINT256 } = require("../utils/constants");
```
franckc, line 2: unused constant. linter should have caught this?

franckc (approved):
Every drop of code from this dripper LGTM! :)

Nice work ser.

contracts/contracts/harvest/Dripper.sol
sparrowDom: Nice thanks a lot for supporting this with the charts! I was expecting a greater difference with calling it more frequently but like you point out (with the polygon analogy) it comes down more to approximation than to fundamentally changing the behaviour.

PR #910, opened 2022-02-18
make it possible to upload the dapp to the IPFS and publish to IPNS

sparrowDom:
- make it possible to deploy the dapp to IPFS


dapp/src/utils/image.js
```diff
@@ -0,0 +1,5 @@
+export function adjustSrcOption(src) {
```
franckc, line 1: nit: naming is hard. but wonder if naming this `rootPath` or `basePath` would represent better what this method does.

franckc (approved):
One nit inline otherwise LGTM

dapp/src/utils/image.js
sparrowDom: So there is `Find in Files (and replace)` option in Sublime :)

Thanks renamed it to `assetRootPath`

PR #915, opened 2022-02-23
fix stake ui contracts

sparrowDom:
Stake UI contracts were missing provider setup when signatures were required


franckc (approved):
LGTM

Issue #906, opened 2022-02-15
Redeem gas estimate failing


sparrowDom:
When redeeming a mix of coins (using the Origin Vault) a large redeem is failing at the ether block: 14212248

PR: https://github.com/OriginProtocol/origin-dollar/pull/924/files

DanielVF:
Was this because there was not enough funds in the default strategies?

sparrowDom:
done via: https://github.com/OriginProtocol/origin-dollar/pull/924

PR #924, opened 2022-03-02
report liquidity error when not enough funds in default strategies

sparrowDom:
report liquidity error when not enough funds in default strategies. This is how the error looks like (next to `Origin Vault`) in the UI

Screenshot 2022-03-02 at 15 00 14

**connected issue:**https://github.com/OriginProtocol/origin-dollar/issues/906


sparrowDom:
Thanks both concerns are valid points. Interestingly when I was measuring with & without the check when doing a redeem on a forked node there was a `40399` gas difference. Not 120k in my case but still substantial increase. 

I see 2 possible solutions: 
- move this calculation into a helper contract and expose it as a non payable function
- add try/catch block around the [withdrawal code](https://github.com/OriginProtocol/origin-dollar/blob/master/contracts/contracts/vault/VaultCore.sol#L156-L157) and assume (sometimes oncorrectly?) if it fails on that line it is a liquidity error

None of the solutions are ideal.. How do you feel about them @DanielVF ?

sparrowDom:
Thanks for the feedback @DanielVF . Have updated the solution where all the error handling is done on the dapp side.

shahthepro (approved):

PR #925, opened 2022-03-02
clarity around approval step

pedrotmr:


micahalcorn:
In the following case, we should not use the phrase "Origin Dollar". That part of the text should indicate which contract is being approved.

Screen Shot 2022-03-02 at 11 24 49 PM

For example:

- Allow the Flipper to move your {DAI/OUSD/USDC/USDT}
- Allow the Origin Vault to move your {DAI/OUSD/USDC/USDT}
- Allow Curve to move your {DAI/OUSD/USDC/USDT}
- Allow Uniswap to move your {DAI/OUSD/USDC/USDT}
- Allow SushiSwap to move your {DAI/OUSD/USDC/USDT}

Note that I don't think we need to specify V2 or V3 on the button for Uniswap and I would recommend preceding the first two with "the".

For mobile, I suggest that we use far fewer words rather than wrapping button text:

- Approve the Flipper
- Approve the Origin Vault
- Approve Curve
- Approve Uniswap
- Approve SushiSwap

I'm also not a fan of putting help icons/tooltips on buttons the way Uniswap does. I recommend leaving those off and just waiting to see if users are confused. Even then, I would prefer some helper text below the buttons rather than stacked click handlers. Since we're not including the help icon, I also don't think we need the token icon to the left of the text. It was supposed to be left-aligned, but I think we should comment it out for now. We can bring it back for symmetry if we decide to add the help icon on the right of the button.

pedrotmr:
just pushed the most recent and required changes

dapp/src/components/buySell/ApproveButtonLogic.js
```diff
@@ -11,6 +11,14 @@ import ContractStore from 'stores/ContractStore'
 import analytics from 'utils/analytics'
 import { assetRootPath } from 'utils/image'
 import { useWeb3React } from '@web3-react/core'
+import Media from 'react-media'
+
+// const GLOBAL_MEDIA_QUERIES = {
```
sparrowDom, line 16: This comments can probably be cleaned up

dapp/src/components/buySell/ApproveButtonLogic.js
```diff
@@ -11,6 +11,14 @@ import ContractStore from 'stores/ContractStore'
 import analytics from 'utils/analytics'
 import { assetRootPath } from 'utils/image'
 import { useWeb3React } from '@web3-react/core'
+import Media from 'react-media'
```
sparrowDom, line 14: Can you use `withIsMobile` instead of `react-media` in order to be consistent. See example here: https://github.com/OriginProtocol/origin-dollar/blob/7074ecfac35b5f362c75afc35fffcce8e15aa63b/dapp/src/components/buySell/SwapHomepage.js#L26

dapp/src/components/buySell/ApproveButtonLogic.js
```diff
@@ -112,6 +120,21 @@ const ApproveButtonLogic = forwardRef(
       },
     }))
 
+    const contractName = (contract) => {
+      if (contract === 'flipper') return 'the Flipper'
```
sparrowDom, line 124: Can you wrap the strings in `fbt` so they are translatable? See example here: https://github.com/OriginProtocol/origin-dollar/blob/7074ecfac35b5f362c75afc35fffcce8e15aa63b/dapp/src/components/buySell/SwapHomepage.js#L285-L287

dapp/src/components/buySell/ApproveButtonLogic.js
```diff
@@ -0,0 +1,188 @@
...
+//   small: "(max-width: 599px)",
+//   medium: "(min-width: 600px) and (max-width: 1199px)",
+//   large: "(min-width: 1200px)"
+// };
+// const matches = useMedia({ queries: GLOBAL_MEDIA_QUERIES });
+
+const ApproveButtonLogic = forwardRef(
+  (
+    {
+      formHasErrors,
```
sparrowDom, line 26: Any specific reason to using `forwardRef` instead of just passing these values as Component properties?

dapp/src/components/buySell/ApproveButtonLogic.js
```diff
@@ -0,0 +1,188 @@
...
+      uniV2Router,
+      sushiRouter,
+      curveOUSDMetaPool,
+      usdt,
+      dai,
+      usdc,
+      ousd,
+    } = useStoreState(ContractStore, (s) => s.contracts || {})
+
+    const contractMap = {
```
sparrowDom, line 58: Names of the contract variables can be changed already deconstructing them from `ContractStore` like so:

```
const {
      vault,
      flipper,
      uniV3SwapRouter: uniswap,
      uniV2Router: uniswapV2,
      sushiRouter: sushiswap,
      curveOUSDMetaPool: curve,
      usdt,
      dai,
      usdc,
      ousd,
    } = useStoreState(ContractStore, (s) => s.contracts || {})
```

without the need for a `contractMap`

dapp/src/components/buySell/SwapHomepage.js
```diff
@@ -1,41 +1,31 @@
...
+import ErrorModal from 'components/buySell/ErrorModal'
+import SettingsDropdown from 'components/buySell/SettingsDropdown'
+import SwapCurrencyPill from 'components/buySell/SwapCurrencyPill'
+import PillArrow from 'components/buySell/_PillArrow'
 import { currencies } from 'constants/Contract'
-import { providersNotAutoDetectingOUSD, providerName } from 'utils/web3'
+import { ethers } from 'ethers'
+import { fbt } from 'fbt-runtime'
+import withIsMobile from 'hoc/withIsMobile'
 import withRpcProvider from 'hoc/withRpcProvider'
```
sparrowDom, line 14: I think @franckc pointed this out some time ago. Can you undo all the changes that are not necessary for this PR? Solely for the reason that it is hard to look at the `diff` of the PR and understand what changed - if most of the imports are re-arranged.

sparrowDom (commented):
@pedrotmr thanks for all the work on this PR. Have taken a look and left a few comments.

dapp/src/components/buySell/ApproveButtonLogic.js
```diff
@@ -0,0 +1,188 @@
...
+      } else if (coin === 'usdt') {
+        setContract(usdt)
+      } else if (coin === 'usdc') {
+        setContract(usdc)
+      } else if (coin === 'ousd') {
+        setContract(ousd)
+      }
+    }, [])
+
+    useImperativeHandle(ref, () => ({
```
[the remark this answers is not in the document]
pedrotmr: that's how I would the approve function on the parent component

dapp/src/components/buySell/ApproveButtonLogic.js
pedrotmr: I tried with that, but it was not working... Something like {withIsMobile ? '' : ''}

dapp/src/components/buySell/ApproveButtonLogic.js
pedrotmr: I used the same way it was used before on the approve modal for a matter of consistence

pedrotmr:
requests updated and comments answered

micahalcorn:
I got into a bad state where the "Allow" button is disabled shows "undefined" when it should be hidden.

**Reproduction steps**:
- Override the default selection (A) by choosing one that has already been approved (B)
- Switch back to the default selection (A)
- Click to generate the MetaMask prompt for the approval
- Reject the transaction
- Note that the button slightly loses some opacity (and shouldn't)
- Switch to the option that does not require approval (B)
- Note that the button does not get hidden and the contract name is "undefined"

https://www.loom.com/share/0828b68af9a34d5b8e56f7ca52be81bb

micahalcorn:
> For example:
> 
> * Allow the Flipper to move your {DAI/OUSD/USDC/USDT}
> * Allow the Origin Vault to move your {DAI/OUSD/USDC/USDT}
> * Allow Curve to move your {DAI/OUSD/USDC/USDT}
> * Allow Uniswap to move your {DAI/OUSD/USDC/USDT}
> * Allow SushiSwap to move your {DAI/OUSD/USDC/USDT}
> 
> Note that I don't think we need to specify V2 or V3 on the button for Uniswap and I would recommend preceding the first two with "the".
> 
> For mobile, I suggest that we use far fewer words rather than wrapping button text:
> 
> * Approve the Flipper
> * Approve the Origin Vault
> * Approve Curve
> * Approve Uniswap
> * Approve SushiSwap

Note that I was suggesting the word "move" rather than "use" on desktop and the word "Approve" rather than "Allow" on mobile.

micahalcorn:
Here's another bad state:

**Setup**
- Uniswap does not require approval (A)
- Flipper requires approval (B)
- Curve requires approval (C)

**Steps**
- See A as the default option
- Override the default by selecting B
- See "Approve" button enabled and "Swap" button disabled
- Click to approve B
- Wait for the transaction to complete
- See "Swap" button enabled
- Change the selection to C
- See both buttons disabled (different opacity) ❌ - can't move forward with option C

dapp/src/components/buySell/ApproveButtonLogic.js
sparrowDom: Maybe this example is better: https://github.com/OriginProtocol/origin-dollar/blob/master/dapp/src/components/buySell/BalanceHeader.js#L24

when you wrap a component in `withIsMobile` it is going to have a `isMobile` property that will communicate if page should render in mobile mode

dapp/src/components/buySell/ApproveButtonLogic.js
sparrowDom: We use an approach to declare functions in the scope of the parent and pass them down to a component like so: https://github.com/OriginProtocol/origin-dollar/blob/master/dapp/src/components/buySell/SwapHomepage.js#L525-L535

The above example functions are all declared inline with the child component, you can also normally declare them on the parent and pass as a reference to child component

dapp/src/components/buySell/ApproveButtonLogic.js
sparrowDom: I see tnx, seems we could improve code there :)

dapp/src/components/buySell/ApproveButtonLogic.js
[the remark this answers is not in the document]
sparrowDom: There is another way to handle this (as described in above comment). Do tell if you will have challenges achieving this, happy to help you

PR #927, opened 2022-03-02
ssfanist/Uniswap V3 support for reward token swap

defigods:

Contract change checklist:
  - [ ] Unit tests pass

### DIFF
--- a/contracts/contracts/harvest/Harvester.sol
+++ b/contracts/contracts/harvest/Harvester.sol
@@ -227,17 +227,25 @@ contract Harvester is Governable {
     /**
      * @dev Swap all supported swap tokens for stablecoins via Uniswap.
      */
-    function swap() external onlyGovernor nonReentrant {
-        _swap(rewardProceedsAddress);
+    function swap(bool _useV3, bytes memory _data)
+        external
+        onlyGovernor
+        nonReentrant
+    {
+        _swap(rewardProceedsAddress, _useV3, _data);
     }
 
     /*
      * @dev Collect reward tokens from all strategies and swap for supported
      *      stablecoin via Uniswap
      */
-    function harvestAndSwap() external onlyGovernor nonReentrant {
+    function harvestAndSwap(bool _useV3, bytes memory _data)
+        external
+        onlyGovernor
+        nonReentrant
+    {
         _harvest();
-        _swap(rewardProceedsAddress);
+        _swap(rewardProceedsAddress, _useV3, _data);
     }
 
     /**
@@ -254,9 +262,13 @@ contract Harvester is Governable {
      *      the caller are sent to the caller of this function.
      * @param _strategyAddr Address of the strategy to collect rewards from
      */
-    function harvestAndSwap(address _strategyAddr) external nonReentrant {
+    function harvestAndSwap(
+        address _strategyAddr,
+        bool _useV3,
+        bytes memory _data
+    ) external nonReentrant {
         // Remember _harvest function checks for the validity of _strategyAddr
-        _harvestAndSwap(_strategyAddr, msg.sender);
+        _harvestAndSwap(_strategyAddr, msg.sender, _useV3, _data);
     }
 
     /**
@@ -266,12 +278,14 @@ contract Harvester is Governable {
      * @param _rewardTo Address where to send a share of harvest rewards to as an incentive
      *      for executing this function
      */
-    function harvestAndSwap(address _strategyAddr, address _rewardTo)
-        external
-        nonReentrant
-    {
+    function harvestAndSwap(
+        address _strategyAddr,
+        address _rewardTo,
+        bool _useV3,
+        bytes memory _data
+    ) external nonReentrant {
         // Remember _harvest function checks for the validity of _strategyAddr
-        _harvestAndSwap(_strategyAddr, _rewardTo);
+        _harvestAndSwap(_strategyAddr, _rewardTo, _useV3, _data);
     }
 
     /**
@@ -279,12 +293,12 @@ contract Harvester is Governable {
      *       rewards to the vault.
      * @param _swapToken Address of the token to swap.
      */
-    function swapRewardToken(address _swapToken)
-        external
-        onlyGovernor
-        nonReentrant
-    {
-        _swap(_swapToken, rewardProceedsAddress);
+    function swapRewardToken(
+        address _swapToken,
+        bool _useV3,
+        bytes memory _data
+    ) external onlyGovernor nonReentrant {
+        _swap(_swapToken, rewardProceedsAddress, _useV3, _data);
     }
 
     /**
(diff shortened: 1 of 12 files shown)


DanielVF:
The biggest thing I see that the choice to use the V3 or V2 interface needs to be a part of the configuration for the token. It should not be passed in from outside. It's okay then if there is only one uniswap address defined for a reward token, it's just how that address is called that will be changing.

contracts/contracts/harvest/Harvester.sol
```diff
@@ -426,4 +432,62 @@ contract Harvester is Governable {
...
+        uint256 _balanceToSwap,
+        uint256 _minExpected
+    ) internal {
+        UniswapV3Router.ExactInputParams memory params = UniswapV3Router
+            .ExactInputParams({
+                path: abi.encodePacked(
+                    _swapToken,
+                    _fee, // Pool fee, swap token -> weth9
+                    wethAddress,
+                    uint24(3000), // Pool fee, weth9 -> usdt
```
DanielVF, line 481: Actually, lets change this to the 0.50% pool, since that new has funds in it.

contracts/contracts/harvest/Harvester.sol
```diff
@@ -355,12 +373,6 @@ contract Harvester is Governable {
         }
     }
 
-    /**
-     * @dev Swap a reward token for stablecoins on Uniswap. The token must have
-     *       a registered price feed with the price provider.
-     * @param _swapToken Address of the token to swap.
-     * @param _rewardTo Address where to send the share of harvest rewards to
-     */
```
tomlinton, line 363: We still need this comment. We can just update it to explain the v3 vs v2 functionality.

ssfanist: I just moved comments to below _swapV2 and _swapV3 functions.
But if you want, I'll put back to _swap function.
So should I remove the comments for _swapV2 and _swapV3 functions? or keep all 3 parts?

tomlinton: All 3 should have comments. They all do something different so they should be updated accordingly.

contracts/contracts/harvest/Harvester.sol
```diff
@@ -112,20 +130,22 @@ contract Harvester is Governable {
      *        Example: 300 == 3% slippage
      * @param _harvestRewardBps uint16 amount of reward tokens the caller of the function is rewarded.
      *        Example: 100 == 1%
-     * @param _uniswapV2CompatibleAddr Address Address of a UniswapV2 compatible contract to perform
```
sparrowDom, line 115: nice catch thanks :)

contracts/contracts/harvest/Harvester.sol
```diff
@@ -137,15 +157,16 @@ contract Harvester is Governable {
         );
         require(
             _uniswapV2CompatibleAddr != address(0),
```
sparrowDom, line 159: I would allow `_uniswapV2CompatibleAddr` to be zero address when the reward token is configured to be swapped using Uniswap V3.

A check that would make sense to me is `_uniswapV2CompatibleAddr ` != address(0) || _v3Fee > 0
meaning either UniswapV2 or Uniswap V3 should be configured. Maybe another check that both of them are not configured.

contracts/contracts/harvest/Harvester.sol
```diff
@@ -171,19 +192,16 @@ contract Harvester is Governable {
             token.safeApprove(oldUniswapAddress, 0);
         }
 
-        // Give Uniswap infinite approval when needed
```
sparrowDom, line 174: Is this ok? I case token still wants to use UniswapV2 this is not going to work? I would: 
- token.safeApprove(0) && token.safeApprove(MAX)(as it was before this PR)  when the token configuration passed to the harvester is using non 0 `_uniswapV2CompatibleAddr `
- token.safeApprove(0) && token.safeApprove(MAX) when uniswap v3 is configured to be used.

contracts/deploy/037_dripper.js
```diff
@@ -108,14 +108,15 @@ module.exports = deploymentWithProposal(
           contract: cHarvester,
           // tokenAddress, allowedSlippageBps, harvestRewardBps, uniswapV2CompatibleAddr, liquidationLimit, doSwapRewardToken
           signature:
-            "setRewardTokenConfig(address,uint16,uint16,address,uint256,bool)",
+            "setRewardTokenConfig(address,uint16,uint16,address,uint256,bool,uint24)",
```
sparrowDom, line 111: These deploys have already been ran and don't need this change

contracts/deploy/036_multiple_rewards_strategies.js
```diff
@@ -251,59 +253,63 @@ module.exports = deploymentWithProposal(
           contract: cHarvester,
           // tokenAddress, allowedSlippageBps, harvestRewardBps, uniswapV2CompatibleAddr, liquidationLimit, doSwapRewardToken
           signature:
-            "setRewardTokenConfig(address,uint16,uint16,address,uint256,bool)",
+            "setRewardTokenConfig(address,uint16,uint16,address,uint256,bool,uint24)",
```
sparrowDom, line 256: as in the 037 migration file, I think this file doesn't need any changes, since it has already been executed on the mainnet

sparrowDom: You will probably need to create a fresh new deploy file (038...) in order to be able to upgrade the harvester.

sparrowDom (commented):
Left a couple of comments inline

contracts/contracts/harvest/Harvester.sol
```diff
@@ -136,16 +156,17 @@ contract Harvester is Governable {
             "Harvest reward fee should not be over 10%"
         );
         require(
-            _uniswapV2CompatibleAddr != address(0),
-            "Uniswap compatible address should be non zero address"
+            (_uniswapV2CompatibleAddr != address(0)) == (_v3Fee == 0),
```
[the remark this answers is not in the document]
ssfanist: 💪

ssfanist:
@DanielVF @tomlinton @sparrowDom @franckc Any more things to update here?

contracts/contracts/harvest/Harvester.sol
```diff
@@ -171,6 +192,8 @@ contract Harvester is Governable {
             token.safeApprove(oldUniswapAddress, 0);
         }
 
+        token.safeApprove(uniswapV3Addr, type(uint256).max);
```
sparrowDom, line 195: If we want to be super clean we should call approve only when current _v3Fee == 0 (or not set) and the current one is not 0.

sparrowDom:
Left 1 minor comment inline, and what is also required is a new completely separate `038` deploy script. All the other deploy scripts have already been executed on the mainnet. And each one executes only once.

Will let @DanielVF stamp the final seal of approval since he is sort of the code owner of this part.

ssfanist:
I'm not sure what to do in the deploy script 38. Would you share with me? @tomlinton

sparrowDom:
I would just copy all the things from script 036 and keep the part that upgrades the Harvester to the new version of the contract

contracts/contracts/harvest/Harvester.sol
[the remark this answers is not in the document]
DanielVF: I am impressed. Worked it out on paper to verify.

Basically this is doing an XOR so that one and only one of the two conditions is true. 

It's a little confusing to understand because one side is true when set, and the other side is false when set. I think it would be clearer if this used the normal XOR substitute "!=" for the middle comparison operator, and then did `(_v3Fee > 0)`. This way both sides of the check are the same kind, with true for this kind is set, and false for this kind is not set.

Thanks, I learned something today.

contracts/contracts/harvest/Harvester.sol
DanielVF: Echoing what Domen said, this whole token approve section should be two independent parts. One that handles uniswap v2 and one that handles uniswap v3. Each can just separately see if it needs to operate, and does not need to know about the other.

Currently this is also missing the second to unapprove uniswapV3.

contracts/test/vault/harvester.js
```diff
@@ -34,16 +34,15 @@ describe("Harvester", function () {
     );
     const mockUniswapRouter = await ethers.getContract("MockUniswapRouter");
```
DanielVF, line 36: The tests don't seem to test the uniswap v3 swapping? Or did I miss it?

DanielVF (changes requested):

PR #928, opened 2022-03-03
Increase code coverage

DanielVF:
Add two unit tests, update code coverage exclusion list to be more accurate.

image

Note that the big red bars on the oracle tests are because it's not testing our production only oracle address list.


franckc (approved):
Nice! LGTM

PR #929, opened 2022-03-04
Edited README to handle runtime error to run dApp setup

jkaihsu:
Ran into this error while setting up my dev. Renaming `dApp/dev.env` to `dApp/.env` fixed the issue

Screen Shot 2022-03-04 at 12 55 35 PM


franckc (approved):
Thanks for adding this to the doc. LGTM

PR #946, opened 2022-03-11
Fix APY discrepancy

rolandpo:
I submitted a fix in utils/contracts.js to fix the discrepancy in APY between the Dapp and the Analytics page. The Dapp was fetching the 30-day trailing average APR and then converting to APY using a 7 day compounding period instead of 30. Recalculating correctly on the Dapp side can introduce some rounding errors so I fetched the already calculated APY from the Analytics side.

I was wondering whether the choice of a 30 day compounding period in calculations was due to the infrequent and unpredictable harvests. Now that this has been smoothed out a bit, would it be a good idea to decrease it?


sparrowDom (approved):
Nice thanks works as expected

PR #954, opened 2022-03-22
[wip] refactor: use react-query for tx history fetching

evgenyboxer:
This is an attempt to refactor the history fetching page by utilising react-query (https://react-query.tanstack.com/)
It's a really powerful library that takes care of fetching/catching/automatic retries/etc... so really great for anything async and comes with its own state management.

Changes in this PR:
1. Introduced a new "query" (a hook) that uses `react-query` under the hood to grab history.
2. Introduced `useMemo` for "computed" props instead of using `useEffect` + `useState`
3. Initiated a little history service that does the fetching (will be handy in the future so we can easily test it).

Now when you go back and forth to the history page it will instantly appear.

Also - it comes with really cool devtools.

Screen Shot 2022-03-23 at 10 48 40 am

Note - I might not be aware of all the edge cases, so this is why its still WIP and needs to be tested.

### DIFF
--- a/dapp/pages/_app.js
+++ b/dapp/pages/_app.js
@@ -4,6 +4,7 @@ import { useWeb3React } from '@web3-react/core'
 import { useRouter } from 'next/router'
 import { useCookies } from 'react-cookie'
 import { useStoreState } from 'pullstate'
+import { QueryClient, QueryClientProvider, useQuery } from 'react-query'
 
 import AccountStore from 'stores/AccountStore'
 import RouterStore from 'stores/RouterStore'
@@ -36,6 +37,8 @@ if (process.browser) {
 }
 
 initSentry()
+
+const queryClient = new QueryClient()
 
 function App({ Component, pageProps, err }) {
   const [locale, setLocale] = useState('en_US')
@@ -161,28 +164,30 @@ function App({ Component, pageProps, err }) {
 
   return (
     <>
-      <AnalyticsProvider instance={analytics}>
-        <AccountListener />
-        <TransactionListener />
-        <UserActivityListener />
-        <WalletSelectModal />
-        <ToastContainer
-          position="bottom-right"
-          autoClose={5000}
-          hideProgressBar={false}
-          newestOnTop={false}
-          closeOnClick
-          rtl={false}
-          pauseOnFocusLoss
-          pauseOnHover
-        />
-        <Component
-          locale={locale}
-          onLocale={onLocale}
-          {...pageProps}
-          err={err}
-        />
-      </AnalyticsProvider>
+      <QueryClientProvider client={queryClient}>
+        <AnalyticsProvider instance={analytics}>
+          <AccountListener />
+          <TransactionListener />
+          <UserActivityListener />
+          <WalletSelectModal />
+          <ToastContainer
+            position="bottom-right"
+            autoClose={5000}
+            hideProgressBar={false}
+            newestOnTop={false}
+            closeOnClick
+            rtl={false}
+            pauseOnFocusLoss
+            pauseOnHover
+          />
+          <Component
+            locale={locale}
+            onLocale={onLocale}
+            {...pageProps}
+            err={err}
+          />
+        </AnalyticsProvider>
+      </QueryClientProvider>
     </>
   )
 }
--- a/dapp/pages/_app.js
+++ b/dapp/pages/_app.js
@@ -4,7 +4,8 @@ import { useWeb3React } from '@web3-react/core'
 import { useRouter } from 'next/router'
 import { useCookies } from 'react-cookie'
 import { useStoreState } from 'pullstate'
-import { QueryClient, QueryClientProvider, useQuery } from 'react-query'
+import { QueryClient, QueryClientProvider } from 'react-query'
+import { ReactQueryDevtools } from 'react-query/devtools'
 
 import AccountStore from 'stores/AccountStore'
 import RouterStore from 'stores/RouterStore'
@@ -187,6 +188,7 @@ function App({ Component, pageProps, err }) {
             err={err}
           />
         </AnalyticsProvider>
+        <ReactQueryDevtools initialIsOpen={false} />
       </QueryClientProvider>
     </>
   )
--- a/dapp/src/components/TransactionHistory.js
+++ b/dapp/src/components/TransactionHistory.js
@@ -1,4 +1,4 @@
-import React, { useEffect, useState } from 'react'
+import React, { useEffect, useState, useMemo } from 'react'
 import { useRouter } from 'next/router'
 import dateformat from 'dateformat'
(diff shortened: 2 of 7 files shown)


dapp/pages/_app.js
```diff
@@ -161,28 +165,31 @@ function App({ Component, pageProps, err }) {
...
+            pauseOnHover
+          />
+          <Component
+            locale={locale}
+            onLocale={onLocale}
+            {...pageProps}
+            err={err}
+          />
+        </AnalyticsProvider>
+        <ReactQueryDevtools initialIsOpen={false} />
```
evgenyboxer, line 191: This is automatically disabled in prod.

dapp/src/components/TransactionHistory.js
```diff
@@ -120,9 +122,6 @@ const TransactionHistory = ({ isMobile }) => {
   const web3react = useWeb3React()
   const router = useRouter()
   const { account: web3Account, active } = web3react
-  const [history, setHistory] = useState(false)
-  const [shownHistory, setShownHistory] = useState(false)
-  const [currentPageHistory, setCurrentPageHistory] = useState(false)
```
evgenyboxer, line 125: Instead of using `setState` I'm computing them using `useMemo`

dapp/src/components/TransactionHistory.js
```diff
@@ -184,63 +183,28 @@ const TransactionHistory = ({ isMobile }) => {
...
-        const json = await response.json()
-        const history = json.history
-        _setFilters([])
-        setHistory(history)
-        setShownHistory(history)
-        setCurrentPage(1)
-      }
-    }
-
-    const attemptToFetchHistory = async () => {
```
evgenyboxer, line 223: The refetch is out of the box for react-query

dapp/src/components/TransactionHistory.js
```diff
@@ -184,63 +183,28 @@ const TransactionHistory = ({ isMobile }) => {
...
-
-  useEffect(() => {
-    if (!active && !account) return
-    const fetchAttempts = 5
-
-    const fetchHistory = async () => {
-      const response = await fetch(
-        `${
-          process.env.ANALYTICS_ENDPOINT
-        }/api/v1/address/${account.toLowerCase()}/history`
```
evgenyboxer, line 210: Moved to a service

dapp/src/queries/useTransactionHistoryQuery.js
```diff
@@ -0,0 +1,15 @@
+import { useQuery } from 'react-query'
+
+import { QUERY_KEYS } from '../constants/queryKeys'
+
+import { transactionHistoryService } from '../services/transaction-history.service'
+
+const useTransactionHistoryQuery = (account, options) => {
+  return useQuery(
+    QUERY_KEYS.TransactionHistory(account),
```
evgenyboxer, line 9: Since react-query keys are global we need to make sure we don't overwrite them. So I'm maintaining a constant list in `queryKeys`.

It's also very handy to have these keys become you can invalidate queries based on their key.

dapp/src/components/TransactionHistory.js
```diff
@@ -184,63 +183,28 @@ const TransactionHistory = ({ isMobile }) => {
...
+  }
+
+  const historyQuery = useTransactionHistoryQuery(account)
+
+  const history = useMemo(
+    () => (historyQuery.isSuccess ? historyQuery.data : []),
+    [historyQuery.isSuccess, historyQuery.data]
+  )
+
+  const shownHistory = useMemo(() => {
```
evgenyboxer, line 195: I have searched the repo and it didn't seem there was a use of `useMemo` - I assume the pattern of `useEffect` + `useState` was used instead.

dapp/src/components/TransactionHistory.js
```diff
@@ -262,29 +226,27 @@ const TransactionHistory = ({ isMobile }) => {
...
-    currentPageHistory &&
     currentPageHistory.filter(
       (tx) => tx.type === 'yield' && parseFloat(tx.amount) > 0.1
     ).length > 0
 
   return (
     <>
       <div className="d-flex holder flex-column justify-content-start">
-        {currentPageHistory && (
+        {historyQuery.isLoading ? (
```
evgenyboxer, line 247: Super easy to see the state of the query.

dapp/src/components/TransactionHistory.js
```diff
@@ -184,63 +183,28 @@ const TransactionHistory = ({ isMobile }) => {
   const setFilters = (filters) => {
     _setFilters(filters)
     setCurrentPage(1)
+  }
+
+  const historyQuery = useTransactionHistoryQuery(account)
```
evgenyboxer, line 188: This hook can actually be used even multiple times in a page and `react-query` is smart enough to dedupe the requests into one!

(so imagine you got multiple sections in the app relying on the history data, you can just use this hook and it won't make an extra network request)

sparrowDom: nice improvement, thanks

dapp/src/components/TransactionHistory.js
sparrowDom: Yeah this is a better approach. I defaulted to a less clean/practical pattern and haven't used `useMemo` where appropriate. Thanks for pointing this out

sparrowDom (approved):
Thanks @evgenyboxer for this PR. It is one of the most clean, well commented and working perfectly contributions I have seen in some time. I really appreciate it.

I like how react-query stores the state in the global context. Makes so much more sense and I like the convenience of its API.  Other areas where we also do fetches could benefit form it. Not over-doing the fetching, automatic retries and storing stuff in global context would all improve the performance and UX of the dapp.

dapp/src/services/transaction-history.service.js
```diff
@@ -0,0 +1,15 @@
+export default class TransactionHistoryService {
+  constructor() {
+    this.baseURL = `${process.env.ANALYTICS_ENDPOINT}/api/v1/address`
+  }
+
+  async fetchHistory(account) {
+    const response = await fetch(
```
TkDodo, line 7: Fyi, this doesn't handle errors correctly. See: https://react-query.tanstack.com/guides/query-functions#usage-with-fetch-and-other-clients-that-do-not-throw-by-default

evgenyboxer: Thanks for spotting this - will send a fix.

sparrowDom: I missed that one as well. Hope you don't mind @evgenyboxer pushed a minor commit to fix it: https://github.com/OriginProtocol/origin-dollar/commit/6d42c191acc0d6e4ce97671dd8d0c6803c059b43

evgenyboxer: looks good! thx.

PR #957, opened 2022-03-25
Wrapped OUSD

### LINKED ISSUE #876: OUSD wrapper contract
We have had several requests for an OUSD wrapper contract. This would provide a non-rebasing version of OUSD that keeps a fixed quantity without rebasing, and yet can be unwrapped to get an ever increasing amount of OUSD.

1. Tax benefits and simplifications. It is possible that in some jurisdictions, each rebase creates a separate taxable event. Also, by wrapping, in certain jurisdictions, yield is taxed at long term rates, rather than the much higher short term rates.

2. This allows off the shelf cross chain bridges to work with OUSD. While this bridged wOUSD would not have the 1-1 stablecoin property that OUSD has, it would still allow cross chain users to use OUSD as an investment vehicle, and to get it for cheap rates.

3. This provides an easy way to build yield bearing OUSD into other projects as a money lego.

In my view, there is little downside to a wrapped version of OUSD. The complexity is small. This empowers users a bit with more options and places to use OUSD. There's a potential supply increase from doing this. And if we can bridge to alt chains, even if it's wrapped, then that's quite a win.

DanielVF:
This adds wrapped OUSD, a non rebasing, but yield earning, contract for putting OUSD into. Closes #876

This is based on the solmate ERC4626 implementation. This implementation is in use by Tribe for their new Turbo Safes, but is not widely used yet. I think using a prebuilt ERC4626 implementation is a fantastic way to go here. The standard is a perfect match for a wrapper token, and the rounding and other details has seen a bit of community scrutiny. According to Joey with tribe, it was audited by a C4 contest and spearbit, however neither report is public at this time. This code has not yet been audited to the same level as OpenZeppelin contracts. That said, I think it's more refined, and has seen more community review, than something I would have written scratch here.

So we should also take a look through the base ERC4626  contract when reviewing this PR

I've tested this by making hundreds of paired test transactions, and ensuring the rounding is sane, and holders earnings follow the rebasing correctly.

Some decisions that are up for debate:

1. I've imported the solmate library contracts into our repo. Their last NPM release was a while ago, and does not have the ERC4626 implementation. I think they will be making an NPM release, possibly before this PR is merged, and we can switch back to using that then.
2. I've gone for a non-upgradable contract. Its cheaper gas, less can go wrong. I'm up for alternative views on this one.
3. I added transferToken to this. One day when someone mistakenly sends 600,000 USDC to this contract, it can be recovered. This also mean that wousd has to have a governor, and governor/ownership methods. Added.

Note 1: Because of the simplicity and elegantness of this implementation, if someone sends this contract OUSD, they aren't getting it back, and it is instantly yield for everyone holding OUSD in the contract.

Note 2: In the past when people have wanted to `rebaseOptIn` from their contracts, we've told them they would need to do it after the constructor. Turns out there's a way around that. :D 

Contract change checklist:
  - [ ] Unit tests pass


contracts/contracts/token/WrappedOusd.sol
```diff
@@ -0,0 +1,43 @@
...
+    ) ERC4626(_underlying, _name, _symbol) Governable() {
+        OUSD(address(_underlying)).rebaseOptOut(); // It's not treated as a contract yet
+        OUSD(address(_underlying)).rebaseOptIn();
+    }
+
+    /**
+     * @notice Show the total amount of OUSD held by the wrapper
+     */
+    function totalAssets() public view override returns (uint256) {
+        return ERC20(asset).balanceOf(address(this));
```
transmissions11, line 27: this is possibly manipulatable in a cream-hack esque manner no? would make wOUSD unsuitable for borrowing in lending markets

contracts/contracts/token/WrappedOusd.sol
```diff
@@ -0,0 +1,43 @@
...
+
+contract WrappedOusd is ERC4626, Governable {
+    using SafeERC20 for IERC20;
+
+    constructor(
+        ERC20 _underlying,
+        string memory _name,
+        string memory _symbol
+    ) ERC4626(_underlying, _name, _symbol) Governable() {
+        OUSD(address(_underlying)).rebaseOptOut(); // It's not treated as a contract yet
```
franckc, line 19: nit: add a comment to explain that calling rebaseOptOut(), followed by rebaseOptIn() allows to opt-in for rebasing in the constructor vs having to call rebaseOptIn() outside of the constructor. nice trick :)

franckc (commented):
Started reviewing. I will dig further in the ERC4626 implementation.

I'd vote for making wOUSD upgradeable even if it means the operations cost a little bit more gas (should be in the order of an extra 1k gas per call, isn't it?). To give us the peace of mind that we can upgrade in case of an emergency.

transmissions11:
note that making it upgradable will require writing your own ERC4626 impl as solmate is not designed for upgradability

contracts/contracts/token/WrappedOusd.sol
DanielVF: The issue is that you can flat out give OUSD to wOUSD, making each wOUSD worth more. This is normal a bad thing for an attacker. wOUSD holders, and OUSD holders both are happy and good here.

The problem is if a lending platform is using wOUSD as collateral, has lended it back to the attacker and suddenly thinks the attacker can borrow a bajillion dollars.


See these two threads:

- https://twitter.com/danielvf/status/1453426746326405130
- https://twitter.com/danielvf/status/1453460110760677379

This isn't a vulnerability in wOUSD, just something a lending platform has to take into account.

contracts/contracts/token/WrappedOusd.sol
DanielVF: With the change to using a proxy, we no longer use this silly technique, and now just rebaseOptIn() in the initializer.

franckc (approved):
Reviewed focusing on the contracts. 
ERC4626's implementation is really clean.
Looks very solid to me.

LGTM

contracts/deploy/039_wrapped_ousd.js
```diff
@@ -0,0 +1,71 @@
+const { deploymentWithProposal, withConfirmation } = require("../utils/deploy");
+
+// Deploy new staking implimentation contract with fix
```
tomlinton, line 3: Typo here and I think this comment is from a previous deploy

contracts/package.json
```diff
@@ -22,7 +22,7 @@
...
     "compute-merkle-proofs-local": "HARDHAT_NETWORK=localhost node scripts/staking/airDrop.js reimbursements.csv scripts/staking/merkleProofedAccountsToBeCompensated.json && cp scripts/staking/merkleProofedAccountsToBeCompensated.json ../dapp/src/constants/merkleProofedAccountsToBeCompensated.json",
     "compute-merkle-proofs-mainnet": "HARDHAT_NETWORK=mainnet node scripts/staking/airDrop.js reimbursements.csv scripts/staking/merkleProofedAccountsToBeCompensated.json && cp scripts/staking/merkleProofedAccountsToBeCompensated.json ../dapp/src/constants/merkleProofedAccountsToBeCompensated.json",
-    "slither": "yarn run clean && slither . --filter-paths \"crytic|mocks|@openzeppelin\" --exclude-low --exclude-informational --exclude conformance-to-solidity-naming-conventions,different-pragma-directives-are-used,external-function,assembly,incorrect-equality",
+    "slither": "yarn run clean && slither . --filter-paths \"crytic|mocks|solmate|@openzeppelin\" --exclude-low --exclude-informational --exclude conformance-to-solidity-naming-conventions,different-pragma-directives-are-used,external-function,assembly,incorrect-equality",
```
tomlinton, line 25: solmate is gone I think?

tomlinton:
Where does this OpenZeppelin ERC4626 implementation come from? I assume it is not from an official release which is why we've added it to the repo instead of just referencing the installed dependency. I think we should add a comment at the time of the OZ files indicating the source and the commit hash if any.

franckc (commented):
How about adding a test to make sure we can upgrade after there has been some activity on the wousd contract. Then checking the balances are right after the upgrade.

DanielVF:
Comes from the draft PR here:

Given that it is a draft, it's worth looking hard at.

https://github.com/OpenZeppelin/openzeppelin-contracts/pull/3171

contracts/package.json
DanielVF: Thanks!

contracts/deploy/039_wrapped_ousd.js
DanielVF: Removed, thanks!

DanielVF:
Added a test for upgrading.

contracts/test/token/wousd.js
```diff
@@ -3,7 +3,7 @@ const { defaultFixture } = require("../_fixture");
 
 const { ousdUnits, daiUnits, isFork, loadFixture } = require("../helpers");
 
-describe("WOUSD", function () {
+describe.only("WOUSD", function () {
```
franckc, line 6: Remove before merging to master :)

DanielVF: image

Fixed.

tomlinton:
Not sure if we are waiting on my approval for this, but if we are, I'm still insisting on that commit hash for the draft ERC4626 implementation to be added as a comment :P

tomlinton (approved):

DanielVF:
> Not sure if we are waiting on my approval for this, but if we are, I'm still insisting on that commit hash for the draft ERC4626 implementation to be added as a comment :P

Done!

Issue #655, opened 2021-08-21
Number/Input Flipping UX


micahalcorn:
There's one slight change I was recommending, but it's very minor. Currently, the bottom known value moves to the top when flipping (it follows the corresponding currency). But it also remains intact in the bottom field prior to the new calculation. I was expecting the top value to follow the flip and be moved into the bottom field. Again, it's not a big deal because the value will quickly be replaced but it caused me to get confused for a second.

**Edit**
Slightly different behavior observed and noted in https://github.com/OriginProtocol/origin-dollar/pull/958#issuecomment-1081462927 🤔

sparrowDom:
done via: https://github.com/OriginProtocol/origin-dollar/pull/958

Issue #732, opened 2021-09-06
Clarity Around ERC-20 Approval Step


micahalcorn:
We often hear from users who are confused by not receiving OUSD after submitting the approval transaction. Many people are either not familiar with the allowance concept or they're submitting the form quickly without noticing that the approval and swap are two different steps. I think we can help with this by doing something more like Uniswap and SushiSwap where there is a clear "1-2" sequence shown before the form is submitted at all.

**Uniswap: two buttons, "Swap" disabled until allowance is granted**

<img width="521" alt="Screen Shot 2021-09-06 at 7 38 45 AM" src="https://user-images.githubusercontent.com/273937/132234006-5b9bf83c-01f8-42e4-9dec-7f5f8eb5a485.png">

**SushiSwap: one "Approve" button, two-step sequence indicated below**

<img width="698" alt="Screen Shot 2021-09-06 at 7 39 05 AM" src="https://user-images.githubusercontent.com/273937/132234089-bdea3df7-2f25-4068-8c4b-4b336d122ec3.png">

**OUSD: single "Swap" button (same as when allowance already exists), conditional modal with approval sequence"**

<img width="574" alt="Screen Shot 2021-09-06 at 7 39 22 AM" src="https://user-images.githubusercontent.com/273937/132234296-90424ae1-d593-4c93-926b-747e76516119.png">
<img width="572" alt="Screen Shot 2021-09-06 at 7 39 36 AM" src="https://user-images.githubusercontent.com/273937/132234384-d0e98504-665b-4425-bc39-f48a90e39544.png">
<img width="576" alt="Screen Shot 2021-09-06 at 7 48 30 AM" src="https://user-images.githubusercontent.com/273937/132234564-8c97b1a7-8c4a-4d9f-8a09-70f3a06bb80d.png">

sparrowDom:
Nice UX improvement. If I had to vote, I would go for the Uniswap variant (a bit more user friendly for my eyes)

micahalcorn:
We might also want to incorporate the cost of allowance approval into the swap. For smaller amounts, it probably makes sense for some users to have their transactions routed through Uniswap instead of OUSD Swap even if the slippage or gas cost on the actual swap transaction is higher. We could also indicate on the table which contracts involve two transactions with some small icon.

joshfraser:
I like this idea. We could put a little tooltip on the icon explaining why they sometimes need to do 2 transactions.

sparrowDom:
[...]

PR #958, opened 2022-03-28
Approval process change

rolandpo:
This removes the approval-related modals from the swap process and adds a two-step sequence, as described in issue #732. If a token-route combination hasn't been approved, then the approval button appears with the swap button disabled. Approval button goes: 'Allow <contract> to use your <coin>' -> 'Waiting for you to confirm' -> 'Approving <coin>' -> '<coin> approved'

After the token gets approved the approval button is disabled and stays that way until a different token is chosen at which point it gets hidden.

This required removing the delay after which the bottom swap field value changes to 'Loading...' after an input change, which I believe also happens to fix issue #655

approval


sparrowDom:
Hey @rolandpo you are on 🔥 with these PRs :) Could you make these adjustments and ping me to take another look at the PR: 
- When approve call fails (wrong nonce for example) no error transaction pops up in the sidebar to the right
- `swap approved` message stays on screen even after swap has finished (user clicks on Swap and wait for transaction to be mined). Should disappear when:
  - the swap has completed OR
  - user clicks on another swap option (Curve, Uniswap, Flipper...) OR
  - swap routes are re-calculated (user has inputted new value into amount input field)

Other than that this PR is almost on the finish line, great work

rolandpo:
Hey, I think that's the second issue fixed.
On the first issue, from what I'm seeing the error transactions do appear in the side panel but they get inserted underneath each other and underneath the latest successful transaction. I might be way off on this, but I think when the transaction sorting mechanism in PrimarySidePanel copies the transaction list in the wrong order and then tries to reorder them, it just sorts by blocknumber, which only increases with a successful transaction in local. Curious to hear your thoughts on this

dapp/src/components/buySell/ApproveSwap.js
```diff
@@ -0,0 +1,257 @@
...
+    const coin = stableCoinToApprove.toUpperCase()
+    return (
+      <>
+        {isMobile
+          ? fbt(
+              'Approve ' +
+                fbt.param('flipper', flipper) +
+                fbt.param('vault', vault) +
+                fbt.param('capatalized', capitalized) +
+                fbt.param('noncapitalized', noncapitalized),
```
sparrowDom, line 102: It will be easier for translators to translate to another language if there is only 1 `fbt.param`. Since what they see in english is "Approve X" (where X is a variable called by the first parameter in function `fbt.param`). So it makes more sense for them to translate: 
`Approve [coin_contract]`
which in code you would express as
` fbt('Approve ' + fbt.param('coin_contract', '....')`

rather than trying to translate `Approve [flipper][vault][capatalized][noncapitalized]` which current code would produce. 

Can you put the whole coin contract in one variable (and do all the capitalisation, "the" preposition) logic in there. And then insert it as one fbt.param called something that will help transaltors understand what it is. like `coin contract` or `transaction route` for example

dapp/src/components/buySell/SwapCurrencyPill.js
```diff
@@ -390,6 +391,11 @@ const SwapCurrencyPill = ({
             <CoinSelect
               selected={showOusd ? 'ousd' : selectedCoin}
               onChange={(coin) => {
+                if (coin !== selectedCoin) {
+                  ContractStore.update((s) => {
+                    s.swapEstimations = 'loading'
```
sparrowDom, line 396: These should be triggered as soon as estimations are being ran: https://github.com/OriginProtocol/origin-dollar/blob/master/dapp/src/hooks/useSwapEstimator.js#L182

Any ideas why also the need for it here?

dapp/src/components/buySell/SwapCurrencyPill.js
```diff
@@ -430,6 +436,9 @@ const SwapCurrencyPill = ({
                 onChange={(e) => {
                   const value = truncateDecimals(e.target.value)
                   const valueNoCommas = removeCommas(value)
+                  ContractStore.update((s) => {
+                    s.swapEstimations = 'loading'
```
sparrowDom, line 440: Same question as the one above...

dapp/src/components/buySell/_PillArrow.js
```diff
@@ -56,6 +57,9 @@ const PillArrow = ({ swapMode, setSwapMode }) => {
         <div
           className="caret-background position-absolute d-flex align-items-center justify-content-center"
           onClick={() => {
+            ContractStore.update((s) => {
+              s.swapEstimations = 'loading'
```
sparrowDom, line 61: Leaving a mark here so this one also gets included in the above question

dapp/src/components/sidePanel/PrimarySidePanel.js
```diff
@@ -54,9 +55,11 @@ const PrimarySidePanel = () => {
         className="primarySidePanel sidepanel-wrapper collapse navbar-collapse"
       >
         <div
-          className={`primary-side-panel d-flex flex-column justify-content-start align-items-center disable-scrollbars ${
-            showingAllContracts ? 'larger' : ''
-          }`}
+          className={`primary-side-panel d-flex flex-column justify-content-start align-items-center disable-scrollbars
```
sparrowDom, line 58: Thanks, amazing attention to detail 🙏

sparrowDom (commented):
Left only a couple of minor comments other than that everything works nicely. Great job with this one 👏

sparrowDom:
To reply on your previous comment, yeah your are right. Ignore my previous comment about errors not showing up.  They do show up, just in local environment not on top, because they are sorted by block number, and those get reset in local environments on node restarts. Not an issue for mainnet.

rolandpo:
commited changes for the first two issues, I think the fbt stuff is better now, but let me know, I might be missing something

micahalcorn:
This works great, thx @rolandpo 👏

I would recommend changing the phrase "{coin} approved" to "{contract} approved" since the coin is already represented by the form state and the allowance is being granted to the contract. Others may disagree and it's not a big deal either way.

Screen Shot 2022-03-28 at 11 19 05 PM

I've also filed a couple of other minor issues that we can leave unresolved as they shouldn't block us from merging these improvements.

https://github.com/OriginProtocol/origin-dollar/issues/963
https://github.com/OriginProtocol/origin-dollar/issues/964

micahalcorn (approved):

micahalcorn:
@rolandpo in regard to https://github.com/OriginProtocol/origin-dollar/issues/655, I'm not sure that it even completely behaves as I thought it did when I filed that issue before. Check out this screen recording:

https://www.loom.com/share/ba2702d6944b4b83ab9a0ebd5e9b7386

**Expected behavior when flipping the form**
- Currency on the bottom moves to the top and its corresponding number (previously the output) becomes the new input
- Currency on the top moves to the bottom and its corresponding number get recalculated immediately
- Mix is changed to DAI when it changes from bottom to top since it's not a supported input
- Input amount remains unchanged if the top currency is changed to OUSD

Again, this is minor and it shouldn't block the merge.

rolandpo:
Ah I see what you mean @micahalcorn, I didn't really look at the how the field values change on a flip. What I did change though, which you can actually see in the video, is that little confusing delay in the bottom field which shows the outdated output for second between the arrow being clicked and 'Loading' appearing.

rolandpo:
Ah I see what you mean @micahalcorn, I didn't really look at the how the field values change on a flip. What I did change though, which you can actually see in the video, is that little confusing delay in the bottom field which shows the outdated output for second between the arrow being clicked and 'Loading' appearing.

rolandpo:
fix for #964

rolandpo:
fix for #964

rolandpo:
changed 'coin approved' to 'contract approved'

rolandpo:
changed 'coin approved' to 'contract approved'

rolandpo:
added some simple states to keep track of the last pending approval and maintain the approval button state between contract changes as described in #963

rolandpo:
added some simple states to keep track of the last pending approval and maintain the approval button state between contract changes as described in #963

dapp/src/components/buySell/ApproveSwap.js
```diff
@@ -0,0 +1,264 @@
...
+  onMintingError,
+  formHasErrors,
+  swappingGloballyDisabled,
+  storeTransaction,
+  storeTransactionError,
+  rpcProvider,
+  isMobile,
+}) => {
+  const [coinApproved, setCoinApproved] = useState(false)
+  const [stage, setStage] = useState(coinApproved ? 'done' : 'approve')
```
sparrowDom, line 27: Since line 26 always starts initial state as `false`. Line 27 initial state should always default to `approve`? Or am I not getting something?

dapp/src/components/buySell/SwapCurrencyPill.js
sparrowDom: Aha I see. That is because there is a 700 ms delay on running the estimations. The reason for that is so that they don't run for each keystroke user does. E.g. you want to swap 2000 stablecoin to OUSD. And they would run for each of those 4 digits. I recommend that you handle this case in this code: 

https://github.com/OriginProtocol/origin-dollar/blob/master/dapp/src/hooks/useSwapEstimator.js#L164

The code in `useSwapEstimator` should recognise that the stablecoin type has been changes or that user has swapped from buying to selling. And in that case not apply 700ms delay. 

The goal is to have all the logic that pertains to the swap delay encapsulated in 1 function in `useSwapEstimator`. It is a lot easier to return to that code in the future and modify it. In the other case if the logic is spread out into various functions it can become very hard to maintain and remember what other parts of the dapp control the logic of delaying estimations.

dapp/src/components/buySell/ApproveSwap.js
```diff
@@ -0,0 +1,264 @@
...
+        formHasErrors ||
+        swappingGloballyDisabled ||
+        !allowancesLoaded ||
+        !needsApproval) &&
+      !coinApproved
+    )
+    ContractStore.update((s) => {
+      s.approvalNeeded = approval
+    })
+  }, [selectedSwap, needsApproval])
```
sparrowDom, line 45: Shouldn't other things also trigger an execution of `useEffect` besides `selectedSwap &needsApproval`?
Things like: formHasErrors, swappingGloballyDisabled, allowancesLoaded, coinApproved?

dapp/src/components/buySell/ApproveSwap.js
```diff
@@ -0,0 +1,264 @@
...
+      </>
+    )
+  }
+
+  return (
+    <>
+      <button
+        className={`btn-blue buy-button mt-4 mt-md-3 w-100`}
+        hidden={
+          (!selectedSwap ||
```
sparrowDom, line 148: Seems very similar code to lines ~33 in this file. Could this be represented as an `approvalNeeded` bool and used in both places?

sparrowDom (commented):
Left a few comments inline. It is getting a bit late here, will look at it tomorrow again with fresh eyes. 

The main thing I would want to simplify with this PR is the way the state of the approve button is stored. In `coinApproved`, `stage`, `isApproving `. Will try to propose a way to simplify it tomorrow.

sparrowDom:
@rolandpo sounds good 👍

dapp/src/components/buySell/ApproveSwap.js
```diff
@@ -0,0 +1,282 @@
...
+    ((!selectedSwap ||
+      formHasErrors ||
+      swappingGloballyDisabled ||
+      !allowancesLoaded ||
+      !needsApproval) &&
+      !coinApproved) ||
+    !show
+
+  useEffect(() => {
+    setShow(false)
```
rolandpo, line 44: this stuff hides the approval button right after switching currencies, otherwise the wrong currency is shown on the button for a split second

dapp/src/components/buySell/ApproveSwap.js
```diff
@@ -0,0 +1,289 @@
...
+    sushiRouter,
+    curveOUSDMetaPool,
+    usdt,
+    dai,
+    usdc,
+    ousd,
+  } = useStoreState(ContractStore, (s) => s.contracts || {})
+
+  const routeConfig = {
+    vault: {
```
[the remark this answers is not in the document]
sparrowDom: Ahh I see that works nicely yes

dapp/src/components/buySell/ApproveSwap.js
sparrowDom: Aha I see thanks for explaining. Made a small change in interest to remove the `show` flag: 
https://github.com/OriginProtocol/origin-dollar/pull/958/commits/a5eeaab5059461f6aba1d619bbe872e10e830860

Moved the global setter of `swapEstimations` out of the `setTimeout` function in case when delay is 0. 

While your approach definitely worked (arguably worked better since in some cases the wrong stablecoin would still be shown for 1 render frame) I really like to have as minimal amount of moving parts (show flag in this case) that can control the rendering of the component.

I realise this might seem like a weird obsession (and it did to me as well in the past), but working on the same codebase for years really makes you appreciate when things are simple. Sometimes code complexity can really grow so big (and I have been guilty of that in the past) that development grinds to a halt, because there are so many interconnected moving parts that making one change breaks 3 other things.

PR #967, opened 2022-03-30
fix ledger address error

rolandpo:
Fix for Ledger address error. The code was setting each derivation path in sequence to extract all addresses, but on choosing an address the path wasn't being changed, so anything other than the last preloaded derivation path wasn't activated.


sparrowDom (approved):
LGTM

Issue #956, opened 2022-03-23
Allow for more granular price tolerance in the DApp


joshfraser:
We currently only allow the user to choose between 1%, 0.5% or 0.25% for their price tolerance. As I understand it, this is telling the AMM to cancel the transaction altogether if they can't guarantee at least X amount. Our current implementation is fine for small amounts but leaves plenty of opportunity for large (7 and 8-figure) traders to get front-run by bots. We should allow custom values & add an explanation bubble about what this is & how it works.

franckc:
@rolandpo Can you take a look? ty :)

PR #970, opened 2022-04-07
adjustable price tolerance

rolandpo:
Adjustable price tolerance from issue #956
User can set a custom slippage percentage, or set to default 0.5%. I capped it to 50% for now. I can add some information/warning later on if that's the play

Screenshot 2022-04-07 at 20 46 47


sparrowDom (approved):
Nicely done thanks 👍

PR #972, opened 2022-04-11
high slippage warning

rolandpo:
High slippage warning when input is larger than 1%.

I've not added any explanation bubbles, to not clutter the settings box further, and with the live 'min received' field sort of serving that purpose, but happy to add some if needed.

Screenshot 2022-04-11 at 13 08 18


sparrowDom (approved):
Nicely done!

PR #974, opened 2022-04-12
fix input flipping

rolandpo:
Small follow-up on what @micahalcorn was talking about in #655. On clicking the flip arrow the previous output becomes the input on either swap direction change.

To be fair I think this change might make changing the swap mode through clicking on OUSD in the dropdown menu less intuitive so appreciate any thoughts. It might not even be necessary


sparrowDom (approved):
Tnx @rolandpo I think this does improve the UX. Have also added the 2 decimal places rounding, so that the number you see in the bottom pill is exactly the same as you see in the top pill. 

IMHO it is an UX improvement though a relatively minor one

Issue #658, opened 2021-08-21
Error Message Placement


micahalcorn:
Per @auregimon, let's use the disabled button copy to communicate issues with the inputs rather than near the inputs themselves.

<img width="520" alt="Screen Shot 2021-08-21 at 12 55 06 PM" src="https://user-images.githubusercontent.com/273937/130333468-778208f6-4f30-4f59-8e09-4fd660987f2d.png">

vs

<img width="616" alt="Screen Shot 2021-08-21 at 12 56 52 PM" src="https://user-images.githubusercontent.com/273937/130333474-92466baa-4594-412e-941f-60bbef6a7a0c.png">

Michael1104Smith:
What disable button do you mean?
So we should add numbers below input box?

micahalcorn:
> What disable button do you mean? So we should add numbers below input box?

No, the disabled button in that screenshot reads "Swap". It should read "Insufficient Balance" and the red text should not appear. The button should also be full-width so that its width doesn't change each time the text changes.

Michael1104Smith:
I have done it and sent "pull" request.

franckc:
@Michael1104Smith It does not look like you are actively working on this issue so un-assigning it....

rolandpo:
done in #975

PR #975, opened 2022-04-13
change error message placement

rolandpo:
Changing the error message placement for insufficient balance and nonavailable routes from the input fields to instead replace the text on the disabled swap button as in #658
Also some general cleanup in the approve/swap files


sparrowDom (approved):
Nicely done, also thanks for the cleanup

Issue #653, opened 2021-08-21
No Estimate Received on Unsupported Contracts


micahalcorn:
When a contract is "Unsupported", its corresponding "Est. received" should be "-" instead of "0.00" to match the other two columns.

<img width="617" alt="Screen Shot 2021-08-21 at 9 35 43 AM" src="https://user-images.githubusercontent.com/273937/130328751-1e87315c-a0ac-4c26-8e23-b5bd7ec973c0.png">

rolandpo:
done in #977

Issue #652, opened 2021-08-21
Max Including Dust


micahalcorn:
If I use the "Max" link, I expect all of that particular currency to be used. It's okay to truncate the input for display purposes, but the remainder should be included in the transaction so that I'm not left with some stablecoin dust.

Michael1104Smith:
Where is "Max" link?

micahalcorn:
<img width="169" alt="Screen Shot 2021-09-28 at 7 31 27 AM" src="https://user-images.githubusercontent.com/273937/135107999-d020df5d-373f-4a86-af3b-5f5a210e1bc9.png">

If I click "Max" and submit this transaction, it will swap 151.552942 DAI but my balance is currently 151.552942421692815265 DAI. I don't want to be left with 0.000000421692815265 DAI after the swap.

Michael1104Smith:
I have done it and sent "commit" request.

franckc:
@Michael1104Smith It does not look like you are actively working on this issue so un-assigning it....

rolandpo:
done in #977

PR #977, opened 2022-04-18
max including dust + unsupported contract estimation

rolandpo:
Changes for including all decimal places when selecting max balance on swap page for issue #652, and estimate display change for issue #653.
The max balance code might seem a bit overly complex, but I think it's needed to cover all the weird formatting edge cases.

### DIFF
--- a/dapp/src/components/buySell/SwapCurrencyPill.js
+++ b/dapp/src/components/buySell/SwapCurrencyPill.js
@@ -348,7 +348,12 @@ const SwapCurrencyPill = ({
   const coinSplits = bottomItem && selectedSwap && selectedSwap.coinSplits
 
   useEffect(() => {
-    if (maxBalanceSet && Number(removeCommas(roundTo2to6Decimals(displayBalance.detailedBalance))) !== Number(removeCommas(roundTo2to6Decimals(coinValue)))) {
+    if (
+      maxBalanceSet &&
+      Number(
+        removeCommas(roundTo2to6Decimals(displayBalance.detailedBalance))
+      ) !== Number(removeCommas(roundTo2to6Decimals(coinValue)))
+    ) {
       setMaxBalanceSet(false)
     }
   }, [coinValue])
@@ -370,7 +375,9 @@ const SwapCurrencyPill = ({
   }
 
   const formatMax = (value) => {
-    return maxBalanceSet && parseFloat(displayBalance.balance) > 0 ? removeCommas(roundTo2to6Decimals(value)) : value
+    return maxBalanceSet && parseFloat(displayBalance.balance) > 0
+      ? removeCommas(roundTo2to6Decimals(value))
+      : value
   }
 
   return (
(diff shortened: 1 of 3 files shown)


rolandpo:
also added a change to align the balance header in the history tab to be consistent with the swap page

sparrowDom (commented):
I think the logic to figure out if max balance is currently set is quite complicated. The more complicated the logic the better chance we will introduce errors to it.  

Also If I input manually the maximum amount the UI doesn't figure out to treat it as a max value. 

All coins we are displaying are stablecoins. I would suggest that we should figure out solely from the value of the input if that input value should be treated as a maxValue one. I was so before, curious what were the edge cases that approach failed to cover?

dapp/src/components/buySell/SwapCurrencyPill.js
```diff
@@ -418,16 +429,24 @@ const SwapCurrencyPill = ({
...
                     onAmountChange(valueNoCommas)
                   }
                 }}
                 onBlur={(e) => {
+                  setMaxBalanceSet(
+                    displayBalance &&
+                      Number(displayBalance.detailedBalance) ===
+                        Number(coinValue)
+                  )
                   const valueRounded = removeCommas(
```
sparrowDom, line 450: Also `onBlur` still seems to set the value of the swap calling `onAmountChange` with the rounded amount. So by: 
- clicking on `Max` button
- focusing into the input field
- focusing out (by clicking `Swap` lets say)
- the `onBlur` will be called, by rounding the amount and leaving the `dust` in the user's wallet

dapp/src/components/buySell/SwapCurrencyPill.js
```diff
@@ -418,16 +429,24 @@ const SwapCurrencyPill = ({
             {topItem && (
               <input
                 type="text"
-                value={coinValue}
+                value={formatMax(coinValue)}
                 placeholder="0.00"
                 onChange={(e) => {
+                  console.log('change')
```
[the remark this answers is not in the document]
rolandpo: yep

dapp/src/components/buySell/SwapCurrencyPill.js
rolandpo: fixed

rolandpo:
I think the idea of swapping one value but displaying a different one is what makes this a bit complicated. Currently when you input a value that happens to be max, then it will be recognised only on onBlur.

Otherwise, if my max is 10, then 10.00 will display as max, but I wouldn't be able to remove trailing zeroes on edit because the dust value gets locked between 2-6 decimals when passed to onChange.

rolandpo:
So to ensure that input editing isn't affected in any way, I'm only asserting maxValueSet either when 'max' button is clicked or on onBlur. And then deasserting whenever the value get edited again (even if it doesn't change the actual value like a trailing zero)

sparrowDom:
Aha I see thanks for explaining the rationale behind it. 

The one nitpick issue I still experience is if: 
- click on the max value, input changes to balance of coin. E.g. `1.981981`
- change value of coin to something else say `1`
- copy paste back the max value of coin: `1.981981`

The `Max` link is still present on the left, but it should be hidden

sparrowDom:
Let me take a stab at this to see if some code simplification is possible

rolandpo:
> * click on the max value, input changes to balance of coin. E.g. `1.981981`
> * change value of coin to something else say `1`
> * copy paste back the max value of coin: `1.981981`
> 
> The `Max` link is still present on the left, but it should be hidden

That would be consistent no? Either balance is exactly 1.981981 and on pasting the value and clicking off the field the Max link hides, or the exact balance has some extra dust and a copy paste only takes up to 6 decimal places so the Max link stays present

sparrowDom:
@rolandpo have tried to simplify code somewhat. Can you check if this PR and see if it achieves all you wanted?
https://github.com/OriginProtocol/origin-dollar/pull/987

My though process was that lets allow the component to set the maximum amount of coins including dust and communicate that to the parent component (`SwapHomepage`). But only display the "nice" value that is floored to show maximum 6 decimals.

sparrowDom:
> 

Figured later why copy/paste hasn't worked but works on current master. Dust usually has more than 6 decimals. And when copy pasting stuff the code needs to compare both amounts rounded to 6 decimals, since user isn't able to input more than 6

Issue #978, opened 2022-04-18
Progressive loading of the history page?


joshfraser:
I have 26 pages of history in my History tab making it quite slow to load.

Is it possible to do progressive loading? It would be nice to load the first page of results before starting to process the following pages. If necessary, we could disable the filters until everything has loaded.

sparrowDom:
This is partly solved via the: https://github.com/OriginProtocol/origin-dollar/pull/980

sparrowDom:
We haven't done any pagination on the analytics or dapp side of this issue. The loading is very slow due to the huge amount of data that needs to go over the network. @rolandpo would you like to take on additional task and: 
- make analytics backend support pagination
- fetch data from the dapp using pagination

rolandpo:
Sure 👍

sparrowDom:
Just a heads up, while time passes and we have more rebase events each day, the performance impact of not using pagination is going to become more apparent. Especially on mobile devices where with slower CPUs.

rolandpo:
done in #998

PR #980, opened 2022-04-19
preloading history

rolandpo:
PR for issue #978. Moved transaction history fetching to AccountListener to preload the data in the background on connecting an account. Bit of a different solution and it refetches all this data more often now, not sure if that's an issue.

### DIFF
--- a/dapp/src/components/AccountListener.js
+++ b/dapp/src/components/AccountListener.js
@@ -21,6 +21,7 @@ import { isProduction, isDevelopment } from 'constants/env'
 import useBalancesQuery from '../queries/useBalancesQuery'
 import useAllowancesQuery from '../queries/useAllowancesQuery'
 import useApyQuery from '../queries/useApyQuery'
+import useTransactionHistoryQuery from '../queries/useTransactionHistoryQuery'
 
 const AccountListener = (props) => {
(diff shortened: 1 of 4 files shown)


dapp/src/components/AccountListener.js
```diff
@@ -344,6 +353,7 @@ const AccountListener = (props) => {
     } else {
       balancesQuery.refetch()
       allowancesQuery.refetch()
+      historyQuery.refetch()
```
sparrowDom, line 356: I would probably rather refetch history only when `account` or `chainId` changes in `useEffect` below the one I am commenting on. To reduce the amount of fetches.

rolandpo:
ok I think that's done, also added a refetch on initial render of transaction history component

dapp/src/components/TransactionHistory.js
```diff
@@ -192,6 +192,10 @@ const TransactionHistory = ({ isMobile }) => {
     [historyQuery.isSuccess, historyQuery.data]
   )
 
+  useEffect(() => {
+    historyQuery.refetch()
```
sparrowDom, line 196: nice touch 👍

sparrowDom (approved):
Very clean and works perfectly. Good job 👍

Issue #992, opened 2022-05-06
Ledger modal doesn't include OUSD in stablecoin balance


joshfraser:
When you connect your ledger, we show the ETH value and stablecoin value under each address to help you select the right wallet.

We currently only include the value of USDT, USDC, and DAI in this calculation. It should include OUSD too.

rolandpo:
done in #994

PR #994, opened 2022-05-10
add ousd to ledger stablecoin balance

rolandpo:
count OUSD in sum of stablecoin balance in Ledger modal as in #992


sparrowDom (approved):
LGTM

Issue #984, opened 2022-04-20
OIP-1 Pause Gaurdian


DanielVF:
There is often a window of 3-7 minutes between an attack contract deployment and the execution of the attack. These attacks are often first noticed by people external to the organization. To pause currently requires:

1. Someone to notice the exploit
2. Find out how to contact us
3. Actually contact someone
4. Convince someone
5. Initial origin responder has to contact another signer
6. We have to create the pause transaction.
7. Second signer has to sign

Instead we could create a contract that has one time use pause actions. After 12 hours OUSD automatically unpauses unless the pause is confirmed by a multisig. These pause authorizations can then be distributed to external security researchers that are monitoring the blockchain. Then if they see an attack come through they can pause immediately in one step. This could cut the reaction time down to 2-3 minutes.

Even if this was only used internally, this would still be quiet the speed difference.

franckc:
I like the idea since it drastically reduces the incident response time.
I assume we would whitelist wallet addresses of the security researchers that would be allowed use this functionality?

DanielVF:
Yes, it would be to specific trusted individuals.

sparrowDom:
Great idea! Does it make sense for this to become an EIP? So that contracts to be paused all implement a common interface. Should be easier for security researchers to interact with those. Especially in such time critical situations. 

I can totally see a need for a service (with website similar to snapshot.org) where other protocols aside from OUSD opt in and give permissions to specific security researchers to pause their contracts. And previous activity of security researchers could be seen, so that trust can be based on historical success. So it is easier for platforms to pick who to trust.

Naturally there should also be monetary incentives when an attack is prevented.

DanielVF:
We have two pause methods, though I wonder if we should combine them into one.

staa99:
Hi @franckc,

The idea is interesting and I'd like to implement the contract.

Considering the requirement that the caller of the pause functions, `pauseRebase` and `pauseCapital`, must be the `governor` or `strategist`, then it might be necessary to extend the `governor` or `strategist` with the desired functionality or upgrade the contract to allow a new address (the `pauser`) to pause/unpause the vault.

What do you think?

franckc:
> Considering the requirement that the caller of the pause functions, `pauseRebase` and `pauseCapital`, must be the `governor` or `strategist`, then it might be necessary to extend the `governor` or `strategist` with the desired functionality or upgrade the contract to allow a new address (the `pauser`) to pause/unpause the vault.

[...]

PR #997, opened 2022-05-16
OIP-1 Implement standard pauser with temporary pausing support

staa99:
# OIP-1: Pause Guardian

Refers to issue: #984 

## Summary

This PR includes an implementation of a standard `Pauser` contract that enables a single-use (per user) temporary pause feature to speed up the response in case of attacks. The users are a whitelist of security researchers granted access by the governor or strategist in the OUSD system. The users have authorization for temporary pausing that can be confirmed by the governor or strategist, or canceled by any user. The users might be incentivized to correctly use this feature to help prevent attacks but that is beyond the scope of the implemented solution.

## Contract change checklist:
  - [ ] Unit tests pass


DanielVF:
Hi! I will be look at this tomorrow morning. Apologies for the delay, I've been out of office.

staa99:
Thanks @DanielVF

DanielVF:
Thanks for writing this, @staa99! Overall, it's good code that's easy to follow. Great work figuring out our deployment and tests setups.

I'll write individual notes as separate comments.

DanielVF:
Since we are changing to always pausing both vault rebasing and capital at the same time, and unpausing them together, we can just use single pause variable do both. (rename old vars as deprecated, add new var to end of storage.  We can also change to using a stand pause events from https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/security/Pausable.sol . 

This should let us have less code in the vault contracts.

contracts/contracts/pausing/Pauser.sol
```diff
@@ -0,0 +1,232 @@
...
+    event ExpiryDurationChanged(uint256 _newExpiry);
+    event PausableChanged(address _newPausable);
+    event StrategistChanged(address _newStrategist);
+    event TempPaused(uint256 _expiry);
+    event Paused();
+    event Unpaused();
+    event Whitelisted(address _account);
+    event DeWhitelisted(address _account);
+
+    uint8 public constant UNPAUSED = 0;
```
DanielVF, line 18: These there state values should probably be an `enum`, since that enforces that not just any number is okay here.

staa99: Good point. Thanks for catching that. I'll update it.

staa99:
Thanks for the review @DanielVF. I'll update the contracts and tests accordingly.

contracts/contracts/pausing/Pauser.sol
```diff
@@ -0,0 +1,232 @@
...
+
+    /**
+     * @dev Address of the strategist
+     */
+    address public strategistAddr = address(0);
+
+    /**
+     * @dev whitelist of addresses allowed to interact with the functionality exposed by this contract.
+     */
+    mapping(address => bool) public whitelist;
```
DanielVF, line 51: Rather than writing our own code to add and remove whitelisted pausers, perhaps we should use the OpenZeppelin access control library.

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/master/contracts/access/AccessControl.sol

contracts/contracts/pausing/Pauser.sol
```diff
@@ -0,0 +1,232 @@
...
+        emit Unpaused();
+
+        // Execute the unpausing action on the contract
+        Pausable(pausable).unpause();
+    }
+
+    /**
+     * @dev Pauses the managed contract indefinitely
+     */
+    function pause() external onlyGovernorOrStrategist {
```
DanielVF, line 172: When someone needs to call pause on this contract, it's going to be very urgent. Because of that, we want to make it as simple and hard to mess up as possible.

So in this case, I think we should collapse the the `tempPause`, `confirmPause`, `cancelTempPause`, `pause`, and `unpause` methods down to just `pause` and `unpause`. This way it's always clear to the user under stress what should be used. They just need to know what state they want the contract to be in, and don't have to worry about the state that the contract is currently in.

`Pause()` moving to a full pause or a temp pause depends on who is calling it.

image

image

image

contracts/contracts/pausing/Pauser.sol
```diff
@@ -0,0 +1,232 @@
...
+        emit DeWhitelisted(_account);
+    }
+
+    /**
+     * @dev Pauses the managed contract for the duration of `expiryDuration`
+     * The pause action can be canceled by any user when the temp-pause expires.
+     * The contract MUST be in an UNPAUSED state.
+     * The function is one-time-use (The contract cannot be temp-paused by the same account twice).
+     */
+    function tempPause() external onlyWhitelistedOrGovernorOrStrategist {
```
DanielVF, line 110: When a whitelisted user calls `pause()` to pause the contract, their role should be removed, blocking their ability to pause again. This prevents a single account from permanently locking up funds by calling unpause/pause over and over again, in the absence of action from the strategist.

This also removes the need for the pauseExpiry mapping variable.

contracts/contracts/pausing/Pauser.sol
DanielVF: Note two differences in this diagram from the current code:


1. Any member of the public can unpause, once the expiration is past. Not just privileged users. The intent is to not have a way that the contract funds could be permanently lost if there were no strategist and a whitelisted user was malicious.

2. Whitelisted users cannot unpause from a temp pause until the expiration time is up, at which point they are like anyone else. This gives us a guaranteed time to react and investigate.

contracts/contracts/pausing/Pauser.sol
staa99: Good idea. Makes it easier to scale in case the requirements change and we need more roles.

contracts/contracts/pausing/Pauser.sol
staa99: Good idea. I think it also makes it easier to maintain. Thanks for the review.

staa99:
Hi @DanielVF,

I have just made the necessary updates. Please review when you're available.

Thank you.

contracts/contracts/vault/VaultCore.sol
```diff
@@ -80,7 +72,7 @@ contract VaultCore is VaultStorage {
         emit Mint(msg.sender, priceAdjustedDeposit);
 
         // Rebase must happen before any transfers occur.
-        if (unitAdjustedDeposit >= rebaseThreshold && !rebasePaused) {
```
DanielVF, line 83: I think we can remove this check because the entire function is blocked by a paused check.

contracts/contracts/pausing/Pauser.sol
```diff
@@ -0,0 +1,172 @@
...
+        // Cannot unpause if already unpaused
+        require(
+            pauseState != PauseState.UNPAUSED,
+            "Contract is already unpaused"
+        );
+
+        // Normal users can unpause when the contract has expired
+        if (!_isGovernorOrStrategist()) {
+            require(
+                block.timestamp > currentExpiry,
```
DanielVF, line 121: Let's be explicit here and also require that the contract is in TEMP_PAUSED. Right now I think anyone could unpause a full pause because the expiry was set to zero.

DanielVF (commented):
First pass looks very good! Very though code changes. I'll take a further look later today.

contracts/contracts/pausing/Pauser.sol
staa99: True. Very important catch. I'll add tests to cover that case as well.

contracts/contracts/vault/VaultCore.sol
staa99: This means I'm still missing something. The variable should already be renamed so that code shouldn't be valid I think.

I'll check later.

PR #998, opened 2022-05-17
transaction history pagination

rolandpo:
This starts by fetching the first page of the paginated transaction history data from analytics. That gets displayed on the history page along with the extracted number of pages. After that, the rest of the data gets fetched in the background in one call.

Paginating data past the first page didn't make much sense to me, as the filters and export button require all the transactions to be fully loaded + looping through multiple calls would take more time. Open to any suggestions otherwise

**Connected PR:** https://github.com/OriginProtocol/ousd-analytics/pull/210


dapp/src/queries/useTransactionHistoryQuery.js
```diff
@@ -4,12 +4,12 @@ import { QUERY_KEYS } from '../constants/queryKeys'
 
 import { transactionHistoryService } from '../services/transaction-history.service'
 
-const useTransactionHistoryQuery = (account, options) => {
+const useTransactionHistoryQuery = (account, transactionItems, options) => {
   return useQuery(
     QUERY_KEYS.TransactionHistory(account),
-    () => transactionHistoryService.fetchHistory(account),
+    () => transactionHistoryService.fetchHistory(account, transactionItems),
```
sparrowDom, line 10: This function should accept items per page as second parameter and not total amount of items I think?

dapp/src/components/AccountListener.js
```diff
@@ -74,7 +76,26 @@ const AccountListener = (props) => {
...
-  const historyQuery = useTransactionHistoryQuery(account)
+  const historyPageQuery = useTransactionHistoryPageQuery(
+    account,
+    transactionHistoryItemsPerPage
+  )
+
+  const transactionItems = useMemo(
+    () =>
+      historyPageQuery.isSuccess
+        ? historyPageQuery.data.page.pages * transactionHistoryItemsPerPage
```
sparrowDom, line 87: Last page might not be complete (have the whole `transactionHistoryItemsPerPage`). Not sure if this plays a role

dapp/src/services/transaction-history-page.service.js
```diff
@@ -0,0 +1,21 @@
+export default class TransactionHistoryPageService {
+  constructor() {
+    this.baseURL = `${process.env.ANALYTICS_ENDPOINT}/api/v1/address`
+  }
+
+  async fetchHistory(account, transactionHistoryItemsPerPage) {
+    const response = await fetch(
+      `${
+        this.baseURL
+      }/${account.toLowerCase()}/history?per_page=${transactionHistoryItemsPerPage}`
```
sparrowDom, line 10: I don't see fetching of specific page number here?

sparrowDom (commented):
I haven't test run the PR, but I am not completely sure the pagination is working here (I could be wrong). What is seems to me is that all of the data is still fetched upfront. React query has some [pagination documentation](https://react-query.tanstack.com/guides/paginated-queries) that could be useful.

rolandpo:
I was planning on having the first page of data display as quick as possible and then loop through the rest of the pages in the background, but just that meant making multiple calls that would load slower where all I wanted is to load all the data, so I ended up ignoring the pagination there.

It might be quite unintuitive now, but I don't think any progressive loading past the first page makes sense because both the filter buttons and the csv export rely on all the transactions being readily available

sparrowDom:
Hmm I see thanks for pointing out the limitations. To implement this *correctly* you would need always load stuff lazily - only when required. Meaning that frontend should pass to the backend:
- current page
- items per page
- filters

and the backend should only respond with the data (the page) requested. Meaning the backend would do the filtering plus pagination and respond only with the data displayed to the user.

For CSV export the frontend would probably need to show some waiting/loading dialog and loop through all the pages to fetch the complete data set. And then perform export on that.

I do realise this further complicates the code, but that is general way how pagination is approached. With this approach we can minimise the network traffic bottleneck and amount of data we pass from the server to the dapp. Also the more data we pass, the more client device needs to store/process and the slower it performs.

rolandpo:
awesome, makes sense

rolandpo:
added full pagination here, single page loads on selecting page number or filter
export button has a separate query to get all data at once

rolandpo:
added caching data by previous page + filter combinations. I also added a simple grey font loading indicator just as a temporary measure, but maybe it's not that bad actually. let me know what you think @sparrowDom

sparrowDom:
lgtm

PR #1002, opened 2022-05-20
SparrowDom/metastrategy

sparrowDom:
**TODO:**
- [x] add unit tests for testing `netOusdMintedForStrategy`
- [x]  document the helper functions in brownie
- [x]  make more unit tests
- [x] make brownie scripts that check on live node nothing can go wrong
- [x] decide which metapools to support and write corresponding deploy scripts

[Deploy spreadsheet link](https://docs.google.com/spreadsheets/d/1phyzOJMmTBPIqTTa0v7HY6XJkjRmbrcdULRZPo_JEoY/edit#gid=1077862851)

This PR presents a strategy to make use of [Curve's Metapool](https://curve.fi/factory/9) by depositing 3CRV + OUSD liquidity into the pool. The OUSD for the deposit is minted without backing and when liquidity is removed it is also burned without receiving stablecoins for the burn. 

The (also some opinionated & up for scrutiny) changes in this PR:
 - `_getTotalPTokens` function is removed from `baseCurveStrategy`. The main problem was that the format of that function has limitation where only 3CRVLP tokens are ever preset in contracts extending it. The change is that base Curve contract doesn't *really* have to know of how many 3CrvLP tokens are staked in various gauges to function. It was pretty convenient when only Convex and 3Pool strategies were using that. Since those could always tell how much 3CRV are used / staked in a gauge. The `Metastrategy` doesn't have that info handy, since it deals with Metapool LP tokens staked in gauge and there is no simple calculation/conversion from `metapoolLP` tokens to `3CRVLP` tokens. Also that can somewhat fluctuate when removing liquidity from metapool in a balanced manner since the pool can be tilted either towards OUSD or towards 3crv. And metapool as its own MetapoolLP tokens that have their own virtual price that is required for calculating `checkBalance`. By removing  `_getTotalPTokens` function: 
   -  the withdrawal function has a different strategy of figuring out how much 3CRV tokens it required. It uses a threshold amount instead of total available LP tokens. This could be problematic in case where the metapool doesn't have the (current 5%) threshold of LP tokens available and we would want to withdraw all. But in that case using `withdrawAll` function is much more convenient and gas effective.
   - it is now a responsibility of each contract extending base Curve to implement `checkBalance`. And to incorporate all the various tokens deposited / locked. 

[This issue](https://github.com/OriginProtocol/origin-dollar/issues/1010) encapsulates the liquidity research, which gets to the bottom of the question "what is the best strategy to add / remove liquidity from the metapool". Results in [this spreadsheet](https://docs.google.com/spreadsheets/d/1DYSyYwHqxRzSJh9dYkY5kcgP_K5gku6N2mQVhoH33vY/edit#gid=0). 

**Convex Generalized Metastartegy**
Merged from the [original PR](https://github.com/OriginProtocol/origin-dollar/pull/1066). It features: 
- creates a common abstract strategy that combines strategies interacting with Curve metapools
- introduces a metapool strategy that can be configured to harvest any Curve metapool
- added more brownie script functionality to test the behaviour in mainnet fork

**Testing on fork:**

Brownie and @DanielVF 's `world` script is your friend here. Also some additional functionality has been added to those helper scripts to more easily interact with Metapool strategy. 

**Setup:**
- set BLOCK_NUMBER in `contracts/.env` to smth you like. I was using `14655959` where metapool is tilted more heavily towards 3CRV (25,238,413 OUSD  28,243,182 3CRV). You can verify this by running `show_metapool_balances()` from brownie console. Run the node in forked mode `yarn run node`. 
- Metastrategy gets freshly deployed on startup.  Look for `META STRATEGY ADDRESS 0x.....` in the node output. And set that address in the top of `brownie/world.py`. If you don't change the block number that contract address doesn't change on node re-run. Once metastrategy is deployed to production that address shall not change. 
- in another terminal go to `/brownie` and run the console `brownie console --network hardhat`:
  - within the console starting always import `from metastrategy import *`. This will fetch 3CRV & OUSD and set all the helper functions to the environment. 
 
**Brownie functions:**
- check `brownie/metapool.py` and possibly `brownie/world.py` for the available functions
- explore scripts under `brownie/metapool/*.py` for ideas on how/what to test

**Open questions:**
- is it ok that we always mint 1 OUSD to 1 3CRV (converted to 3CRV dollar values). Should we make it more effective depending on how the pool is tilted?
- Is there a possible flash loan exploit of manipulating curve pool / minting / redeeming using our Vault?

**Watch out before merging:**
- set `forceDeploy` to false before merging the deploy script


DanielVF (changes requested):

contracts/contracts/vault/VaultCore.sol
```diff
@@ -315,6 +377,8 @@ contract VaultCore is VaultStorage {
 
         // Only rachet OUSD supply upwards
         ousdSupply = oUSD.totalSupply(); // Final check should use latest value
+        console.log("OUSD SUPPLY", ousdSupply / 10**18);
```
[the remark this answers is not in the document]
sparrowDom: yup, removed

contracts/contracts/vault/VaultStorage.sol
```diff
@@ -107,6 +108,9 @@ contract VaultStorage is Initializable, Governable {
     // Deprecated: Tokens that should be swapped for stablecoins
     address[] private _deprecated_swapTokens;
 
+    // Meta strategy that is allowed to mint/burn OUSD without changing collateral
```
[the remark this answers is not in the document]
sparrowDom: sounds good 👍

contracts/contracts/utils/InitializableSecondary.sol
```diff
@@ -0,0 +1,38 @@
+// SPDX-License-Identifier: agpl-3.0
+pragma solidity ^0.8.0;
+
+abstract contract InitializableSecondary {
```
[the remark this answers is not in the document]
sparrowDom: Looks like I had a brain fart moment here... Have just put a lot of array type parameters inside an array to circumvent the limitation here.

contracts/contracts/vault/VaultCore.sol
```diff
@@ -127,16 +161,16 @@ contract VaultCore is VaultStorage {
...
-            // Allow a max difference of maxSupplyDiff% between
-            // backing assets value and OUSD total supply
-            uint256 diff = _totalSupply.divPrecisely(_backingValue);
-            require(
-                (diff > 1e18 ? diff.sub(1e18) : uint256(1e18).sub(diff)) <=
-                    maxSupplyDiff,
-                "Backing supply liquidity error"
-            );
-        }
+        // if (maxSupplyDiff > 0) {
```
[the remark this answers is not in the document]
sparrowDom: had some brownie console stuff failing becuase of this so temporarily commented it out. Can not reproduce ATM which is great... so un-commenting it

contracts/contracts/vault/VaultCore.sol
```diff
@@ -96,6 +105,31 @@ contract VaultCore is VaultStorage {
...
+     *
+     * Notice: can't use nonReentrant modifier since BaseCurveStrategy's deposit
+     * already has that modifier present
+     */
+    function mintForStrategy(uint256 _amount)
+        external
+        whenNotCapitalPaused
+        onlyOusdMetaStrategy
+    {
+        require(_amount > 0, "Amount must be greater than 0");
```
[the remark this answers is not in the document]
sparrowDom: One safeguard that comes to mind is the ratio of `net_OUSD_in_circulation/dollar_value_of_gauge_OUSD3CRVLP` In most cases I would expect that value to be close to 0.5. Adding a configurable safeguard so that the ratio doesn't go beyond - lets say 0.7 would make sense to me.

contracts/contracts/strategies/ConvexMetaStrategy.sol
```diff
@@ -0,0 +1,343 @@
...
+        vault.mintForStrategy(threePoolLpDollarValue);
+
+        uint256 ousdBalance = ousd.balanceOf(address(this));
+        uint256[2] memory _amounts = [ousdBalance, threePoolLpBalance];
+
+        // TODO: figure out what the best slippage guard is. Also minReceived is in
+        // OUSD3Pool LP tokens so need to account for that
+        //
+        // Important(!) we need to be sure there are no flash loan attack possibilities here
+        uint256 minReceived = ((ousdBalance + threePoolLpBalance) * 985) / 1000;
```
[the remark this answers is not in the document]
sparrowDom: I think there are 2 separate issues here: 
1. bad math around max slippage check
2. what ratio of OUSD to 3crvLP to mint when depositing. 

I have corrected the first issue by converting OUSD & 3LP being deposited to dollar value and then converting that to metapoolLP value and applying 1% max slippage [as we do here](https://github.com/OriginProtocol/origin-dollar/blob/master/contracts/contracts/strategies/BaseCurveStrategy.sol#L52). 

As for the second issue one of the things we can check is how the metapool is tilted before we are depositing funds. We can add some safeguards here, but I need to research first which tilts are potentially dangerous to us. Is there any other knowledge you think we should be cognisant of here that could prove dangerous?

sparrowDom:
@franckc added documentation on how to test metastrategies in forked mode under `brownie/Readme.md` and also in the ticket description under `Brownie functions:`. In case you find some documentation info lacking do tell

contracts/contracts/strategies/ConvexOUSDMetaStrategy.sol
```diff
@@ -0,0 +1,221 @@
...
+import { IRewardStaking } from "./IRewardStaking.sol";
+import { IConvexDeposits } from "./IConvexDeposits.sol";
+import { ICurvePool } from "./ICurvePool.sol";
+import { IERC20 } from "./BaseCurveStrategy.sol";
+import { BaseConvexMetaStrategy } from "./BaseConvexMetaStrategy.sol";
+import { StableMath } from "../utils/StableMath.sol";
+import { IVault } from "../interfaces/IVault.sol";
+
+contract ConvexOUSDMetaStrategy is BaseConvexMetaStrategy {
+    using StableMath for uint256;
```
sparrowDom, line 21: We don't need this one anymore right @DanielVF ?

DanielVF: I think "StableMath" is different from "SafeMath". I think we use this on for division. We look hard at it and see what it provides us.

contracts/contracts/vault/VaultAdmin.sol
```diff
@@ -133,6 +133,19 @@ contract VaultAdmin is VaultStorage {
         assetDefaultStrategies[_asset] = _strategy;
     }
 
+    /**
+     * @dev Set a minimum amount of OUSD in a mint or redeem that triggers a
```
DanielVF, line 137: Comments don't match here.

contracts/contracts/vault/VaultCore.sol
```diff
@@ -13,12 +13,14 @@ pragma solidity ^0.8.0;
...
 import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
 import { SafeMath } from "@openzeppelin/contracts/utils/math/SafeMath.sol";
+import "@openzeppelin/contracts/utils/Strings.sol";
 
 import { StableMath } from "../utils/StableMath.sol";
 import { IOracle } from "../interfaces/IOracle.sol";
 import { IVault } from "../interfaces/IVault.sol";
 import { IBuyback } from "../interfaces/IBuyback.sol";
 import "./VaultStorage.sol";
+import "hardhat/console.sol";
```
DanielVF, line 23: We need to remove both this and the Strings.sol

contracts/contracts/vault/VaultCore.sol
```diff
@@ -97,6 +107,46 @@ contract VaultCore is VaultStorage {
...
+            _rebase();
+        }
+
+        netOusdMintedForStrategy += int256(_amount);
+
+        require(
+            abs(netOusdMintedForStrategy) < netOusdMintForStrategyThreshold,
+            string(
+                bytes.concat(
+                    bytes("Attempting to mint too much OUSD. Threshold: "),
```
DanielVF, line 138: We should probably save contract size and not build these errors with numbers in them.

contracts/contracts/vault/VaultCore.sol
```diff
@@ -97,6 +107,46 @@ contract VaultCore is VaultStorage {
...
+
+        emit Mint(msg.sender, _amount);
+
+        // Rebase must happen before any transfers occur.
+        // TODO: double check the relevance of this
+        if (_amount >= rebaseThreshold && !rebasePaused) {
+            _rebase();
+        }
+
+        netOusdMintedForStrategy += int256(_amount);
```
DanielVF, line 132: Is this an unchecked conversion to int256()? If so, you might be able to pass in a very large mint of OUSD and have it get converted to a very small negative number, passing the require.

contracts/contracts/vault/VaultCore.sol
```diff
@@ -188,6 +238,47 @@ contract VaultCore is VaultStorage {
...
     }
 
+    /**
+     * @dev Burn OUSD for OUSD Meta Strategy
+     * @param _amount Amount of OUSD to burn
+     *
+     * Notice: can't use nonReentrant modifier since BaseCurveStrategy's deposit
+     * already has that modifier present
+     */
+    function redeemForStrategy(uint256 _amount)
```
DanielVF, line 248: This function might be more of a "burn" than a "redeem"? There's nothing coming back from doing it, name might be confusing.

contracts/contracts/strategies/ConvexOUSDMetaStrategy.sol
sparrowDom: Oh right we need it for things like `mulTruncate`. Thx

contracts/contracts/vault/VaultAdmin.sol
sparrowDom: tnx fixed

contracts/contracts/vault/VaultCore.sol
sparrowDom: 🙃 removed

contracts/contracts/vault/VaultCore.sol
sparrowDom: Cool, removed. Influence of UniswapV2Pair Tech makes it even more compelling to compact the code.

contracts/contracts/vault/VaultCore.sol
sparrowDom: good point, renaming

contracts/contracts/vault/VaultCore.sol
sparrowDom: Have done a check that amount value isn't too high. Will also add a test for that

sparrowDom: test added here: https://github.com/OriginProtocol/origin-dollar/pull/1002/commits/976868115c511f5a510f0a6a8ecfb4aed9f2e3c3

sparrowDom:
@DanielVF when you have a chance can you eyeball this commit. The code should do the same thing and should be simplified: https://github.com/OriginProtocol/origin-dollar/pull/1002/commits/e0063ebb8c9b8c6132d1c00e4ce358e12cfb1293

contracts/contracts/strategies/ConvexGeneralizedMetaStrategy.sol
```diff
@@ -0,0 +1,153 @@
...
+     * @param num3CrvTokens Number of Convex LP tokens to remove from gauge
+     */
+    function _lpWithdraw(uint256 num3CrvTokens) internal override {
+        ICurvePool curvePool = ICurvePool(platformAddress);
+
+        uint256 gaugeTokens = IRewardStaking(cvxRewardStakerAddress).balanceOf(
+            address(this)
+        );
+        /**
+         * Convert 3crv tokens to metapoolLP tokens and double it. Doubling is required because aside
```
DanielVF, line 74: I think these comments aren't correct for this file?

contracts/contracts/strategies/ConvexGeneralizedMetaStrategy.sol
```diff
@@ -0,0 +1,153 @@
...
+            gaugeTokens,
+            true
+        );
+
+        uint128 metapool3CrvCoinIndex = _getMetapoolCoinIndex(
+            address(pTokenAddress)
+        );
+        // always withdraw all of the available metapool LP tokens (similar to how we always deposit all)
+        // slither-disable-next-line unused-return
+        metapool.remove_liquidity_one_coin(
```
DanielVF, line 147: This WithdrawAll method is super dangerous. There's no check on the minimum amount returned, which means that a MEV sandwich on a low liquidly pool could lose most of our funds, unless we wrapped the transaction in something that checks the funds after the withdrawAll(). 

Note that our existing curve / convex strategies don't have this same risk on withdrawAll(), because everything they are withdrawing is a mix of stablecoins that we can return to users. If an attacker bends the pool, we might not get the coins we were expecting, but we would be getting more coins total and a positive rebase. But here, we could purely lose big money.

A further danger is that this kind of MEV sandwich would be totally automated, and not even require anyone to know about this contract.

Even without a MEV attack, we could take arbitrary slippage losses from low pool liquidity with this method. If we ended up having  2/3 of a pools liquidity, us leaving would be _very_ painful here.

In the past, we've regarded withdrawAll() on our strategies as a reasonably safe step to take take in an emergency. Now here it's the most dangerous method on any strategy.

(I think all the above may apply to the regular _lpWithdraw method as well.)

### Options for fixing it

1. We revert if there is more than 1% slippage on deposits, however this is unacceptable on withdraws I think because if something depegs to 97%, we may need to just take the loss and get out.

2. Pass in a required minimum swap result amounts. Maybe two of them. This is might the best long term solution. However, our vault and our vault to strategy API's don't currently support this. We would need to alter all strategies.

3. Add code to the vault withdraw and reallocate methods to total the Vault before and after the withdraw, to check against a passed in minimum loss amount. This has the advantage of not altering the vault to strategy API's, and the disadvantage of adding 400K gas to each inner strategist action that touches withdraws. Still, the only big disadvantage is gas.

4. Up our ops game another level on strategist actions and emergency response. Make sure we always run these methods wrapped with checkLosses. (We have always been doing this for curve related transactions, this just ups the stakes a lot on getting it wrong)

Open for other thoughts too.

contracts/contracts/strategies/BaseConvexMetaStrategy.sol
```diff
@@ -0,0 +1,180 @@
...
+     */
+    function initialize(
+        address[] calldata _rewardTokenAddresses, // CRV + CVX
+        address[] calldata _assets,
+        address[] calldata _pTokens,
+        /**
+         * in the following order: _platformAddress(3Pool address), _vaultAddress, _cvxDepositorAddress
+         * _metapoolAddress, _metapoolMainToken, _cvxRewardStakerAddress, _metapoolLPToken
+         */
+        address[] calldata _initAddresses,
```
DanielVF, line 57: Was using individual arguments hitting the stack too deep error?

I'm wondering about the tradeoffs here between an array vs a struct for `_initAddresses`. Seems a little prone to confusion to use an array with each element being a completely different kind of thing, and the only visible difference being at the documentation level.

contracts/contracts/strategies/ConvexStrategy.sol
```diff
@@ -78,37 +79,27 @@ contract ConvexStrategy is BaseCurveStrategy {
         require(success, "Failed to deposit to Convex");
     }
 
-    function _lpWithdraw(uint256 numPTokens) internal override {
+    function _lpWithdraw(uint256 numCrvTokens) internal override {
+        uint256 gaugePTokens = IRewardStaking(cvxRewardStakerAddress).balanceOf(
+            address(this)
+        );
+
+        // Not enough in this contract or in the Gauge, can't proceed
```
DanielVF, line 87: Why do we do this additional check? If we don't have enough tokens, the withdraw will revert regardless?

DanielVF:
One of the biggest lifetime gas costs of each non-default strategy is the cost to determine the balance of holdings. This is because moving funds in or out might be a once a month thing, while rebasing happens at least daily, and during mints and during redeems, and during allocates, and even lots of times during funds reallocations. So from a gas point of view, it's all about the cost of checking a strat's balance .

This becomes more important as we look to move to larger numbers of strategies holding smaller amounts of funds.

I think we we could reduce the complexity and gas usage of checking the balance by ensuring that we deposits and withdraws always result in a clean state in which funds are completely stored back in Convex at the end. This will mean that we can drop all the external calls, logic, and math around working with intermediate level tokens both when balance checking and when withdrawing.

contracts/contracts/strategies/BaseConvexMetaStrategy.sol
```diff
@@ -0,0 +1,180 @@
...
+    function checkBalance(address _asset)
+        public
+        view
+        virtual
+        override
+        returns (uint256 balance)
+    {
+        require(assetToPToken[_asset] != address(0), "Unsupported asset");
+        balance = 0;
+        uint256 assetDecimals = Helpers.getDecimals(_asset);
```
DanielVF, line 100: Super tiny: This isn't used until the very last line of the function. Would be clearer and less mental overhead to move it to the second last line.

contracts/contracts/strategies/ConvexOUSDMetaStrategy.sol
```diff
@@ -0,0 +1,222 @@
...
+         * and apply the max slippage
+         */
+        uint256 minReceived = (ousdBalance + threePoolLpDollarValue)
+            .divPrecisely(metapoolVirtualPrice)
+            .mulTruncate(uint256(1e18) - maxSlippage);
+
+        // slither-disable-next-line unused-return
+        metapool.add_liquidity(_amounts, minReceived);
+
+        uint256 metapoolLp = metapoolErc20.balanceOf(address(this));
```
DanielVF, line 80: Maybe a comment here that we are depositing our whole balance into convex, not just what added, which has the effect of also depositing anything that the strategy received through other means.

contracts/contracts/strategies/ConvexOUSDMetaStrategy.sol
```diff
@@ -0,0 +1,222 @@
...
+
```
DanielVF, line 81: Super tiny: The balanceOf and the deposit line are tightly linked since the amount is used in the deposit and never used elsewhere. This means that we probably don't need a new line separating these two lines.

contracts/contracts/strategies/ConvexOUSDMetaStrategy.sol
```diff
@@ -0,0 +1,222 @@
...
+        IERC20 metapoolErc20 = IERC20(address(metapool));
+        ICurvePool curvePool = ICurvePool(platformAddress);
+
+        uint256 threePoolLpBalance = threePoolLp.balanceOf(address(this));
+        uint256 curve3PoolVirtualPrice = curvePool.get_virtual_price();
+        uint256 threePoolLpDollarValue = threePoolLpBalance.mulTruncate(
+            curve3PoolVirtualPrice
+        );
+
+        uint128 crvCoinIndex = _getMetapoolCoinIndex(pTokenAddress);
```
DanielVF, line 38: We use so many addresses and index in these strategies. Feels like it might a huge gas savings if we set them at constructor time and used immutable values. Might save a lot of storage reads.

contracts/contracts/strategies/ConvexGeneralizedMetaStrategy.sol
sparrowDom: ah yes thanks, corrected the comment

contracts/contracts/strategies/ConvexGeneralizedMetaStrategy.sol
sparrowDom: Oh my thanks for catching this, and yes you make a good point. Thinking about the options you lay out: 

1. As you say this one doesn't cover all the use cases
2. This is one thing that would definitely work, though with the downside you point out, that the API changes and we need to re-deploy almost everything. 
3. I like this one a a lot. 
4. Would work though might be risky, especially since is something we need to manually keep in mind. 

I am leaning towards option 3. Though I have another proposal: 

5. Have a maxWithdrawalSlippage that is set to 1% at initialisation. And that rate is calculated when withdrawing liquidity. Also have a public setter, so that amount can be changed. This way no API would require changing and not much additional gas should be used. (too bad we forgot to add a `int256[99] private _reserved;` to BaseCurveStrategy, since it would fit nicely in there.

contracts/contracts/strategies/ConvexOUSDMetaStrategy.sol
```diff
@@ -0,0 +1,223 @@
...
+        uint128 ousdCoinIndex = _getMetapoolCoinIndex(
+            address(metapoolMainToken)
+        );
+
+        uint256 ousdToAdd = toPositive(
```
DanielVF, line 43: When OUSD is low in the pool, this method can lose up to 0.04% of minted OUSD.

This is because curve is designed to have a large area where prices are near 1 to 1. As add catchup OUSD into this area, we end up not making a profit, but still paying a swap fee because our deposit is not balanced.

This can still happen on the other end of the situation as well. If the pool was balanced with a little too much 3pool, and we were putting more three pool tokens in, the ideal strategy probably would not be to put no OUSD in.

DanielVF: Brownie console test run to show issue:

```python
from world import *
ousd_metapool = load_contract("ousd_metapool", OUSD_3CRVF_POOL)
threepool_virtual_price = threepool.get_virtual_price()

# NONFREN = '0x6c3F90f043a72FA612cbac8115EE7e52BDe6E490'
CONVEX = '0x25f0cE4E2F8dbA112D9b115710AC297F816087CD'


def pool_balances():
	print("%s 3pool, %s OUSD" % (
		c18(ousd_metapool.balances(1) * threepool_virtual_price // int(1e18)),
		c18(ousd_metapool.balances(0))
		)
	)

def amount_check():
	with TemporaryFork():
		ousd_metapool.remove_liquidity(ousd_metapool.balanceOf(flipper),[0,0], {'from': flipper})
		return history[-1].events['Transfer'][1]['_value'] * threepool_virtual_price // int(1e18) + history[-1].events['Transfer'][0]['value']

ousd.approve(ousd_metapool, 1e70, {'from': flipper})

amount = amount_check()
for i in range(0,10):
	ousd_metapool.add_liquidity([100000*1e18,0], 0, {'from':flipper}) 	
	new = amount_check()
	print(c18((new - amount) * 1000))
	amount = new


ousd_metapool.remove_liquidity_one_coin(1*1e6*1e18, 0, 0, governor, {'from': CONVEX})
ousd_metapool.remove_liquidity_one_coin(1*1e6*1e18, 0, 0, governor, {'from': CONVEX})
ousd_metapool.remove_liquidity_one_coin(1*1e6*1e18, 0, 0, governor, {'from': CONVEX})
ousd_metapool.remove_liquidity_one_coin(1*1e6*1e18, 0, 0, governor, {'from': CONVEX})



amount = amount_check()
for i in range(0,100):
	ousd_metapool.add_liquidity([10000*1e18,0], 0, {'from':flipper}) 	
	new = amount_check()
	print(c18(new - amount))
	amount = new
	pool_balances()
```

DanielVF: I guess how this method works comes to which we prioritize more, peg holding, trading gain/loss, or long term APY. The current method is pure peg holding. Makes money if things are wildly out of peg, looses money if near peg, and does not prioritize APY/liquidity on one side (while it does on the other)

sparrowDom: Some time ago we have been discussing that we would want to maintain the metapool as balanced as possible. As in to provide stability to the pool. If we are to optimise for APY it would be better to print more OUSD to result in maximum amount of LP tokens. Similar things have been figured out in [this research](https://docs.google.com/spreadsheets/d/1DYSyYwHqxRzSJh9dYkY5kcgP_K5gku6N2mQVhoH33vY/edit#gid=0).

One way to modify the strategy to optimise APY (and consequently amount of LP deployed to Convex) would be to:
- always print at least the same amount of OUSD as counterpart 3crv
- if metapool low on OUSD (higher in 3pool) print so much OUSD that it is 50/50

contracts/contracts/strategies/ConvexOUSDMetaStrategy.sol
sparrowDom: Great idea, Done

contracts/contracts/strategies/ConvexOUSDMetaStrategy.sol
sparrowDom: 👍

contracts/contracts/strategies/ConvexOUSDMetaStrategy.sol
sparrowDom: added

contracts/contracts/strategies/BaseConvexMetaStrategy.sol
sparrowDom: 👍 moved

sparrowDom:
Regarding the gas optimisation I've made a proposal for solution [here](https://github.com/OriginProtocol/origin-dollar/pull/1122). My biggest concern is getting generalizedConvexStrategy [_lpWithdrawal calculation wrong](https://github.com/OriginProtocol/origin-dollar/pull/1122/files#diff-b6689626806597cf7b5dbdfd270ee14ab9e9ca194115f1031faad22610b207e8R134) and withdrawing too much 3Crv. If the math on how much 3crv to withdraw isn't super solid, those funds will be hidden to `checkBalance` function. 

Since this is more of a gas optimisation and not a security risk I would propose to mark this as a `P1` and do it in the near future.

DanielVF:
Yes, we can skip the gas optimization

contracts/contracts/strategies/ConvexStrategy.sol
sparrowDom: This is just to get a nicer error message instead of the "reverted without a reason" message. You think it is better to optimise for gas here?

contracts/contracts/strategies/ConvexGeneralizedMetaStrategy.sol
```diff
@@ -0,0 +1,152 @@
...
+
+        require(success, "Failed to deposit to Convex");
+    }
+
+    /**
+     * Withdraw the specified amount of tokens from the gauge. And use all the resulting tokens
+     * to remove liquidity from metapool
+     * @param num3CrvTokens Number of Convex LP tokens to remove from gauge
+     */
+    function _lpWithdraw(uint256 num3CrvTokens) internal override {
```
DanielVF, line 67: It's possible the math in here could get much simpler, and the coin transfers more efficient.

1) By using `remove_liquidity_imbalance` instead of `remove_liquidity_one_coin`, we can specify what we want out in terms of output tokens, not in input tokens. This allows us to get precisely the amount out that we want, rather than a big percentage extra.

2) In order to withdraw from the convex pool, instead of guessing with a huge factor, we be within two hundredths of a percent or so by using `calc_token_amount` then multiplying by the worst case `fee()` from the strategy. We'll get slightly less fee then this, but it will still be very close.

sparrowDom:
Things done: 
- make initialisation a struct: https://github.com/OriginProtocol/origin-dollar/pull/1002/commits/053590b8049e896dbe13baf43fc0374087bf9f72
- MEV guard PR with tests: https://github.com/OriginProtocol/origin-dollar/pull/1123
- print at least the same amount OUSD as 3crv. And at max twice the OUSD: https://github.com/OriginProtocol/origin-dollar/pull/1002/commits/8925168a0c6b6cc90b0d014cd22fc7fa8cba6657


Stuff to do :
- Change the way we calculate how to withdraw liquidity

contracts/contracts/strategies/ConvexOUSDMetaStrategy.sol
```diff
@@ -45,6 +45,13 @@ contract ConvexOUSDMetaStrategy is BaseConvexMetaStrategy {
                 int256(threePoolLpDollarValue)
         );
 
+        /* Add so much OUSD so that the pool ends up being balanced. But at minimum
+         * add at least the same amount of OUSD as 3poolLP and at maximum at twice as
```
franckc, line 49: super nit. typo "... and at a maximum add twice as much OUSD."

sparrowDom: better yes, tnx

sparrowDom:
@DanielVF @franckc Simplification of the removal of liquidity logic: 
https://github.com/OriginProtocol/origin-dollar/pull/1002/commits/b553302165994639ccd4db5b25ae3b5ebbe338c0

contracts/contracts/strategies/ConvexOUSDMetaStrategy.sol
```diff
@@ -0,0 +1,176 @@
...
+
+        require(success, "Failed to deposit to Convex");
+    }
+
+    /**
+     * Withdraw the specified amount of tokens from the gauge. And use all the resulting tokens
+     * to remove liquidity from metapool
+     * @param num3CrvTokens Number of Convex LP tokens to remove from gauge
+     */
+    function _lpWithdraw(uint256 num3CrvTokens) internal override {
```
DanielVF, line 98: I think this withdraw can still get a bit simpler.


```python
needed = (num3CrvTokens + 1) * metapool.totalSupply()  // metapool.balances(1)

convex.withdrawAndUnwrap(needed, true);
ousd_metapool.remove_liquidity(needed, [0,0])
ousd.burn(ousd.balanceOf(address(this))
```


Since there are no fees when removing with remove_liquidity, we don't have to worry about fees.

Since remove_liquidity removes both sides in the proportion between pool balances, lp tokens burnt, and total supply, we can perfectly calculate how much LP tokens we need to use (outside of solidity rounding error.). Add a one and we don't have to worry about rounding. (Check this though.)

Because we are withdrawing both coins, we will have OUSD unless we are withdrawing less than a couple raw LP tokens. We might as well just keep the code clean and always burn.

sparrowDom: Ran this code 
```python
#############
## setup metapool
from metastrategy import *
threepool_lp.approve(ousd_metapool.address, 0, {"from": me})
threepool_lp.approve(ousd_metapool.address, 1e50, {"from": me})
ousd_metapool.add_liquidity([0, 1e24], 0, {"from":me}) #add 1mio liquidity

num3CrvTokens = 1e23
needed = (num3CrvTokens + 1) * ousd_metapool.totalSupply() // ousd_metapool.balances(1)
user_3crv_balance = threepool_lp.balanceOf(me)
ousd_metapool.remove_liquidity(needed, [0,0], {"from":me})
user_3crv_diff = threepool_lp.balanceOf(me) - user_3crv_balance
user_3crv_diff
```
and it outputs 99999999999999990439768

Good news is that like you say fee is not included in liquidity removal.  Bad news is there seems to be slightly more calculation error than expected. I am also slightly worried that different pool implementations might have different error rates. I would suggest to add a ful extra coin to required 3poolRemoval
```
needed = (num3CrvTokens + 1e18) * ousd_metapool.totalSupply() // ousd_metapool.balances(1)
```
This should give plenty of room for rounding errors and shouldn't cause any serious problems with funds being left behind on contracts

sparrowDom: Oh ignore the rambling regarding different pool implementations. Looks like an extra 1e7 of 3crvLP should make up for the calculation error. For safe measure rather adding 1e10 of extra 3crv LP

sparrowDom: thanks @DanielVF great suggestions. Have further simplified the code see this commit: https://github.com/OriginProtocol/origin-dollar/pull/1002/commits/61f506beffcd3e13a3cc0586dd60e310b2e5a03c

sparrowDom: Re-ran the above test code with depositing 1e26 tokens and withdrawing 1e25. The output of `user_3crv_diff` was: 
10000000000000000554802370

Interesting that the calculation sometimes overshoots and sometimes undershoots.

sparrowDom: I wonder if 1e10 is enough of a margin for error.

DanielVF: Another calculation approach, using a higher precision division:

```python
k = int(1e36) * total_supply // balance
diff = (balance) * k - (balance - amount - 1) * k
to_burn = diff // int(1e36)
```

Haven't thought it all the way through though. Works with what I've played with.

contracts/utils/generalizedConvexStratDeployment.js
```diff
@@ -0,0 +1,179 @@
...
+      const dConvexTokenMetaStrategyProxy = await deployWithConfirmation(
+        proxyName
+      );
+      const cConvexTokenMetaStrategyProxy = await ethers.getContractAt(
+        proxyName,
+        dConvexTokenMetaStrategyProxy.address
+      );
+
+      // 2. Deploy new implementation
+      const dConvexTokenMetaStrategyImpl = await deployWithConfirmation(
```
DanielVF, line 56: I don't know that we need to deploy a new implementation for each strategy? Can we just reuse the same one?

DanielVF:
# Code Review

#### Requirements

This PR introduces new a new OUSD pool AMO, and generic 3CRV single-sided curve pool strategy. It also refactors the current curve/convex related code, and updates the previous 3pool convex and curve strategies.

🟠 The singled sided curve strategies have no guarantee that we can withdraw what we put in. The strategies are much more like trading - timing and prices matter. These strategies can result in **considerable losses** if a coin goes below peg and or liquidity is reduced. Because of the nature of curve's bonding curve, **our actual withdraw losses can be multiples higher than the apparent pool price peg loss**.

❌ The hardcoded slippage restrictions in these contracts means that in an increasing peg loss situation we may not be able to withdraw funds for the days that it takes us to deploy new contracts and go though the governance process.

🟠 The OUSD AMO pool may loose small amounts of money each deposit as deposits are focused on balancing the pool over profitability.

🟠 Once used at scale, these strategies will expect us to have a roughly equal amount of each stablecoins available for deposit/withdrawal in our default strategies. Otherwise redeems will drain higher percentages of the lowest funded default strategy.

❌ In use the OUSD AMO is likely to mostly operate in one direction, with either long periods of mostly minting OUSD, or long periods of mostly redeeming OUSD. Because of this, governance will need keep extending the netOusdMintForStrategyThreshold. However, this high netOusdMintForStrategyThreshold will mean that that while it might only allow a small change in one direction in the usual way, it could allow a truly massive change back the other way. For example, if it had mostly been used for redeems, and had done a net 105 million burns so far, we might want to set the limit to 115 million to enable 10 million more burns. However, this would mean that it could also be permitted to mint 220 million OUSD. I think the solution to this is that whenever governance changes the limit, we should reset the `netOusdMintForStrategyThreshold` back to zero. This will mean that the new limit based on where things currently are.

- [ ] Will need analytics site updated for new strategies, both crawling and UI.
- [ ] Must have alerting of peg loss thresholds or going below liquidity thresholds.
- [ ] Will need to update docs

#### Internal State

_All internal strategy contract state are configuration variables._

#### Failure Consequences

In the mid-term these strategies may hold almost all backing funds for OUSD, and thus the impact of failure could be massive.

The OUSD contract can mint OUSD. This could cause a loss of funds if these were not equaled out with a corresponding gain in assets.


#### Attack

_What conditions could cause this code to fail if they were not true_

➟ _In progress_

MEV attacks on pools.

Peg loss.

OUSD cycle.


#### Logic

➟ _In progress_


#### Tests

➟ _In progress_

#### Flavor

➟ _In progress_

_Could this code be simpler?_

#### Overflow

- [x] Uses v8
- [ ] Check that all for loops use uint256
  - ❌ See discussion on `_getMetapoolCoinIndex`. It won't overflow, but is less gas efficient.

#### Black magic

#### Authentication

- [x] Never use tx.origin

#### Cryptographic code

-[x] _No Cryptographic code_

#### Gas problems

-[x] _All for loops are over lists under admin control, and of small size. All contracts are upgradable._

#### External calls

- [x] Reentrancy guards on all state changing functions (_no publicly callable state changing functions_.)
- [x] Malicious behaviors (_we are only using trusted contracts._)
- [ ] No slippage attacks (we need to validate expected tokens received) 🟠 _Inherent to the design of these curve and 3pool contracts are risks here._
- [ ] Oracles, one of: 🟠 _Inherent to the design of these curve and 3pool contracts are pricing risks._
  - [ ] No oracles
  - [ ] Oracles can't be bent


#### Ethereum

_Contract does not handle ethereum._

contracts/contracts/strategies/BaseCurveStrategy.sol
```diff
@@ -108,53 +117,71 @@ abstract contract BaseCurveStrategy is InitializableAbstractStrategy {
...
+         * Add a 5% threshold to help calculate required amount of crv3Tokens:
+         *  - convert asset to 18 decimals
+         *  - divide by virtual price to get 3Crv equivalent
+         *  - add a 5% threshold to it
+         */
+        // slither-disable-next-line divide-before-multiply
+        uint256 crv3TokensTreshold = (((_amount.scaleBy(
+            18,
+            Helpers.getDecimals(_asset)
+        ) * 105) * 1e18) / virtual_price) / 100;
```
DanielVF, line 143: Will something bad happen here because of this 105% and a peg loss on 3crv?

DanielVF: This 105 is basically a somewhat non-intuitive hardcoded slippage limit. We should probably use our normal 3pool slippage limit here.

contracts/contracts/strategies/BaseConvexMetaStrategy.sol
```diff
@@ -0,0 +1,218 @@
...
+abstract contract BaseConvexMetaStrategy is BaseCurveStrategy {
+    using StableMath for uint256;
+    using SafeERC20 for IERC20;
+    event MaxWithdrawalSlippageUpdated(
+        uint256 _prevMaxSlippagePercentage,
+        uint256 _newMaxSlippagePercentage
+    );
+
+    // used to circumvent the stack too deep issue
+    struct InitState {
```
DanielVF, line 27: `InitConfig` is probably a better name for this than `InitState`.

contracts/contracts/strategies/BaseConvexMetaStrategy.sol
```diff
@@ -0,0 +1,218 @@
...
+        metapoolMainToken.safeApprove(address(metapool), type(uint256).max);
+    }
+
+    /**
+     * @dev Get the index of the coin
+     */
+    function _getMetapoolCoinIndex(address _asset)
+        internal
+        view
+        returns (uint128)
```
DanielVF, line 165: If this returns an int128, then all the places that use this function will not need to do a cast on the result.

contracts/contracts/strategies/BaseConvexMetaStrategy.sol
```diff
@@ -0,0 +1,218 @@
...
+     * IMPORTANT Minimum maxWithdrawalSlippage should actually be 0.1% (1e15)
+     * for production usage. Contract allows as low value as 0% for confirming
+     * correct behavior in test suite.
+     */
+    function setMaxWithdrawalSlippage(uint256 _maxWithdrawalSlippage)
+        external
+        onlyVaultOrGovernorOrStrategist
+    {
+        require(
+            _maxWithdrawalSlippage <= 1e17,
```
DanielVF, line 188: I'm concerned that require this will result in us being stuck in the pool in case of a depeg.

contracts/contracts/strategies/BaseConvexMetaStrategy.sol
```diff
@@ -0,0 +1,218 @@
...
+    {
+        // Collect CRV and CVX
+        IRewardStaking(cvxRewardStakerAddress).getReward();
+        _collectRewardTokens();
+    }
+
+    /**
+     * @dev If x a negative number return 0 else return x
+     */
+    function toPositive(int256 x) internal pure returns (uint256) {
```
DanielVF, line 215: I don't think this is a clear name, because there's many possible ways to convert a number to a positive number (The most obvious being to flip the sign.)

contracts/contracts/strategies/ConvexGeneralizedMetaStrategy.sol
```diff
@@ -0,0 +1,155 @@
...
+            .divPrecisely(metapool.get_virtual_price())
+            .mulTruncate(num3CrvTokens);
+
+        int128 metapool3CrvCoinIndex = int128(
+            _getMetapoolCoinIndex(address(pTokenAddress))
+        );
+
+        // add 10% margin to the calculation of required tokens
+        // slither-disable-next-line divide-before-multiply
+        uint256 estimatedMetapoolLPWithMargin = (estimationRequiredMetapoolLpTokens *
```
DanielVF, line 86: I'm concerned that this is in effect a hardcoded slippage check. Perhaps we should be using something that runs off our metapool slippage config.

contracts/contracts/strategies/ConvexGeneralizedMetaStrategy.sol
```diff
@@ -0,0 +1,155 @@
...
+
+        uint256 metapoolVirtualPrice = metapool.get_virtual_price();
+        /**
+         * First convert all the deposited tokens to dollar values,
+         * then divide by virtual price to convert to metapool LP tokens
+         * and apply the max slippage
+         */
+        uint256 minReceived = threePoolLpDollarValue
+            .divPrecisely(metapoolVirtualPrice)
+            .mulTruncate(uint256(1e18) - MAX_SLIPPAGE);
```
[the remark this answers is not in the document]
sparrowDom: Good point fixed here: https://github.com/OriginProtocol/origin-dollar/pull/1002/commits/ab1625fd5fb43c113642802f6c3e799ce4b49c3b

contracts/utils/generalizedConvexStratDeployment.js
sparrowDom: good point fixed here: https://github.com/OriginProtocol/origin-dollar/pull/1002/commits/ab1625fd5fb43c113642802f6c3e799ce4b49c3b

contracts/contracts/strategies/BaseCurveStrategy.sol
sparrowDom: Addressing this issue in this PR: https://github.com/OriginProtocol/origin-dollar/pull/1161

contracts/contracts/strategies/BaseConvexMetaStrategy.sol
sparrowDom: agreed, fix: https://github.com/OriginProtocol/origin-dollar/pull/1002/commits/b6b5a3b00d524a651ba517c5f1cb261552607456

contracts/contracts/strategies/BaseConvexMetaStrategy.sol
sparrowDom: we can change this to int128, but then we need casts on all the places that expect it to be uint256. like: 
```
- metapool.balances(crvCoinIndex) ( 3 occurrences)
- IVault(vaultAddress).burnForStrategy(_removedAmounts[mainCoinIndex])
- IVault(vaultAddress).burnForStrategy(_removedAmounts[mainCoinIndex])
```
Looks like uint128 is implicitly cast to uint256 (in all above places) while int128 needs to be explicitly cast. 

I would rather leave it as is.

contracts/contracts/strategies/ConvexGeneralizedMetaStrategy.sol
sparrowDom: Wow I missed this comment somehow. Point 2 you suggested in this comment is what has been done in this PR: https://github.com/OriginProtocol/origin-dollar/pull/1161

contracts/contracts/strategies/ConvexGeneralizedMetaStrategy.sol
```diff
@@ -0,0 +1,155 @@
...
+         * removing liquidity in the balanced manner (only removing 3CrvLP tokens)
+         */
+        // slither-disable-next-line divide-before-multiply
+        uint256 estimationRequiredMetapoolLpTokens = curvePool
+            .get_virtual_price()
+            .divPrecisely(metapool.get_virtual_price())
+            .mulTruncate(num3CrvTokens);
+
+        int128 metapool3CrvCoinIndex = int128(
+            _getMetapoolCoinIndex(address(pTokenAddress))
```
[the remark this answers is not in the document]
sparrowDom: Yup nice catch thanks... fix: https://github.com/OriginProtocol/origin-dollar/pull/1002/commits/ef022b40a47de9f785df484ce879dc9870ca44bf

contracts/contracts/strategies/BaseConvexMetaStrategy.sol
sparrowDom: yes agree, better to have more flexibility than this sanity check... fix: https://github.com/OriginProtocol/origin-dollar/pull/1002/commits/48011870bea50ef45b101f623dd5ddf139914c03

contracts/contracts/strategies/BaseConvexMetaStrategy.sol
sparrowDom: good point, renamed to more conventional `max` function here: https://github.com/OriginProtocol/origin-dollar/pull/1002/commits/b826c7c58550e6b274bfe869e4a4a48f704623d0

contracts/contracts/strategies/BaseConvexMetaStrategy.sol
```diff
@@ -0,0 +1,218 @@
...
+    address internal cvxRewardStakerAddress;
+    uint256 internal cvxDepositorPTokenId;
+    ICurveMetaPool internal metapool;
+    IERC20 internal metapoolMainToken;
+    IERC20 internal metapoolLPToken;
+    // Ordered list of metapool assets
+    address[] internal metapoolAssets;
+    // Max withdrawal slippage denominated in 1e18 (1e18 == 100%)
+    uint256 public maxWithdrawalSlippage;
+    uint128 internal crvCoinIndex;
```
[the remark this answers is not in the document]
sparrowDom: Addressed a couple of comments above why that opens the problems on another end

contracts/contracts/strategies/ConvexGeneralizedMetaStrategy.sol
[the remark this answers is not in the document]
sparrowDom: So both Generalized and OUSD meta strategies use hardcoded MAX_SLIPPAGE in their deposit functions. I think that is the right behaviour since I understand this hardcoded slippage as "the amount of loss we are willing to take when entering curve pool". If the loss is higher I think not entering the pool is the right approach.

contracts/contracts/strategies/ConvexGeneralizedMetaStrategy.sol
sparrowDom: This code is altered and has another approach in this PR: https://github.com/OriginProtocol/origin-dollar/pull/1161

franckc (approved):
No further comments on my end. I reviewed the migration 43 and it looks good.
Ok to merge on my side after:
 - PR 1161 is in
 - all comments from @DanielVF have been addressed
 - we have checked that the CI/CD fork tests failure are not a real issue

contracts/contracts/strategies/BaseConvexMetaStrategy.sol
DanielVF: Fix confirmed. Comment resolved.

contracts/contracts/strategies/BaseConvexMetaStrategy.sol
DanielVF: I'm good leaving it as it is.

contracts/contracts/strategies/BaseConvexMetaStrategy.sol
DanielVF: Looks good.

contracts/contracts/strategies/BaseConvexMetaStrategy.sol
DanielVF: Looks good.

contracts/contracts/strategies/ConvexGeneralizedMetaStrategy.sol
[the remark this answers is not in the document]
DanielVF: Fix looks good.

contracts/contracts/strategies/BaseConvexMetaStrategy.sol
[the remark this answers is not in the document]
DanielVF: Understood.

contracts/utils/generalizedConvexStratDeployment.js
```diff
@@ -52,10 +54,19 @@ module.exports = ({
         dConvexTokenMetaStrategyProxy.address
       );
 
-      // 2. Deploy new implementation
-      const dConvexTokenMetaStrategyImpl = await deployWithConfirmation(
-        "ConvexGeneralizedMetaStrategy"
-      );
+      let convexTokenMetaStrategyImpl;
+
+      if (deployStrategyImplementation) {
```
DanielVF, line 59: The thing you got back after a deploy command is the transaction, I think - a different kind of object than what you got from `ethers.getContract`, which is why we always made the distinction between them, with one of them being prepended with a `d`.

This code sets `convexTokenMetaStrategyImpl` with either kind. 

It looks like only the address is used? In which case perhaps we call this variable `implementationAddress`?

Or if we need to use it more, we could do the deploy in the if statement, then always ethers.getContract the contract regardless of if we have deployed it or not.

contracts/contracts/vault/VaultAdmin.sol
```diff
@@ -133,6 +133,19 @@ contract VaultAdmin is VaultStorage {
...
+     * @dev Set maximum amount of OUSD that can at any point be minted and deployed
+     * to strategy (used only by ConvexOUSDMetaStrategy for now).
+     * @param _threshold OUSD amount with 18 fixed decimals.
+     */
+    function setNetOusdMintForStrategyThreshold(uint256 _threshold)
+        external
+        onlyGovernor
+    {
+        netOusdMintForStrategyThreshold = _threshold;
+        emit NetOusdMintForStrategyThresholdChanged(_threshold);
```
DanielVF, line 146: In use, the OUSD AMO is likely to mostly operate in one direction for either long periods of time, either mostly minting OUSD, or mostly redeeming OUSD. Because of this, governance will need keep extending the netOusdMintForStrategyThreshold. However, this high netOusdMintForStrategyThreshold will mean that that while it might only allow a small change in one direction in the typical direction, it could allow a truly massive change back the other way. For example, if it had mostly been used for redeems, and had done a net 105 million burns so far, we might want to set the limit to 115 million to enable 10 million more burns. However, this would mean that it could also be permitted to mint 220 million OUSD. I think the solution to this is that whenever governance changes the limit, we should reset the `netOusdMintForStrategy` back to zero. This will mean that the new limit based on where things currently are.

contracts/contracts/vault/VaultCore.sol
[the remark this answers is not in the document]
DanielVF: Confirmed this has been uncommented.

contracts/utils/generalizedConvexStratDeployment.js
sparrowDom: Yup good point it is a different object, both seem to expose the address property. Have implemented optimisation you suggest: https://github.com/OriginProtocol/origin-dollar/pull/1002/commits/1301af1f2b690c40bacfcce948c3c70a75611d65

contracts/contracts/vault/VaultAdmin.sol
sparrowDom: Thanks @DanielVF! And sorry for overlooking this in your CR sum up comment. You have already laid all this out there. Fix for it here: https://github.com/OriginProtocol/origin-dollar/pull/1002/commits/302ea687c19c910dff8202086a7b8d7f2ddc2de2

Also another thing we gain with this is that we can actually reduce the threshold amount and not worry that the current netOusdMinted counter is too much near one or the other limit. Which would make it hard to make the threshold smaller.

sparrowDom:
@DanielVF merging this in for now... lets talk if there is smth urgent that we'd still need to address

Issue #865, opened 2021-12-17
Improve connecting wallet UX on mobile


sparrowDom:
This is still in a suggestion / work in progress phase type of issue: 

Visiting ousd.com from a mobile browser can be confusing. Wallet connection options and error messages when connecting the not supported wallets. If we were to copy Uniswap's approach we would: 

- show only the Wallet connect option when visiting with a mobile browser (No ledger / Metamask options)
- autoconnect to a wallet if using Metamask mobile browser
- on desktop behaviour is not changed. We should still change the "No Ethereum Wallet detected" to something that can not be confused with "There is no ethereum in your wallet"

micahalcorn:
In my experience, MetaMask and Coinbase Wallet already skip the modal in their mobile browsers. So I don't think we would need to change anything there. I was also surprised to see that WalletConnect already works quite well from mobile Safari.

micahalcorn:
Another option is to leave the buttons intact on mobile and just link to each vendor if there is no web3 on the window. So, for example, the MetaMask button would point to https://metamask.io in mobile Safari.

franckc:
@micahalcorn Is it something we'd like to prioritize?

micahalcorn:
Coinbase Wallet and WalletConnect work well for me. For MetaMask, I would look into [deeplinking](https://docs.metamask.io/guide/mobile-best-practices.html#deeplinking). If it's possible to detect that the user does not have a given app installed, then it would be nice to link to the app's website when the button is clicked. I don't think I would hide any buttons like Ledger just because it isn't supported on mobile. People often use these buttons as indicators of what we do support. So I would just prompt the user if she taps Ledger and say that it's not supported on phones. Otherwise, yes, these would be nice improvements.

sparrowDom:
Agree, deeplinking would be a great addition, especially since some ( no idea how many) users are already used to it from the other apps. 
As for the issue of hiding unsupported wallets I would still show them but grey them out (make them seem disabled). We can still show the prompt as @micahalcorn suggests in case users click on them.

rolandpo:
done in #1008

PR #1008, opened 2022-05-31
mobile wallet options

rolandpo:
issue #865

- grey out Ledger option on mobile wallet modal
- on clicking Ledger prompt user to use WalletConnect with Ledger Live instead
- add MetaMask deeplink that takes user to app or appstore depending on whether it's installed


dapp/src/components/WalletSelectContent.js
```diff
@@ -107,8 +114,19 @@ const WalletSelectContent = ({}) => {
...
               key={name}
-              className="connector-button d-flex align-items-center"
-              onClick={() => onConnect(name)}
+              className={`connector-button d-flex align-items-center ${
+                isMobile && name === 'Ledger' ? 'grey' : ''
+              }`}
+              onClick={() => {
+                if (isMobile && name === 'MetaMask') {
+                  setError(null)
+                  window.location.href = process.env.METAMASK_DEEPLINK
```
sparrowDom, line 123: staging / prod / local environment should have this env var available

sparrowDom (commented):
Left 1 comment please take a look. Also curious have you found a convenient way to test this? Maybe deploy to staging and test with a mobile device?

dapp/src/components/WalletSelectContent.js
rolandpo: yep, forgot to add

rolandpo:
added MetaMask deeplink to dapp .env file. I was testing the mobile stuff with localtunnel to generate a link for mobile access

sparrowDom:
Tested this a couple of comments: 
 - when on mobile device can you change the order of connections to: WalletConnect, CoinbaseWallet, MyEtherWallet, Metamask, grayed out Leger. It makes more sense to me to put the popular/convenient option on top. And those differ between desktop / mobile devices
 - when clicking on metamsk link on mobile device this error page opens for me
Screenshot_20220617-131855

rolandpo:
not sure about this error page, it works for me on iphone on safari/chrome

sparrowDom:
This is usual behaviour where Android / iPhone devices behave differently. Do you have an Android device you can use to test this? If not there should be Android simulators available.

rolandpo:
changed the order of wallet options for mobile

sparrowDom:
Issue still persists for me on Android, but it is not a blocker to get this merged in. Great work thanks 🙏

sparrowDom (approved):
LGTM

PR #1011, opened 2022-06-01
wrapping form fixes

rolandpo:
simple wrapping form fixes, consistent with swap page. disabling wrapping on 0 amount and clearing input on initial load


sparrowDom (approved):
Works nicely thanks!

PR #1012, opened 2022-06-02
apy link on main page

rolandpo:
add link to APY analytics on main page as requested by Hasagio


sparrowDom (approved):
Nice little gem thanks

Issue #963, opened 2022-03-29
Maintain approval button state between contract selections


micahalcorn:
**Steps to reproduce**
- Need to approve the Flipper to swap DAI for OUSD
- Submit an approval transaction
- See the top button change state from "Allow..." to "Approving..."
- Before the transaction is confirmed, override the contract selection to Curve
- Switch the contract selection back to Flipper
- See the top button state return to "Allow..." while an approval transaction is still in progress

Note several things:
1. Once the approval transaction is confirmed, the top button does change state to be disabled since approval is not longer necessary, which is great.
2. It's unfortunate having redundant transaction indicators. As discussed in https://github.com/OriginProtocol/origin-dollar/issues/732#issuecomment-926695606, it might make sense to do away with the transaction feed on the right. But I generally like the pattern where we reset the form state when a swap transaction has been successfully submitted instead of waiting for it to be confirmed. This necessitates having a transaction history/status outside of the form. To be consistent, we could also unblock the user from submitting the swap transaction if she has a pending approval transaction. It's a little different in this case though because the form state is still applicable to the swap button after the approval transaction has been submitted.
3. If the page is refreshed, all bets are off. I don't think we should try to maintain form/button state in that case.

PR #1013, opened 2022-06-02
introduce global state for approval stage and update accordingly

jastroud:
original issue: https://github.com/OriginProtocol/origin-dollar/issues/963

- Maintain the approval button status while switching between other contracts. 
- Disable the approval button while waiting for the network or user's approval.

### DIFF
--- a/dapp/src/components/buySell/ApproveSwap.js
+++ b/dapp/src/components/buySell/ApproveSwap.js
@@ -24,11 +24,12 @@ const ApproveSwap = ({
   rpcProvider,
   isMobile,
 }) => {
-  const [stage, setStage] = useState('approve')
   const [contract, setContract] = useState(null)
   const [isApproving, setIsApproving] = useState({})
   const web3react = useWeb3React()
   const { library, account } = web3react
+  const approvalStage = useStoreState(ContractStore, (s) => s.approvalStage)
+  const stage = approvalStage?.[stableCoinToApprove]?.[selectedSwap?.name];
   const coinApproved = stage === 'done'
   const isWrapped =
     selectedSwap &&
@@ -47,7 +48,6 @@ const ApproveSwap = ({
       s.approvalNeeded = approvalNeeded
     })
   }, [approvalNeeded])
-
   const {
     vault,
     flipper,
@@ -61,6 +61,18 @@ const ApproveSwap = ({
     ousd,
     wousd,
   } = useStoreState(ContractStore, (s) => s.contracts || {})
+
+  const updateApprovalStage = (newStage) => {
+    ContractStore.update((s) => {
+      s.approvalStage = {
+        ...s.approvalStage,
+        [stableCoinToApprove]: {
+          ...s.approvalStage?.[stableCoinToApprove],
+          [selectedSwap?.name]: newStage
+        }
+      }
+    })
+  }
 
   const routeConfig = {
     vault: {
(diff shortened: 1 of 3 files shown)


tomhirst (commented):
Great work @jastroud, I've tested this as working and the code looks fine.

A couple of UX/UI points:

1. Could we disable the approval button when it's in the approval state? IE, "Approving Uniswap...". There's no reason to allow the user to click again at this stage.

2. Referencing back to @micahalcorn's original issue: https://github.com/OriginProtocol/origin-dollar/issues/963, could we disable the contract switcher while an existing approval is being processed?

jastroud:
@tomhirst Thanks for your review.
For #2 feedback, how are we gonna disable the contract switcher exactly?
Are we gonna grey out (disable) contract buttons while an existing approval is being processed?

jastroud:
@tomhirst Alright, I disabled the other contract buttons when transaction approval is in progress for now.
You can review it again when you are available.

tomhirst:
> @tomhirst Alright, I disabled the other contract buttons when transaction approval is in progress for now. You can review it again when you are available.

Hi @jastroud, yep that's what I meant. My thinking is that it makes sense to solve the problem described in the issue by making it impossible for the user to select another contract while we're waiting for an existing approval to go through.

I can see that the contract selectors aren't buttons, so one option could be to grey them out globally, like when the modal is opened:

Screenshot 2022-06-06 at 10 57 13

I've tested the disabling of the approval button. The disable flag doesn't persist when we're waiting for the network (post user approval). I've left a comment in the code that should solve this.

1 - Waiting for user
Screenshot 2022-06-06 at 10 53 05

2 - Waiting for network
Screenshot 2022-06-06 at 10 53 09

3 - Approved
Screenshot 2022-06-06 at 10 53 19

Thanks for taking the time to look at my suggestions!

tomhirst:
> @tomhirst Thanks for your review, I updated the PR and disabled the button while waiting for the network. Disabling contract selection while the existing approval is being processed is already included in the PR.
> 
> You should be good to review the PR again, thanks!

Nice, the approval button state feels more consistent now. I couldn't see the contract selection items being disabled in my testing yesterday, though @rolandpo makes a good point below...

> If I'm being honest I can imagine a scenario in which a user wants to exchange multiple tokens, approval gets stuck in limbo for a bit due to gas price swings so would then move on to the next token, while waiting for the previous transactions to get sorted. Bit of an edge case for sure, but might be frustrating to have the dapp freeze essentially on a pending approval

Yeah, this could potentially happen. IE, a user wants to queue approvals, which may require them to switch contracts between transactions.

Disabling the whole contract switching panel would prevent the problem in https://github.com/OriginProtocol/origin-dollar/issues/963 from even being seen, and it wouldn't freeze the entire dApp, but I guess the global state changes are enough to close the issue because they persist approval state (button text) on a per token/contract basis.

Let's remove the disabling of the contract selection buttons to prevent any potential frustration caused by network delays for users who might want to queue transactions, then I'll run a final test and merge this PR.

Thanks, @jastroud and @rolandpo 🤝

jastroud:
@tomhirst @rolandpo I updated the PR based on your feedback, you can review it again when you are available. Thanks!

jastroud:
@tomhirst To be a bit more clear, my below commit was to disable the contract selection while the existing approval is being processed.
https://github.com/OriginProtocol/origin-dollar/pull/1013/commits/7d63a9e735d1a5015f7d5fec3868e5106ce07e6c

sparrowDom:
old PR closing...

shahthepro:
Closing this since we no longer have the DApp in this repo

Issue #961, opened 2022-03-29
Consider prompting the user when contract override is abandoned


micahalcorn:
**Steps to reproduce**
- Enter an amount of stablecoins to swap for OUSD
- See Flipper chosen as the default option (for example)
- Override the default selection by clicking on Curve
- Change the input amount slightly
- Contract selection is changed back to default

I _think_ this is the desired behavior because the risk would be greater if we maintained the override and then the new amount resulted in a much worse outcome than it would with the (new) default selection. Also, the contract section changes to a loading state anytime an input is changed. The problem though is that it is often out of view below the fold and the user may not realize that it has happened. We might want to prompt the user with an alert similar to the warning that we display then the default selection is initially overridden. We could also wait until the user attempts to submit the transaction and then require confirmation if (1) the default selection was overridden, (2) the input was changed after the initial warning was displayed, _and_ (3) the current selection is different from the last override.

ritikBhandari:
Hi,
Can I take up this issue if you dont miind?

sparrowDom:
Posting a suggestion to the problem here: 
<img width="688" alt="Screenshot 2022-06-06 at 10 38 15" src="https://user-images.githubusercontent.com/579910/172126896-56bb2f44-cd4e-42ac-9a50-ef140fcdff7f.png">

IMHO this is a relatively elegant solution that is above the fold and is not hard to switch from a blocking to a non blocking one.

PR #1019, opened 2022-06-06
introduce lastOverride global state to display prompt

jastroud:
Original issue: https://github.com/OriginProtocol/origin-dollar/issues/961

We prompt the users when they submit the transactions if (1) the default selection was overridden, (2) the input was changed after the initial warning was displayed, and (3) the current selection is different from the last override

I introduced a new global state (lastOverride) to save the last selection from the users and compare it with the best/current selection when they try to submit the transaction (approval) and based on the comparison result, display the confirmation modal.


sparrowDom:
Screenshot 2022-06-09 at 10 25 01

Hey when testing I got this message even though I have not manually changed anything.

Also would be cool to reference the original issue in the PR description and add some description so that whoever checks this PR out knows what is going on.

jastroud:
@sparrowDom My bad, I missed the case when lastOverride is undefined (first time). I updated the PR and you are good to review it again. Good catch, Thanks!

sparrowDom:
Hey still having issues with this one. Now the prompt doesn't show up when it should. See behaviour here: 
https://www.loom.com/share/101f7b3b03144d56a04c1cafbb8bc320

jastroud:
https://www.loom.com/share/5dff6bc062fa4678aa583e7bc2151f8f

@sparrowDom I am not able to replicate the issue on my local, above is what I am seeing on my end.
I think caches might cause the issue.

rolandpo:
One minor thing is that if you switch contracts while the modal is shown then it becomes a bit confusing

Screenshot 2022-06-09 at 15 45 36

jastroud:
Posting screenshots on various devices.

image
image
image

jastroud:
@sparrowDom @rolandpo Updated the PR based on our discussion and agreement. You can review it again when you are available, Thanks!

sparrowDom (approved):
Nice job thanks!

Issue #1015, opened 2022-06-05
Immediate unwrapping appears to result in a loss


micahalcorn:
As a user who is hesitant about wrapping (and confused by the exchange rate math), I want to know that I can unwrap my OUSD and get the same amount back. Due to some rounding, I suspect, flipping the form causes the OUSD output to be lower than my original input.

https://www.loom.com/share/344dcb5a92064e739cfc7779ac0ef88e

rolandpo:
done in #1020

PR #1020, opened 2022-06-06
fix input flipping on wrap page

rolandpo:
Fix downward spiral on flipping fields on wrap form. When the inputs are flipped twice the top field shows a slightly smaller floored value compared to original due to the nature of the contract, but that's the extent of the spiral.
The actual value decreases in the background, but that's negligible. In the case of the user wanting to wrap their max balance and then flipping a couple times where this would matter, the actual value returns to max balance, making sure no dust is left behind

related to #1015


sparrowDom (approved):
works great, thanks

PR #1009, opened 2022-05-31
allow access to dapp without connected wallet

rolandpo:
issue #1007

- swap/wrap forms show even if no wallet is connected
- contract table on swap page also shows, and assumes no allowances have been granted
- swap button greyed out and display 'Connect Wallet' message
- coin balances are hidden
- transaction history page stays the same as before

PR #1027, opened 2022-06-21
add wallet connect functionality to swap button when not connected

rolandpo:
wrapping up #1009


micahalcorn:
I'll ask @sparrowDom or @tomhirst to review this next week, but I think we should go ahead and merge this prior to the launch where we hope to drive substantial traffic to OUSD.

micahalcorn (approved):

Issue #1026, opened 2022-06-20
Selectable APY time interval on homepage


micahalcorn:
We've updated the OUSD homepage to more clearly state that the APY shown is actual and historical rather than current or forward-looking. Let's take this one step further by allowing the viewer to choose a different historical measure of the protocol's performance. My suggestion is to make "year" obviously clickable and make the other options "week" and "month". We could underline it with a little gear icon to the right or just use an ugly, native `select` element instead of reinventing it. Either way, let's default to the 30-day/monthly choice. A week is often not long enough to include a representative sample of yield-generating events, but a year is arguably a worse indicator of possible future performance.

Also, it's a little awkward to say "month" when the calculation is actually the trailing 30-day period. I think we should be consistent about using numbers or words, so here are the two options in my mind:

**Option A** (good)
- "year"
- "month"*
- "week"

**Option B** (good)
- "365 days"
- "30 days"*
- "7 days"

~~**Other**~~ (bad)
- "year"
- "30 days"*
- "week"

\* default selection

<img width="488" alt="Screen Shot 2022-06-20 at 9 54 00 AM" src="https://user-images.githubusercontent.com/273937/174653906-51243f13-37fd-4438-9e6a-56bb5bf0a023.png">

staa99:
Hi @micahalcorn,

I think using `Option B` is great. That way, the calculation feels more accurate and conveys the right information to the user as well. For styling, I think underlining the `year` and making it clickable to show a little component that displays the options is a good idea. Defaulting to the `month` is probably best since it provides a closer estimate of the current situation. I can pick this up if you don't mind. It'd be my first contribution to the dApp so I might need some help figuring some stuff out.

micahalcorn:
> Hi @micahalcorn,
> 
> I think using `Option B` is great. That way, the calculation feels more accurate and conveys the right information to the user as well. For styling, I think underlining the `year` and making it clickable to show a little component that displays the options is a good idea. Defaulting to the `month` is probably best since it provides a closer estimate of the current situation. I can pick this up if you don't mind. It'd be my first contribution to the dApp so I might need some help figuring some stuff out.

Sure, thanks 👍 If we go with option B, let's also just change the existing "year" text to "30 days" to match the default selectable option.

staa99:
Hi @micahalcorn, 

I am having an issue running the dApp locally. I'm guessing the problem is from some missing ABIs. My branch is based on `master`.

[...]

PR #1033, opened 2022-07-08
selectable APY on homepage

rolandpo:
This adds the option to choose between APY sample periods on the homepage. Options are 7, 30, 60, 90, 365 days, similar to what has already been implemented on the swap page, but can cut these down. Default is set to 30 days
also localStorage is shared between the selection on the homepage and swap page

issue #1026

Screenshot 2022-07-08 at 18 18 58

Screenshot 2022-07-08 at 18 19 21


micahalcorn:
@rolandpo I actually meant for this to be applied to the text under the big number below the fold. Do your changes address that section too or just the little banner in the nav? It probably belongs in both places but let's be sure to include it where we use the words and have the big call-to-action.

Screen Shot 2022-07-08 at 12 44 40 PM

rolandpo:
removed the apy selection from the top (and set it to 30 days) and added it down below

Screenshot 2022-07-12 at 20 17 56

Screenshot 2022-07-12 at 20 18 06

micahalcorn (approved):

Issue #1040, opened 2022-07-13
Ledger account derivation mismatch


micahalcorn:
I have a Ledger Nano X that I'm testing with. The accounts that I'm generating don't match up with any of the options that are shown by our DApp. The first account is the same but the rest are not. It looks like this might be related to https://github.com/OriginProtocol/origin-dollar/issues/21#issuecomment-694997222.

**Ledger Live**
Ethereum 1: 0x**c...** `"freshAddressPath": "44'/60'/0'/0/0"`
Ethereum 2: 0x**9...** `"freshAddressPath": "44'/60'/1'/0/0"`

**ousd.com**
<img width="476" alt="Screen Shot 2022-07-12 at 10 42 04 PM" src="https://user-images.githubusercontent.com/273937/178659361-08eba052-9c7f-4bbe-863d-d457fc3dbdd8.png">

micahalcorn:
As another data point, I connected the Ledger to MetaMask and they offer the same accounts as Ledger Live (also not matching our DApp).

sparrowDom:
done via: https://github.com/OriginProtocol/origin-dollar/pull/1041

PR #1041, opened 2022-07-13
draft: change method of iterating through BIP-44 addresses

rolandpo:
This provides a fix for issue #1040. From what I'm seeing the problem isn't with the derivation path itself but the way the library iterates through available addresses. Currently 44'/60'/0'/x instead of 44'/60'/x'/0.

I added a hacky fix that handles the Ledger Live option separately by iterating through its addresses correctly, there might be a simpler solution
There's also a question of whether to replace the existing legacy path or add another one


sparrowDom (approved):
Sweet thanks. Works as expected ( I see the correct 2nd account generated by the Ledger Live)

PR #1044, opened 2022-07-13
quick fix for navigation links

rolandpo:
- spread out dapp navigation links like it was before, now that earn is gone
- fix `get ousd` button disappearing on medium screens


sparrowDom (approved):
LGTM

PR #1058, opened 2022-08-02
draft: enable fee distribution to OGV stakers

zskolnik:
Request for feedback.  Notes:
- OGV doesn't have an oracle yet
- need to figure out how to pull in dependencies from ousd-governance (e.g. RewardsSource, MockOGV)
- a couple unit tests are commented out
- also see update to RewardsSource.sol in ousd-governance


DanielVF (commented):
Hey, good start here.

I think we'll definitely want to test that the funds have arrived.

I don't think this would go live without an oracle, so probably want to go ahead and use a mock oracle for the tests.

shahthepro:
Closing this in favor of #1068

Issue #792, opened 2021-10-25
Error: cannot estimate gas


micahalcorn:
I saw this error when attempting to swap from the modal after approving the Curve contract for the first time. For some reason, I can't find a corresponding error in Sentry.

[ousd.com-1635136024489.log](https://github.com/OriginProtocol/origin-dollar/files/7407210/ousd.com-1635136024489.log)

sparrowDom:
Fixes the issue: https://github.com/OriginProtocol/origin-dollar/pull/810

micahalcorn:
A user is reporting this error when attempting to swap OUSD for either USDT or USDC via the Flipper.

![revert](https://user-images.githubusercontent.com/273937/178628525-0dc69e66-4a21-4edc-ab46-bdadc95ecdf1.png)

franckc:
I think @micahalcorn has a Nano X

sparrowDom:
@micahalcorn have you ever tried swapping using the NanoX? From console log I assume the user had ledger connected via Metamsk

sparrowDom:
So the problem was that Dapp wasn't detecting that Flipper doesn't have enough funds to perform the swap. User was trying to swap 10k OUSD using flipper. flipper didn't have 10k USDC. Dapp should say "amount too high", but didn't rule Flipper out and was trying to perform the swap.

micahalcorn:
> So the problem was that Dapp wasn't detecting that Flipper doesn't have enough funds to perform the swap. User was trying to swap 10k OUSD using flipper. flipper didn't have 10k USDC. Dapp should say "amount too high", but didn't rule Flipper out and was trying to perform the swap.

Interesting, was that unique to the Ledger? I tried checking that with regular MetaMask connection and Flipper did not show up as an option with larger amounts.

sparrowDom:
Another use has been experiencing this issue this time Firefox/Metamask. So it looks like it is not a Ledger specific issue. 
<img width="1164" alt="Screenshot 2022-08-08 at 09 09 53" src="https://user-images.githubusercontent.com/579910/183360155-5079dc54-1064-485f-8ff4-f95c7be353bb.png">

rolandpo:
done in #1063

micahalcorn:
There's some concern that #1063 doesn't completely resolve this, but none of us have been able to reproduce the issue lately. We'll reopen this with more details if we find anything.

PR #1063, opened 2022-08-09
fix for Flipper error on redeem

rolandpo:
fix for issue #792

The `swapMode` variable was being read from local storage regardless of whether it had a value, so that was `null` for anyone visiting for the first time or after deleting cookies if they didn't change the swap mode manually. That then caused both the swap dapp to default to redeem mode and `coinInfoList[swapMode === 'redeem' ? selectedCoin : 'ousd']` to choose OUSD, which had the Flipper estimate looking at levels of OUSD in the contract instead of the stables being redeemed.
Also the swap page is in mint mode by default now

### DIFF
--- a/dapp/src/components/buySell/SwapHomepage.js
+++ b/dapp/src/components/buySell/SwapHomepage.js
@@ -39,7 +39,7 @@ const SwapHomepage = ({
 
   // mint / redeem
   const [swapMode, setSwapMode] = useState(
-    process.browser ? localStorage.getItem(lastSelectedSwapModeKey) : 'mint'
+    process.browser && localStorage.getItem(lastSelectedSwapModeKey) !== 'null' ? localStorage.getItem(lastSelectedSwapModeKey) : 'mint'
   )
   const previousSwapMode = usePrevious(swapMode)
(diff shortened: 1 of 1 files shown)


dapp/src/components/buySell/SwapHomepage.js
```diff
@@ -135,7 +137,7 @@ const SwapHomepage = ({
 
     // currencies flipped
     if (previousSwapMode !== swapMode) {
-      localStorage.setItem(lastSelectedSwapModeKey, swapMode)
+      if (swapMode) localStorage.setItem(lastSelectedSwapModeKey, swapMode)
```
shahthepro, line 140: Not related to this PR, but this `useEffect` hook is listening to the changes on the `swapMode` var. So, I don't see why there's a need for `previousSwapMode !== swapMode` check here.

Also, `previousSwapMode` and `selectedSwap` aren't being passed to this hook. Unless the `SwapHomepage` component rerenders everytime there is a change in the `ContractStore` state, this may not work as intended

rolandpo: Right, removed previousSwapMode. I think the selectedSwap check prevents the component from breaking on initial render so left that there

shahthepro (approved):

PR #1066, opened 2022-08-16
Convex generalized Meta strategy

sparrowDom:
Convex generalized meta strategy
- creates a common abstract strategy that combines strategies interacting with Curve metapools
- introduces a metapool strategy that can be configured to harvest any Curve metapool
- added more brownie script functionality to test the behaviour in mainnet fork

**PR merges into metastrategy PR instead of master**


sparrowDom:
Experienced something really weird on Hardhat mainnet fork on block `14655959`. The curve frax_metapool exchange coins transactions were failing and when tracing the transaction it failed in the step of transfering curve 3pool_lp token. 

So when running node on said block number and executing the `metapool/frax_pool_test.py` script via `brownie run metapool/frax_pool_test.py --network hardhat` I get this output: 
Screenshot 2022-08-16 at 19 09 56
Which indicates that frax_metapool though it had 1202 million 3crv_lp tokens but in reality when manually inspecting the amount of those tokens it turned out to be 0. 

Double checking that data [using Dune analytics](https://dune.com/queries/1173969) it seems as if the frax metapool actually had the 3CurveLp tokens in the vicinity where hardhat indicated they were 0. 

Which makes me think this could be a hardhat problem, but would still be cool if someone else can verify - reproduce the problem.

DanielVF:
This contract acts fairly differently than our 3pool strategy. While these difference might be unavoidable, it's good to know what they are.

Primarily this is because we are only ever depositing or withdrawing from one side of the target curve meta strategy.

1. Deposits and withdraws will always have swap fees. This because if we deposit just one side to a balanced pool, we are in effect swapping a portion of that for the other side of the pool, and paying fees to do so. Because this is only half of the deposit, the fee works out to be about 0.02%, though this depends on the pool. This is on top of whatever fee and slippage we already have with 3Pool.

2. When the vault asks the strategy for the total value of its funds, the existing convex strategy will always have that at or more than that of stablecoins that it can withdraw. This is no longer true.

3. If a pool is imbalanced, we could take trading losses withdrawing, and if imbalanced far enough the actual amount withdrawable could be practically zero. (With three pool along, we can always withdraw tokens that are valid for giving to OUSD users, even in impossibly imbalanced situations.)

sparrowDom:
These are all great points, very good to keep in mind *and make it more obvious in strategy yield calculator* that entering / exiting these sorts of strategies is more costly. So their yield should be that much higher for the strategy to make sense. 

In regards to point 3. I am thinking we should really do a good enough due diligence on the coin that is paired to 3pool in the metapool. IMO if the pool turns out imbalanced so bad that we take loss when removing liquidity it was either because: 
 - our liquidity in that pool represents a too great share of total pool liquidity
 - we haven't done due diligence / coin surprisingly crashes -> then the stablecoin will loose its 1$ peg anyway. So even if we were able to remove the liquidity in a balanced manner we would take a loss.

brownie/README.md
```diff
@@ -22,7 +22,20 @@ Or connect to a direct forked of mainnet:
...
+OUSD & Generalized metastrategy both have their corresponding deploy scripts that need to be
+have `forceDeploy` set to `true`. Node will emit addresses of the deployed OUSD & FRAX strategy
+in console. Copy those and set them in the `brownie/metastrategy.py` configuration section under
+`OUSD_META_STRATEGY` & `FRAX_STRATEGY` variables. Also set which one of those you want to act as a
+default USDT asset strategy under `USDT_DEFAULT_META_STRATEGY`. This way minting (using USDT asset) will
+directly fund configured strategy.
+
+(Note most of the configuration section shall be deleted once these strategies are deployed to mainnet.) 
+
+#### Notes
```
sparrowDom, line 38: @franckc @DanielVF @shahthepro added some more configuration info to metastrategy Readme

sparrowDom:
@shahthepro @DanielVF heads up this has been merged to Metastrategy PR ( https://github.com/OriginProtocol/origin-dollar/pull/1002 )

PR #1067, opened 2022-08-17
fix for minting on forked environment

sparrowDom:
fix for minting on forked environment


franckc (approved):
LGTM

crypto-ticky:
sorry, mint task will work only on forked environment?
That won't work on Fork=false --network localhost.
I am not sure that is your requirement with mint task.

sparrowDom:
@crypto-ticky we use the localhost network only in test suite for now and that one doesn't require the mint task. We used to support local/test nets in the past but not anymore.

PR #1071, opened 2022-08-23
Withdraw batch

crazybuster:

Contract change checklist:
  - [ ] Unit tests pass

### DIFF
--- a/contracts/contracts/vault/VaultAdmin.sol
+++ b/contracts/contracts/vault/VaultAdmin.sol
@@ -239,6 +239,24 @@ contract VaultAdmin is VaultStorage {
         strategyTo.depositAll();
     }
 
+    function withdraw(
+      address _strategyFromAddress,
+      address[] calldata _assets,
+      uint256[] calldata _amounts
+    ) external onlyGovernorOrStrategist {
+      require(
+            strategies[_strategyFromAddress].isSupported,
+            "Invalid from Strategy"
+        );
+      require(_assets.length == _amounts.length, "Parameter length mismatch");
+
+      IStrategy strategyFrom = IStrategy(_strategyFromAddress);
+      for (uint256 i = 0; i < _assets.length; i++) {
+          // Withdraw from Strategy and Vault  as recipient
+          strategyFrom.withdraw(address(this), _assets[i], _amounts[i]);
+      }
+    }
+
     /**
      * @dev Sets the maximum allowable difference between
      * total supply and backing assets' value.
--- a/contracts/contracts/vault/VaultAdmin.sol
+++ b/contracts/contracts/vault/VaultAdmin.sol
@@ -239,6 +239,12 @@ contract VaultAdmin is VaultStorage {
         strategyTo.depositAll();
     }
 
+    /**
+     * @notice Withdraw multiple assets from one strategy to the vault
+     * @param _strategyFromAddress Address of Strategy to withdraw assets from.
+     * @param _assets Array of asset address that will be withdrawn from the strategy.
+     * @param _amounts Array of amounts of each corresponding asset to withdraw.
+     */
     function withdraw(
       address _strategyFromAddress,
       address[] calldata _assets,
--- a/contracts/contracts/interfaces/IVault.sol
+++ b/contracts/contracts/interfaces/IVault.sol
@@ -111,6 +111,12 @@ interface IVault {
         uint256[] calldata _amounts
     ) external;
 
+    function withdraw(
+        address _strategyFromAddress,
+        address[] calldata _assets,
+        uint256[] calldata _amounts
+    ) external;
+
     // VaultCore.sol
     function mint(
         address _asset,
--- a/contracts/test/abi/IVault.json
+++ b/contracts/test/abi/IVault.json
@@ -4,924 +4,17 @@
   "sourceName": "contracts/interfaces/IVault.sol",
   "abi": [
     {
-      "constant": false,
-      "inputs": [],
-      "name": "unpauseRebase",
-      "outputs": [],
-      "payable": false,
-      "stateMutability": "nonpayable",
-      "type": "function"
-    },
-    {
-      "constant": true,
-      "inputs": [],
-      "name": "redeemFeeBps",
-      "outputs": [
-        {
-          "internalType": "uint256",
-          "name": "",
-          "type": "uint256"
-        }
-      ],
-      "payable": false,
-      "stateMutability": "view",
-      "type": "function"
-    },
-    {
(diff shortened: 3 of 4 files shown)


sparrowDom:
@shahthepro @DanielVF @franckc this is ready to be reviewed

contracts/contracts/vault/VaultAdmin.sol
```diff
@@ -246,25 +246,97 @@ contract VaultAdmin is VaultStorage {
...
+    ) external onlyGovernorOrStrategist {
+        _withdrawFromStrategy(
+            address(this),
+            _strategyFromAddress,
+            _assets,
+            _amounts
+        );
+    }
+
+    function _withdrawFromStrategy(
```
franckc, line 323: nit: document that _receiver can either be another strategy or the vault.

contracts/contracts/vault/VaultAdmin.sol
```diff
@@ -246,25 +246,97 @@ contract VaultAdmin is VaultStorage {
...
+    ) internal {
+        require(
+            strategies[_strategyFromAddress].isSupported,
+            "Invalid from Strategy"
+        );
+        require(_assets.length == _amounts.length, "Parameter length mismatch");
+
+        IStrategy strategyFrom = IStrategy(_strategyFromAddress);
+        for (uint256 i = 0; i < _assets.length; i++) {
+            // Withdraw from Strategy and Vault  as recipient
```
franckc, line 337: nit: update comment. receiver may not be the Vault. could be a strategy.

contracts/deploy/045_deposit_withdraw_tooling.js
```diff
@@ -0,0 +1,36 @@
...
+    // Current contracts
+    const cVaultProxy = await ethers.getContract("VaultProxy");
+
+    const dVaultAdmin = await deployWithConfirmation("VaultAdmin");
+    const cVault = await ethers.getContractAt("Vault", cVaultProxy.address);
+
+    // Governance Actions
+    // ----------------
+    return {
+      name: "Deploy new Morpho Compound strategy",
```
franckc, line 25: update

franckc (commented):
Took a pass. A few small nits inline otherwise looks great.
Please take a look at the fork tests failures - they seem relevant.

contracts/contracts/vault/VaultAdmin.sol
sparrowDom: tnx fixed

contracts/deploy/045_deposit_withdraw_tooling.js
sparrowDom: tnx fixed

sparrowDom:
with the latest commit only 1 unrelated Morpho fork test is failing.

sparrowDom:
@DanielVF a friendly ping

franckc:
Friendly ping @shahthepro @DanielVF for the review - not urgent at all but now that we have more strategies it would be nice to have that to save on gas and simplifies our weekly allocations.

@sparrowDom as part of this PR, do we also need to update any tooling (like scripts) that we use for generating the allocation transactions?

shahthepro (approved):
Looks good to me, LGTM

sparrowDom:
good thinking @franckc! I think @DanielVF uses the brownie `allocations.py` to build the strategist allocations. @DanielVF do you want me to do the changes there as well, or will you do them as a part of next strategist action?

contracts/test/vault/vault.fork-test.js
```diff
@@ -152,6 +154,82 @@ forkOnlyDescribe("ForkTest: Vault", function () {
       expect(balancePreMint).to.approxEqualTolerance(balancePostRedeem, 1);
     });
 
+    it("should withdraw from and deposit to strategy", async () => {
```
DanielVF, line 157: These tests could be much simpler if it just used `checkBalance` on the strategy, instead of converting to and from the underlying cTokens.

Also, since the changes are only to the vault, we shouldn't need to mint OUSD? I think the code might be clearer if we just transfer the USDC and DAI to the vault, rather than mint.

DanielVF, line 157: Given that these are dangerous write methods, we should have tests for each that verify that the public cannot call each of these, and that the correct revert message happens.

contracts/contracts/vault/VaultAdmin.sol
```diff
@@ -246,25 +246,100 @@ contract VaultAdmin is VaultStorage {
...
         IStrategy strategyTo = IStrategy(_strategyToAddress);
 
         for (uint256 i = 0; i < _assets.length; i++) {
             require(strategyTo.supportsAsset(_assets[i]), "Asset unsupported");
-            // Withdraw from Strategy and pass other Strategy as recipient
-            strategyFrom.withdraw(address(strategyTo), _assets[i], _amounts[i]);
+            // Send required amount of funds to the strategy
+            IERC20(_assets[i]).safeTransfer(_strategyToAddress, _amounts[i]);
+            // Deposit into the Strategy from the vault
+            strategyTo.deposit(_assets[i], _amounts[i]);
```
DanielVF, line 300: 1. Rather than calling deposit() on the strategy each time we move a coin over, let's just call depositAll() after we have moved the funds. This is more gas efficient, but mostly importantly it can save a lot of real dollars, since depositing only a single coin into a curve pool gets costs you exchange rate fees. If we do three single coin deposits, we pay fees on each. If we deposit three coins in the same action, the fees are smaller to zero.

DanielVF:
Once we've used this in production for a few weeks, we'll probably want to remove the reallocate function. The gas savings is three transfers vs using deposit withdraw, but in the actual practice with our funds allocations, I don't think it would get used very often. Having two almost duplicate ways to do the same thing in the code isn't a good thing.

We won't make a change for this in this PR though.

contracts/test/vault/vault.fork-test.js
sparrowDom: yup great point adding that

contracts/contracts/vault/VaultAdmin.sol
sparrowDom: that is a great point thanks

contracts/test/vault/vault.fork-test.js
sparrowDom: Yup agreed, have greatly simplified this function also by adding some helper methods to more easily ready strategy balance differences. The test should read much better now.

DanielVF (approved):
I have reviewed these changes, and they are good.

PR #1084, opened 2022-09-13
Shah/fork tests r2

sparrowDom:
Major changes in this PR:
- **Fork tests**: 
  - Runs unit tests against mainnet fork
  - Uses hardhat fixtures and mocha for the tests
  - Stores hardhat's fixture and fork caches using GitHub `actions/cache` workflow for faster runtimes
- **CI**: 
   - Every job is now on the same workflow
   - Workflow runs when a PR is opened, re-opened or a branch has commits pushed to it
   - Yarn dependencies are now cached for faster setup
   - Yarn install is run under `frozen` mode (to ensure that the yarn.lock is always up to date)

--------

Contract change checklist:
  - [ ] Unit tests pass


sparrowDom:
@shahthepro I hope you don't mind... have created a PR out of this one. It is easier to review and leave comments on

sparrowDom:
convenience change: https://github.com/OriginProtocol/origin-dollar/pull/1085

.github/workflows/slither.yml
```diff
@@ -1,34 +0,0 @@
-name: Slither
```
franckc, line 1: What is the reason from not running slither anymore in CI/CD?

franckc: nevermind. I see it only got move to ousd.yml!

.github/workflows/ousd.yml
```diff
@@ -0,0 +1,147 @@
...
+          cache-dependency-path: contracts/yarn.lock
+
+      - run: yarn install --frozen-lockfile
+        working-directory: ./contracts
+
+      - name: Test with Slither
+        working-directory: ./contracts
+        run: yarn run slither
+
+  snyk:
```
franckc, line 136: Is the Snyk integration still working? we stopped paying for the subscription a while ago because our team did not find it useful.
Perhaps we could get rid of it and enable GitHub dependabot as a replacement? Those two tools are slightly different but if we keep up with updating dependencies recommended by dependabot I think that will go a long way in terms of keeping the repo safe against package related security vulnerability. Thoughts?

franckc (approved):
This is great! Nic work. Having those mainnet fork tests running locally and in CI/CD will go a long way towards improving our tooling. Just a couple of comments inline.

contracts/test/_fixture.js
```diff
@@ -494,42 +532,148 @@ async function convexVaultFixture() {
...
+      threepoolSwapAbi,
+      addresses.mainnet.ThreePool
+    );
+    // const curveFactory = await ethers.getContractAt(curveFactoryAbi, curveFactoryAddress)
+
+    // Get some 3CRV from most loaded contracts/wallets
+    await transferTokensFromContract(
+      addresses.mainnet.ThreePoolToken,
+      [
+        "0xceaf7747579696a2f0bb206a14210e3c9e6fb269",
```
sparrowDom, line 558: These addresses are Curve metapools (Frax, USD, alUsd, Lusd & another one). Would be cool not transfering all the 3crv tokens from that contract -- and I am only saying this because I fell into that pitfall before :) Maybe transfering 30% of the tokens out would be enought.

The issues is when you start fork testing those contract and they have no 3Crv (and we might fork test Frax & Lusd soon for example) then unexpected errors happen because those contracts expect to have 3Crv balance but in reality have none.

contracts/test/_fixture.js
```diff
@@ -494,42 +532,148 @@ async function convexVaultFixture() {
...
-    );
+  return fixture;
+}
 
-  await fixture.vault
-    .connect(sGovernor)
-    .setAssetDefaultStrategy(
-      fixture.usdc.address,
-      fixture.OUSDmetaStrategy.address
+async function impersonateAndFundContract(address) {
```
sparrowDom, line 619: nit: `impersonateAndFundAddress` maybe since it also works on `EOA`s.

contracts/test/_metastrategies-fixtures.js
```diff
@@ -0,0 +1,221 @@
...
+}
+
+async function tiltTo3CRV(fixture, amount) {
+  const { vault, domen, ousdMetaPool } = fixture;
+
+  amount = amount || ousdUnits("500000");
+
+  // Tilt to 3CRV by half a million
+  const exchangeSign = "exchange(int128,int128,uint256,uint256)";
+  await ousdMetaPool.connect(domen)[exchangeSign](1, 0, amount, 0);
```
sparrowDom, line 69: I think this will double the tilt? Also the pool should be balanced before the tilt happens. Since if pool has 10 OUSD and 10 3crv and you exchange 5 CRV for OUSD pool will have 15CRV and 5 OUSD. End difference is 10. For that reason I would first: 
- balance pool
- exchange half of the tilt

contracts/test/_metastrategies-fixtures.js
```diff
@@ -0,0 +1,221 @@
...
+}
+
+async function tiltToOUSD(fixture, amount) {
+  const { vault, domen, ousdMetaPool } = fixture;
+
+  amount = amount || ousdUnits("500000");
+
+  // Tilt to 3CRV by half a million
+  const exchangeSign = "exchange(int128,int128,uint256,uint256)";
+  await ousdMetaPool.connect(domen)[exchangeSign](0, 1, amount, 0);
```
sparrowDom, line 90: same would apply here as above comment

contracts/test/strategies/meta.fork-test.js
```diff
@@ -0,0 +1,283 @@
...
+            1
+          );
+        } else {
+          // 1x for DAI
+          expect(supplyDiff).to.approxEqualTolerance(ousdUnits(amount), 1);
+        }
+
+        // // TODO: Check why this is zero on fork for USDT??
+        // const { ousdMetaPool, cvxBooster } = fixture
+        // console.log(await ousdMetaPool.connect(user).balanceOf(cvxBooster.address))
```
sparrowDom, line 59: I think the ousdMetaStrategy instead of metapool is the one that should be having the tokens deposited in the booster

contracts/test/strategies/meta.fork-test.js
```diff
@@ -0,0 +1,283 @@
...
+        const amount = "10000";
+
+        // Mint with all three assets
+        for (const asset of [usdt, usdc, dai]) {
+          await vault
+            .connect(anna)
+            .mint(asset.address, await units(amount, asset), 0);
+        }
+
+        // Total supply should be up by (10k x 2) + (10k x 2) + 10k = 50k
```
sparrowDom, line 91: can we do an approximate `expect` instead of above comment? To be extra careful

contracts/test/strategies/meta.fork-test.js
```diff
@@ -0,0 +1,283 @@
...
+        const balanceDiff = newBalance.sub(currentBalance);
+        expect(balanceDiff).to.approxEqualTolerance(ousdUnits(amount), 1);
+
+        // Supply checks
+        const newSupply = await ousd.totalSupply();
+        const supplyDiff = newSupply.sub(currentSupply);
+        if ([usdt.address, usdc.address].includes(asset.address)) {
+          // Ensure at least 1.5x OUSD has been added to supply
+          // (in case of USDT/USDC)
+          expect(supplyDiff).to.be.gte(ousdUnits(amount).mul(3).div(2));
```
sparrowDom, line 143: Would be really cool to get these amounts to approximate correctness. For example if metapool is tilted to 3crv for 50k minting using 80k should result in a balanced metapool. The contract should add 80k worth of 3crv & 80k OUSD + 50k OUSD (balance the pool)

contracts/test/strategies/meta.fork-test.js
sparrowDom, line 59: I think the metastrategy should be having some balance in cvxBooster instead of metapool

contracts/test/strategies/meta.fork-test.js
```diff
@@ -0,0 +1,283 @@
...
+forkDescribe("Convex 3pool/OUSD Meta Strategy", function () {
+  this.timeout(0);
+
+  describe("Balanced metapool", () => {
+    let fixture;
+    beforeEach(async () => {
+      fixture = await loadFixture(withBalancedMetaPool);
+    });
+
+    describe("Mint", function () {
```
sparrowDom, line 24: Reading this function for the first time almost brings tear to my eyes. Being able to run these types of checks on every commit is beyond awesome!

contracts/test/strategies/meta.fork-test.js
```diff
@@ -0,0 +1,283 @@
...
+const { isForkTest, units, ousdUnits } = require("../helpers");
+const {
+  withBalancedMetaPool,
+  withCRV3TitledMetapool,
+  withOUSDTitledMetapool,
+} = require("../_metastrategies-fixtures");
+
+// Ugly hack to avoid running these tests when running `npx hardhat test` directly.
+// A right way would be to add suffix to files and use patterns to filter
+const forkDescribe = isForkTest ? describe : describe.skip;
```
sparrowDom, line 13: nit: maybe rename this to `forkOnlyDescribe` and you could move it to helpers since it is used in a couple of files

contracts/test/helpers.js
```diff
@@ -12,6 +12,19 @@ chai.Assertion.addMethod("approxEqual", function (expected, message) {
   chai.expect(actual, message).lte(expected.mul("100001").div("100000"));
 });
 
+chai.Assertion.addMethod(
+  "approxEqualTolerance",
```
sparrowDom, line 16: nice addition 👍

contracts/test/_metastrategies-fixtures.js
```diff
@@ -0,0 +1,221 @@
...
+  await addLiquidity(fixture, ousdAmount, crv3Liquidity, fixture.daniel);
+
+  // Franck has added imbalanced liquidity
+  await addLiquidity(fixture, ousdAmount, crv3Liquidity.div(2), fixture.franck);
+
+  // Balance metapool
+  await balanceMetaPool(fixture);
+
+  // Tilt to OUSD
+  await tiltToOUSD(fixture);
```
sparrowDom, line 174: You could add `balanceMetapool` call inside `tiltToOUSD` function. This way calling that function on its own will be correct without needing the precondition of balancing the metapool

contracts/test/_metastrategies-fixtures.js
```diff
@@ -0,0 +1,221 @@
...
+  await addLiquidity(fixture, ousdAmount, crv3Liquidity, fixture.daniel);
+
+  // Franck has added imbalanced liquidity
+  await addLiquidity(fixture, ousdAmount, crv3Liquidity.div(2), fixture.franck);
+
+  // Balance metapool
+  await balanceMetaPool(fixture);
+
+  // Tilt to 3CRV
+  await tiltTo3CRV(fixture);
```
sparrowDom, line 196: same idea would apply here as above comment

contracts/test/_metastrategies-fixtures.js
```diff
@@ -0,0 +1,221 @@
...
+async function removeAllLiquidity(fixture, user) {
+  const { ousdMetaPool } = fixture;
+
+  const address = user.getAddress();
+  const removeSign = "remove_liquidity(uint256,uint256[2])";
+  const removeMethod = await ousdMetaPool.connect(user)[removeSign];
+  removeMethod(ousdMetaPool.balanceOf(address), ["0", "0"]);
+}
+
+async function removeLiquidityImbalanced(fixture, user, crv3Liquidity) {
```
sparrowDom, line 121: Maybe add comment: 
// removing liquidity in a manner where we first remove required amount of 3crv. And with whatever LP remains we remove OUSD

contracts/test/strategies/meta.fork-test.js
```diff
@@ -0,0 +1,283 @@
...
+        const amount = "10000";
+
+        // Mint with all three assets
+        for (const asset of [usdt, usdc, dai]) {
+          await vault
+            .connect(anna)
+            .mint(asset.address, await units(amount, asset), 0);
+        }
+
+        // Total supply should be up by (10k x 2) + (10k x 2) + 10k = 50k
```
sparrowDom, line 179: would be cool to do supply check here via code (from what is in the commet)

contracts/test/strategies/meta.fork-test.js
sparrowDom, line 91: would be cool to do supply check here via code (from what is in the commet)

sparrowDom (commented):
somewhere 60% done with the review. Still super excited about this

sparrowDom:
Also have noticed that `forceDeploy` doesn't work anymore - and I know for obvious reasons of needing to run on CI. Is there a simple way where we could still use `forceDeploy` when running on local machines? Maybe using smth like `USE_FORCE_DEPLOY_FLAG` env var or smth similar

.github/workflows/ousd.yml
shahthepro: I do see it working [on the workflow](https://github.com/OriginProtocol/origin-dollar/actions/runs/3097657322/jobs/5014583055). Looks like their free plan works with OpenSource projects, there's a limit though.

I'll look into adding dependabot as a separate PR, will leave this until we have that done

shahthepro: Here's [an issue](https://github.com/OriginProtocol/origin-dollar/issues/1115) to keep track of it

shahthepro:
> Also have noticed that `forceDeploy` doesn't work anymore - and I know for obvious reasons of needing to run on CI. Is there a simple way where we could still use `forceDeploy` when running on local machines? Maybe using smth like `USE_FORCE_DEPLOY_FLAG` env var or smth similar

`forceDeploy` shouldn't be affected at all. It still overrides everything else: <https://github.com/OriginProtocol/origin-dollar/blob/shah/fork-tests-r2/contracts/utils/deploy.js#L323>. 

I guess it has to do with something else, happy to debug and fix it if it can be reproduced

contracts/contracts/strategies/ConvexGeneralizedMetaStrategy.sol
```diff
@@ -123,12 +123,15 @@ contract ConvexGeneralizedMetaStrategy is BaseConvexMetaStrategy {
...
-        metapool.remove_liquidity_one_coin(
-            metapoolErc20.balanceOf(address(this)),
-            metapool3CrvCoinIndex,
-            num3CrvTokens
-        );
+        uint256 burnAmount = metapoolErc20.balanceOf(address(this));
+        if (burnAmount > 0) {
+            // slither-disable-next-line unused-return
+            metapool.remove_liquidity_one_coin(
+                metapoolErc20.balanceOf(address(this)),
```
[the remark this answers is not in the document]
shahthepro: Yep, totally missed it

contracts/node.sh
```diff
@@ -7,8 +7,13 @@ RED='\033[0;31m'
 NO_COLOR='\033[0m'
 
 main()
-{
-    rm -rf deployments/localhost
+{   
+    if [ "$IS_TEST" == "true" ] && [ "$FORK" == "true" ];
```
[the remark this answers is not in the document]
shahthepro: Actually, I think the change to this script is not necessary. 

This line here: <https://github.com/OriginProtocol/origin-dollar/blob/shah/fork-tests-r2/contracts/fork-test.sh#L59>

The `fork-test.sh` script copies deployments from `localhost` network when using a local forked node and from `mainnet` network otherwise

contracts/test/_fixture.js
shahthepro: Yep, sounds good to me, making that change

contracts/fork-test.sh
```diff
@@ -0,0 +1,80 @@
...
+        # Check if any node is running on port 8545
+        defaultNodeUrl=http://localhost:8545
+
+        # If local node is running, $resp is a non empty string
+        resp=$(curl -X POST --connect-timeout 3 -H "Content-Type: application/json" --data '{"jsonrpc":"2.0","method":"web3_clientVersion","params":[],"id":67}' "$defaultNodeUrl")
+
+        if [ ! -z "$resp" ]; then
+            # Use the running node, if found
+            echo "Found node running on $defaultNodeUrl"
+            export LOCAL_PROVIDER_URL="$defaultNodeUrl"
```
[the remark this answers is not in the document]
shahthepro: Right now, if you set `LOCAL_PROVIDER_URL` env variable, it's gonna skip the node availability check and use it, like an override to the default `localhost:8545`.

Happy to change it though

contracts/test/_metastrategies-fixtures.js
shahthepro: This function (and also `tiltToOUSD()`) is only called by `withLiquidityOn3CRVTitledPool()` or `withCRV3TitledMetapool()`. Both of which balance the pool before doing the tilt.

> End difference is 10
I'm sorry, I'm a little off the loop here. Do you mean to say the difference shouldn't be high or should it be higher than this?

contracts/test/strategies/meta.fork-test.js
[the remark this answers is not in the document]
shahthepro: Yep, will move it to helpers

contracts/test/_metastrategies-fixtures.js
```diff
@@ -0,0 +1,221 @@
...
+  const crv3Liquidity = await get3CRVLiquidity(fixture, ousdAmount);
+
+  // Daniel has added balanced liquidity
+  await addLiquidity(fixture, ousdAmount, crv3Liquidity, fixture.daniel);
+
+  // Franck has added imbalanced liquidity
+  await addLiquidity(fixture, ousdAmount, crv3Liquidity.div(2), fixture.franck);
+
+  // Balance metapool
+  await balanceMetaPool(fixture);
```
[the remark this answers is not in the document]
shahthepro: My reasoning was not adding `loadFixture` inside those methods is because the tiltTo3Crv function is also used by `withCRV3TitledMetapool` where we load the balanced metapool. But yeah, I could rewrite this

contracts/test/_metastrategies-fixtures.js
```diff
@@ -0,0 +1,221 @@
...
+
+async function withBalancedMetaPool() {
+  const fixture = await loadFixture(withDefaultStrategiesSet);
+
+  await balanceMetaPool(fixture);
+
+  return fixture;
+}
+
+async function balanceMetaPool(fixture) {
```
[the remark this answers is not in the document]
shahthepro: Good point, since everything involved is a stablecoin, I just assumed the difference would be negligible (if not equal). But totally agree that we should be using dollar values instead, will look into it

contracts/test/strategies/meta.fork-test.js
shahthepro: Thanks for catching this. It's actually the reward pool that should have balance of meta strategy contract, since the balance is staked when deposited. Will update the tests

shahthepro: It should be the CVXRewardPool that has some balance of LP tokens staked on behalf of OUSDMetaStrategy, have updated the tests

contracts/test/vault/vault.fork-test.js
```diff
@@ -0,0 +1,221 @@
...
+  });
+
+  describe("Assets & Strategies", () => {
+    it("Should NOT have any unknown assets", async () => {
+      const { vault } = fixture;
+      const assets = await vault.getAllAssets();
+
+      const knownAssets = [
+        // TODO: Update this every time a new asset is supported
+        "0xdAC17F958D2ee523a2206206994597C13D831ec7", // USDT
```
sparrowDom, line 149: these could also be read from addresses ( and others in this file as well)

sparrowDom (commented):
left a couple of more comments

contracts/fork-test.sh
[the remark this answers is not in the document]
sparrowDom: don't feel strongly about it, this way also works fine. Lets keep it

contracts/test/strategies/metapool-liquidity.fork-test.js
```diff
@@ -0,0 +1,310 @@
...
+  removeAllLiquidity,
+  removeLiquidityImbalanced,
+  get3CRVLiquidity,
+  withOUSDTitledMetapool,
+  withLiquidityOn3CRVTitledPool,
+  withCRV3TitledMetapool,
+  withLiquidityOnOUSDTitledPool,
+} = require("../_metastrategies-fixtures");
+
+forkOnlyDescribe("OUSD MetaPool Liquidity Test", function () {
```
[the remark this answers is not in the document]
shahthepro: Ah, makes sense. I assumed we also wanted to test if OUSDMetaPool works as intended without failing. If that isn't a problem, can totally remove this

contracts/test/vault/harvester.fork-test.js
```diff
@@ -0,0 +1,113 @@
...
+  describe("Rewards Config", () => {
+    it("Should have correct reward token config for CRV", async () => {
+      const { harvester, crv } = fixture;
+
+      const config = await harvester.rewardTokenConfigs(crv.address);
+
+      expect(config.allowedSlippageBps).to.equal(300);
+      expect(config.harvestRewardBps).to.equal(100);
+      expect(config.uniswapV2CompatibleAddr).to.equal(
+        "0xd9e1cE17f2641f24aE83637ab66a2cca9C378B9F"
```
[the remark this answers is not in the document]
shahthepro: Yep, I guess, we should aim to add tests for any OUSD contracts that interact with other external contracts on mainnet. I saw Harvester was using Uniswap, just wanted to make sure it works on fork

contracts/node.sh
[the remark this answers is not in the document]
sparrowDom: Yeah that works correctly when spinning up node via `fork-test.sh` but didn't work for me when using 2 terminal setup. Since the node is already running in the first terminal and has already read from deployment files before the `fork-test.sh` script executes.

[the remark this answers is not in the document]
sparrowDom: Thinking more on this I am not sure copying deployments from localhost to hardhat is a good idea. Since both actually fork from mainnet right? I mean hardhat network doesn't fork its state from the localhost network (both within hardhat configuration). Rather both fork mainnet. (do correct me if I am wrong)

I was under the impression: 
- hardhat network solely used for running tests in forked mode. Forks its state from mainnet
- localhost network used when running node in local - forked development mode. Forks its state from mainnet

There is smth going on where if I run `FORK=true IS_TEST=true yarn run node` it reads some deployment info from `deployments/localhost`. You can verify this by: 
- delete `deployments/hardhat` && `deployments/localhost`
- run `FORK=true IS_TEST=true yarn run node`

It will fail when trying to fetch OUSDProxy address. If you then `cp -r deployments/mainnet deployments/localhost` the running of the node will succeed. 

This leaves me under the impression that part of the deployments are being read from hardhat folder and part from the localhost folder.

contracts/test/_metastrategies-fixtures.js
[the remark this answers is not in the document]
sparrowDom: Cool, [the virtual price](https://curve.readthedocs.io/factory-pools.html#StableSwap.get_virtual_price) on each metapool is a 1e18 denominated value. You can use StableMath's mulTruncate to get dollar value: 
```
dollarValue = lpAmount.mulTruncate(virtualPrice)
```

contracts/test/_metastrategies-fixtures.js
```diff
@@ -0,0 +1,221 @@
...
+  const { domen, ousdMetaPool } = fixture;
+
+  receiver = receiver ? receiver.getAddress() : domen.getAddress();
+
+  const addSignature = "add_liquidity(uint256[2],uint256,address)";
+  const addMethod = await ousdMetaPool.connect(domen)[addSignature];
+  addMethod([ousdAmount, crv3Liquidity], 0, receiver);
+}
+
+async function removeAllLiquidity(fixture, user) {
```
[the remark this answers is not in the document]
sparrowDom: resolving since we don't need this for tests. (had this function in brownie to explore different ways of deploying / removing liquidity)

contracts/test/strategies/metapool-liquidity.fork-test.js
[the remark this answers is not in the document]
sparrowDom: Yup I think we can remove this bit

contracts/test/vault/vault.fork-test.js
```diff
@@ -0,0 +1,221 @@
...
+  let fixture;
+  beforeEach(async () => {
+    fixture = await loadFixture(defaultFixture);
+  });
+
+  describe("Admin", () => {
+    it("Should have the correct governor address set", async () => {
+      const { vault } = fixture;
+      expect(await vault.governor()).to.equal(
+        "0x72426BA137DEC62657306b12B1E869d43FeC6eC7"
```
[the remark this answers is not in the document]
sparrowDom: we decided it was better not to do that... resolving

contracts/node.sh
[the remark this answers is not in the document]
shahthepro: @sparrowDom I believe this is cleared now. I'm updating the readme so that it's clear, marking this as resolved. Feel free to ping me if you have any further queries regarding this

contracts/test/vault/vault.fork-test.js
shahthepro: As per our discussion, leaving them as-is for now, (with a comment about it at the top of the file). Happy to change it anytime

contracts/test/vault/vault.fork-test.js
```diff
@@ -0,0 +1,221 @@
...
+      expect(await vault.ousdMetaStrategy()).to.equal(
+        "0xEB85123eb2992296a835B422EfB0275C8A6C9745"
+      );
+    });
+  });
+
+  describe("Rebase", () => {
+    it(`Shouldn't be paused`, async () => {
+      const { vault } = fixture;
+      expect(await vault.rebasePaused()).to.be.false;
```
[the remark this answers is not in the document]
shahthepro: Yep, I added that as a defensive check, can remove it if it's unnecessary

contracts/test/_metastrategies-fixtures.js
shahthepro: Ah, yeah, that makes sense. Will change it

contracts/test/_metastrategies-fixtures.js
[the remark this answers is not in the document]
shahthepro: This is done

PR #1085, opened 2022-09-13
make it more convenient running fork tests

sparrowDom:
Instead of user configuring, fork tests check on their own if another local node is running and attach to it.


sparrowDom:
@shahthepro had a convenience idea, where the fork tests attach to the local node in case one is running, and spin up a new one when it is not. Feel free to merge if you like the idea / approach

franckc (approved):
LGTM

shahthepro:
This is actually a good idea too, thanks. Merging this in

shahthepro (approved):

Issue #1080, opened 2022-09-12
Rinkeby -> Goerli migration


franckc:
With The Merge happening in just a few days from now, we should migrate from Rinkeby to Goerli.
Alchemy mentioned they'll stop supporting Rinkeby by October 5th.

We should look for references and use of Rinkeby in our 3 repos and replace with Goerli:
 - [x] OUSD: https://github.com/OriginProtocol/origin-dollar/
 - [x] OUSD governance: https://github.com/OriginProtocol/ousd-governance
 - [x] OUSD analytics: https://github.com/OriginProtocol/ousd-analytics

Non exhaustive list of things to look for:
 - Chain ID 4
 - string "rinkeby" or "Rinkeby"

micahalcorn:
Related: https://github.com/OriginProtocol/ousd-governance/issues/286

tomhirst:
@ivanzhelyazkov Loose process I had planned for OUSD governance:

1. Deploy contracts to Goerli
1. Set up a new Heroku environment for Goerli (with database)
1. Set it to auto-deploy from the `staging` branch
1. Update the environment vars as appropriate (`https://eth-goerli.alchemyapi.io/v2/...` etc.)

ivanzhelyazkov:
OUSD governance contracts PR for the migration:
https://github.com/OriginProtocol/ousd-governance/pull/333
OUSD repo don't use a testnet environment - Master and Staging are connected to the mainnet deployment (there are deprecated Rinkeby references though)
OUSD analytics repo doesn't have any Rinkeby references

ivanzhelyazkov:
For OUSD-governance deployed new governance contracts on Goerli, connected the UI to the new contracts and replaced all Rinkeby references with Goerli, as well as tested the entire workflow from submitting a proposal to voting.
For Origin-dollar removed all Rinkeby references and tested that staging works correctly.

PR #1086, opened 2022-09-14
Remove rinkeby references

ivanzhelyazkov:
This PR removes all Rinkeby references from origin-dollar.
Rinkeby is not currently used for staging or stable, and there is no plan to support a testnet in the near future, since there are no good mocks for most of the protocols which OUSD interacts with.
References issue:
https://github.com/OriginProtocol/origin-dollar/issues/1080


dapp/src/utils/web3.js
```diff
@@ -1,7 +1,6 @@
 const networkInfo = {
   1: 'Mainnet',
   3: 'Ropsten',
```
franckc, line 3: Let's also cleanup those other testnet and only keep Mainnet and Localhost

dapp/src/utils/web3.js
```diff
@@ -26,7 +25,6 @@ export function getEtherscanHost(web3React) {
   const chainIdToEtherscan = {
     1: 'https://etherscan.io',
     3: 'https://ropsten.etherscan.io',
```
franckc, line 27: let's clean up ropsten since it is not used

franckc (commented):
Nice! Great cleanup. Couple small comments inline.

franckc:
@sparrowDom I took a pass but could you please also review this PR? Ty ser!

dapp/src/utils/web3.js
ivanzhelyazkov: Removed in the latest commit

dapp/src/utils/web3.js
ivanzhelyazkov: Removed in the latest commit

shahthepro (approved):
LGTM

franckc:
@sparrowDom Can you take a final look at this PR and then merge? ty :)

dapp/src/utils/web3.js
```diff
@@ -1,9 +1,5 @@
 const networkInfo = {
   1: 'Mainnet',
-  3: 'Ropsten',
```
sparrowDom, line 3: I think we can keep this in. From what I remember these map nework IDS to names. Could still be useful for the dapp to be able to tell the name of the network when a user switches to a network the dapp does not support.

sparrowDom (approved):
Sweet thanks for doing this massive cleanup 👏 . One minor comment inline

dapp/src/utils/web3.js
sparrowDom: Have checked the code, currently we fetch the name of the network only when successfully connected. So for current functionality of the dapp it is ok that they stay removed.

PR #1089, opened 2022-09-14
Add sentry on staging

ivanzhelyazkov:
The PR enables sentry on staging environment

### DIFF
--- a/dapp/src/utils/sentry.js
+++ b/dapp/src/utils/sentry.js
@@ -14,6 +14,7 @@ export const initSentry = () => {
   const distDir = `${config.serverRuntimeConfig.rootDir}/.next`
   Sentry.init({
     enabled: isProduction || isStaging,
+    environment: isProduction ? 'production' : 'staging',
     integrations: [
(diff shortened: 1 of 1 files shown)


dapp/src/utils/sentry.js
```diff
@@ -10,7 +13,7 @@ export const initSentry = () => {
   const config = getConfig()
   const distDir = `${config.serverRuntimeConfig.rootDir}/.next`
   Sentry.init({
-    enabled: process.env.NODE_ENV === 'production',
+    enabled: isProduction || isStaging,
     integrations: [
```
franckc, line 17: I think we are missing the environment. See https://docs.sentry.io/platforms/javascript/configuration/environments/
Perhaps something like this:
```
      environment: isProduction ? 'production' : 'staging',
```

ivanzhelyazkov: Ah yes, my bad - added

franckc (approved):
LGTM. Let's try this out!

PR #1100, opened 2022-09-16
Remove translucency on OUSD input

micahalcorn:
We previously ghosted OUSD as an optional input when OUSD is the swap output. However, the option is still selectable. This causes confusion for @matt-bullock and likely others.

Screen Shot 2022-09-15 at 9 30 01 PM

becomes

Screen Shot 2022-09-15 at 9 31 16 PM

### DIFF
--- a/dapp/src/components/buySell/SwapCurrencyPill.js
+++ b/dapp/src/components/buySell/SwapCurrencyPill.js
@@ -16,8 +16,8 @@ import {
 import { assetRootPath } from 'utils/image'
 import { _ } from 'fbt-runtime/lib/fbt'
 
-const CoinImage = ({ small, coin, isSemiTransparent = false }) => {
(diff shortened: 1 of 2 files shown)


dapp/src/components/buySell/SwapCurrencyPill.js
```diff
@@ -16,8 +16,8 @@ import {
 import { assetRootPath } from 'utils/image'
 import { _ } from 'fbt-runtime/lib/fbt'
 
-const CoinImage = ({ small, coin, isSemiTransparent = false }) => {
-  const className = `coin-image ${isSemiTransparent ? 'transparent' : ''}`
+const CoinImage = ({ small, coin }) => {
+  const className = `coin-image`
```
wonjun-opensource, line 20: We no longer need to use backtick here because this is not a template literal any more.

We can also make use of Eslint quotes rule in this project?

rolandpo (approved):
lgtm

PR #1104, opened 2022-09-19
Fix sentry on staging

ivanzhelyazkov:
Staging env variables are setup as NODE_ENV=production and STAGING=true, so this configures sentry to send notifications to the proper environment.

### DIFF
--- a/dapp/src/utils/sentry.js
+++ b/dapp/src/utils/sentry.js
@@ -3,7 +3,8 @@ import { RewriteFrames } from '@sentry/integrations'
 import getConfig from 'next/config'
 
 const isStaging = process.env.STAGING === 'true'
-const isProduction = process.env.NODE_ENV === 'production'
+// in staging environment NODE_ENV is also set to 'production'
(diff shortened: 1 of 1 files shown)


dapp/src/utils/sentry.js
```diff
@@ -14,7 +14,7 @@ export const initSentry = () => {
   const distDir = `${config.serverRuntimeConfig.rootDir}/.next`
   Sentry.init({
     enabled: isProduction || isStaging,
-    environment: isProduction ? 'production' : 'staging',
```
sparrowDom, line 17: I would leave this line as it was before and change the line 6 to (and add a comment): 
```
// in staging environment NODE_NEV is also set to 'production'
const isProduction = process.env.NODE_ENV === 'production' &&  process.env.STAGING !== 'true'
```

sparrowDom (commented):
Left 1 comment inline

sparrowDom (approved):
Nice thanks!

PR #1113, opened 2022-09-21
add inline assembly to solhint

sparrowDom:
In light of [Arbitrum hack](https://medium.com/@0xriptide/hackers-in-arbitrums-inbox-ca23272641a2) where assembly gas optimisation code was carelessly used we should be extra cautious when using assembly code as well. While this solhint rule change by itself won't fix much, it should serve as a nice reminder to be extra careful when using assembly. 

All our current assembly usages look ok and have added solhint exceptions.


franckc (approved):
Nice. I like how we are systematically learning from security attacks in the ecosystem.

PR #1116, opened 2022-09-22
Remove unused deploy files

shahthepro:
From what I get, `deploy.sh` script is no longer used since build/deploy happens on CloudBuild. I don't think it works either, since it's not updated to use `deploy.env` file change.

Feel free to close this, if this is still being used


franckc (approved):
LGTM but would be good for @sparrowDom to chime in.

sparrowDom (approved):
nice thanks for cleanup!

Issue #823, opened 2021-11-16
Some data not reset when wallet is disconnected


joshfraser:
Disconnecting your wallet in MetaMask resets your balance, but not your pending yield or lifetime earnings

<img width="1006" alt="Screen Shot 2021-11-15 at 11 27 41 PM" src="https://user-images.githubusercontent.com/982234/141918531-b53aeaa5-12f0-44a2-b477-1605ba22d609.png">

micahalcorn:
I had the opposite experience today: my balance remained but the other two numbers were removed.

talentdev0807:
This is working well on my side just have some delay
I think this comes from calling the apis asynchronously at the same time

franckc:
@talentdev0807 Unassigned the issue since it does not look like you are working on it?

PR #1119, opened 2022-09-25
Update ContractStore.walletConnected in AccountListener when the user disconnects the wallet

wonjun-opensource:
Issue https://github.com/OriginProtocol/origin-dollar/issues/823

`Disconnecting your wallet in MetaMask resets your balance, but not your pending yield or lifetime earnings`
and
`I had the opposite experience today: my balance remained but the other two numbers were removed.`

occurred because we were waiting until the fetch queries were completed in `AccountListener` after a wallet is disconnected. 

A good way to reproduce some of this behavior is to connect the wallet first. Then, open Chrome DevTools -> Network -> set it to "offline" to force fetch queries to fail. And then, disconnect the wallet and see that "Pending Yield" still has the value set on the header.

A better approach is to update `ContractStore` and set `ContractStore.walletConnected = false` when `useWeb3React` tells us that the account is no longer `active` as soon as the user disconnects their wallet.

### DIFF
--- a/dapp/src/services/apy.service.js
+++ b/dapp/src/services/apy.service.js
@@ -12,7 +12,7 @@ export default class ApyService {
           throw new Error(`Unexpected days param: ${days}`)
         }
         const response = await fetch(endpoint)
-        if (!response.ok) {
+        if (!response || !response.ok) {
           throw new Error(`Failed to fetch ${days} day APY`, err)
         }
         const json = await response.json()
--- a/contracts/test/vault/rebase.js
+++ b/contracts/test/vault/rebase.js
@@ -51,7 +51,7 @@ describe("Vault rebase pausing", async () => {
     );
   });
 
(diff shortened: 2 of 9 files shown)


dapp/src/services/apy.service.js
```diff
@@ -12,7 +12,7 @@ export default class ApyService {
           throw new Error(`Unexpected days param: ${days}`)
         }
         const response = await fetch(endpoint)
-        if (!response.ok) {
+        if (!response || !response.ok) {
```
shahthepro, line 15: nit: Could use the shorthand `!response?.ok`

shahthepro (approved):
Looks good to me, LGTM

sparrowDom (approved):
thanks for this LGTM

franckc:
@sparrowDom Can we merge this PR?

sparrowDom:
yes thanks for nudge @franckc

PR #1122, opened 2022-09-29
Gas optimisation proposal

sparrowDom:
Proposal to perform gas optimisation on check balance calls.


franckc (approved):
LGTM

sparrowDom:
We've decided to hold off on this one for now. While it does provide gas optimisation it also introduces possible new risks. We would like to minimise risk for now and the cost(risk)/benefit doesn't seem attractive.

PR #1123, opened 2022-09-29
Metastrategy - MEV guard

sparrowDom:
Guard agains MEV exploit by setting a withdrawal threshold on the Metapool strategy that can be set by Governor or Strategist. By default have it set to 1% so that there is no possibility of MEV. In case there is a crisis and we want to withdraw funds on a de-pegged asset, strategist can override that 1% to a higher value and withdraw all funds


sparrowDom:
@DanielVF @franckc could you check this PR out please? The tests are a bit flaky and will need additional work, but the good news is that the MEV test succeeds for alUsd. Confirming that the withdrawal slippage guard works as intended.

contracts/contracts/strategies/BaseConvexMetaStrategy.sol
```diff
@@ -160,6 +163,22 @@ abstract contract BaseConvexMetaStrategy is BaseCurveStrategy {
         revert("Invalid Metapool asset");
     }
 
+    /**
+     * @dev Sets max withdrawal slippage that is considered when removing
```
franckc, line 167: nit: document what unit is used for _maxWithdrawalSlippage. It's a percentage and looks like 100% = 1e18, 1% = 1e16

contracts/contracts/strategies/BaseConvexMetaStrategy.sol
```diff
@@ -18,6 +18,7 @@ import { Helpers } from "../utils/Helpers.sol";
 abstract contract BaseConvexMetaStrategy is BaseCurveStrategy {
     using StableMath for uint256;
     using SafeERC20 for IERC20;
+    event MaxWithdrawalSlippageUpdated(uint256 _previousAmount, uint256 _newAmount);
```
franckc, line 21: super nit: it's not really an amount but rather a percentage. so for code clarity could name these arguments something like `_prevMaxSlippagePercentage` and `_newMaxSlippagePercentage`.

Or just keep things brief since the event name is by itself very descriptive. like `_prev` and `_new`.

naming is hard :)

franckc (approved):
Couple nits but otherwise LGTM to merge after the unit tests pass.

contracts/contracts/strategies/ConvexGeneralizedMetaStrategy.sol
```diff
@@ -140,12 +140,15 @@ contract ConvexGeneralizedMetaStrategy is BaseConvexMetaStrategy {
 
         uint256 burnAmount = metapoolLPToken.balanceOf(address(this));
         if (burnAmount > 0) {
+            uint256 burnDollarAmount = burnAmount.mulTruncate(metapool.get_virtual_price());
```
shahthepro, line 143: What happens when the price is either too high or too low? Should we be worried about burning too much or too less OUSD?

Wondering if a min/max slippage tolerance should be added here too

shahthepro (approved):

contracts/contracts/strategies/BaseConvexMetaStrategy.sol
sparrowDom: yup good point, adding that

contracts/contracts/strategies/BaseConvexMetaStrategy.sol
sparrowDom: tnx, changed

contracts/contracts/strategies/ConvexGeneralizedMetaStrategy.sol
sparrowDom: This strategy is for non OUSD metapools. Here we burn metapoolLP tokens to receive only one of the coins in the pool - 3crvLP. That is done below with the `metapool.remove_liquidity_one_coin`  call

sparrowDom:
Merging this in and will fix tests in the main branch later. @DanielVF would still be cool that you can 👀 this one

sparrowDom:
Thanks for reviews @franckc & @shahthepro

contracts/contracts/strategies/BaseConvexMetaStrategy.sol
```diff
@@ -160,6 +165,27 @@ abstract contract BaseConvexMetaStrategy is BaseCurveStrategy {
...
+     * liquidity from Metapools.
+     *
+     * 1e18 == 100%, 1e16 == 1%
+     */
+    function setMaxWithdrawalSlippage(uint256 _maxWithdrawalSlippage)
+        external
+        onlyVaultOrGovernorOrStrategist
+    {
+        require(
+            _maxWithdrawalSlippage >= 1e15 && _maxWithdrawalSlippage <= 1e17,
```
DanielVF, line 179: This is one of those two edged swords. In a depeg, the max slippage could go way below 10%, and we might want to pay that. On the other hand, it's giving the strategist a lot of power to go further than that.

sparrowDom: Hmm good point, in a serious de-peg this isn't very useful. What do you think the sweet spot is? maybe 30%?

For someone to take advantage of this they would need to compromise the strategist wallet and perform a flash loan attack.

Issue #1126, opened 2022-10-04
Account history override improvements


micahalcorn:
We have a nice feature that allows anyone to view the OUSD account history of a given address by supplying an `account_override` query parameter to the URL.

Example: https://ousd.com/history?override_account=0xfA75aFe16bD3802D7Cd59900D33038db238A90A6

There are two improvements that would be helpful here:

1) Similar to https://github.com/OriginProtocol/origin-dollar/issues/1007, the account history cannot be viewed without having a wallet connected. This is not only inconvenient, but it's a little confusing if you're required to connect account B to be able to see the history for account A. Sites like [Zerion](https://app.zerion.io/0xfa75afe16bd3802d7cd59900d33038db238a90a6) and [Zapper](https://zapper.fi/account/0xfa75afe16bd3802d7cd59900d33038db238a90a6?) allow you to view any account just by typing in the address.
2) The big numbers at the top of the page should reflect the same account as the history does. Currently, it reflects the connected account regardless of whether or not the override is in effect.
<img width="688" alt="Screen Shot 2022-10-04 at 6 16 35 AM" src="https://user-images.githubusercontent.com/273937/193829200-f7f4b6a7-d4a2-469e-bf08-604db7768ae6.png">

joshfraser:
I like how Zapper and Zerion add your wallet address to the URL. This makes it really intuitive that you can change the URL to view any wallet you want.

alfredlouisw8:
hi, can I work on this?

micahalcorn:
> hi, can I work on this?

Sure, I've added you to the organization so that you can push a branch instead of forking. Let us know if you have any questions. 🙂

alfredlouisw8:
thank you! I've created a [PR](https://github.com/OriginProtocol/origin-dollar/pull/1130) for this, please kindly review it and tell me if I did anything wrong :)

alfredlouisw8:
Hi sorry to ping you @micahalcorn, I just want to check if I need to do something else here, I have created a PR for this issue :)

PR #1130, opened 2022-10-10
account history override improvements

alfredlouisw8:

Contract change checklist:
  - [ ] Unit tests pass

### DIFF
--- a/dapp/pages/history.js
+++ b/dapp/pages/history.js
@@ -1,6 +1,7 @@
 import React, { useEffect, useState } from 'react'
 import { useWeb3React } from '@web3-react/core'
 import { fbt } from 'fbt-runtime'
+import { useRouter } from 'next/router'
 
 import Layout from 'components/layout'
 import Nav from 'components/Nav'
@@ -11,6 +12,8 @@ import { assetRootPath } from 'utils/image'
 
 export default function History({ locale, onLocale }) {
   const { active } = useWeb3React()
+  const router = useRouter()
+  const overrideAccount = router.query.override_account
 
   return (
     <>
@@ -18,8 +21,8 @@ export default function History({ locale, onLocale }) {
         <Nav dapp page={'history'} locale={locale} onLocale={onLocale} />
         <div className="home d-flex flex-column">
           <BalanceHeader />
-          {active && <TransactionHistory />}
-          {!active && (
+          {(overrideAccount || active) && <TransactionHistory />}
+          {!overrideAccount && !active && (
             <div className="empty-placeholder d-flex flex-column align-items-center justify-content-start">
               <img src={assetRootPath('/images/wallet-icons.svg')} />
               <div className="header-text">
--- a/dapp/pages/history.js
+++ b/dapp/pages/history.js
@@ -1,7 +1,6 @@
 import React, { useEffect, useState } from 'react'
 import { useWeb3React } from '@web3-react/core'
 import { fbt } from 'fbt-runtime'
-import { useRouter } from 'next/router'
 
 import Layout from 'components/layout'
 import Nav from 'components/Nav'
@@ -9,14 +8,36 @@ import BalanceHeader from 'components/buySell/BalanceHeader'
 import TransactionHistory from 'components/TransactionHistory'
 import GetOUSD from 'components/GetOUSD'
 import { assetRootPath } from 'utils/image'
+import { useOverrideAccount } from '../src/utils/hooks'
+import ErrorModal from '../src/components/buySell/ErrorModal'
 
 export default function History({ locale, onLocale }) {
   const { active } = useWeb3React()
-  const router = useRouter()
-  const overrideAccount = router.query.override_account
+  const { overrideAccount, isValid } = useOverrideAccount()
+  const [showErrorModal, setShowErrorModal] = useState(!isValid)
 
+  const errorMap = [
+    {
+      errorCheck: (err) => {
+        return err === 'invalidAddress'
+      },
+      friendlyMessage: fbt(
+        "Overridden account's address is invalid",
+        "Overridden account's address is invalid"
+      ),
+    },
+  ]
   return (
     <>
+      {showErrorModal && (
+        <ErrorModal
+          error="invalidAddress"
+          errorMap={errorMap}
+          onClose={() => {
+            setShowErrorModal(false)
+          }}
+        />
+      )}
       <Layout locale={locale} onLocale={onLocale} dapp>
         <Nav dapp page={'history'} locale={locale} onLocale={onLocale} />
         <div className="home d-flex flex-column">
--- a/dapp/pages/history.js
+++ b/dapp/pages/history.js
@@ -11,25 +11,26 @@ import { assetRootPath } from 'utils/image'
 import { useOverrideAccount } from '../src/utils/hooks'
 import ErrorModal from '../src/components/buySell/ErrorModal'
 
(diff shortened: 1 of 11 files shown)


sparrowDom:
Hey I don't see any of the 2 issues addressed. When wallet disconnected and account override active the page doesn't look like it has the active account connected. Also the "big numbers" do not show the balance of the overridden account

Screenshot 2022-10-14 at 15 13 38

Regarding the malformed account, an error modal could be displayed.

alfredlouisw8:
Hi sorry if I misunderstood, I thought one of the issues was to show the transaction history even though the wallet is not connected if there is an account override
this is before
image

so are you saying the account should appear connected if there is an override account?

I will fix the big numbers issue

sparrowDom:
@alfredlouisw8 actually you are right. I think I was confused, because when I had an active wallet connected, and with account override, the active wallet on the top right got replaced with an overriden account. The transaction data shows nicely when account is overriden and wallet disconnected as you point out. 

I think it would be cool for behaviour of this PR to work as follows: 
- Wallet connected on the top right gets replaced with overriden account address. The green dot changes to a white one. And before the address there is a `readonly:` text. So end text in the top right wallet connection component would be: `[white_dot] reaonly: 0x[overriden_address]`. This would be shown only when dapp is transaction history page. Otherwise overriden_account UI behaviour of the wallet connection component should be ignored. Overriden account display on the transaction history page should override UI whether the user's wallet is connected or not. 
  - bonus item: When there is an overriden wallet active the dropdown of the wallet connected UI should be changes an only have a `Clear` button instead of Disconnect one. This should refresh the page without the `override_account` property. 
- Big numbers (the one in [the image of the issue](https://github.com/OriginProtocol/origin-dollar/issues/1126)) should also be overriden. That behaviour should be limited to transaction history page.

alfredlouisw8:
Hi I've updated my PR, please review it thanks!

dapp/pages/history.js
```diff
@@ -8,18 +8,42 @@ import BalanceHeader from 'components/buySell/BalanceHeader'
...
 import { assetRootPath } from 'utils/image'
+import { useOverrideAccount } from '../src/utils/hooks'
+import ErrorModal from '../src/components/buySell/ErrorModal'
 
 export default function History({ locale, onLocale }) {
   const { active } = useWeb3React()
+  const { overrideAccount, isValid } = useOverrideAccount()
+  const [showErrorModal, setShowErrorModal] = useState(!isValid)
 
+  const errorMap = [
```
shahthepro, line 19: All values on this object seem to be independent of the component's props. Let's move it outside the component?

dapp/src/utils/hooks.js
```diff
@@ -145,3 +147,18 @@ export function usePrevious(value) {
   })
   return ref.current
 }
+
+export function useOverrideAccount() {
+  const router = useRouter()
+  const isValid =
```
shahthepro, line 153: Going by this logic, `override_account` can have a value but it won't be always valid. But I don't see the `isValid` being used anywhere before using the `overrideAccount`.

Perhaps, make sure the function doesn't return `overrideAccount` if `isValid` is false

dapp/src/components/_AccountStatusContent.js
```diff
@@ -85,7 +87,19 @@ const AccountStatusContent = ({ className, onOpen }) => {
             </>
           )}
         </div>
-        {active && (
+        {overrideAccount && (
+          <div className="disconnect-box d-flex">
+            <a
+              className="btn-clear-blue w-100"
+              onClick={() => {
+                window.location = window.location.pathname
```
shahthepro, line 95: What's the purpose of using this? To not use the `overrideAccount`? If that's the case, why not use React Router's hooks and functions to do it?

alfredlouisw8: this is just to refresh the page without the override_account params. do you think we should use `router.push('/history)` instead?

dapp/src/components/AccountStatusDropdown.js
```diff
@@ -60,15 +63,26 @@ const AccountStatusDropdown = ({ className, showLogin, dapp }) => {
           )}
           {/* What causes !active && account? */}
           {dapp && !active && account && <div className="dot" />}
-          {active && !correctNetwork && (
+          {active && overrideAccount && (
+            <>
+              <div className="dot white" />
+              <div className="address">
+                {`${fbt('readonly', 'readonly')}: ${overrideAccount.substring(
```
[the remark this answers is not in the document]
alfredlouisw8: because there is a `readonly` text, the space is not enough
image

maybe we can pass a params to the function and make it look like this
image

dapp/src/utils/hooks.js
```diff
@@ -145,3 +147,18 @@ export function usePrevious(value) {
   })
   return ref.current
 }
+
+export function useOverrideAccount() {
+  const router = useRouter()
```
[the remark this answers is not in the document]
alfredlouisw8: sure

dapp/src/utils/hooks.js
alfredlouisw8: ` const overrideAccount =
    pathname === '/history' && override_account && isValid
      ? override_account
      : null`

There is this logic below so only it will only return `override_account` if `isValid` is true else it will return null

dapp/src/components/AccountStatusDropdown.js
```diff
@@ -60,15 +63,26 @@ const AccountStatusDropdown = ({ className, showLogin, dapp }) => {
...
               <div className="address">
                 {fbt('Wrong network', 'Wrong network')}
               </div>
             </>
           )}
-          {dapp && active && correctNetwork && (
+          {dapp && active && correctNetwork && !overrideAccount && (
             <>
               <div className="dot green" />
               <div className="address">{shortenAddress(account)}</div>
```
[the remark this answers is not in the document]
alfredlouisw8: what about this popup?
image

any changes to this popup when account is overridden?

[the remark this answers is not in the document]
shahthepro: @alfredlouisw8 Not sure about that, I'll defer it to @micahalcorn and @sparrowDom

dapp/src/components/AccountStatusDropdown.js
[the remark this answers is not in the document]
shahthepro: Ah, gotcha

dapp/src/utils/hooks.js
shahthepro: Ah, yes. I missed it

dapp/src/components/_AccountStatusContent.js
shahthepro: Yeah but `replace` rather than `push`, I'd say that'd be a lot cleaner

dapp/pages/history.js
alfredlouisw8: sure

dapp/src/components/_AccountStatusContent.js
alfredlouisw8: sure

franckc:
oops. Looks like we dropped the ball on reviewing the latest set of changes from you @alfredlouisw8 - really sorry about that!

@shahthepro do you want to take another pass at reviewing and if it looks good, let's merge! :)

shahthepro (approved):
LGTM

PR #1137, opened 2022-10-19
Metastrategy  M-01

sparrowDom:
**Issue description**
There are several instances throughout the code base where the strategies use their token
balance without validating it matches the expected value. This has two implications:
- It does not account for tokens that could be sent to the contract directly.
- It assumes the contract has received the expected number of tokens.
The most important example is [both](https://github.com/OriginProtocol/origin-dollar/blob/bfe0ac8e5d7c05b9bf1021fafb25e0aed8a6ed45/contracts/contracts/strategies/ConvexOUSDMetaStrategy.sol#L154) [instances](https://github.com/OriginProtocol/origin-dollar/blob/bfe0ac8e5d7c05b9bf1021fafb25e0aed8a6ed45/contracts/contracts/strategies/ConvexOUSDMetaStrategy.sol#L176) when the ConvexOUSDMetaStrategy
attempts to burn all its OUSD tokens. This would [fail](https://www.google.com/url?q=https://github.com/OriginProtocol/origin-dollar/blob/bfe0ac8e5d7c05b9bf1021fafb25e0aed8a6ed45/contracts/contracts/vault/VaultCore.sol%23L254&sa=D&source=apps-viewer-frontend&ust=1666249263576262&usg=AOvVaw1U9BHZtY4FLtXqHsRRU5p7&hl=en) if its balance exceeds the acceptable
range, which could occur if someone sends OUSD directly to the strategy.

Additionally, in the interest of predictability:
- When the [amount of tokens to deposit](https://github.com/OriginProtocol/origin-dollar/blob/bfe0ac8e5d7c05b9bf1021fafb25e0aed8a6ed45/contracts/contracts/strategies/ConvexOUSDMetaStrategy.sol#L55) is already known, that value can be used instead
of the [contract's balance](https://github.com/OriginProtocol/origin-dollar/blob/bfe0ac8e5d7c05b9bf1021fafb25e0aed8a6ed45/contracts/contracts/strategies/ConvexOUSDMetaStrategy.sol#L69). Alternatively, only the difference [needs to be minted](https://github.com/OriginProtocol/origin-dollar/blob/bfe0ac8e5d7c05b9bf1021fafb25e0aed8a6ed45/contracts/contracts/strategies/ConvexOUSDMetaStrategy.sol#L65) in the first
place.
- When the [amount of tokens to receive](https://github.com/OriginProtocol/origin-dollar/blob/bfe0ac8e5d7c05b9bf1021fafb25e0aed8a6ed45/contracts/contracts/strategies/ConvexOUSDMetaStrategy.sol#L98) is already known, it can be used as [the minimum
withdrawal amount](https://github.com/OriginProtocol/origin-dollar/blob/bfe0ac8e5d7c05b9bf1021fafb25e0aed8a6ed45/contracts/contracts/strategies/ConvexOUSDMetaStrategy.sol#L151).

Lastly, in the interest of local reasoning and robustness, consider explicitly confirming that the
Curve protocol returns the expected number of tokens, rather than assuming it respects the
stated minimum thresholds.


DanielVF:
I'm really not sure about this one. Somewhat inherent in the way we withdraw from these contracts generates hundreds to thousands of dollars of "dust". I think we do want to have an easy way to have this invested back in on the next deposit.

Also, if people send us money, that's a good thing from the protocol's point of view. If we are right up against a mint or burn limit, then we could rerun the transaction with a different amount or alter the limit later.

What do you think?

sparrowDom:
To sum up the changes this PR makes: 
1. [Generalized metapool strat]: Deposit to Gauge the exact amount of LP received by adding metapool liquidity 
2. [Generalized metapool strat]: Withdraw metapoolLP amount is exact amount removed from Gauge
3. [OUSD metapool strat]: Only add amount of OUSD that has been minted for strategy
4. [OUSD metapool strat]: Deposit to Gauge the exact amount of LP received by adding metapool liquidity 
5. [OUSD metapool strat]: Add min OUSD & 3Crv amounts when withdrawing metapool liquidity
6. [OUSD metapool strat]: Burn exact amount of OUSD removed from Metapool
7. [OUSD metapool strat][already in base branch]: Withdraw metapoolLP amount is exact amount removed from Gauge

I think that 1&2 together and 4&7 should make sure that there are is no metapoolLP dust lying around on the contract as a result of contract operations (this doesn't include someone sending OUSD & metapoolLP to the contract). Not sure if the same holds for 3poolLP dust, though changes in this PR don't affect that. 

3&6 should take care of no OUSD dust being on the contract. 

5 is for extra security. 

I am personally fine with these changes, since any dust laying around on the contract would indicate an either unexpected contract behaviour or some other entity sending funds to the contract.

DanielVF:
Not sure if its this change or from the main PR, but the `withdrawAllFromStrategy()` seems to be failing for me on the meta strategy "Transaction reverted without a reason string"

```python
vault_admin.reallocate(COMP_STRAT, META_STRAT, [usdc], [10000000e6], {'from': STRATEGIST})
vault_admin.reallocate(META_STRAT, COMP_STRAT, [usdc], [5000000e6], {'from': STRATEGIST})
vault_admin.withdrawAllFromStrategy(META_STRAT, {'from': STRATEGIST}) # Revert
```

DanielVF:
I'm seeing 3crv dust left on the strategy after doing a withdrawal from it:

```python
vault_admin.reallocate(COMP_STRAT, BM_STRAT, [usdc], [10000000e6], {'from': STRATEGIST})
vault_admin.reallocate(BM_STRAT, COMP_STRAT, [usdc], [5000000e6], {'from': STRATEGIST})
print(threecrv.balanceOf(BM_STRAT)) # 6757660433475890581
```

sparrowDom:
Will look into this

sparrowDom:
> I'm seeing 3crv dust left on the strategy after doing a withdrawal from it:
> 

The changes in this PR shouldn't have any effect on the 3crv dust. Have confirmed this by running the script on base metastrategy branch and this branch. The same amount of `3CRVLP` dust is left on the strategy contract. Have also confirmed that there is no `metapoolLP` left on the contract: 

```python
from metastrategy import *

with TemporaryFork():
  vault_admin.reallocate(COMP_STRAT, FRAX_STRATEGY, [usdc], [10000000e6], {'from': STRATEGIST})
  vault_admin.reallocate(FRAX_STRATEGY, COMP_STRAT, [usdc], [5000000e6], {'from': STRATEGIST})
  print(frax_metapool.balanceOf(FRAX_STRATEGY)) # 0
  print(threepool_lp.balanceOf(FRAX_STRATEGY)) # 59138978035229957838 -> same on both branches
```

sparrowDom:
Figured out what the issue with withdrawal was. `maxWithdrawalSlippage` was set to 0 on all contracts. Fix it here: https://github.com/OriginProtocol/origin-dollar/pull/1151

sparrowDom:
@DanielVF good catch! Have removed the metapoolLP contract token balance from the `checkBalance` calculation. 

I guess someone sending metapoolLP tokens to the strategy could inflate its perceived balance which wouldn't all be liquid.

contracts/contracts/strategies/ConvexOUSDMetaStrategy.sol
```diff
@@ -148,12 +143,19 @@ contract ConvexOUSDMetaStrategy is BaseConvexMetaStrategy {
...
         );
 
-        // always withdraw all of the available metapool LP tokens (similar to how we always deposit all)
-        // slither-disable-next-line unused-return
-        metapool.remove_liquidity(lpToBurn, [uint256(0), uint256(0)]);
-        IVault(vaultAddress).burnForStrategy(
-            metapoolMainToken.balanceOf(address(this))
+        // calculate the min amount of OUSD expected for the specified amount of LP tokens
+        // slither-disable-next-line divide-before-multiply
+        uint256 minOUSDAmount = lpToBurn *
```
DanielVF, line 148: This line really doesn't feel right to me.

I assume it's supposed to work like:

`min dollar value of pool lp tokens` -`3pool dollar value withdrawn` - 1 < `ousd dollar value withdrawn`

However, the `lpToBurn * metapool.get_virtual_price()` is multiplying two 1e18 numbers together without division. This should mean that this results in a crazy high number, which I would expect would break withdraws due to not meeting the minimum.

Secondly, the `- num3CrvTokens` line feels like it's trying to determine the dollar value of the 3pool LP tokens, but since it's not adjusted for the virtual price, this would make the subtraction side too small, which should make it break with false alarms when the exchange rates are close to equal.

Given that it looks like there are two reasons that this code would not work, I wonder if I'm wrong or if the fork tests are.

Stepping back, it looks like this code is just trying to duplicate what the curve calculations are doing. Is this code gaining us anything other than validating curve math? If not, I would lean toward switching back to using a zero for the OUSD minimum.

DanielVF: I've confirmed with fork testing that withdraws are broken. If I changed it to `uint256 minOUSDAmount = 0`, it can withdraw funds.

DanielVF: image

PR #1138, opened 2022-10-19
Metastrategy L-01

sparrowDom:
**Issue description:**
The `withdraw` function of the `BaseCurveStrategy` determines the number of 3CRV
tokens to burn by using the price to [slightly overestimate](https://github.com/OriginProtocol/origin-dollar/blob/bfe0ac8e5d7c05b9bf1021fafb25e0aed8a6ed45/contracts/contracts/strategies/BaseCurveStrategy.sol#L134-L137) the desired value, [determining the
corresponding amount](https://github.com/OriginProtocol/origin-dollar/blob/bfe0ac8e5d7c05b9bf1021fafb25e0aed8a6ed45/contracts/contracts/strategies/BaseCurveStrategy.sol#L141-L144) of stablecoin, and then [scaling down the LP amount](https://github.com/OriginProtocol/origin-dollar/blob/bfe0ac8e5d7c05b9bf1021fafb25e0aed8a6ed45/contracts/contracts/strategies/BaseCurveStrategy.sol#L146) linearly to match
the required stablecoin value. Instead, consider directly querying the amount of LP tokens to
burn with the [calc_token_amount function](https://curve.readthedocs.io/exchange-pools.html#StableSwap.calc_token_amount), and then adjusting for fees. If desired, the
amount could be validated with the [calc_withdraw_one_coin function](https://curve.readthedocs.io/exchange-pools.html#StableSwap.calc_withdraw_one_coin).

### DIFF
--- a/contracts/contracts/strategies/BaseCurveStrategy.sol
+++ b/contracts/contracts/strategies/BaseCurveStrategy.sol
@@ -118,41 +118,19 @@ abstract contract BaseCurveStrategy is InitializableAbstractStrategy {
         );
 
         uint256 coinIndex = _getCoinIndex(_asset);
-        int128 curveCoinIndex = int128(uint128(coinIndex));
-        // Calculate the max amount of the asset we'd get if we withdrew all the
-        // platform tokens
         ICurvePool curvePool = ICurvePool(platformAddress);
 
-        uint256 virtual_price = curvePool.get_virtual_price();
-        /**
-         * Add a 5% threshold to help calculate required amount of crv3Tokens:
-         *  - convert asset to 18 decimals
-         *  - divide by virtual price to get 3Crv equivalent
-         *  - add a 5% threshold to it
-         */
-        // slither-disable-next-line divide-before-multiply
-        uint256 crv3TokensTreshold = (((_amount.scaleBy(
-            18,
-            Helpers.getDecimals(_asset)
-        ) * 105) * 1e18) / virtual_price) / 100;
-
-        // Calculate how many platform tokens we need to withdraw the asset
-        // amount in by adding 5% of overhead to required withdrawn amount
-        uint256 thresholdAmountReceived = curvePool.calc_withdraw_one_coin(
-            crv3TokensTreshold,
-            curveCoinIndex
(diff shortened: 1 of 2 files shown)


DanielVF:
This is MUCH cleaner!

My only concern is that this always overestimates the fees. While that's always "safe" in terms of always working, it does mean that there is always dollars of "dust" left over after this.

This is because when withdrawing imbalanced liquidity you are not charged fees on the funds you are withdrawing that already match your desired coin types/amounts. Fees are only on the funds that must be "swapped" to match your target withdraw.

At current rates, withdrawing 9,788,809 DAI from 3pool, would estimate the fees at 957 3crv, and but actually only cost 469. If we withdrew in perfect match with the balances in the pool, the actual fees would be zero.

I'm not sure at the moment if/how we should handle this. 

```python
import random
from world import *

WINTER="0xe74b28c2eae8679e3ccc3a94d5d0de83ccb84705"
threecrv = Contract.from_explorer("0x6c3f90f043a72fa612cbac8115ee7e52bde6e490")

threecrv.approve(threepool, 1e70, {'from': WINTER})
for i in range(0, 10):
	amount = random.randint(0, 1e7 * 1e18)
	expected = threepool.calc_token_amount([amount, 0, 0], False)
	fee = (expected * threepool.fee()) / 1e10
	expected += fee
	with TemporaryFork():
		before = threecrv.balanceOf(WINTER)
		threepool.remove_liquidity_imbalance([amount, 0,0], 1e70, {'from': WINTER})
		actual = before - threecrv.balanceOf(WINTER)
		diff = expected - actual
		print("Removed %d DAI (%s), %d diff (%s), expected LP cost %d (%s), actual %d (%s), fee %d (%s)" %(
			amount,
			c18(amount),
			diff,
			c18(diff),
			expected,
			c18(expected),
			actual,
			c18(actual),
			fee,
			c18(fee)
			))
```

sparrowDom:
I've tried a couple of ways to figure out how much of that fee is applicable to the amount we are withdrawing, but the end calculation gets more complicated than the original codebase

sparrowDom:
I suggest we close this one for now.

PR #1143, opened 2022-10-20
Metastrategy L-05 [Misleading comments]

sparrowDom:
The code base contains several misleading comments:
- The `collectRewardTokens` function of the `BaseConvexMetaStrategy` contract
[claims](https://github.com/OriginProtocol/origin-dollar/blob/bfe0ac8e5d7c05b9bf1021fafb25e0aed8a6ed45/contracts/contracts/strategies/BaseConvexMetaStrategy.sol#L190) to send rewards to the vault, but it sends them to the `harvester` address.
- The `depositAll` function of the `BaseCurveStrategy` contract [does not accurately
describe](https://github.com/OriginProtocol/origin-dollar/blob/bfe0ac8e5d7c05b9bf1021fafb25e0aed8a6ed45/contracts/contracts/strategies/BaseCurveStrategy.sol#L92-L93) how all descendant contracts retrieve the 3CRV token address.
- The [comment](https://github.com/OriginProtocol/origin-dollar/blob/bfe0ac8e5d7c05b9bf1021fafb25e0aed8a6ed45/contracts/contracts/strategies/ConvexOUSDMetaStrategy.sol#L57-L62) explaining how much OUSD to add in the `ConvexOUSDMetaStrategy`
states that it could mint less OUSD if the metapool has too much, but it always adds [at
least as much OUSD as 3CRV](https://github.com/OriginProtocol/origin-dollar/blob/bfe0ac8e5d7c05b9bf1021fafb25e0aed8a6ed45/contracts/contracts/strategies/ConvexOUSDMetaStrategy.sol#L54). Moreover, it claims the metapool will end up balanced.
However [it won't add more than twice as much OUSD](https://github.com/OriginProtocol/origin-dollar/blob/bfe0ac8e5d7c05b9bf1021fafb25e0aed8a6ed45/contracts/contracts/strategies/ConvexOUSDMetaStrategy.sol#L55) (by value) than 3CRV, even if
that's necessary to balance the metapool.
- The `mintForStrategy` function of the `VaultCore` contract [claims](https://github.com/OriginProtocol/origin-dollar/blob/bfe0ac8e5d7c05b9bf1021fafb25e0aed8a6ed45/contracts/contracts/vault/VaultCore.sol#L115-L116) that it cannot use
the `nonReentrant` modifier because it clashes with the `BaseCurveStrategy
contract's modifier. However, the two modifiers do not interact with each other. Instead,
it is the `nonReentrant` modifier on the `allocate` [function](https://github.com/OriginProtocol/origin-dollar/blob/bfe0ac8e5d7c05b9bf1021fafb25e0aed8a6ed45/contracts/contracts/vault/VaultCore.sol#L286) that would cause the
conflict.
- The `@param` [comment](https://github.com/OriginProtocol/origin-dollar/blob/bfe0ac8e5d7c05b9bf1021fafb25e0aed8a6ed45/contracts/contracts/strategies/ConvexOUSDMetaStrategy.sol#L98) describing the `_lpWithdraw` function of the
`ConvexOUSDMetaStrategy` contract claims it is the number of Convex LP tokens to
redeem, but it is the number of 3CRV tokens to retrieve.

Consider correcting these comments.


contracts/contracts/vault/VaultCore.sol
```diff
@@ -112,8 +112,9 @@ contract VaultCore is VaultStorage {
      * @dev Mint OUSD for OUSD Meta Strategy
      * @param _amount Amount of the asset being deposited
      *
-     * Notice: can't use nonReentrant modifier since BaseCurveStrategy's deposit
-     * already has that modifier present
+     * Notice: can't use `nonReentrant` modifier since the `mint` function can
+     * be called within the same transaction as `mintForStrategy` and can cause
```
DanielVF, line 116: Being called multiple times in the same transaction would be okay? The problem is that it is being called from _inside_ an already nonReentrant method by the vault admin to reallocate/deposit/withdraw funds?

sparrowDom: Thanks, corrected

contracts/contracts/vault/VaultCore.sol
```diff
@@ -112,8 +112,9 @@ contract VaultCore is VaultStorage {
      * @dev Mint OUSD for OUSD Meta Strategy
      * @param _amount Amount of the asset being deposited
      *
-     * Notice: can't use nonReentrant modifier since BaseCurveStrategy's deposit
-     * already has that modifier present
+     * Notice: can't use `nonReentrant` modifier since the `mint` function can
```
DanielVF, line 115: If we ever do the `mint -> allocate -> mintForStrategy` route we are doomed from allowing users to control funds moving into or out of a manipulable strategy.

I think `reallocate -> mintForStrategy` is the more likely path.

sparrowDom: Agree and that though did cross my mind. I agree and expect this to never happen on mainnet, because of the reason you point out. Test suite on the other hand is using this exact path. And once we roll out the `strategy testing framework` I expect all strategies tested will need to support `mint -> allocate -> mintForStrategy`.

If we were to only use `reallocate -> mintForStrategy` path we could actually have the non reentrant modifier on `mintForStrategy` since `reallocate` doesn't have non reentrancy guard. 

@DanielVF could you check the comment on `burnForStrategy` as well? That one potentially collides with redeem's non reentrancy. (that again probably won't be part of any mainnet setups).

DanielVF: I'm very concerned about having comments in the code mentioning the necessity of supporting a series of action that are a near guarantee of a protocol funds loss. If some future person is reading through the code looking to understand the contracts, they could think this was an important good thing. That's why I would rather these two comments have the slightly longer, but expected in actual use situtation.

sparrowDom:
@DanielVF corrected the comment. Decided not to merge yet since any possible post PR changes would mess up the clean commit hashes we pass on to OZ audit team

DanielVF (approved):

PR #1145, opened 2022-10-20
Metastrategy L-08 [Inconsistent storage gaps]

sparrowDom:
**Issue description:**
When using the proxy pattern for upgrades, it is common practice to include storage gaps on
parent contracts to reserve space for potential future variables. The size is typically chosen so
that all contracts have the same number of variables (usually 50). However, the code base uses
inconsistent sizes and does not always include a gap at all. In particular:
- The `InitializableAbstractStrategy` contract [reserves 98 slots](https://github.com/OriginProtocol/origin-dollar/blob/bfe0ac8e5d7c05b9bf1021fafb25e0aed8a6ed45/contracts/contracts/utils/InitializableAbstractStrategy.sol#L59), bringing the total
storage usage up to 106.
- The `BaseCurveStrategy` has no storage gaps.
The `BaseConvexMetaStrategy` contract [reserves 30 slots](https://github.com/OriginProtocol/origin-dollar/blob/bfe0ac8e5d7c05b9bf1021fafb25e0aed8a6ed45/contracts/contracts/strategies/BaseConvexMetaStrategy.sol#L51), bringing the total storage
usage up to 39.

Consider using consistently sized storage gaps in all contracts that have not yet been
deployed. For contracts that cannot be changed, because they are ancestors of live contracts
in the code base, consider documenting the unusual storage size to facilitate safe upgrades.

### DIFF
--- a/contracts/contracts/strategies/BaseConvexMetaStrategy.sol
+++ b/contracts/contracts/strategies/BaseConvexMetaStrategy.sol
@@ -48,7 +48,7 @@ abstract contract BaseConvexMetaStrategy is BaseCurveStrategy {
     uint128 crvCoinIndex;
     uint128 mainCoinIndex;
 
-    int256[41] private __reserved;
+    int256[41] private ___reserved;
 
     /**
      * Initializer for setting up strategy internal state. This overrides the
--- a/contracts/contracts/strategies/BaseConvexMetaStrategy.sol
+++ b/contracts/contracts/strategies/BaseConvexMetaStrategy.sol
@@ -48,7 +48,7 @@ abstract contract BaseConvexMetaStrategy is BaseCurveStrategy {
     uint128 crvCoinIndex;
     uint128 mainCoinIndex;
 
-    int256[30] private __reserved;
+    int256[41] private __reserved;
 
     /**
      * Initializer for setting up strategy internal state. This overrides the
--- a/contracts/contracts/strategies/ThreePoolStrategy.sol
+++ b/contracts/contracts/strategies/ThreePoolStrategy.sol
@@ -19,7 +19,7 @@ import { Helpers } from "../utils/Helpers.sol";
  * IMPORTANT(!) If ThreePoolStrategy needs to be re-deployed, it requires new
  * proxy contract with fresh storage slots. Changes in `BaseCurveStrategy`
  * storage slots would break existing implementation.
- * 
+ *
  * Remove this notice if ThreePoolStrategy is re-deployed
  */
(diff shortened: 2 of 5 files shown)


contracts/contracts/strategies/ThreePoolStrategy.sol
```diff
@@ -15,6 +15,11 @@ import { IERC20, BaseCurveStrategy } from "./BaseCurveStrategy.sol";
 import { StableMath } from "../utils/StableMath.sol";
 import { Helpers } from "../utils/Helpers.sol";
 
+/*
+ * IMPORTANT(!) If TreePoolStrategy is every to be re-deployed, it needs new
```
DanielVF, line 19: "TreePool"

DanielVF:
How do these storage slot changes affect our currently deployed convex strategy?

contracts/contracts/strategies/BaseConvexMetaStrategy.sol
```diff
@@ -48,7 +48,7 @@ abstract contract BaseConvexMetaStrategy is BaseCurveStrategy {
     uint128 crvCoinIndex;
     uint128 mainCoinIndex;
 
-    int256[30] private __reserved;
```
DanielVF, line 51: Why is this changed to 41?

contracts/contracts/strategies/ThreePoolStrategy.sol
sparrowDom: tnx fixed

contracts/contracts/strategies/BaseConvexMetaStrategy.sol
sparrowDom: As per audit recommendation that base contracts take up 50 storage slots combining local variables with the __reserve gap.

sparrowDom:
> How do these storage slot changes affect our currently deployed convex strategy?

Hmm good point this will break it. I see 3 things we could do: 
1. Do not refactor functionality of `BaseCurveStrategy` out of `ConvexStrategy`. The metastrategies still utilise `BaseCurveStrategy`, but `ConvexStrategy` does not. Downside duplicated functionality 
2. As with ThreePool strategy add a notice that re-deploying ConvexStrategy also requires deployment of a new proxy. 
3. Remove the `_reserved` gap in `BaseCurveStrategy` and risk not having the ability to add local variables if needed. 

I would go for option `2.` thoughts?

DanielVF:
Yes, we can add the comment to that strategy.

sparrowDom:
Done, merging....

PR #1148, opened 2022-10-20
Metastrategy N-03 [Unindexed events]

sparrowDom:
**Issue description:**
Consider [indexing](https://docs.soliditylang.org/en/latest/contracts.html#events) the VaultCore [events](https://github.com/OriginProtocol/origin-dollar/blob/bfe0ac8e5d7c05b9bf1021fafb25e0aed8a6ed45/contracts/contracts/vault/VaultStorage.sol#L28-L50) to assist the task of off-chain services searching
and filtering for specific events.

**Note:**
I have not changed existing events to indexed. Would probably be confusing since events already emitted would have not shown up in the index.

### DIFF
--- a/contracts/contracts/vault/VaultStorage.sol
+++ b/contracts/contracts/vault/VaultStorage.sol
@@ -37,7 +37,7 @@ contract VaultStorage is Initializable, Governable {
     event RebasePaused();
     event RebaseUnpaused();
     event VaultBufferUpdated(uint256 _vaultBuffer);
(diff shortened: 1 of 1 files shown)


DanielVF:
Yes, good to not change the others. That would mess up our monitoring.

I don't think we need to make this change. Events like Transfer are good candidates for indexes, because a lot of the use of the events is from the "users" point of view. However here the use of this event all about config changes to this contract. I would rather keep consistency, and everything we use the event for is tracked based off the contract emitting the event.

sparrowDom:
Agreed, closing this one

PR #1150, opened 2022-10-21
Metastrategy L-03 [Misusing safeApprove]

sparrowDom:
**Issue description:**
The `BaseCurveStrategy` [contract](https://github.com/OriginProtocol/origin-dollar/blob/bfe0ac8e5d7c05b9bf1021fafb25e0aed8a6ed45/contracts/contracts/strategies/BaseCurveStrategy.sol#L261-L262) and the `BaseConvexMetaStrategy` [contract](https://github.com/OriginProtocol/origin-dollar/blob/bfe0ac8e5d7c05b9bf1021fafb25e0aed8a6ed45/contracts/contracts/strategies/BaseConvexMetaStrategy.sol#L138) provides
functionality to grant addresses an infinite allowance. However, they bypass the safety
mechanism in `safeApprove` that prevents changing the allowance between two non-zero
values, which is intended to prevent front-running attacks that spend both allowances. In this
case, since the intended allowance is unlimited, the possibility of front-running is irrelevant.
Nevertheless, we consider it bad practice to use a safety mechanism while bypassing its
additional requirements. Consider using the standard `approve` function and validating that it
succeeds.

### DIFF
--- a/contracts/contracts/strategies/BaseConvexMetaStrategy.sol
+++ b/contracts/contracts/strategies/BaseConvexMetaStrategy.sol
@@ -138,17 +138,17 @@ abstract contract BaseConvexMetaStrategy is BaseCurveStrategy {
     function _approveBase() internal override {
         IERC20 pToken = IERC20(pTokenAddress);
         // 3Pool for LP token (required for removing liquidity)
-        pToken.safeApprove(platformAddress, 0);
-        pToken.safeApprove(platformAddress, type(uint256).max);
+        pToken.approve(platformAddress, 0);
+        pToken.approve(platformAddress, type(uint256).max);
         // Gauge for LP token
-        metapoolLPToken.safeApprove(cvxDepositorAddress, 0);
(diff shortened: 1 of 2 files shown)


DanielVF:
I think I didn't work into the details of what safeApprove does. We may actually want to keep it.

SafeApprove is a local OZ function that does three things:

- Requires you to go back to zero allowance before changing an allowance. Don't need this.
- Handles USDT's non standard non-boolean return on approve. We do need this.
- Checks that the boolean version of approve returns true. We do need this.

Given that, we probably want to keep it.

https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v4.7.3/contracts/token/ERC20/utils/SafeERC20.sol#L46-L59

sparrowDom:
Ahh right I hadn't realised it also handles the special USDT return type and also requires return types being true if existing. Closing this issue than and marking it `wont-do`

PR #1151, opened 2022-10-22
Metastrategy Custom-01

sparrowDom:
**Origin identified issue:**
`maxWithdrawalSlippage` in `BaseConvexMetaStrategy` had its value initialised on contract creation. That poses a problem with proxy contracts since the value is set in the storage slot space of the implementation contract instead of the proxy contract: 
Screenshot 2022-10-22 at 17 32 43

Meaning that all generalized metastrategy would have `maxWithdrawalSlippage` set to 0 instead of 1e16. 

This PR moves setting of the value to initialise function which is always called in the storage slot space of the proxy contract.


shahthepro (approved):
LGTM. 

One question that may be related though, In [this line](https://github.com/OriginProtocol/origin-dollar/blob/sparrowDom/metastrategy-C-01/contracts/deploy/043_convexOUSDMeta.js#L63), we could pass the data to call the `initialize` method with its args (that is done on the next few lines as a separate call). Any reason why it's not done that way and is split into two calls?

sparrowDom:
@shahthepro yup I see that. You are right, we could do it all inside one call. The end effect would be the same (since the deployer is already the owner of the contract and no one else can call initialize) and either of the 2 ways are OK I think.

DanielVF:
One thing I'm concerned with here is that the existing fork tests did not catch this issue. I think we should make sure that beyond just testing that this value is set, we should have a test that would have failed given that `withdrawAll()` was not working.

sparrowDom:
> One thing I'm concerned with here is that the existing fork tests did not catch this issue. I think we should make sure that beyond just testing that this value is set, we should have a test that would have failed given that `withdrawAll()` was not working.

That is a good point. Also adding that test in

sparrowDom:
The test that has failed is not a big issue since the fix for that one is in this PR: https://github.com/OriginProtocol/origin-dollar/pull/1152

DanielVF (changes requested):
-

DanielVF (approved):

PR #1152, opened 2022-10-24
Remove lower bound -> C-02

sparrowDom:
Remove lower bound of `maxWithdrawalSlippage`

Main reason for doing this is so that in test environment we can set the `maxWithdrawalSlippage` to 0% and confirm the correct behaviour of curve StableSwap contracts in combination with generalized metastrategy contract. When doing `withdrawAll` with too small `maxWithdrawalSlippage` the transaction should revert.

### DIFF
--- a/contracts/test/strategies/generalized-meta.fork-test.js
+++ b/contracts/test/strategies/generalized-meta.fork-test.js
@@ -234,7 +234,7 @@ metastrategies.forEach(
 
             expect(error).to.be.oneOf([
               "Transaction reverted without a reason string",
-              `'Not enough coins removed'`,
+              `\\'Not enough coins removed\\'`,
             ]);
 
             // should not revert when slippage tolerance set to 10%
--- a/contracts/test/strategies/generalized-meta.fork-test.js
+++ b/contracts/test/strategies/generalized-meta.fork-test.js
@@ -27,7 +27,7 @@ const metastrategies = [
     rewardPoolAddress: "0xDBFa6187C79f4fE4Cda20609E75760C5AaE88e52",
     // metapool implementation wont allow tilting of the pools the way this test does it
     // and then withdrawing liquidity
-    skipMewTest: true,
+    skipMewTest: false,
   },
   {
     token: "USDD",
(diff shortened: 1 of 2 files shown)


DanielVF:
@sparrowDom Could you explain why this set of changes?

sparrowDom:
> @sparrowDom Could you explain why this set of changes?

Sorry my bad. Have edited original description with more info.

contracts/test/strategies/generalized-meta.fork-test.js
```diff
@@ -218,16 +218,24 @@ metastrategies.forEach(
 
             await fixture.metaStrategy
               .connect(sGovernor)
-              .setMaxWithdrawalSlippage(ousdUnits("0.001"));
+              .setMaxWithdrawalSlippage(ousdUnits("0"));
             await tiltToMainToken(fixture);
 
-            await expect(
-              vault
+            let error = false;
```
DanielVF, line 224: Is there a reason for this longer form of catching a revert and checking the error message than using the hardhat helper, `to.be.revertedWith()`?


https://github.com/OriginProtocol/origin-dollar/blob/1bb1a410c364149980af815b536ef89328c01df6/contracts/test/vault/index.js#L57-L59

DanielVF:
Overall concept of this PR looks good. Have an inline question on one of the ways a test is written.

contracts/test/strategies/generalized-meta.fork-test.js
sparrowDom: Yeah, when slippage errors occur different StableSwap contract implementations fail with different error messages. For `alUsd`, `USDD` and `BUSD` transaction fails without an error message, and `mUSD` fails with `VM Exception while processing transaction: reverted with reason string 'Not enough coins removed' `

So wanted to cover both cases and didn't find `chai` having some sort of expect.to.fail.with.**or** functionality

DanielVF: Makes sense! Maybe a comment to explain why we are doing it differently.

DanielVF (approved):

contracts/test/strategies/generalized-meta.fork-test.js
sparrowDom: good point adding that

PR #1183, opened 2022-11-10
make fork tests more resilient

sparrowDom:
Which accounts have certain ERC20 on mainnet constantly changes. Make tests that require initial funding of fork test accounts more resilient by utilising a list of possible erc20 bag addresses and pick the one with the most funds.


contracts/utils/addresses.js
```diff
@@ -18,7 +18,7 @@ addresses.mainnet.Binance = "0xf977814e90da44bfa03b6295a0616a897441acec";
  * https://etherscan.io/accounts/label/binance?subcatid=3-0&size=100&start=0&col=2&order=desc
  */
 addresses.mainnet.BinanceAll =
```
shahthepro, line 20: Another idea, could probably use Etherscan's API to fetch the top holders, rather than hardcoding.

But we might end up over-engineering, so this is perfect for now

shahthepro (approved):
LGTM

contracts/utils/addresses.js
sparrowDom: Hmm that is very good thinking.

If we need to adjust this list too often your suggestion is probably the perfect solution

sparrowDom:
Thanks for super quick review @shahthepro !

PR #1189, opened 2022-11-21
Buyback tooling for brownie

DanielVF:
- Could be further optimized by downloading the ABI for ogv, so it doesn't need to load from etherscan


DanielVF:
Thanks for finding these @sparrowDom. Updated.

Issue #1197, opened 2022-11-30
Broken image redirect


micahalcorn:
The earn (staking) page is not rendering an image because https://**www**.ousd.com/images/splines.png is working but https://**app**.ousd.com/images/splines.png is not.

<img width="1512" alt="Screen Shot 2022-11-30 at 11 50 01 AM" src="https://user-images.githubusercontent.com/273937/204894816-234ca6a0-af38-4706-8301-61d6dc3f05ba.png">

PR #1199, opened 2022-12-03
add image to earn page

rolandpo:
add splines image to /earn page banner, issue #1197
also move redirect from /pages to config file


micahalcorn (approved):
Nice cleanup in addition to the image ✨

PR #1205, opened 2022-12-20
add LUSD deployment script and tests

sparrowDom:
- add LUSD deployment script and tests
- fixes slither github-actions workflow setup

**Deployment PR:** https://github.com/OriginProtocol/origin-dollar/pull/1206

PR #1206, opened 2022-12-20
Deploy 45 - LUSD

sparrowDom:
LUSD deployment artifacts from this pr: https://github.com/OriginProtocol/origin-dollar/pull/1205 

Contracts

|Name|Address|
|------|------|
| ConvexLUSDMetaStrategyProxy | 0x7a192dd9cc4ea9bdedec9992df74f1da55e60a19 |
| ConvexGeneralizedMetaStrategy |0xB6764c2Cc8F1fDCd89Af1C3e246f886579746233|


.github/workflows/ousd.yml
```diff
@@ -102,7 +102,9 @@ jobs:
 
   slither:
     name: "Slither"
-    runs-on: ubuntu-latest
+    # As long as we need Python 3.6 here in the test, we can only use up to Ubuntu 20.
```
shahthepro, line 105: Thanks for updating this. We've a low-priority issue to replace Slither with dependabot in future, until then this should work

shahthepro (approved):
Took a quick pass. The contracts are verified on Etherscan and the source matches codebase. Also, the order of deployment transactions seem to be right

DanielVF:
- [ ] Deployed contract verified code (and dependancies) match the final PR commit's code
- [ ] 🛑The transactions that interacted with the newly deployed contract match the deploy script.
- [x] Governance proposal matches the deploy script: [Proposal 41 TX](https://etherscan.io/tx/0xd6f57e7b07a3c38a5fbf253dfd0ff6c7c04f2ee816ef371bba6a5c163d4cfa4a#eventlog)
- [ ] Update Docs


### 🛑Transactions Check

The transfer governance command transfers ownership of the proxy to deployer address rather than to the governor contract. https://etherscan.io/tx/0xf99cc2433a0d2984e67c44dfc1c3d8d6749f8b2bd17ec9b8f9d200ba143b8999

This also causes the governance proposal to fail, given that the governor cannot accept ownership.

image


### Governance Proposal Check

```python
from world import *
v = governor.getActions(41)
[print(v[0][i],v[1][i],v[2][i],) for i in range(0, len(v)+1)]
```

franckc:
Great catch @DanielVF ! That could have been a serious security vulnerability.

@sparrowDom How abou
 - we clean up the playbook to be clear about what needs to be set?
 - also, is there any safety check we should add to the deploy script so that it issues a warning or does not run at all if it detects the network is mainnet and the deployer address is also being used as governor address?

Some additional context on the discord discussion: https://discord.com/channels/404673842007506945/965649970269143120/1057400565719638046

Also CC @shahthepro so that we don't make the same mistake when deploying the upcoming Morpho Aave strategy

DanielVF:
Perhaps we could have the deployer address set by the deployer private key when it is set. This way there's no config file to change during deployments, and there's not really anything to mess up.

sparrowDom:
Proposed solution that adds a safeguard: https://github.com/OriginProtocol/origin-dollar/pull/1211

sparrowDom:
[This transaction](https://etherscan.io/tx/0xfccc990068b1d6c426341d18dea9b76970940a5a92512c796c934a4e47ea904a) correct the wrong pending governor on the LUSD strategy contract

DanielVF (approved):

PR #1207, opened 2022-12-20
Add Morpho Aave strategy

shahthepro:
Adds Morpho Aave strategy. Morpho Aave doesn't distribute any reward tokens, so the rewards related functions don't do anything right now.

The deploy script uses Aave v2 tokens and have tested it on fork.

Uses migration ID 046 instead of 045 (since that's used in #1205 for LUSD).

-----------

Contract change checklist:
  - [ ] Unit tests pass


contracts/contracts/strategies/MorphoAaveStrategy.sol
```diff
@@ -0,0 +1,239 @@
...
+    /**
+     * @dev Deposit asset into Morpho
+     * @param _asset Address of asset to deposit
+     * @param _amount Amount of asset to deposit
+     */
+    function _deposit(address _asset, uint256 _amount) internal {
+        require(_amount > 0, "Must deposit something");
+
+        IMorpho(MORPHO).supply(
+            address(_getPTokenFor(_asset)),
```
DanielVF, line 127: `_deposit` calls `_getPTokenFor` twice. This does two storage reads. 

It could be more efferent to do one read into a temporary variable (as `_withdraw` does), then use that in the rest of the function.

contracts/contracts/strategies/MorphoAaveStrategy.sol
```diff
@@ -0,0 +1,239 @@
...
+
+    function _withdraw(
+        address _recipient,
+        address _asset,
+        uint256 _amount
+    ) internal {
+        require(_amount > 0, "Must withdraw something");
+        require(_recipient != address(0), "Must specify recipient");
+
+        address pToken = assetToPToken[_asset];
```
DanielVF, line 168: I think it would be good to use `_getPTokenFor` here to load it. This way we get a revert in case the asset is not found.

contracts/contracts/strategies/MorphoAaveStrategy.sol
```diff
@@ -0,0 +1,239 @@
...
+    /**
+     * @dev Collect accumulated rewards and send them to Harvester.
+     */
+    function collectRewardTokens()
+        external
+        override
+        onlyHarvester
+        nonReentrant
+    {
+        // Morpho Aave-v2 doesn't distribute reward tokens
```
DanielVF, line 89: Interestingly enough, AAVE reward collection and distrobution is in the current Morpho AAVE contract, but it's not configure to be active, and contract addresses needed are not set.

Because of incentivized harvesting, these two methods have no gas cost unless called, which would mean that we could go ahead and implement them.

However, since we can't test them currently, I'd much rather not have untested code that is live. I agree with you making these functions basically empty.

contracts/contracts/strategies/MorphoAaveStrategy.sol
```diff
@@ -0,0 +1,239 @@
...
+ */
+import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import { IERC20, InitializableAbstractStrategy } from "../utils/InitializableAbstractStrategy.sol";
+import { IMorpho } from "../interfaces/morpho/IMorpho.sol";
+import { ILens } from "../interfaces/morpho/ILens.sol";
+import { StableMath } from "../utils/StableMath.sol";
+import "../utils/Helpers.sol";
+
+contract MorphoAaveStrategy is InitializableAbstractStrategy {
+    address public constant MORPHO = 0x777777c9898D384F785Ee44Acfe945efDFf5f3E0;
```
DanielVF, line 17: Verified

contracts/contracts/strategies/MorphoAaveStrategy.sol
```diff
@@ -0,0 +1,239 @@
...
+    address public constant LENS = 0x507fA343d0A90786d86C7cd885f5C49263A91FF4;
```
DanielVF, line 18: Verified address

contracts/utils/addresses.js
```diff
@@ -30,9 +30,9 @@ addresses.mainnet.AAVE_ADDRESS_PROVIDER =
   "0xb53c1a33016b2dc2ff3653530bff1848a515c8c5"; // v2
 addresses.mainnet.Aave = "0x7Fc66500c84A76Ad7e9c93437bFc5Ac33E2DDaE9"; // v1-v2
 addresses.mainnet.aTUSD = "--"; // Todo: use v2
-addresses.mainnet.aUSDT = "--"; // Todo: use v2
+addresses.mainnet.aUSDT = "0x3Ed3B47Dd13EC9a98b44e6204A523E766B225811"; // v2
```
DanielVF, line 33: Verified Contract address.

contracts/utils/addresses.js
```diff
@@ -30,9 +30,9 @@ addresses.mainnet.AAVE_ADDRESS_PROVIDER =
...
 addresses.mainnet.aDAI = "0x028171bca77440897b824ca71d1c56cac55b68a3"; // v2
-addresses.mainnet.aUSDC = "--"; // Todo: use v2
+addresses.mainnet.aUSDC = "0xBcca60bB61934080951369a648Fb03DF4F96263C"; // v2
```
DanielVF, line 35: Verified contract address

contracts/contracts/strategies/MorphoAaveStrategy.sol
```diff
@@ -0,0 +1,239 @@
+// SPDX-License-Identifier: agpl-3.0
+pragma solidity ^0.8.0;
+
+/**
+ * @title OUSD Morpho Compound Strategy
```
sparrowDom, line 5: nit: Aave

contracts/contracts/strategies/MorphoAaveStrategy.sol
```diff
@@ -0,0 +1,239 @@
...
+            pToken,
+            address(this)
+        );
+    }
+
+    /**
+     * @dev Retuns bool indicating whether asset is supported by strategy
+     * @param _asset Address of the asset
+     */
+    function supportsAsset(address _asset)
```
sparrowDom, line 219: [MorphoCompound strategy](https://github.com/OriginProtocol/origin-dollar/blob/master/contracts/contracts/strategies/MorphoCompoundStrategy.sol#L17) extends [BaseCompoundStrategy](https://github.com/OriginProtocol/origin-dollar/blob/master/contracts/contracts/strategies/BaseCompoundStrategy.sol#L23-L42) that basically has  `supportsAsset` and `_getP(C)TokenFor ` already implemented. 

You could do the same thing here, and not needing to implement 2 additional functions.

But I get it that base contract is Compound and not Aave...

I wonder if we should rename `_getCTokenFor` to `_getPTokenFor` in `BaseCompoundStrategy` and move it to `InitializableAbstractStrategy`. There is some code duplication that might be cool to clean up... though not having a strong opinion on this one

sparrowDom (commented):
Great job @shahthepro. Seems solid!

I got 1 small consideration whether we should refactor the `supportsAsset`, `_getCTokenFor` into something compound and avve strategies might use. Since it seems they currently differ only in the names of the contracts. 

Otherwise looks solid 💪

contracts/contracts/strategies/MorphoAaveStrategy.sol
shahthepro: > However, since we can't test them currently, I'd much rather not have untested code that is live

Agreed. We can always upgrade the contract when Morpho enables distributing rewards

contracts/deploy/046_morpho_aave_strategy.js
```diff
@@ -0,0 +1,117 @@
...
+          await getTxOpts()
+        )
+    );
+
+    // 4. Init and configure new Morpho strategy
+    const initFunction = "initialize(address,address[],address[],address[])";
+    await withConfirmation(
+      cMorphoAaveStrategy.connect(sDeployer)[initFunction](
+        cVaultProxy.address,
+        // [assetAddresses.AAVE], // reward token addresses
```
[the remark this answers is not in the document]
shahthepro: Ah yes, my bad. I was using these for testing. Will remove it

contracts/contracts/strategies/MorphoAaveStrategy.sol
shahthepro: Changed, thanks for catching this

contracts/contracts/strategies/MorphoAaveStrategy.sol
shahthepro: Yep, have updated it

contracts/contracts/strategies/MorphoAaveStrategy.sol
shahthepro: Agreed, have changed it

contracts/contracts/strategies/MorphoAaveStrategy.sol
shahthepro: I actually started off by inheriting the `BaseCompoundStrategy` but the `cToken` nomenclature and some compound-specific functions (like `_convertUnderlyingToCToken`) were what caused me to remove that inheritance.

But I agree that we should avoid code redundancy of some methods like `supportsAsset` and `_getPTokenFor`. I guess we could have a new abstract contract `BaseGenericStrategy` (ik, I'm bad at naming :P) or something, which `BaseCompoundStrategy` and these Morpho strategies can inherit from. Let me explore what options we have

DanielVF: I think we would be getting into the realm of too much inheritance. The methods duplicated in this contract are small and focused. I don't think there's harm in it. Security, and via extension, understandability are are the biggest things we need to prioritize. Flatter contracts trees are much easier to review and understand, and sometimes some duplication is a worthwhile tradeoff for what's really important.

shahthepro: Yep, totally agree with that

sparrowDom: agree, sounds good to me

contracts/contracts/strategies/MorphoAaveStrategy.sol
MerlinEgalite: Currently, there are no rewards on Aave (and it's not planned). We'll remove them in a future upgrade

contracts/contracts/strategies/MorphoAaveStrategy.sol
```diff
@@ -0,0 +1,241 @@
...
+        address _recipient,
+        address _asset,
+        uint256 _amount
+    ) internal {
+        require(_amount > 0, "Must withdraw something");
+        require(_recipient != address(0), "Must specify recipient");
+
+        address pToken = address(_getPTokenFor(_asset));
+
+        IMorpho(MORPHO).withdraw(pToken, _amount);
```
MerlinEgalite, line 172: Note that by using type(uint256).max you can withdraw the whole position

contracts/contracts/strategies/MorphoAaveStrategy.sol
```diff
@@ -0,0 +1,241 @@
...
+ * @title OUSD Morpho Aave Strategy
+ * @notice Investment strategy for investing stablecoins via Morpho (Aave)
+ * @author Origin Protocol Inc
+ */
+import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import { IERC20, InitializableAbstractStrategy } from "../utils/InitializableAbstractStrategy.sol";
+import { IMorpho } from "../interfaces/morpho/IMorpho.sol";
+import { ILens } from "../interfaces/morpho/ILens.sol";
+import { StableMath } from "../utils/StableMath.sol";
+import "../utils/Helpers.sol";
```
MerlinEgalite, line 14: I don't think it's used

MerlinEgalite (approved):
Left some comments, but LGTM overall

contracts/contracts/strategies/MorphoAaveStrategy.sol
DanielVF: @shahthepro. We should go ahead and remove this line if it's not used. Will keep the codebase clean.

DanielVF (approved):

contracts/contracts/strategies/MorphoAaveStrategy.sol
shahthepro: Sure, will remove and redeploy the contracts today

contracts/contracts/strategies/MorphoAaveStrategy.sol
sparrowDom: thanks for info @MerlinEgalite

Issue #1204, opened 2022-12-19
Consistent meta images


micahalcorn:
We currently have different content in https://github.com/OriginProtocol/ousd.com vs this repo. Let's make everything match the first example here.

https://ousd.com

<img width="528" alt="Screen Shot 2022-12-19 at 2 56 04 PM" src="https://user-images.githubusercontent.com/273937/208543626-bd80983f-c036-4524-a910-8fd7aab93aaa.png">

https://app.ousd.com (empty)
<img width="530" alt="Screen Shot 2022-12-19 at 2 59 54 PM" src="https://user-images.githubusercontent.com/273937/208543810-3a7b4ef2-ef65-434d-8c96-34a789014e53.png">

https://app.ousd.com/earn
<img width="526" alt="Screen Shot 2022-12-19 at 2 56 57 PM" src="https://user-images.githubusercontent.com/273937/208543696-ee818d12-c3ab-4818-b930-741339a4d14d.png">

PR #1210, opened 2022-12-29
add meta images

rolandpo:
set meta images for #1204


shahthepro (approved):

dapp/pages/_document.js
```diff
@@ -39,6 +39,7 @@ class MyDocument extends Document {
             type="text/javascript"
           ></script>
 
+          <meta name="image" content={"https://cmsmediaproduction.s3.amazonaws.com/meta_9121c5630d.jpeg"} />
```
shahthepro, line 42: Why not have it as a local file as opposed to using a remote URL?

rolandpo: i added a file and local path thanks. do you think that will work as is?

sparrowDom:
we already have a meta image in OUSD

PR #1212, opened 2023-01-04
Sparrow dom/deploy additions

sparrowDom:
- remove generalised strategies that we have no plans for deploying
- exchange alUSD with LUSD in unit tests
- make it easier to execute proposals that are in new/queued state and run fork tests on top of them:
  - if proposalId is present in deploy script and proposal exists in forked node queue and/or execute it and skip deployment
  - if proposalId is/Isn't present and forked node has lower block number (proposal doesn't exist on chain yet) execute deployment
  - if proposalId is present and proposal has already been executed on forked node skip deployment


sparrowDom:
The one fork test failing is not relevant... I think this is good to merge CI wise

franckc (approved):
Good cleanup and improvements! LGTM

PR #1217, opened 2023-01-23
Migrate DApp to Heroku

shahthepro:

Contract change checklist:
  - [ ] Unit tests pass


.github/workflows/ousd.yml
```diff
@@ -146,4 +146,51 @@ jobs:
...
+
+  build-image:
+    name: "Deploy App"
+    runs-on: ubuntu-latest
+    needs: [contracts-lint, dapp-lint]
+    env:
+      HEROKU_TOKEN: ${{ secrets.HEROKU_TOKEN }}
+      CURR_BRANCH: ${{ github.ref_name }}
+    # TODO: Remove `shah/heroku-deploy` branch after merging PR
+    if: ${{ contains(fromJSON('["shah/heroku-deploy", "master", "staging", "stable"]'), github.ref_name) }}
```
sparrowDom, line 159: Why does pushing `master` also result in the building of the image? Is it an early warning signal when stuff gets merged into master. And that we are not surprised at build step?

sparrowDom (approved):
great work LGTM

.github/workflows/ousd.yml
shahthepro: Yes, I intentionally made it a way such that the image is built  on pushed to master but isn't pushed to Heroku or deployed. It usually a good indicator to know if someone on master is broken or not.

Another reason is caching. I'm also looking at exploring caching these built images. The idea is that once we build an image on master, we should probably reuse it (or at least some layers of it) on other branches like `staging` and `stable`. I gotta spend a bit more time, so I would split that up into a separate task. But it'd be nice to have and would also make deployments a bit faster since we don't have to rebuild the same image again and again

PR #1218, opened 2023-01-24
add gtm

rolandpo:
add gtm to app.ousd.com

there is some existing code for google analytics already, I can edit the pr depending on whether it can be replaced or not

### DIFF
--- a/dapp/pages/_app.js
+++ b/dapp/pages/_app.js
@@ -21,6 +21,7 @@ import { logout, login } from 'utils/account'
 import WalletSelectModal from 'components/WalletSelectModal'
 import { ToastContainer } from 'react-toastify'
 import { getConnector, getConnectorImage } from 'utils/connectors'
+import { GTM_ID, pageview } from '../lib/gtm'
 
 import analytics from 'utils/analytics'
 import { AnalyticsProvider } from 'use-analytics'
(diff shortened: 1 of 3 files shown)


dapp/pages/_app.js
```diff
@@ -189,6 +197,19 @@ function App({ Component, pageProps, err }) {
             pauseOnFocusLoss
             pauseOnHover
           />
+          <script
```
sparrowDom, line 200: I think this sort of script tag would fit better in the _document file

sparrowDom (approved):
aside from 1 comment LGTM

dapp/pages/_app.js
rolandpo: moved, thanks

jonathansnow: Late question here. What is the benefit/purpose of having this script in the _document file? We have the script in the _app file for every other deployment, and ideally, we keep this consistent across our stack.

sparrowDom: We have an unwritten convention where we put any javascript links or vanilla javascript tags in the `_document` as you can see [here for example](https://github.com/OriginProtocol/origin-dollar/blob/master/dapp/pages/_document.js#L63-L100) and _app.js contains react javascript components and modules imported via package manager. 

I don't see any technical benefits to this approach. And things would work equally well if they were in the `_app.js` file

jonathansnow: Sounds good. Thanks for the explanation!

shahthepro (approved):

PR #1222, opened 2023-01-26
update IPFS link and condition to display

sparrowDom:
update IPFS link and condition to display


sparrowDom:
@shahthepro this changes the staging and prod's env var to `IPFS_DAPP_URL=https://ousd.eth.limo/` . This will probably need to be changed in Github environment secrets that I think you have been working on. 

Have already changed this on the Heroku staging app

PR #1243, opened 2023-02-10
OGV governance deploy tooling

sparrowDom:
This PR updates our deploy tooling so it is able to: 
- on **mainnet** build out the proposal and output `to` and `data` that is to be copied to gnosis Safe to create a multisig transaction. With the old governance we were able to create a proposal directly on the governance contract, but OGV governance requires the proposer to have 10m veOGV balance. For that reason it is done via 5/8 multisig. 
- on **forked net** if governance proposal is already `pending/Active/Succeeded/Queued` the tooling is able to drive it to the end of its life-cycle and execute it. The only time this would fail is when there would be too much `against` votes already present on the proposal in forked environment. 
- on **forked net** if governance proposal hasn't been submitted yet. The tooling submits its, and does all the intermediate steps to execute it.
- **local node environment** is not supported by this deploy tooling

To test that the tooling works correctly I have 
- simulated the execution of currently [active proposal 40434364243407050666554191388123037800510237271029051418887027936281231737485](https://governance.ousd.com/proposals/40434364243407050666554191388123037800510237271029051418887027936281231737485) 
- and executed the #48 deploy script that changes the vaultAdmin. This step would fail if I didn't first execute the first step (currently active proposal) with the error that underlying Timelock transaction has failed. Which is correct since the governance needs to first accept the governance of `VaultProxy` before it is able to change the `VaultAdmin` implementation. 

I will also be using this tooling to submit the governance proposal to upgrade `VaultAdmin` to enable the new withdraw/deposit strategy functionality.


sparrowDom:
Current fork test CI is failing because [the active proposal](https://governance.ousd.com/proposals/40434364243407050666554191388123037800510237271029051418887027936281231737485) has no been executed yet

DanielVF (approved):

PR #1244, opened 2023-02-13
Remove unused env var

shahthepro:
This part of the code didn't make any sense to me, we were using Defi Pulse API key to query from ethgasstation 😅 

Looks like ethgasstation API works without any keys


sparrowDom:
I remember that they had in in the docs a long time ago and being confused when reading them. Can't find those docs anymore, so it looks like as you say, this key is no longer needed.

sparrowDom (approved):
LGTM

PR #1258, opened 2023-03-02
Uniswap V3 Strategy

shahthepro:
Things to decide:
  - [ ] decide if we are ok Strategist being 2/8 is having so much control over strategy configuration. 1 possible solution is to set the UniswapStrategist to be the old 5/8 governance. So we can change things quickly (no timelock) and still be relatively safe. 

Contract change checklist:
  - [ ] Unit tests pass


contracts/contracts/vault/VaultAdmin.sol
```diff
@@ -184,8 +194,29 @@ contract VaultAdmin is VaultStorage {
...
+
+    /**
+     * @dev Add a strategy to the Vault
+     * @param _addr Address of the strategy to add
+     * @param isUniswapV3 Set to true, if the strategy is an instance of GeneralizedUniswapV3Strategy
+     */
+    function _approveStrategy(address _addr, bool isUniswapV3) internal {
         require(!strategies[_addr].isSupported, "Strategy already approved");
-        strategies[_addr] = Strategy({ isSupported: true, _deprecated: 0 });
+        strategies[_addr] = Strategy({
```
sparrowDom, line 215: Leaving a note here: Strategy structs are mapped via:
```
mapping(address => Strategy) internal strategies
```

the [bool, uint256, bool] cause the Strategy structure to take up 3 storage slots. The new variable `isUniswapV3Strategy` has correctly been added to the end of the struct, ensuring that no old mappings are affected. 

nitpick: in theory because of a struct taking up 3 storage slots there could be collisions but it is really really highly unlikely.

DanielVF:
## Security goals

From a security design point of view, we should assume that the operator account is completely controlled by North Korea a couple times each year.

One style of attack would be for a malicious operator to almost infinitely bend the pool (a feature of Uniswap v3),  then use the operator to rebalance to a tick near that bent position. The amount would be set to pull all USDC from the matching reserve strategy on that side into the position to be lost. When the attacker trades back through that new position, they would pick up our USDC for basically free.  The attack would then repeat on the other side of pool to get all of our USDT.

We must have a way to cap the maximum damage that a malicious operator can do, it must be at the contract level, and it must be able to cover all operator operations.

We also need to be able to disable the operator immediately from the strategist multisig. Don't want to have to wait seven days for governance.

## Gas usage

A major component of the profitability of this strategy is likely the gas costs. We don't have to go all ultra MEV bot assembly level, but we do need to know what are gas usage is, and make sure we aren't doing anything silly. One thing this might rule out is any safety checks at the vault total value level, since that's crazy expensive.

## Swap

It looks like we may need a way to swap from one coin to another to build our position. Let's say we only have USDC, and we need a mix of USDC and USDT to go into active position.

contracts/contracts/strategies/GeneralizedUniswapV3Strategy.sol
```diff
@@ -0,0 +1,850 @@
...
+    ) external override onlyVault nonReentrant {
+        require(_asset == token0 || _asset == token1, "Unsupported asset");
+
+        IERC20 asset = IERC20(_asset);
+        uint256 selfBalance = asset.balanceOf(address(this));
+
+        if (selfBalance < amount) {
+            // Try to pull remaining amount from reserve strategy
+            // This might throw if there isn't enough in reserve strategy as well
+            IVault(vaultAddress).withdrawForUniswapV3(
```
DanielVF, line 293: These strategies are already under the vault's control, and could be withdrawn directly from. The part that the vault can't get to is the uniswap position. We'd need to withdraw from the currently active position instead. This might give us some extra coins on the other side, but that would be okay. They can sit on the strategy, because the vault might be asking for them next.

contracts/contracts/strategies/GeneralizedUniswapV3Strategy.sol
```diff
@@ -0,0 +1,850 @@
...
+    /**
+     * @notice Closes active LP position, if any, and transfer all token balance to Vault
+     * @inheritdoc InitializableAbstractStrategy
+     */
+    function withdrawAll() external override onlyVault nonReentrant {
+        if (currentPositionTokenId > 0) {
+            _closePosition(currentPositionTokenId);
+        }
+
+        IERC20 cToken0 = IERC20(token0);
```
DanielVF, line 322: These aren't ctokens? should this just be token0 and token1? We should probably fix throughout the code.

shahthepro: I used the names `token0` and `token1` for the addresses. Since this is an instance of the contract, I was thinking `Contract of Token 0 and 1` when I named it `cToken0`. But I agree it's confusing with Compound Tokens, will change it

contracts/contracts/strategies/GeneralizedUniswapV3Strategy.sol
shahthepro: Got it, will change that part of the code

contracts/contracts/strategies/GeneralizedUniswapV3Strategy.sol
```diff
@@ -0,0 +1,850 @@
...
+        (uint160 sqrtRatioX96, , , , , , ) = IUniswapV3Pool(platformAddress)
+            .slot0();
+
+        if (currentPositionTokenId > 0) {
+            require(
+                tokenIdToPosition[currentPositionTokenId].exists,
+                "Invalid token"
+            );
+
+            (uint256 amount0, uint256 amount1) = uniswapV3Helper.positionValue(
```
DanielVF, line 454: Calculating a strategies balance is the place where gas efficiency matters most. We go through every coin, in every strategy, every rebase, redeem, large mint. And we do it at least twice during allocations. 

Given that the operator is probably rebalancing every day or few, it's might be worth considering how much extra gas taking into account fees in this calculation will cost us.

contracts/contracts/strategies/GeneralizedUniswapV3Strategy.sol
```diff
@@ -0,0 +1,850 @@
...
+        _depositAll();
+    }
+
+    /**
+     * @notice Deposits all undeployed balances of the contract to the reserve strategies
+     */
+    function _depositAll() internal {
+        uint256 token0Bal = IERC20(token0).balanceOf(address(this));
+        uint256 token1Bal = IERC20(token1).balanceOf(address(this));
+        if (token0Bal > 0) {
```
DanielVF, line 266: This could actually be greater than zero. If depositing dust costs $20 in gas, we do it once a day, and we earn 3% on the dust, then my math (which may be wrong) says we would still be better off keeping up to $240,000 dust in a Strat. This may be excessive, but something like $40K dust should be fine.

contracts/contracts/strategies/GeneralizedUniswapV3Strategy.sol
```diff
@@ -0,0 +1,850 @@
...
+    /**
+     * @notice Returns a unique ID based on lowerTick and upperTick
+     * @dev Basically concats the lower tick and upper tick values. Shifts the value
+     *      of lowerTick by 24 bits and then adds the upperTick value to avoid overlaps.
+     *      So, the result is smaller in size (int48 rather than bytes32 when using keccak256)
+     * @param lowerTick Lower tick index
+     * @param upperTick Upper tick index
+     * @param key A unique identifier to be used with ticksToTokenId
+     */
+    function _getTickPositionKey(int24 lowerTick, int24 upperTick)
```
DanielVF, line 482: This is clever, and I like it.

contracts/contracts/strategies/GeneralizedUniswapV3Strategy.sol
```diff
@@ -0,0 +1,850 @@
...
+        INonfungiblePositionManager.MintParams memory params = INonfungiblePositionManager
+            .MintParams({
+                token0: token0,
+                token1: token1,
+                fee: poolFee,
+                tickLower: lowerTick,
+                tickUpper: upperTick,
+                amount0Desired: desiredAmount0,
+                amount1Desired: desiredAmount1,
+                amount0Min: maxSlippage == 0
```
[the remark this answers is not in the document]
sparrowDom: was thinking of the same thing. The operator should have a clear view of the state of Uniswap pool without the dangers of front running bots altering it just before the transaction executes. It would make sense to pass the min amounts as parameters to the external function

contracts/contracts/strategies/GeneralizedUniswapV3Strategy.sol
```diff
@@ -0,0 +1,850 @@
...
+    /// Callback function for whenever a NFT is transferred to this contract
+    // solhint-disable-next-line max-line-length
+    /// Ref: https://docs.openzeppelin.com/contracts/3.x/api/token/erc721#IERC721Receiver-onERC721Received-address-address-uint256-bytes-
+    function onERC721Received(
+        address,
+        address,
+        uint256,
+        bytes calldata
+    ) external returns (bytes4) {
+        // TODO: Should we reject unwanted NFTs being transfered to the strategy?
```
sparrowDom, line 791: is there any downside to leaving unexpected NFTs on the contract?

sparrowDom (commented):
**Security**
Similar to what @DanielVF points out my biggest fear is that Uniswap strategy gives the Operator the ability to override the Governance allocation and pull all the funds of the reserve strategies for token0 & token1 assets. 

One way to mitigate the issue would be to keep internal accounting of how much of each asset has been deposited to reserve strategies and how much withdrawn. So Uniswap strategy could never withdraw more than it has deposited. 

Another way would be to have a strategy TVL threshold that gets updated every time allocation is called. This way when Vault would be `withdrawingForUniswapV3` it would be able to check that `uniswapV3.checkBlalance() + assetRequested < uniswapV3TVL`


**Rebalance approaches**

Might be cool that we test out some more rebalancing approaches: 
- if we end up with much more of one token than another we could simulate various approaches to rebalancing:
  - Swapping those tokens to have 50/50 share of tokens and rebalancing
  - Deploying the overhead of the tokens into reserve strategy 
  - Deploying liquidity in imbalanced manner (the way strategy tokens are imbalanced)
- have we though about copying (with doing sanity checks) Sommelier or other successful rebalancing protocols?

sparrowDom:
Also awesome work on this massive PR!

contracts/contracts/strategies/GeneralizedUniswapV3Strategy.sol
shahthepro: Agreed, I guess we could also probably make it a configurable variable, if we wanna tweak it in future

contracts/contracts/strategies/GeneralizedUniswapV3Strategy.sol
shahthepro: None as far as I can tell, but just wanted to know if anyone else thinks otherwise

shahthepro:
Thanks for the comments @DanielVF @sparrowDom 

I have locally made the changes to remove the `maxSlippage` and made it as an passable arg. I'll also add a threshold on the amount of tokens Uniswap Strategy can pull from reserve strategies

contracts/contracts/strategies/GeneralizedUniswapV3Strategy.sol
```diff
@@ -0,0 +1,850 @@
...
+        address[] memory _assets = new address[](2);
+        _assets[0] = token0;
+        _assets[1] = token1;
+
+        super._initialize(
+            _poolAddress,
+            _vaultAddress,
+            new address[](0), // No Reward tokens
+            _assets, // Asset addresses
+            _assets // Platform token addresses
```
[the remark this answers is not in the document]
shahthepro: `pTokens` aren't needed for this strategy, yeah. `InitializableAbstractStrategy.initialize()` would throw if I pass a empty array here. Couldn't pass address(0) either since `_setPTokenAddress()` (called in `initialize()` method) would throw. 

I thought of copying relevant code from `InitializableAbstractStrategy` and getting rid of `pToken` and `rewardToken` logics but that'd involve a lot of code duplication and might be harder to maintain in the long run. So, as of now, the Uniswap strategy just ignores the `pToken` and `rewardTokenAddresses` fields

contracts/contracts/strategies/GeneralizedUniswapV3Strategy.sol
```diff
@@ -0,0 +1,1144 @@
...
+    /**
+     * @notice Deposits all undeployed balances of the contract to the reserve strategies
+     */
+    function _depositAll() internal {
+        uint256 token0Bal = IERC20(token0).balanceOf(address(this));
+        uint256 token1Bal = IERC20(token1).balanceOf(address(this));
+        if (
+            token0Bal > 0 && token0Bal >= poolTokens[token0].minDepositThreshold
+        ) {
+            IVault(vaultAddress).depositForUniswapV3(token0, token0Bal);
```
sparrowDom, line 373: Maybe we could rename this to `depositToUniswapV3ReserveStrategy`.

I keep forgetting what this function does and need to check it out

shahthepro: Noted. Will change it

contracts/contracts/strategies/GeneralizedUniswapV3Strategy.sol
```diff
@@ -0,0 +1,1144 @@
...
+                token1Needed > 0 && token1Needed < t1ReserveBal,
+                "Cannot swap when the asset is available in reserve"
+            );
+            // Additional amount of token0 required for swapping
+            token0Needed += swapAmountIn;
+            // Subtract token1 that we will get from swapping
+            token1Needed -= swapMinAmountOut;
+
+            // Approve for swaps
+            t0Contract.safeApprove(address(swapRouter), swapAmountIn);
```
sparrowDom: should we do this only once (maybe on contract initialisation) rather than on every swap

contracts/contracts/strategies/uniswap/UniswapV3LiquidityManager.sol
```diff
@@ -0,0 +1,830 @@
...
+            );
+            // Additional amount of token0 required for swapping
+            token0Needed += swapAmountIn;
+            // Subtract token1 that we will get from swapping
+            token1Needed = (swapMinAmountOut >= token1Needed)
+                ? 0
+                : (token1Needed - swapMinAmountOut);
+
+            // Approve for swaps
+            t0Contract.safeApprove(address(swapRouter), swapAmountIn);
```
sparrowDom, line 664: (sorry commented on outdated code a few minutes earlier)...
should we call the safe approve only once instead on every swap?

sparrowDom: We would need to do it with max amount though...

contracts/contracts/strategies/GeneralizedUniswapV3Strategy.sol
shahthepro: Yep. I added that as a comment in the new contract that I pushed last night

contracts/contracts/strategies/uniswap/UniswapV3LiquidityManager.sol
```diff
@@ -0,0 +1,962 @@
...
+            uint256 minAmount1
+        ) = _calculateLiquidityToWithdraw(position, asset, amount);
+
+        // Liquidiate active position
+        _decreasePositionLiquidity(
+            position.tokenId,
+            liquidity,
+            minAmount0,
+            minAmount1
+        );
```
sparrowDom, line 144: @shahthepro should we deposit the leftover assets to reserve strategies at this point by calling `_depositAll`?

shahthepro: The funds (of the requested asset) removed from the active position are then transferred to the Vault. This function is also actually delegate called from another one in UniswapV3Startegy. So, we cannot call depositAll here. If we add that at the end of the withdraw on the other contract. It'll add up the gas costs of redeem. That's the reason I didn't wanna add it

sparrowDom: oh I see thanks for explanation

contracts/contracts/vault/VaultStorage.sol
```diff
@@ -70,6 +76,8 @@ contract VaultStorage is Initializable, Governable {
     struct Strategy {
         bool isSupported;
         uint256 _deprecated; // Deprecated storage slot
+        // Set to true if the Strategy is an instance of `UniswapV3Strategy`
+        bool isUniswapV3Strategy;
```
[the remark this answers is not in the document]
naddison36: I'd like to get rid of the `_deprecated` property in the next Vault deploy. I've raise PR https://github.com/OriginProtocol/origin-dollar/pull/1653 against collateral swaps to change this.

naddison36 (commented):
I've done a first pass of the code. I haven't gone deep into the logic yet.

contracts/contracts/utils/UniswapV3Helper.sol
```diff
@@ -0,0 +1,302 @@
+// SPDX-License-Identifier: agpl-3.0
+pragma solidity =0.7.6;
```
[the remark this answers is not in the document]
shahthepro: Uniswap v3 is still on v0.7.6. That's the reason I couldn't directly import these helper function in the other contracts (since we use v0.8.6)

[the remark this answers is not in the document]
naddison36: ok, leave the Uniswap helper on Solidity 0.7.6

sparrowDom:
just going through all the open PRs. We are still planning to complete this, but has been on the back-burner for now?

naddison36:
It's on the back-burner for now

shahthepro:
Closing this PR for now, can always reopen this branch if it becomes a priority again in future

PR #1263, opened 2023-03-17
possible contract moves to circumvent delegate calls

sparrowDom:
these changes: 
- remove the need for uniswap library
- move the `withdraw` and the `withdrawAll` from strategy contract to liquidityManager

pros: 
- no more delegate call
- no more need for library

cons: 
size increase of Liquidity Manager from
Screenshot 2023-03-17 at 09 55 52
to
Screenshot 2023-03-17 at 10 42 25

### DIFF
--- a/contracts/contracts/strategies/uniswap/UniswapV3LiquidityManager.sol
+++ b/contracts/contracts/strategies/uniswap/UniswapV3LiquidityManager.sol
@@ -787,15 +787,10 @@ contract UniswapV3LiquidityManager is UniswapV3StrategyStorage {
      */
     function withdrawAll() external override onlyVault nonReentrant {
         if (activeTokenId > 0) {
-            // TODO: This method is only callable from Vault directly
-            // and by Governor or Strategist indirectly.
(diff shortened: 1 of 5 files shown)


contracts/contracts/strategies/uniswap/UniswapV3LiquidityManager.sol
```diff
@@ -15,22 +14,9 @@ import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.s
...
-            vaultAddress,
-            poolTokens[token0].minDepositThreshold,
-            poolTokens[token1].minDepositThreshold
-        );
-    }
-
     // TODO: Intentionally left out non-reentrant modifier since otherwise Vault would throw
     function withdrawAssetFromActivePosition(address _asset, uint256 amount)
-        external
+        public
```
shahthepro, line 19: Does this work on fork? Since it has a onlyVault modifier, this might not work. We could probably have an `internal` function without any modifier and one `external` with necessary modifiers?

shahthepro (approved):

PR #1267, opened 2023-03-27
retire analytics events

rolandpo:
remove some analytics event tracking code + clean up some old marketing pages content


dapp/src/components/buySell/SwapHomepage.js
```diff
@@ -218,12 +217,6 @@ const SwapHomepage = ({
     const metadata = swapMetadata()
 
     try {
-      analytics.track('Before Swap Transaction', {
```
sparrowDom, line 221: curious why are we removing this one?

rolandpo:
For the analytics tracking event I'm going off @jonathansnow's recommendations here https://docs.google.com/spreadsheets/d/1iWfCSBpxxAFu8yObAVKZCnlfk-d903RpgGCggT2Ex1A/edit#gid=0. I think for `Before swap` events, those are covered by the other swap events triggered on success/failure etc. In general though, I think the current analytics wrapper library might be retired eventually and whichever event tracking is still needed would be reimplemented in GTM directly, consistent with the rest of the codebases

jonathansnow:
Yes, what Roland mentioned above is true. We're migrating everything to GTM/GA4 and moving away from third-party libraries. UA EOL is on July 1st.

We've removed items we're not concerned about right now while noting them to be added back in the future possibly.

PR #1268, opened 2023-03-28
OETH Strategy Interface

sparrowDom:
Proposal for new Strategy to Vault interface. Though process behind this proposal: 
- platformTokens and assets are not necessarily mapped 1:1. That approach is applicable to Aave and Compound strategies and no so much to Curve / Uniswap pool. If a strategy will require that type of mapping it should implement it in its own class. 
- withdrawals and deposits are now possible with a list of assets and corresponding amounts. This way strategies can deposit and withdraw funds in a more capital efficient manner.
- there are 2 functions to checkBalance. One's intention is to check the `checkTotalBalance` -> so that Vault knows the total amount of assets in the strategy (more gas efficient). The other one returns the total amount of platform token that is withdraw-able from the strategy. 
- `platformAddress` removed from the abstract strategy. It could be multiple addresses that pertain to the specific strategy implementation and those can have a more descriptive and accurate address variable names. 

**To Be Decided**: 
- should checkBalance(_asset) always return total amount of _asset that is withdraw-able from the strategy. Meaning that looping through all the assets and summing balances together will result in larger amount that is deposited in the strategy. Or should it return a `balanced` result where looping through the assets will always return the same amount as `checkTotalBalance`. 
  - [Domen]I am leaning toward reporting total _asset withdraw-able from the strategy 
- should `transferToken` always transfer to the governor? TimelockController might not  have the functionality to do anything with ERC20 tokens
- do we need a flavour of Deposit & Withdrawal functions that check min/max value of platform tokens minted/burned. Would probably be useful for the vault when doing re-allocation. And not as much when users will mint/burn OETH and deposit funds into the strategy.


DanielVF:
Thanks

✅ PlatformTokens and assets are not necessarily mapped 1:1. That approach is applicable to Aave and Compound strategies and no so much to Curve / Uniswap pool. If a strategy will require that type of mapping it should implement it in its own class.
✅ 💚 Withdrawals and deposits are now possible with a list of assets and corresponding amounts. This way strategies can deposit and withdraw funds in a more capital efficient manner.

there are 2 functions to checkBalance:

✅ One's intention is to check the checkTotalBalance -> so that Vault knows the total amount of assets in the strategy (more gas efficient). _Yes. This needs to be denominated in ideal "units" -ETH for OETH, USD for OUSD.

❓ The other one returns the total amount of platform token that is withdraw-able from the strategy. _A strategy like uniswap might not have one single platform token? Do we really need this?"

✅ platformAddress removed from the abstract strategy. It could be multiple addresses that pertain to the specific strategy implementation and those can have a more descriptive and accurate address variable names.


🟦 ✅ yes reporting total _asset withdraw-able from the strategy. checkBalance(_asset) always return total amount of _asset that is withdraw-able from the strategy. Meaning that looping through all the assets and summing balances together will result in larger amount that is deposited in the strategy. _Let's do it_

🟦 should transferToken always transfer to the governor? _Should be fine to send to timelock. timelock should always be able to run arbitrary commands.

🟦 do we need a flavour of Deposit & Withdrawal functions that check min/max value of platform tokens minted/burned. Would probably be useful for the vault when doing re-allocation. And not as much when users will mint/burn OETH and deposit funds into the strategy. _we could do this, but I'd rather just have one set of methods, not duplicates_

❗ _The one think I'm seeing missing from this is a way to withdraw "units" worth from the strategy, in whatever balance it currently has. For example, the vault would want to withdraw 5ETH from the strategy and so would call something like `withdrawUnits(5*1e18)`_

sparrowDom:
❓ *checkBalance(address _asset) that would return total amount of LSD withdraw-able? 
I am not sure if you are for it or against?

✅ transferToken will send funds to TimelockController
✅ Not adding `Deposit` & `Withdrawal` with min/max LP tokens
✅ added the `withdrawUnits` function

DanielVF:
Yes, I'm for checkBalance(address _asset) would returning the amount of LSD that LSD that would be withdrawn if we did a withdrawAll()

naddison36:
Is this PR still required? Or has it since been replaced?

sparrowDom:
We can table this for now. Since there is no short to middle term plan of us implementing this in either OUSD or OETH protocol. 

Feel free to re-open if anyone disagrees.

PR #1271, opened 2023-03-31
OETH - Full Changes

DanielVF:

**TODO:**
  - [ ] PriceUnitMint & PriceUnitRedeem need to be updated in the dapp

Contract change checklist:
  - [ ] Unit tests pass


contracts/contracts/vault/VaultCore.sol
```diff
@@ -74,25 +76,22 @@ contract VaultCore is VaultStorage {
...
         }
         require(price >= MINT_MINIMUM_ORACLE, "Asset price below peg");
-        uint256 assetDecimals = Helpers.getDecimals(_asset);
         // Scale up to 18 decimal
-        uint256 unitAdjustedDeposit = _amount.scaleBy(18, assetDecimals);
-        uint256 priceAdjustedDeposit = _amount.mulTruncateScale(
-            price.scaleBy(18, 8), // Oracles have 8 decimal precision
-            10**assetDecimals
-        );
+        uint256 priceAdjustedDeposit = (_toUnits(_amount, _asset) * price) /
```
sparrowDom, line 80: much cleaner 👍

contracts/contracts/vault/VaultCore.sol
```diff
@@ -576,7 +567,7 @@ contract VaultCore is VaultStorage {
         uint256 assetCount = getAssetCount();
         uint256[] memory assetPrices = _getAssetPrices();
         uint256[] memory assetBalances = new uint256[](assetCount);
-        uint256[] memory assetDecimals = new uint256[](assetCount);
+        uint256[] memory assetUnits = new uint256[](assetCount);
```
sparrowDom, line 570: since this var is used only inside the for loop we can remove it as a local variable inside the function

contracts/contracts/vault/VaultCore.sol
```diff
@@ -671,6 +659,24 @@ contract VaultCore is VaultStorage {
         return assets[_asset].isSupported;
     }
 
+    function _toUnits(uint256 raw, address _asset)
```
sparrowDom, line 662: I love how this function simplifies a bunch of the code above. 💪 
nit: would add documentation to it

contracts/test/vault/exchangeRate.js
```diff
@@ -0,0 +1,80 @@
...
+
+const {
+  ousdUnits,
+  daiUnits,
+  loadFixture,
+  setOracleTokenPriceUsd,
+  isFork,
+} = require("../helpers");
+
+describe.only("Vault Redeem", function () {
```
sparrowDom, line 12: `.only` sneaked in there :)

contracts/contracts/vault/VaultCore.sol
```diff
@@ -671,6 +659,24 @@ contract VaultCore is VaultStorage {
...
+        internal
+        view
+        returns (uint256)
+    {
+        if (assets[_asset].hasExchangeRate) {
+            return (raw * IExchangeRateToken(_asset).exchangeRate()) / 1e18;
```
sparrowDom, line 668: For [rEth](https://etherscan.io/address/0xae78736cd615f374d3085123a210448e74fc6393#readContract) it is `getExchangeRate` 

Screenshot 2023-04-03 at 14 39 45

contracts/contracts/vault/VaultCore.sol
sparrowDom: Hmm also for `rEth` that is a token with an `getExchangeRate` wont this erroneously adjust the price twice? Since the `_toUnits` function already multiplies the raw amount with the exchange rate of the token lower in this file -> `return (raw * IExchangeRateToken(_asset).exchangeRate()) / 1e18; `

sparrowDom (commented):
LGTM I am surprised by how little changes to the Vault are required for it to support a token with an exchange rate. Elegantly solved! Left a couple of comments inline

contracts/test/vault/exchangeRate.js
DanielVF: Thanks!

image

contracts/contracts/strategies/Generalized4626Strategy.sol
```diff
@@ -0,0 +1,141 @@
...
+    function withdraw(
+        address _recipient,
+        address _asset,
+        uint256 _amount
+    ) external override onlyVault nonReentrant {
+        require(_amount > 0, "Must withdraw something");
+        require(_recipient != address(0), "Must specify recipient");
+        require(_asset == address(assetToken), "Unexpected asset address");
+
+        IERC4626(platformAddress).withdraw(_amount, _recipient, address(this));
```
DanielVF, line 71: Does this allow you to always withdraw the amount listed by checkBalance?

contracts/contracts/strategies/Generalized4626Strategy.sol
```diff
@@ -0,0 +1,141 @@
...
+     * @return balance    Total value of the asset in the platform
+     */
+    function checkBalance(address _asset)
+        external
+        view
+        override
+        returns (uint256 balance)
+    {
+        require(_asset == address(assetToken), "Unexpected asset address");
+        return IERC4626(platformAddress).totalAssets();
```
DanielVF, line 117: @sparrowDom, This seems wrong. TotalAssets isn't how many assets we are owed by the downstream vault, but how many assets it contains in total, for all users.

contracts/test/strategies/frax-ETH.fork-test.js
```diff
@@ -21,34 +23,127 @@ forkOnlyDescribe("ForkTest: Frax ETH Strategy", function () {
     fixture = await loadFixture(fraxETHStrategyForkedFixture);
   });
 
-  describe.only("Mint", function () {
-    it("Should deploy fraxETH in Frax ETH Strategy", async function () {
-      const { daniel, frxETH } = fixture;
-      await mintTest(fixture, daniel, frxETH, "10");
-    });
+  it("Should deposit fraxETH in Frax ETH Strategy", async function () {
+    const { daniel, frxETH } = fixture;
```
DanielVF, line 27: frxETH isn't used here. Through this file there's a few things lint complains about.

contracts/contracts/strategies/Generalized4626Strategy.sol
```diff
@@ -114,7 +114,15 @@ contract Generalized4626Strategy is InitializableAbstractStrategy {
         returns (uint256 balance)
     {
         require(_asset == address(assetToken), "Unexpected asset address");
-        return IERC4626(platformAddress).totalAssets();
+        /* We are intentionally not counting the amount of assetToken parked on the
```
DanielVF, line 117: 👍

contracts/deploy/051_oeth.js
```diff
@@ -0,0 +1,273 @@
...
+  const sDeployer = await ethers.provider.getSigner(deployerAddr);
+
+  const assetAddresses = await getAssetAddresses(hre.deployments);
+  console.log(
+    `Using asset addresses: ${JSON.stringify(assetAddresses, null, 2)}`
+  );
+
+  // Proxies
+  await deployWithConfirmation("OETHVaultProxy");
+  await deployWithConfirmation("OETHProxy");
```
DanielVF, line 66: I thought we had already deployed the OETH proxy?

contracts/contracts/vault/VaultCore.sol
```diff
@@ -92,7 +90,7 @@ contract VaultCore is VaultStorage {
         emit Mint(msg.sender, priceAdjustedDeposit);
 
         // Rebase must happen before any transfers occur.
-        if (unitAdjustedDeposit >= rebaseThreshold && !rebasePaused) {
```
DanielVF, line 95: These thresholds are going to be different for OETH.

DanielVF:
Some personal helpers for playing with the contract:

```python
FRAX_ETH_STRATEGY = '0x8E294b1FE62a746371FDA21FAD1C69294B252819'


oeth = load_contract('ousd', '0x050c4FcA28725d975c2896682eBD2905D2E58E84')
OETH_VAULT_PROXY = '0x617F960A19f0f4d560256E04B5F4B92E187C512D'

oeth_vault_admin = load_contract('vault_admin', OETH_VAULT_PROXY)
oeth_vault_core = load_contract('vault_core', OETH_VAULT_PROXY)

DUDE = '0x9026A229b535ecF0162Dfe48fDeb3c75f7b2A7AE'
unlock(DUDE)


# TODO
oeth_vault_admin.setStrategistAddr(DUDE, {'from': GOV_MULTISIG})


frxeth.approve(oeth_vault_admin, 1e70, {'from': DUDE})
oeth_vault_core.mint(frxeth, 2e18, 0, {'from': DUDE})




frax_eth_strat = load_contract('aave_strat', FRAX_ETH_STRATEGY)

oeth_vault_admin.depositToStrategy(FRAX_ETH_STRATEGY, [frxeth], [1e18], {'from': DUDE})
print(frax_eth_strat.checkBalance(frxeth))
oeth_vault_admin.withdrawFromStrategy(FRAX_ETH_STRATEGY, [frxeth], [1e18-1e3], {'from': DUDE})
print(frax_eth_strat.checkBalance(frxeth))
oeth_vault_admin.withdrawAllFromStrategy(FRAX_ETH_STRATEGY, {'from': DUDE})
print(frax_eth_strat.checkBalance(frxeth))


print(oeth.totalSupply())
print(oeth_vault_core.totalValue())
oeth_vault_core.rebase({'from': DUDE})
print(oeth.totalSupply())
print(oeth_vault_core.totalValue())
oeth_vault_core.redeem(1e18, 0, {'from': DUDE})
print(oeth.totalSupply())
print(oeth_vault_core.totalValue())
```

contracts/contracts/strategies/Generalized4626Strategy.sol
sparrowDom: Yup this was very wrong replaced with: 
```
return
            IERC4626(platformAddress).convertToAssets(
                shareToken.balanceOf(address(this))
            );
```

contracts/test/strategies/frax-ETH.fork-test.js
sparrowDom: thanks fixef

contracts/contracts/vault/VaultCore.sol
sparrowDom: Yup I've added a TODO some time ago to double check these values here: 
https://github.com/OriginProtocol/origin-dollar/pull/1271/files#diff-a164c283475bded9c293a6f65d3256cf31a805c1fbf900425336b5095b2a08e0R118-R127

contracts/deploy/051_oeth.js
sparrowDom: You're right nice catch thanks. Removing this line

contracts/contracts/strategies/Generalized4626Strategy.sol
sparrowDom: will add a test for this to confirm

sparrowDom:
oh nice thanks those helpers are going to come in handy!

contracts/contracts/strategies/Generalized4626Strategy.sol
sparrowDom: Yes it allows the withdrawal of the whole balance. Test here: 
https://github.com/OriginProtocol/origin-dollar/pull/1271/commits/6c490c66d709636ecf0ba1970fe580b03ef70797

contracts/contracts/strategies/Generalized4626Strategy.sol
```diff
@@ -0,0 +1,149 @@
...
+    function withdraw(
+        address _recipient,
+        address _asset,
+        uint256 _amount
+    ) external override onlyVault nonReentrant {
+        require(_amount > 0, "Must withdraw something");
+        require(_recipient != address(0), "Must specify recipient");
+        require(_asset == address(assetToken), "Unexpected asset address");
+
+        IERC4626(platformAddress).withdraw(_amount, _recipient, address(this));
```
DanielVF, line 71: Perhaps we should check that the return value from calling `withdraw` here matches the _amount.

contracts/contracts/strategies/Generalized4626Strategy.sol
```diff
@@ -0,0 +1,149 @@
...
+    /**
+     * @dev Deposit assets by converting them to shares
+     * @param _asset Address of asset to deposit
+     * @param _amount Amount of asset to deposit
+     */
+    function _deposit(address _asset, uint256 _amount) internal {
+        require(_amount > 0, "Must deposit something");
+        require(_asset == address(assetToken), "Unexpected asset address");
+
+        IERC4626(platformAddress).deposit(_amount, address(this));
```
DanielVF, line 42: I don't think it's worth checking the return value here. We should just tell slither to ignore this.

sparrowDom: Thank added an ignore

contracts/contracts/strategies/Generalized4626Strategy.sol
sparrowDom: Slither returns amount of shares that were burned (sfrxETH tokens) in this case and not the amount of assets that have been withdrawn (frxETH token). 

From the [source code](https://github.com/FraxFinance/frxETH-public/blob/7f7731dbc93154131aba6e741b6116da05b25662/flattened.sol#L688) it does seem that the correct amount of `assets` gets transferred. Though this applies only for `sfrxToken` contract of course. 

We could run `shares = previewWithdraw(assets)` with the withdrawing assets and confirm that the `shares` amount match between what `previewWithdraw` returns and what `withdrawal` returns.

DanielVF: Probably good just ignoring then

contracts/contracts/vault/VaultCore.sol
```diff
@@ -588,57 +561,171 @@ contract VaultCore is VaultStorage {
...
+     * Note: since we are returning the price of the unit and not the one of the
+     * asset (see comment above how 1 rETH exchanges for 1.2 units) we need
+     * to make the Oracle price adjustment as well since we are pricing the
+     * units and not the assets.
+     *
+     * The price also snaps to a "full unit price" in case a mint or redeem
+     * action would be unfavourable to the protocol.
+     *
+     */
+    function _toUnitPrice(address _asset, bool isMint)
```
DanielVF, line 682: Since this is no longer a conversion, it should just be named `_unitPrice` now.

contracts/contracts/vault/VaultStorage.sol
```diff
@@ -120,6 +125,13 @@ contract VaultStorage is Initializable, Governable {
     // How much net total OUSD is allowed to be minted by all strategies
     uint256 public netOusdMintForStrategyThreshold = 0;
 
+    // Cheaper to read decimals locally than to call out each time
+    mapping(address => uint256) internal decimalsCache; // TODO: Move to Asset struct
+
+    uint256 constant MIN_UNIT_PRICE_DRIFT = 0.7e18;
+    uint256 constant MAX_UNIT_PRICE_DRIFT = 1.3e18;
+    uint256 constant MINT_MINIMUM_ORACLE = 99800000;
```
DanielVF, line 133: We'll need to resolve `MINT_MINIMUM_ORACLE` and `MINT_MINIMUM_UNIT_PRICE`

DanielVF:
# OETH Zapper Code Review

#### Requirements

We want to allow users with raw ETH to being able to send it to an address and get OETH. We also want to allow users with sFXETH to mint OETH. By having this as a separate contract, we keep the vault simple, and avoid handling ETH in the vault.

### Deployment Considerations

Monitoring would be nice, but not necessary. Good to see how much this is used.

Will need to check constructor addresses when deployed.

#### Internal State

No stored state. Intended to only ephemerally hold balances.

#### Attack

If this contract failed, the worst case would a transaction's worth of user's funds (not protocol funds).

All publicly callable methods also use a minimum amount which will revert if the user does not get back that much OETH.

Could funds could become stuck in this contract? Every publicly callable method ends by sending all OETH on the contract to the caller. If a someone accidentally sends mintable funds or OETH to this contract, they will be folded into the next deposit and returned to the next depositor. Any other ERC20 would become stuck, which is acceptable.

If OETH had a rounding error, and returned less funds than expected, it could break the deposit. `depositSFRXETH` has a user controlled minimum. This could be adjusted in the UI. Depositing ETH insists on that zapper receiving exactly 1:1. Given that the current version of the contract is non-rebasing, this should always be true.

#### Logic

There are no if statements, and all external paths end with funds going to the user.

#### Tests

🔅 This code has no unit test or fork tests. Don't need unit tests since this hardcodes live contract addresses for simplicity. Fork tests would be good.

#### Flavor

Could this code be simpler? Not that I see.

🔅✔︎ The repeated `slither-disable-next-line unused-return` are a bit ugly. I've removed them and replace them with entries in the triage file.

🔅✔︎ No document strings on methods. Added.

Could this code be less vulnerable to other code behaving weirdly? In the end the requires ensure that they user gets what they expect. If something breaks one day in a called token or contract, we can just deploy a new zapper, and there is no state that needs to carry over.

#### Overflow

_No math!_

#### Proxy
_No proxy!_

#### Black magic

#### Dependencies
_Only interfaces are used._

#### Deploy
- [x] Check that any deployer permissions are removed after deploy. _No deployer permissions.

#### Authentication

- [x] Never use tx.origin

#### Cryptographic code

_No crypto._

#### Gas problems

_No Loops._

#### External calls

- [x] Unsafe external calls are handled
- [x] Reentrancy guards on all state changing functions. _No state_
- [x] Oracles, one of:
  - [ ] No oracles
  - [ ] Oracles can't be bent

#### Ethereum

This contract sends and receives ethereum.

The only receiving is in `receive()`, which calls deposit. The vault's reentrancy guards should be protect the vault. We do not allow fallback calls with eth to arbitrary addresses.

We send eth, but only to the trusted WETH contract, and we have minimum amount checks afterwords that prevent any funny business.

DanielVF:
# Code Review: Frax ETH Strategy

#### Requirements

This stakes FRXETH tokens into the 4626 sFRXETH contract. This also works as a generic 4626 strategy to wrap any token into a 4626.

As with any Strategy, the three key concerns are:

- Money goes in
- Money goes out
- Balances are correct, and cannot be manipulated below what we can withdraw.

### Deployment Considerations

Should be in our monitoring / event logging.

#### Internal State

_This contract only holds configuration._

`shareToken` and `assetToken` are somewhat duplicates of the `InitializableAbstractStrategy`'s built in pToken storage. However, having real names adds a lot of clarity to the code, as well as gas savings. The only thing I could see that could go wrong here would be that the contract could be initialized with multiple assets set, the last would win, and then you would have to upgrade the proxy in order to go back. I don't think this is worth attempting to prevent. If you want to misconfigure a strat when setting it up, there's a lot of ways to do this. Maybe a comments note. 

#### Attack

This strategy could be holding a large portion of our backing funds. It's important that it works.

The most common danger with these strategies is a balance checking problem, with an attacker being able to force an incorrect high balance, rebasing to catch the  gains, and then withdrawing.

🔆 Check balance uses `convertToAssets` to get our maximum withdraw. This gets the correct answer with stfxeth, but it would feel better to use `previewRedeem` in case redeeming in some other 4626 contract was implemented differently.

4626 contracts can sometimes be vulnerable to rounding attacks on initial deposits. stfxeth has protection against loosing all your money (might only lose half?). However this would be very gas expensive to check check for. Perhaps a comment on only going into pools that have existing funds. stfxeth also uses stored asset amounts, which should be donation proof.

The canonical 4626, and stfxeth, correctly rounds `previewRedeem` amounts down, as does redeeming. This means that our balance check should match our worst case.

Deposit and withdraw look correct.

#### Logic

_Looks good_


#### Tests

#### Flavor

_Code is elegant._

#### Overflow

_No math_

#### Proxy
- [x] Make sure proxy implementation contracts don't initialise variable state on variable declaration and do it rather in initialize function.

#### Black magic

#### Dependencies
_No change to standard strategy dependancies._

#### Deploy

#### Authentication

- [x] Never use tx.origin

#### Cryptographic code

_no crypto_

#### Gas problems

_no loops, other than one in our control in initialization._

#### External calls

- [x] Oracles, one of:
  - [ ] No oracles
  - [x] Oracles can't be bent

I notice that we have a reentrancy checks on the state changing methods, but not on check `checkBalance`. It's probably not worth adding a read only reentrancy check on this method, since the vault should have it's own reentrancy checks that should prevent a rebase from happening in the middle of a deposit or a withdraw. Oh hey! The vault core does have these checks, but the vault admin does not!

📛 Vault Admin needs to have reentrancy checks on the strategist moving funds. If an attacker somehow gains execution in the middle of a funds move, this will prevent the attacker from being able to make a rebase. ( Issue created on this https://github.com/OriginProtocol/origin-dollar/issues/1366 )

#### Ethereum

DanielVF:
# Governance Ownership Checks

✅ All looks good

### Multisig owned

OETH Proxy
0xbe2ab3d3d8f6a32b96414ebbd865dbd276d3d899

VaultProxy
0xbe2ab3d3d8f6a32b96414ebbd865dbd276d3d899

Vault
0xbe2ab3d3d8f6a32b96414ebbd865dbd276d3d899

VaultAdmin
0xbe2ab3d3d8f6a32b96414ebbd865dbd276d3d899

VaultCore
0xbe2ab3d3d8f6a32b96414ebbd865dbd276d3d899

FraxEthStrategyProxy
0xbe2ab3d3d8f6a32b96414ebbd865dbd276d3d899

Frax Strategy Implementation
0xbe2ab3d3d8f6a32b96414ebbd865dbd276d3d899

Woeth Proxy*
0xbe2ab3d3d8f6a32b96414ebbd865dbd276d3d899

Dripper Proxy
0xbe2ab3d3d8f6a32b96414ebbd865dbd276d3d899


### Deployer owned:


OETH Impl
0xfd9e6005187f448957a0972a7d0c0a6da2911236

Woeth Impl
0xfd9e6005187f448957a0972a7d0c0a6da2911236

Dripper IMPL
0xfd9e6005187f448957a0972a7d0c0a6da2911236

### No Owner:

OETH Oracle Router
No Owner

Zapper
No Owner

OUSD Oracle Router
No Owner

DanielVF:
# Code Review: OETH Vault

#### Requirements

This change is designed to allow the vault to work with rebasing tokens that are not always worth 1 ETH (or 1 USD). The vault does this by internaly converting these coins into 1e18 units that correspond to an ETH (or USD). This actually simplifies a lot of the code, since before the vault constantly had to do conversions in and out of the token's decimals.

This change is intended to be backwards compatible (outside of the OracleRouter, and price views) with the old code. This means that all existing tests should pass.

### Deployment Considerations

🔆 Requires new OracleRouter

#### Attack

These new changes affect the three most critical parts of the vault - minting, redeeming, and balance calculations for rebasing. A calculation failure in any one of these areas would be GG for the protocol.

An attacker's goal would be to create the impression that their assets are worth too much during mints, worth too much during rebase, or worth too little during a redeem.

There's a discussion of oracles on the OracleRouter review. Even with bad oracles, the total number of stables should only go up -- as long as the  exchange rate does not go down.

The fact that an exchange rate could go down is a new twist that we didn't have in OUSD. We also have rebasing tokens that could balance down. To some degree, this "we have less backing funds now" is a core risk we are taking with OETH.

📛 I belive there is a profitable attack that would require several things to line up. Attack goes like this:

Attacker knows that a bigger than redeem fee exchange rate drop on reth is coming. Attacker can force the chainlink oracle to update to reflect the new exchange rate, but the Attacker can buy/sell reth at approximately the same price before and after the exchange rate drop.

In this case, the attacker would wrap the exchange rate and oracle update transactions, acquiring a large amount of RETH before the updates and depositing into OETH. Then after the update, they would redeem. Given the new exchange rate, OETH would treat rETH as less valuable, and send more of it to make up an ETH. In the right circumstances then, the attacker would net more than they put in.

If the chainlink oracle doesn't update at the same time, this attack does not work. If the oracle is at the new low price, the mint gets subtracted from, if the oracle is at the old high price, the redeem gets subtracted from. If the oracle is in between, it affects both sides.

We do have defenses intended to limit the scope of such an attack. If the vault value is too low of a percentage of the ousd total supply, then the mints will revert. However this vault check happened before the redeem happened. Here's why this is bad:

|What|Total Supply|Vault Value|
|----|----|----|
|Initial|10|10|
|Attacker moves in|110|110|
|Bad thing reduces vault value|110|106|
|Attacker moves 100 worth back out|10|6|

The vault value check should happen after the redeem, not before.

I've added this change here:

https://github.com/OriginProtocol/origin-dollar/pull/1271/commits/508921bd39f988fa61b60cea372b910725ff7bd0


#### Logic

I've simmed out the math on paper and it seems to work. Old OUSD tests pass, and new exchange rates test pass.

#### Tests

  - [x] Exchange rates tests pass
  - [x] Past OUSD tests pass

#### Flavor

Why do we have all the virtuals in VaultCore? Those should not be needed?

#### Overflow

#### Proxy

#### Black magic

🔆 We don't have our vault core proxy fix merged in. This means that we MUST have multisig ownership of the VaultCore implementation contract. We do have this in the current live contract.

#### Dependencies

_No new dependancies, and no change in dependancies._

#### Deploy

#### Authentication

- [x] Never use tx.origin

#### Cryptographic code

_No crypto_

#### Gas problems

    - [ ] A way to remove items

#### External calls

- [x] Reentrancy guards on all state changing functions _(True in vault core, issue in vault noted elsewhere.)_
- [x] Oracles, one of:
  - [ ] No oracles
  - [ ] Oracles can't be bent

As in the long discussion in the OracleRouters review, our current oracles are bad. However, we should still keep at least the same assets.


#### Ethereum

sparrowDom (approved):
LGTM

Lets merge this to master since it is already live on mainnet for some time

PR #1277, opened 2023-04-03
Increase performance fee and update Buyback for increased Treasury

shahthepro:
- [x] Increase performance fee to 20%
- [x] Add support for transferring a part of the fee to the strategist
- [x] Add deployment script for the new contract
- [x] Move funds out of the old Buyback contract
---------

Contract change checklist:
  - [x] Unit tests pass

### DIFF
--- a/contracts/deploy/051_upgrade_buyback.js
+++ b/contracts/deploy/051_upgrade_buyback.js
@@ -1,3 +1,4 @@
+const { isFork } = require("../test/helpers");
 const { deploymentWithGovernanceProposal } = require("../utils/deploy");
 const { getTxOpts } = require("../utils/tx");
 // const addresses = require("../utils/addresses");
@@ -5,7 +6,7 @@ const { getTxOpts } = require("../utils/tx");
 module.exports = deploymentWithGovernanceProposal(
   {
     deployName: "051_upgrade_buyback",
-    forceDeploy: true,
+    forceDeploy: isFork,
   },
   async ({
     withConfirmation,


DanielVF (changes requested):
# Buyback Review


#### Requirements

Rather than spending all protocol fees on buying back OGV, we can take a 0-100% percentage and directly send it to a treasury account. 

This PR will also configure the vault to increase the percentage sent to the buyback from 10% to 20%

- [ ] We need to check and update our current monitoring. 
- [ ] We'll need to update the expected limits in our current automation around buybacks. It should error at the moment, since the amount received is 50% less.
- [ ] (🟠 Change needed) I don't think there's any OGN on the "old" buyback contract - I think it's on an even older one. We should probably actually move that very old OGN to the strategist address. 
- [ ] We'll need to not do buybacks using the old contract once we deploy this contract.

#### Internal State

_All state is config_

#### Attack

If attacked, this contract has a limited amount of damage that can be done, since it only holds funds in flight.

We assume the min OGV received amount is set honestly, since low numbers there could provide MEV opportunity.

The funds recipients are all loaded from config variables. If they could be changed, that would be an attack. It looks like all config set methods have the correct authentication on them.

#### Logic

The math involves splitting up the funds between swapping and transferring. This looks correct.


#### Tests

#### Flavor

- [ ] (🟠 Change optional) We could deduplicate the uniswap sets by making them into a private method like we do with the other config methods.

#### Overflow

_No problems_

#### Proxy

_No proxy_

#### Black magic

_No black magic._

#### Dependencies

_No dependency changes_

#### Deploy

#### Authentication

- [x] Never use tx.origin

#### Cryptographic code

_No Crypto_

#### Gas problems

_No loops_

#### External calls

- [x] Oracles, one of:
  - [x] No oracles
  - [ ] Oracles can't be bent


#### Ethereum

shahthepro:
@DanielVF Regarding the OGN, [this is the Buyback contract](https://etherscan.io/address/0x77314eb392b2be47c014cde0706908b3307ad6a9) that has about 938k OGN balance. 

The deployment and script files use that address for OGN as well. If you run the deployment on fork and run the script, it'll print the balances it has. Sorry if my naming convention was confusing but I used `oldBuybackContract` for the the one holding OGN and `recentBuybackContract` for the currently active one 😅

contracts/deploy/052_upgrade_buyback.js
```diff
@@ -0,0 +1,148 @@
...
+          args: [cOGN.address, ognBalance],
+        },
+        {
+          // 6. Transfer OUSD balance from Governor to New Contract
+          contract: cOUSD,
+          signature: "transfer(address,uint256)",
+          args: [cBuyback.address, ousdBalance],
+        },
+        {
+          // 7. Transfer OGN balance from Governor to New Contract
```
DanielVF, line 140: Since the community has voted on spending the OGN, let's have the old OGN end up on the strategist multisig, rather than the new buyback.

DanielVF (approved):

Issue #1259, opened 2023-03-07
Clean up usage of fixtures and optimize unit tests


shahthepro:
Right now, the contracts use `hardhat-deploy` plugin for deployments. However, we have been wrongly using `ethereum-waffle` package's `createFixtureLoader()` method. This essentially means that right now, none of the unit/fork tests are using fixtures at all. Changing them to use `hardhat-deploy` packages [`deployments.createFixture`](https://github.com/wighawag/hardhat-deploy#creating-fixtures) fastens the tests by quite a significant time

naddison36:
I've worked on this today in PR https://github.com/OriginProtocol/origin-dollar/pull/1745

PR #1317, opened 2023-04-11
Unit test fixes

sparrowDom:
Unit test fixes

### DIFF
--- a/contracts/test/vault/harvester.js
+++ b/contracts/test/vault/harvester.js
@@ -28,10 +28,37 @@ describe("Harvester", function () {
     await comp.connect(governor).transfer(compoundStrategy.address, compAmount);
   };
 
+  let fixture;
+
+  beforeEach(async function () {
+    fixture = await loadFixture(compoundVaultFixture);
+
+    /* Ethereum Waffle caches fixtures and uses evm snapshot and evm revert:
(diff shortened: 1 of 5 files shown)


sparrowDom:
Merging this in so I can apply the test fixes to the router branch.

contracts/test/vault/harvester.js
```diff
@@ -28,10 +28,37 @@ describe("Harvester", function () {
     await comp.connect(governor).transfer(compoundStrategy.address, compAmount);
   };
 
+  let fixture;
+
+  beforeEach(async function () {
+    fixture = await loadFixture(compoundVaultFixture);
```
shahthepro, line 34: nit: We aren't using the fixtures the way it's supposed to be. So, using `loadFixture` has zero effect on the performance. We should switch to use `deployments.createFixture` instead. I already started doing that on the Uniswap V3 branch. It'd be nice if we can do that for all new tests as well: <https://github.com/OriginProtocol/origin-dollar/pull/1258/files#diff-42bf690234c6c24cf49e6eee681c1d54e88995355d811b66366a640e03e935dbR1253>

Related issues: https://github.com/OriginProtocol/origin-dollar/issues/1259

PR #1319, opened 2023-04-12
move decimals cache to asset struct

sparrowDom:
move decimals cache to asset struct


sparrowDom:
merging this in will address issues later if needed

PR #1321, opened 2023-04-13
support oracles reporting feeds with different decimal format

sparrowDom:
support oracles reporting feeds with different decimal format


contracts/contracts/oracle/OracleRouter.sol
```diff
@@ -34,14 +38,35 @@ abstract contract OracleRouterBase is IOracle {
...
         }
         return uint256(_price);
     }
 
+    function getDecimals(address _asset)
+        internal
+        view
+        returns (uint8)
+    {
+        require(decimalsCache[_asset] > 0, "Oracle: Decimals not cached");
```
DanielVF, line 56: This does two reads. Let's use a temp var

sparrowDom: yes, thanks... done

sparrowDom:
Merging this in. 4 tests are failing but are not related to the changes in this branch. (well 1 is but we will update the oracle address once that gets deployed on the mainnet)

PR #1454, opened 2023-05-11
add ledger live connector

rolandpo:
add connector for integration into Ledger Live Dapp Browser, which will make it possible down the road to view the dapp directly through the Ledger Live iframe and connect automatically to the active account

### DIFF
--- a/dapp-oeth/ledger_manifest.json
+++ b/dapp-oeth/ledger_manifest.json
@@ -3,7 +3,7 @@
   "name": "OETH",
   "url": "https://dapp-browser.apps.ledger.com",
   "params": {
-    "dappUrl": "http://localhost:3000/?embed=true",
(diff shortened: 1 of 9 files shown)


shahthepro (approved):
LGTM

rafaelugolini:
I tried here and got this:
image

Am I doing something wrong?

rolandpo:
Hey, the way it's meant to work is through the Ledger Live Discover section. Hopefully it gets added there once I finish writing the dapp plugin, getting some other necessary files and we can officially submit a request.

Right now, the way to test it would be to run the dapp locally, open Ledger Live in development mode and import the ledger_manifest.json file. The dapp would then open inside the iframe and connect automatically to the Ledger user's account, without the option to disconnect or switch users

PR #1471, opened 2023-05-16
Fix sfrxeth -> oeth receive amount estimate

smitch88:
The previous `sfrxETH` estimate was effectively hardcoded to 1. This fix calls into the `sfrxETH` contract to get the `previewDeposit` value and multiplies it by the OETH vault mint price estimate for `frx` to get a conversion. This conversion is applied to the requested swap amount for display purposes.

`Fixed sfrxETH -> OETH`
Screenshot 2023-05-16 122135


dapp-oeth/src/hooks/useSwapEstimator.js
```diff
@@ -374,6 +370,21 @@ const useSwapEstimator = ({
       }
     }
 
+    let amountReceived = amount
+
+    if (coinToSwap === 'sfrxeth') {
+      const [estimatedDeposit, frxEthMintPrice] = await Promise.all([
```
shahthepro, line 376: Should we add some error handling?

shahthepro (approved):

dapp-oeth/src/hooks/useSwapEstimator.js
smitch88: I think by default if this errors in the promise it would already throw an `unexpected_error`, do we want to do something different?

shahthepro: Not needed in that case

Issue #1366, opened 2023-04-27
Vault Admin add reentrancy protection


DanielVF:
📛 Vault Admin needs to have reentrancy checks on the strategist moving funds. If an attacker somehow gains execution in the middle of a funds move, this will prevent the attacker from being able to make a rebase.

Note that this is an additional prevention layer. Currently nether OUSD nor OETH handle off execution in the middle of deposits / withdraws.

PR #1486, opened 2023-05-16
Add reentrancy guards to our strategist allocation methods

DanielVF:
Our user facing, state changing vault methods have reentrancy protection guards on them. This PR adds the same protection to the vault admin's `depositTo()` and `withdrawFrom()`. In this way if an attacker gets execution during an admin reallocation, they cannot make writes to the vault during that.

Note that in theory we don't need reentrancy guards in the vault (currently none of our strategies allow handing off execution beyond the strategies they invest in), but they provide a second layer of protection against reentrancy in case something is missed somewhere during future development, or a compromise of an underlying strategy.

This PR also makes the vault `totalVault()` method revert if read while the vault is inside reentrant protected code. This will prevent a third party from getting wrong data if read while inside our vault executing code. https://twitter.com/danielvf/status/1657019677544001536

Closes [1366](https://github.com/OriginProtocol/origin-dollar/issues/1366)

### DIFF
--- a/contracts/contracts/governance/Governable.sol
+++ b/contracts/contracts/governance/Governable.sol
@@ -134,6 +134,26 @@ contract Governable {
         }
     }
 
+    /**
+     * @dev Prevents a view contract function from being called
+     * while the contract is inside a reentrant protected method.
+     * This can prevent third party contracts from being attacked
+     * via working with partially updated data.
+     */
+    modifier nonReentrantView() {
+        bytes32 position = reentryStatusPosition;
+        uint256 _reentry_status;
+        // solhint-disable-next-line no-inline-assembly
+        assembly {
+            _reentry_status := sload(position)
+        }
+
+        // On the first call to nonReentrant, _notEntered will be true
+        require(_reentry_status != _ENTERED, "Reentrant call");
+
+        _;
+    }
+
     function _setPendingGovernor(address newGovernor) internal {
         bytes32 position = pendingGovernorPosition;
         // solhint-disable-next-line no-inline-assembly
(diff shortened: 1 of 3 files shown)


naddison36:
There are gas costs implications for adding this precautionary protection.

Here's the change in average gas costs of the OETH Vault from running the unit tests

| Method | Old | New | Difference |
|---|---|---|---|
| depositToStrategy | 145,291 | 148,861 | 3,570 |
| withdrawFromStrategy | 139,857 | 142,794 | 2,937 |

contracts/contracts/governance/Governable.sol
```diff
@@ -134,6 +134,26 @@ contract Governable {
         }
     }
 
+    /**
+     * @dev Prevents a view contract function from being called
+     * while the contract is inside a reentrant protected method.
+     * This can prevent third party contracts from being attacked
+     * via working with partially updated data.
+     */
+    modifier nonReentrantView() {
```
naddison36, line 143: Contracts will be smaller if the modifier calls an internal function. Modifiers are like macros in they copy the code to all functions that are applied to. Copying a call to an internal function is a lot less duplicate code to copying the whole function.

DanielVF: This sounds worth doing then. Particularly our main non-reentrant modifier.

DanielVF: Sure

naddison36:
@DanielVF you are correct. It's better to not split the `nonReentrant` modifier into two internal calls. It added 660 gas to the `redeem` function but only reduce the contract size by 0.019 KB.

I've rebased your branch with `master`

naddison36 (approved):

sparrowDom (approved):
Love this extra layer of protection and future proof thinking

PR #1500, opened 2023-05-17
OETH L-09

sparrowDom:
Use safe cast


DanielVF:
What's your thinking on copying the code in from OpenZeppilin vs importing it? I can see pros and cons on each side, just curious on how you see it.

sparrowDom:
I though the whole library gets included into the bytecode by the compiler. Have done a little digging and the compiler seems to include only library functions that are actually consumed by the contract: [documentation](https://ethereum.stackexchange.com/questions/96853/does-solidity-embed-only-the-library-functions-that-are-consumed) 

So my original reason isn't valid. Changing to library import

DanielVF (approved):

PR #1501, opened 2023-05-17
OETH N-01 - Scope

sparrowDom:
Add scope to variables

### DIFF
--- a/contracts/contracts/vault/VaultCore.sol
+++ b/contracts/contracts/vault/VaultCore.sol
@@ -27,9 +27,9 @@ contract VaultCore is VaultStorage {
     using StableMath for uint256;
     using SafeMath for uint256;
     // max signed int
-    uint256 private constant MAX_INT = 2**255 - 1;
+    uint256 internal constant MAX_INT = 2**255 - 1;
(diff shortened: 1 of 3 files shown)


contracts/contracts/vault/VaultCore.sol
```diff
@@ -27,9 +27,9 @@ contract VaultCore is VaultStorage {
     using StableMath for uint256;
     using SafeMath for uint256;
     // max signed int
-    uint256 constant MAX_INT = 2**255 - 1;
+    uint256 private constant MAX_INT = 2**255 - 1;
```
DanielVF, line 30: Why are these two private and the others internal?

sparrowDom: My thinking is no Contract is going to be extending the VaultCore.

DanielVF: I'd rather just stick with internal, it shouldn't matter gas or contract size wise.

sparrowDom: ok sounds good

PR #1506, opened 2023-05-18
OETH N 05

sparrowDom:
Remove unneeded SafeMath in favour of build int 0.8.0 overflow protection


contracts/contracts/vault/VaultCore.sol
```diff
@@ -218,7 +216,7 @@ contract VaultCore is VaultStorage {
             // backing assets value and OUSD total supply
             uint256 diff = oUSD.totalSupply().divPrecisely(totalUnits);
             require(
-                (diff > 1e18 ? diff.sub(1e18) : uint256(1e18).sub(diff)) <=
+                (diff > 1e18 ? diff - 1e18 : uint256(1e18) - diff) <=
```
DanielVF, line 219: Do we need this cast still?

DanielVF (approved):

contracts/contracts/vault/VaultCore.sol
sparrowDom: thanks good catch

PR #1524, opened 2023-05-19
Update Buyback scripting

sparrowDom:
update buyback scripting for the new buyback implementation


brownie/buyback.py
```diff
@@ -45,6 +44,11 @@ def build_buyback_tx(max_dollars=5000, max_slippage=2.0):
       Percentage of slippage from current prices to allow.
     """
     buyback = BUYBACK_CONTRACT
+    treasuryBps = BUYBACK_CONTRACT.treasuryBps()
+    ousd_for_treasury = max_dollars * treasuryBps / 1e4
```
DanielVF, line 48: I've been doing this wrong all this time, but 1e4 / 1e18 return floats. For the precise int's we should use 10**4 or 10**18 in python.

image

sparrowDom: Oh wow me too -.-. Changing all to ints

PR #1535, opened 2023-05-23
Test fixes

sparrowDom:
Curve OUSD Metapool is right now on the mainnet so heavily imbalanced that the funds within it are not sufficient enough for it to balance itself. In such cases the tests have an additional feature where an account with enough CRV deposits liquidity so pool has enough funds.


sparrowDom:
Merging this in for the tests on master to pass. Happy to address any concerns subsequently

Issue #1429, opened 2023-05-06
Support multi-step swaps with Curve


micahalcorn:
When a user has ETH and wants OETH, there are two options: minting with the zapper or swapping through the Curve ETH/OETH pool. However, if she has stETH or another LSD and wants OETH, minting is currently the only option. Due to the frequency of stETH (and others) being off-peg, there's a high likelihood that users won't be able to acquire OETH from our dapp. Let's add support for additional Curve swaps beyond just ETH/OETH.

In order of priority...
- [ ] stETH
- [ ] WETH
- [ ] rETH
- [ ] frxETH

rafaelugolini:
@smitch88 isn't this merged?

smitch88:
Yup, closing

PR #1538, opened 2023-05-23
OETH Dapp, curve swap support for stETH, WETH, rETH, and frxETH

smitch88:
Code changes to support #1429 

Add support for curve swaps for mint/redeem of OETH
 - stETH, WETH, rETH, frxETH

This implementation is more of a hardcoded route approach with the structure of the routes defined in `curveRoutes` as a constant. This could be fetched dynamically but I believe it would be easier to do this if we integrate `@curvefi/api` rather than deriving this ourselves from contract calls.

* Updates `dapp-oeth/abis/CurveRegistryExchange.json` to include `get_exchange_multiple_amount` for estimate support and `exchange_multiple` for actual swap support.
* Updates `useCurrencySwapper` to integrate the new curve quotes
* Updates allowances on the initial mount to retrieve token allowances from the curve router contract for each of stETH, WETH, rETH, and frxETH


HrikB (changes requested):
For me, I am unable to make the swaps happen from OETH -> [Collateral of Choice]

When I click the approve button, I provide an approval to the OETH/ETH pool contract at `0x94B17476A93b3262d87B9a326965D1E91f9c13E7`. However, the route requires calling `transferFrom` from the address `0x99a58482bd75cbab83b27ec03ca68ff489b5788f` which is the registry exchange contract. So my transaction fails because that approval has not been provided.
image

HrikB:
Seems like `0x99a58482bd75cbab83b27ec03ca68ff489b5788f` is the right one to approve. On a OETH -> stETH trade on curve.fi, that is the address needing to be approved.

smitch88:
Interesting, looking at this now

smitch88:
@HrikB should be good to go now

HrikB (changes requested):
Ok I've got the approval working!

The only things that need work right now are as per our discussions on discord:

- Fixing the loading state
- Figuring out bug that is causing OETH -> rETH swap to fail

HrikB (approved):

PR #1552, opened 2023-05-27
banner

HrikB:


dapp/src/components/layout.js
```diff
@@ -97,6 +97,18 @@ const Layout = ({
           </>
         )}
       </Head>
+
+      <div
+        className={classnames('notice text-white text-center p-3 dapp d-none')}
+      >
+        <div className="container d-flex flex-column flex-md-row align-items-center justify-content-center">
+          <span className="text">
+            OETH has been paused as a security precaution. Minting and redeeming
```
shahthepro, line 106: Shouldn't this be OUSD

dapp-oeth/src/components/layout.js
```diff
@@ -68,7 +68,18 @@ const Layout = ({
           content="https://ousd.com/images/share-twitter.png"
         />
       </Head>
-
+      <div
```
shahthepro, line 71: Can we put these alerts behind a ENV var so that we don't have to submit a PR and deploy it everytime?

shahthepro (approved):

dapp-oeth/src/components/layout.js
shahthepro: Looking at this again, I guess it makes more sense to read from the contracts directly. This should never require deployment or ENV change

rafaelugolini: I think we are over using ENV variables. The usage of const is also good :)

PR #1557, opened 2023-05-29
Proxy init to change governor at the end

naddison36:
- Proxy `initialize` to do governor change at the end so the implementation contract can be initialised
- Simplified the OETH Morpho deploy script by setting the governor on proxy initialize

### DIFF
--- a/contracts/test/harvest/ousd-harvest-crv.fork-test.js
+++ b/contracts/test/harvest/ousd-harvest-crv.fork-test.js
@@ -34,7 +34,7 @@ forkOnlyDescribe("ForkTest: Harvest OUSD", function () {
       const crvHarvested = balanceAfterHarvest.sub(balanceBeforeHarvest);
       expect(crvHarvested).to.be.gt(parseUnits("20000", 18));
     });
-    it("should harvest and swap", async function () {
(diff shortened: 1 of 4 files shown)


contracts/test/harvest/ousd-harvest-crv.fork-test.js
```diff
@@ -64,7 +64,15 @@ forkOnlyDescribe("ForkTest: Harvest OUSD", function () {
...
-    it("should not harvest and swap", async function () {
+    /*
+     * Skipping this test as it should only fail on a specific block number, where
+     * there is:
+     *  - no liquidation limit
+     *  - strategy has accrued a lot of CRV rewards
+     *  - depth of the SushiSwap pool is not deep enough to handle the swap without
+     *    hitting the slippage limit.
+     */
+    it.skip("should not harvest and swap", async function () {
```
sparrowDom, line 75: @naddison36 these tests were failing because they would only work on a block number where our strategy had a great amount of rewards accrued. Added comments with explanation.

I would love to re-enable them, but we would need to upgrade our fork test environment in a way where some tests could be ran against a specific block number. The CI just runs them against the latest block number

sparrowDom (approved):

Issue #1540, opened 2023-05-24
off-chain component that interacts with the 1Inch API and the Vault


naddison36:
# Scope

This task is to implement the off-chain collateral asset swap functionality for the OETH Vault.

The on-chain contract component are being implemented in https://github.com/OriginProtocol/origin-dollar/issues/1539

## Out of scope

Implementing collateral swaps on the OUSD Vault can be done separately at a later point.

## Requirements

The relevant assumptions from the [design decisions](https://www.notion.so/Analysis-of-OETH-Collateral-Swapper-1d9d8b2d2b104968ad81c734d161fe58) are:

- The strategist mutlisig will execute the collateral swaps.
- 1Inch Pathfinder V5 will be used as the swap provider.
- There will be two layers of MEV protection
  - The strategist will pass in a min amount to be returned from the swap.
  - The vault will check that the return amount is not outside the Oracle deviation limits of the from and to asset exchange rates.
- The strategist will decide if partial fills are allowed or not for each swap. There will be no on-chain enforcement of this.

## Process

This is the happy path for the off-chain command line interface (cli) component.

- A signer of the strategist multisig inputs the required swap details including the from asset, from asset amount, to asset and slippage amount as a percentage. eg 0.1 is 0.1% which is 10 basis points.
- An optional parameter will if partial fills will be allowed or not. This is a parameter of the 1Inch swap API call. The default will be partial fills will not be allowed
- The off-chain component will get the Oracle exchange rates from the chain.
- The off-chain component will warn if the exchange rates are stale??
- The off-chain component uses the Oracle exchange rates and the entered slippage to calculate a minimum to asset amount.
- The off-chain component calls the 1Inch Pathfinder API /v5.0/1/swap
- The off-chain component outputs the details of the swap including final to asset amount and swap routes
- The off-chain component gets the `tx.data` from the response
- The off-chain component builds a tx that calls the `collateralSwap` function is called on the Vault with the from asset, from asset amount, to asset and 1Inch router data.
- The off-chain component outputs the `collateralSwap` tx details so they can be copied into the strategist multisig's tx builder.

## Testing Strategy

The collateral swap will be fork tested using the production 1Inch API and a local mainnet fork.

## Resources

- [1Inch V5 Pathfinder introduction](https://docs.1inch.io/docs/aggregation-protocol/introduction)
- [1Inch V5 Pathfinder Swapper docs](https://docs.1inch.io/docs/aggregation-protocol/api/swagger)
[...]

Issue #1539, opened 2023-05-24
Add collateral swap function to Vault


naddison36:
# Scope

This task is to implement the on-chain collateral asset swap functionality on the OETH Vault contract.

The off-chain component that calls the 1Inch API is being implemented in https://github.com/OriginProtocol/origin-dollar/issues/1540

## Out of scope

- Implementing collateral swaps on the OUSD Vault can be done separately at a later point.

# Requirements

The relevant assumptions from the [design decisions](https://www.notion.so/Analysis-of-OETH-Collateral-Swapper-1d9d8b2d2b104968ad81c734d161fe58) are:
- The strategist decides off-chain what assets should be swapped and how many.
- The strategist executes the collateral swaps.
- As the strategist is executing the swap, there is no need for gas compensation.
- As the executor of the collateral swap is the strategies, it will not be compensated for gas.
- There will be two layers of MEV protection
   1.  The strategist will pass in a min amount to be returned from the swap.
   2. The vault will check that the return amount is not outside the Oracle deviation limits of the from and to asset exchange rates.
- 1Inch Pathfinder V5 will be used as the swap provider.
- The vault will approve the swap provider to transfer the from asset on each swap.
- The strategist will decide if partial fills are allowed or not for each swap. There will be no on-chain enforcement of this.
- Collateral swaps will only be done against assets in the vault. The strategist can construct a transaction that withdraws assets from a strategy, does a swap and then deposits assets to another strategy.
- Swaps can not be executed when the Vault has paused capital.
- The vault will only allow approved collateral assets to be swapped.
- The vault will prevent swapped assets to leave the vault.
- The vault will check the before and after to asset balance has increased enough post swap.
- The collateral swap will only support one-to-one asset swaps. If multiple swaps are required, the swap function be called multiple times.

# Testing Strategy

- The collateral swap will be unit tested by mocking the 1Inch Router.
- The collateral swap will be fork tested using the production 1Inch API and a local mainnet fork.

# Resources

- [1Inch V5 Pathfinder introduction](https://docs.1inch.io/docs/aggregation-protocol/introduction)
- [1Inch V5 Pathfinder contract code](https://vscode.blockscan.com/ethereum/0x1111111254EEB25477B68fb85Ed929f73A960582)
- [1Inch V5 Pathfinder  swap parameters](https://docs.1inch.io/docs/aggregation-protocol/api/swap-params)
[...]

PR #1559, opened 2023-05-30
OETH Collateral Swaps

naddison36:
# Changes

See the following issue for the specifications
- Add collateral swap function to Vault [#1539](https://github.com/OriginProtocol/origin-dollar/issues/1539)
- off-chain component that interacts with the 1Inch API and the Vault [#1540](https://github.com/OriginProtocol/origin-dollar/issues/1540)

Notion resources
- [Solution analysis](https://www.notion.so/OETH-Collateral-Swaps-1d9d8b2d2b104968ad81c734d161fe58)
- [Design decisions](https://www.notion.so/813478cd16c84f3bb869a5e6ff0a2eda?v=20b5d1c1681f4647b6099ea1ea6251cf)

Significant contract changes
- [x] New `Swapper1InchV5` contract that implements the `ISwapper` interface.
- [x] `swapCollateral` function added to the `VaultCore`
- [x] `setSwapper`, `setOracleSlippage` and `setSwapAllowedUndervalue` added to `VaultAdmin`.
- [x] New Vault view functions `swapper` and `allowedSwapUndervalue`.
- [x]  The vault's `assets` mapping to `Assets` has been repacked. It was taking up 2 slots as the `decimals` was type uint256. Adding the new `allowedSwapSlippageBps` property would mean 3 slots would be used for each asset. The `decimals` has been retyped to a `unit8` which means all fields now fit in one slot. That means all the assets in the vault need their decimals recached as the decimals location has moved. It also means the deployment of the the new `OETHVaultCore` and `OETHVaultAdmin` contracts need to override the storage safety check.
- [x] Merge Vault gas savings https://github.com/OriginProtocol/origin-dollar/pull/1578 and https://github.com/OriginProtocol/origin-dollar/pull/1613
- [x] `nonReentrantView` removed from `Governable`
- [x] Removed `SafeMath` from `InitializableAbstractStrategy`
- [x] Removed `SafeMath` and `StableMath` from `VaultStorage`

Other significant changes

- [x] The `ethereum-waffle` dependency has been upgraded so the `withNamedArgs` chai matcher can be used for event testing.
- [x] Added a logger for unit and fork tests. See `./utils/logger.js`
- [x] `axios` added as a dependency to call the 1Inch swap API.

# Review

Contract change checklist:
  - [x] Copy & paste code review [security checklist](https://github.com/OriginProtocol/security/blob/master/templates/Contract-Code-Review-Example.md) below this checklist.
  - [x] Unit tests pass

Dependent change requests

-  [x] https://github.com/OriginProtocol/origin-dollar/pull/1637
-  [x] https://github.com/OriginProtocol/origin-dollar/pull/1633

# Deployment

## 1Inch Swapper

 - [x] 065 deployment script executed
    the `Swapper1InchV5` contract was deployed to [0xcD0fcF8a31Bc78ec07752e9CCD3960E936D18366](https://etherscan.io/address/0xcD0fcF8a31Bc78ec07752e9CCD3960E936D18366#code)

## OETH Vaults
 - [x]  Rebase `master` to `nicka/collateral-swaps`
 - [ ] Contract deployment
   - [ ] Propose governance changes to the OETH Governor contract for the 66 deployment script.
   - [ ] Raise separate Governor txs to `queue` and `execute` the proposal
   - [ ] 5 of the Governor multisig signers sign the Governor txs
   - [ ] Execute the `queue` proposal tx
   - [ ] After 24 hours, execute the `execute` proposal tx
 - [ ] Commit deployment artifacts and add block number to `contracts/deployments/mainnet/.migrations.json`
 - [ ] Merge to `master`
 - [ ] GitBook updates
   - [ ] OETH Registry docs updated with new Swapper contract and updated vault implementation contract addresses
   - [ ] Governance docs updated to include `swapCollateral` and `pauseRebase` as functions that the Strategist can execute.
   - [ ] Remove `reallocate` as a function that the Strategist can execute.
 - [ ] Swapper contract added to Immunefi bounty program
 - [ ] Resubmit OETH Vault contract for Dune decoding of new functions and events
 
# Post deployment

- [ ] Change PR https://github.com/OriginProtocol/origin-dollar/pull/1631 to merge back to `master`


contracts/contracts/vault/VaultCore.sol
```diff
@@ -390,6 +395,76 @@ contract VaultCore is VaultStorage {
...
+            "Strategist slippage limit"
+        );
+
+        // Check the slippage against the Oracle in case the strategist made a mistake or has become malicious.
+        // to asset amount = from asset amount * from asset price / to asset price
+        uint256 minOracleToAssetAmount = (_fromAssetAmount *
+            (1e4 - fromAssetConfig.allowedSwapSlippageBps) *
+            IOracle(priceProvider).price(_fromAsset)) /
+            (IOracle(priceProvider).price(_toAsset) *
+                (1e4 + toAssetConfig.allowedSwapSlippageBps));
```
sparrowDom, line 458: I don't think we should consider the slippage twice - for each of the assets present in the exchange. Normally when I think of slippage when it comes to a transaction I think of the price movement in a pool. Which would mean that slippage pertains to a pair of assets rather than 1 single asset. 

1. Since here we configure it per asset I think we should either rename it to `allowedSwapSlippagePortionBps` or `allowedHalfSwapSlippageBps` to try to communicate that the configured slippage represents only a portion of the total slippage in transaction. (though I like none of the name suggestions 🙂  )

2. Or leave it as is and change the calculation to: 
```
uint256 minOracleToAssetAmount = (_fromAssetAmount *
            IOracle(priceProvider).price(_fromAsset)) /
            (IOracle(priceProvider).price(_toAsset)) * 
            (1e4 - fromAssetConfig.allowedSwapSlippageBps);
```

3. If you want to account for different assets having different liquidity and the swap to account for both slippages when considering min amount
```
uint256 minOracleToAssetAmount = (_fromAssetAmount *
            (1e4 - fromAssetConfig.allowedSwapSlippageBps / 2) *
            IOracle(priceProvider).price(_fromAsset)) /
            (IOracle(priceProvider).price(_toAsset) *
                (1e4 + toAssetConfig.allowedSwapSlippageBps / 2));
```

I would go with option 3

sparrowDom (commented):
Great work @naddison36 . Made a pass of the core part of the contracts, and haven't gotten to tests yet.

contracts/contracts/swapper/Swapper1InchV5.sol
```diff
@@ -0,0 +1,108 @@
...
+                (bytes4, address, bytes)
+            );
+            SwapDescription memory swapDesc = SwapDescription({
+                srcToken: IERC20(_fromAsset),
+                dstToken: IERC20(_toAsset),
+                srcReceiver: payable(executer),
+                dstReceiver: payable(msg.sender),
+                amount: _fromAssetAmount,
+                minReturnAmount: _minToAssetAmount,
+                flags: 4 // _REQUIRES_EXTRA_ETH is second bit. _PARTIAL_FILL is first bit
```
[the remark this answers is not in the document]
naddison36: yes, but I don't know what the third bit is. I know it's returned from the swap API call.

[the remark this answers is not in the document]
sparrowDom: 3rd bit it `SHOULD_CLAIM` : https://github.com/1inch/1inch-v2-contracts/blob/master/contracts/OneInchExchange.sol#L23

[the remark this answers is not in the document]
sparrowDom: Which seems to mean that it is an ERC20 token that the OneInchExchange transfers to itself. That 3rd bit should be 0 if fromToken would be vanilla ETH.

contracts/deploy/060_harvest_crv_limit_2.js
```diff
@@ -9,6 +9,7 @@ module.exports = deploymentWithGovernanceProposal(
     forceDeploy: false,
     proposalId:
       "19517057494793169051021794342486874975141107655269915104229644719251875808935",
+    forceSkip: true,
```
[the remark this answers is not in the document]
sparrowDom: has been merged

contracts/contracts/vault/VaultCore.sol
```diff
@@ -390,6 +395,76 @@ contract VaultCore is VaultStorage {
...
+
+        // Transfer from assets to the swapper contract
+        IERC20(_fromAsset).safeTransfer(swapper, _fromAssetAmount);
+
+        // Call to the Swapper contract to do the actual swap
+        // slither-disable-next-line unused-return
+        ISwapper(swapper).swap(
+            _fromAsset,
+            _toAsset,
+            _fromAssetAmount - 1,
```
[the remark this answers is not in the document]
sparrowDom: Ah right. We could do the conditional check and subtract 1 only in cases of stETH, but I guess that would be more gas wasteful than leaving it as it is.

A comment in the code to explain the meaning of -1 would be welcome IMO.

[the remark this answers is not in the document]
naddison36: good point. I'll add a comment

contracts/contracts/vault/VaultAdmin.sol
```diff
@@ -275,12 +293,15 @@ contract VaultAdmin is VaultStorage {
...
+        uint256 assetCount = _assets.length;
+        for (uint256 i = 0; i < assetCount; ++i) {
+            require(
+                IStrategy(_strategyToAddress).supportsAsset(_assets[i]),
+                "Asset unsupported"
+            );
         }
         // Tell new Strategy to deposit into protocol
-        strategyTo.depositAll();
+        IStrategy(_strategyToAddress).depositAll();
```
DanielVF, line 304: Many times in this PR, we've moved from something like:

```
foos[I].bar();
```

To something like:

```
foo = foos[i];
foo.bar();
```

I think this is great.

However here we're moving the other way, and I think this makes the code harder to read. Same thing a few more times in this file with strategies.

I've love to be consistent on it. I don't think the separate variable form uses more gas?

sparrowDom: Nick as done these as a separate PR which would be easier to review:
https://github.com/OriginProtocol/origin-dollar/pull/1578
https://github.com/OriginProtocol/origin-dollar/pull/1613

The thing is (that I've also learned via above PRs) that if you do smth like foos[I].length in as a stopping condition in for loop, it gets evaluated every loop performing an SLOAD.

.vscode/launch.json
```diff
@@ -7,13 +7,14 @@
         {
             "type": "node",
             "request": "launch",
-            "name": "Unit tests",
+            "name": "Unit tests - OETH Swapper",
```
DanielVF, line 10: Is this the correct name? Sounds like it's just running OETH Swapper tests.

brownie/collateral_swap.py
```diff
@@ -0,0 +1,285 @@
...
+console_colors = {}
+console_colors["ENDC"] = '\033[0m'
+console_colors["FAIL"] = '\033[91m'
+console_colors["WARNING"] = '\033[93m'
+console_colors["OKGREEN"] = '\033[92m'
+console_colors["OKCYAN"] = '\033[96m'
+
+
+@contextmanager
+def silent_tx():
```
DanielVF, line 29: Lol, we could promote this one up to world.py. Good stuff.

contracts/contracts/oracle/OracleRouter.sol
```diff
@@ -297,3 +297,27 @@ contract OracleRouterDev is OracleRouterBase {
         maxStaleness = fm.maxStaleness;
     }
 }
+
+// @notice Oracle Router that denominates all prices in ETH
+contract OETHOracleRouterDev is OETHOracleRouter {
```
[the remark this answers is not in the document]
DanielVF: Agree

contracts/contracts/swapper/Swapper1InchV5.sol
```diff
@@ -0,0 +1,108 @@
...
+        if (swapSelector == SWAP_SELECTOR) {
+            // Decode the executer address and data from the RLP encoded _data param
+            (, address executer, bytes memory executerData) = abi.decode(
+                _data,
+                (bytes4, address, bytes)
+            );
+            SwapDescription memory swapDesc = SwapDescription({
+                srcToken: IERC20(_fromAsset),
+                dstToken: IERC20(_toAsset),
+                srcReceiver: payable(executer),
```
DanielVF, line 66: Why is this from using a user passed in variable rather just `address(this)`?

contracts/contracts/swapper/Swapper1InchV5.sol
```diff
@@ -0,0 +1,108 @@
...
+     * @param _data RLP encoded executer address and bytes data. This is re-encoded tx.data from 1Inch swap API
+     */
+    function swap(
+        address _fromAsset,
+        address _toAsset,
+        uint256 _fromAssetAmount,
+        uint256 _minToAssetAmount,
+        bytes calldata _data
+    ) external override returns (uint256 toAssetAmount) {
+        require(
```
DanielVF, line 44: Do we need these two checks? The swap will fail if we don't have enough money or approval, or if the min amount was not reached afterwords.

contracts/contracts/swapper/Swapper1InchV5.sol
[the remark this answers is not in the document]
DanielVF: If transfers are cheaper than transferFrom's, we could just transfer the asset to the one inch router before the swap and set this to zero.

contracts/contracts/vault/VaultCore.sol
```diff
@@ -390,6 +395,77 @@ contract VaultCore is VaultStorage {
...
+    /**
+     * @notice Strategist swaps collateral assets sitting in the vault.
+     * @param _fromAsset The token address of the asset being sold by the vault.
+     * @param _toAsset The token address of the asset being purchased by the vault.
+     * @param _fromAssetAmount The amount of assets being sold by the vault.
+     * @param _minToAssetAmount The minimum amount of assets to be purchased.
+     * @param _data implementation specific data. eg 1Inch swap data
+     * @return toAssetAmount The amount of toAssets that was received from the swap
+     */
+    function swapCollateral(
```
DanielVF, line 407: Given that this is a admin function, can it go in VaultAdmin.sol? If there's not room, we can remove reallocate.

naddison36: Good point. I'll move

contracts/contracts/swapper/Swapper1InchV5.sol
naddison36: No, these are not needed. I had them in for testing to get better error messages. I'll remove now

contracts/contracts/swapper/Swapper1InchV5.sol
naddison36: `executer` is the 1Inch contract that will execute the swap. It is not the `Swapper1InchV5` or `AggregationRouterV5` contracts. An example is the `0x92F3f71CeF740ED5784874B8C70Ff87ECdF33588` contract.

.vscode/launch.json
naddison36: The name is correct as it has the `test/vault/oeth-swapper.js` param. I'll add another launcher for all unit tests

contracts/contracts/vault/VaultAdmin.sol
naddison36: The main reason I prefer the `IStrategy(_strategyToAddress)` style is it uses one less variable which helps avoid stack too deep errors.

The gas saving in this example is very small (5 gas). 
using `strategyTo` style
```
|  Contract                               ·  Method                              ·  Min        ·  Max        ·  Avg        ·  # calls      ·  eur (avg)  │
|  OETHVault                              ·  reallocate                          ·     246075  ·     248303  ·     247189  ·            2  ·          -  │
```

using `IStrategy(_strategyToAddress)` style
```
|  Contract                               ·  Method                              ·  Min        ·  Max        ·  Avg        ·  # calls      ·  eur (avg)  │
|  OETHVault                              ·  reallocate                          ·     246070  ·     248298  ·     247184  ·            2  ·          -  │
```

Note in this case there is no gas saving in using the `assetCount` variable as `_assets` is already in memory.
```Solidity
uint256 assetCount = _assets.length;
IStrategy strategyTo = IStrategy(_strategyToAddress);
for (uint256 i = 0; i < assetCount; ++i) {
}
```

The following uses the same gas and one less variable in the stack.
```Solidity
IStrategy strategyTo = IStrategy(_strategyToAddress);
for (uint256 i = 0; i < _assets.length; ++i) {
}
```

But if `_assets` was a storage variable, it would definitely be worth using the `assetCount` variable as each loop will do an SLOAD of the length.

As there is no gas different, I've gone for style consistency in signing the array length to a variable.

contracts/contracts/vault/VaultCore.sol
naddison36: @sparrowDom I think we need different slippages for different assets. WETH swaps should be very close to the Oracle price while stETH and rETH may not.

I agree the name is misleading. How about `allowedOracleSlippageBps`? This makes it clearer that it it slippage from the Oracle price and not slippage on an AMM pool from the swap size.

contracts/contracts/oracle/OracleRouter.sol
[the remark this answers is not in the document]
naddison36: I've removed `OracleRouterDev`

brownie/addresses.py
```diff
@@ -32,6 +32,14 @@
 VAULT_CORE_IMPL = '0xf00d4b19458c594d4ea9d0b9861edfd2c444fa9a'
 VAULT_VALUE_CHECKER = '0xEEcD72c99749A1FC977704AB900a05e8300F4318'
 OUSD_METASTRAT = '0x89eb88fedc50fc77ae8a18aad1ca0ac27f777a90'
+SWAPPER_1INCH = '0xcD0fcF8a31Bc78ec07752e9CCD3960E936D18366'
+ROUTER_1INCH_V5 = '0x1111111254EEB25477B68fb85Ed929f73A960582'
+
+# OETH Contracts 
+OETH = '0x856c4Efb76C1D1AE02e20CEB03A2A6a08b0b8dC3'
```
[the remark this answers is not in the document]
naddison36: yes, they are problem from a messed up merge. I'll fix

contracts/contracts/vault/VaultCore.sol
sparrowDom: `allowedOracleSlippage` is much better. Maybe also `allowedAssetSlippage` isn't bad since it similarly communicates that it pertains to 1 asset rather than a pool (pair of assets)

contracts/contracts/vault/VaultAdmin.sol
DanielVF: FYI: I've confirmed with test contracts, that when compiled with 'via-ir' optimizations, these two contracts result in identical gas usage AND identical max stack depth during evm execution.

```soliditiy
contract Variable {
	function foo(address barAddress) external {
		IBar bar = IBar(barAddress);
		bar.bazz();
	}
}

contract Cast {
	function foo(address barAddress) external {
		IBar(barAddress).bazz();
	}
}
```

contracts/contracts/vault/VaultAdmin.sol
```diff
@@ -250,6 +280,77 @@ contract VaultAdmin is VaultStorage {
...
+     */
+    function swapCollateral(
+        address _fromAsset,
+        address _toAsset,
+        uint256 _fromAssetAmount,
+        uint256 _minToAssetAmount,
+        bytes calldata _data
+    )
+        external
+        whenNotCapitalPaused
```
DanielVF, line 300: I realized I did not communicate clearly in the design decisions on this one. We want this method to match the other _strategist_ allocation methods like `depositTo` and `withdrawFrom` which can operate even when the capital is paused.

We should remove `whenNotCapitalPaused` from this one.

contracts/contracts/vault/VaultCore.sol
naddison36: The field is on the `Asset` structure so I think it's clear it relates to an asset. eg `assets[i].allowedOracleSlippageBps`.

I'll go ahead with `allowedOracleSlippageBps`

contracts/contracts/vault/VaultAdmin.sol
naddison36: Yes, I think that makes sense. The strategist can swap out a depegging asset. I'll do the change

contracts/contracts/vault/VaultAdmin.sol
naddison36: Did your `IBar` implementation do anything?

I put this in remix using Solidity 0.8.7 with the optimizer enabled
```Solidity
// SPDX-License-Identifier: MIT
pragma solidity 0.8.7;

interface IBar {
    function bazz() external;
}


contract Bar is IBar {
    uint256 counter = 1;
    function bazz() external override {
        counter += 1;
    }
}

contract Variable {
	function foo(address barAddress) external {
		IBar bar = IBar(barAddress);
		bar.bazz();
	}
}

contract Cast {
	function foo(address barAddress) external {
		IBar(barAddress).bazz();
	}
}
```

Cast execution cost 8,315
Variable execution cost 8,328
So a 13 gas saving.

naddison36: Also, I can get one more variable using cast rather than the variable method before hitting stack too deep

```Solidity
contract Cast {
    function foo(
        address addr1,
        address addr2,
        address addr3,
        address addr4,
        address addr5,
        address addr6,
        address addr7,
        address addr8,
        address addr9,
        address addr10,
        uint256 num1,
        uint256 num2,
        uint256 num3,
        uint256 num4,
        uint256 num5,
        uint256 num6
    ) external {
        IBar(addr1).bazz();
    }
}
```

```Solidity
contract Variable {
    function foo(
        address addr1,
        address addr2,
        address addr3,
        address addr4,
        address addr5,
        address addr6,
        address addr7,
        address addr8,
        address addr9,
        address addr10,
        uint256 num1,
        uint256 num2,
        uint256 num3,
        uint256 num4,
        uint256 num5
    ) external {
        IBar bar = IBar(addr1);
        bar.bazz();
    }
}
```

DanielVF: The lack of EVM stack size difference in the test case was when compiled with `via-ir` - without those optimizations turned on, it does use another stack slot.

bar() was empty.

I'm okay using the casts here.

contracts/contracts/vault/VaultAdmin.sol
```diff
@@ -2,34 +2,23 @@
...
 import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
 
-import { StableMath } from "../utils/StableMath.sol";
 import { IOracle } from "../interfaces/IOracle.sol";
+import { ISwapper } from "../interfaces/ISwapper.sol";
 import "./VaultStorage.sol";
 
 contract VaultAdmin is VaultStorage {
     using SafeERC20 for IERC20;
-    using StableMath for uint256;
```
DanielVF, line 18: Yay!

contracts/contracts/vault/VaultAdmin.sol
```diff
@@ -2,34 +2,23 @@
...
 
     /**
-     * @dev Verifies that the caller is the Vault, Governor, or Strategist.
+     * @dev Verifies that the caller is the Governor, or Strategist.
      */
-    modifier onlyVaultOrGovernorOrStrategist() {
```
DanielVF, line 23: Love seeing dead code go away.

contracts/contracts/vault/VaultAdmin.sol
```diff
@@ -504,7 +553,7 @@ contract VaultAdmin is VaultStorage {
         if (tokenAsset.decimals != 0) {
             return;
         }
-        uint256 decimals = IBasicToken(token).decimals();
+        uint8 decimals = IBasicToken(token).decimals();
```
DanielVF, line 556: Does cause a gas regression? Sometimes it's cheaper to work with uint256s since they match up to a full slot.

That said, this method is only rarely going to be called, so it's probably not worth gas optimizing.

naddison36: Yes, it is cheaper to do arithmetic and pass numbers around between functions using uint256.
But in this case we are getting ready to write the decimal value to storage. We get huge gas savings in other functions by packing the asset data into a single slot.

contracts/contracts/vault/VaultAdmin.sol
naddison36: I felt cleaner after doing it :-)

contracts/test/vault/oeth_vault.fork-test.js
```diff
@@ -1,46 +1,440 @@
...
+          toAsset.address,
+          fromAmount,
+          minToAssetAmount,
+          swapData
+        );
+
+      await expect(tx).to.be.revertedWith(error);
+    };
+
+    describe("Collateral swaps", async () => {
```
sparrowDom, line 206: awesome converage with this test 👍

sparrowDom (approved):
Awesome job. This is a massive PR that is also very clean and complete. 

Left a couple of minor comments inline

contracts/contracts/vault/VaultAdmin.sol
```diff
@@ -164,6 +163,32 @@ contract VaultAdmin is VaultStorage {
         emit NetOusdMintForStrategyThresholdChanged(_threshold);
     }
 
+    /**
+     * @dev Set the contract the performs swaps of collateral assets.
```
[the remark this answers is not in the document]
naddison36: good spot. I'll fix

contracts/contracts/vault/VaultAdmin.sol
```diff
@@ -253,6 +279,76 @@ contract VaultAdmin is VaultStorage {
...
+        // Check the to assets returned is above slippage amount specified by the strategist
+        require(
+            toAssetAmount >= _minToAssetAmount,
+            "Strategist slippage limit"
+        );
+
+        // Check the slippage against the Oracle in case the strategist made a mistake or has become malicious.
+        // to asset amount = from asset amount * from asset price / to asset price
+        uint256 minOracleToAssetAmount = (_fromAssetAmount *
+            (1e4 - fromAssetConfig.allowedOracleSlippageBps) *
```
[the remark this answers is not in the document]
naddison36: The transaction will still work if the swaps are at parity to the Oracle prices. And in some cases, the swaps will be better than the Oracle price. 0% allowed slippage means 1:1 price as (1e4 - 0) / 1e4 = 1.

contracts/contracts/vault/VaultAdmin.sol
```diff
@@ -253,6 +279,76 @@ contract VaultAdmin is VaultStorage {
...
+        uint256 _minToAssetAmount,
+        bytes calldata _data
+    )
+        external
+        nonReentrant
+        onlyGovernorOrStrategist
+        returns (uint256 toAssetAmount)
+    {
+        // Check fromAsset and toAsset are valid
+        Asset memory fromAssetConfig = assets[address(_fromAsset)];
```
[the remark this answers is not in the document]
naddison36: good spot

contracts/test/_fixture.js
```diff
@@ -597,6 +620,67 @@ function oethDefaultFixtureSetup() {
...
+    return fixture;
+  });
+}
+
+function oeth1InchSwapperFixtureSetup() {
+  return deployments.createFixture(async () => {
+    const fixture = await oethDefaultFixture();
+    const { mock1InchSwapRouter } = fixture;
+
+    const swapRouterAddr = "0x1111111254EEB25477B68fb85Ed929f73A960582";
```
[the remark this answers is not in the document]
naddison36: Its at the bottom `addresses.mainnet.oneInchRouterV5`

contracts/deploy/066_oeth_oeth_vault_swaps.js
```diff
@@ -0,0 +1,106 @@
...
+      [],
+      null,
+      true
+    );
+
+    // 2. Connect to the OETH Vault as its governor via the proxy
+    const cVaultProxy = await ethers.getContract("OETHVaultProxy");
+    const cVault = (
+      await ethers.getContractAt("OETHVault", cVaultProxy.address)
+    ).connect(addresses.mainnet.OldTimelock);
```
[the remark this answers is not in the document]
naddison36: nice. I'll change

contracts/test/vault/oeth-swapper.js
```diff
@@ -0,0 +1,415 @@
...
+
+    it("Should swap assets using 1inch router", async () => {
+      const { swapper1Inch, strategist, weth, frxETH, mock1InchSwapRouter } =
+        fixture;
+
+      const deadAddr = "0x1111111111222222222233333333334444444444";
+
+      const data = utils.defaultAbiCoder.encode(
+        ["bytes4", "address", "bytes"],
+        [utils.arrayify("0x12aa3caf"), deadAddr, utils.arrayify("0xdead")]
```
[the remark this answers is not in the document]
naddison36: good point. I have then in 1Inch.js so I will import them from there

contracts/test/vault/oeth-swapper.js
```diff
@@ -0,0 +1,415 @@
...
+      const toAmount = utils.parseEther("100");
+
+      await weth
+        .connect(strategist)
+        .mintTo(swapper1Inch.address, fromAmount.mul(2));
+      await frxETH
+        .connect(strategist)
+        .mintTo(mock1InchSwapRouter.address, toAmount.mul(2));
+
+      await swapper1Inch.approveAssets([weth.address]);
```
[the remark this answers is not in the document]
naddison36: good question. It's because `Mock1InchSwapRouter` was only transferring assets in `swap` and not the Uniswap functions `unoswapTo` and `uniswapV3SwapTo`.
I'll add the transfers to the Uniswap functions to be consistent.

contracts/test/_fixture.js
```diff
@@ -597,6 +620,67 @@ function oethDefaultFixtureSetup() {
...
+
+    const { weth, reth, stETH, frxETH, matt, strategist, oethVault } = fixture;
+    const zeroAddr = "0x0000000000000000000000000000000000000000";
+
+    for (const token of [weth, reth, stETH, frxETH]) {
+      const defaultStrategy = await oethVault.assetDefaultStrategies(
+        token.address
+      );
+
+      if (defaultStrategy != zeroAddr) {
```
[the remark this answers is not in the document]
naddison36: @shahthepro what do you think?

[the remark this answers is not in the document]
shahthepro: That’s a good suggestion as well. Will change it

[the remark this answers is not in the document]
shahthepro: Have pushed this change, works as expected

contracts/contracts/vault/VaultAdmin.sol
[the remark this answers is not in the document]
sparrowDom: ok makes sense

contracts/test/vault/oeth-swapper.js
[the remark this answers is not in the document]
sparrowDom: aha I see thanks

DanielVF (approved):
TLDR: one change I'd like to see, two very optional things for consideration.

#### Requirements

We to be able to swap from one collateral type to another, and do so with the operational flexibility to get a good price.

We want to be able to use a massive number of swap platforms via 1inch. 

These swap should be atomic, so that we can verify that they have completed successfully in the same swap function execution.

In addition we want a hard cap on the amount of damage a malicious strategist could do.

### Deployment Considerations

(See naddison's notes)

- [x] allowedOracleSlippageBps for each
- [x] swapConfig set

#### Internal State

✅ The swap contract has no internal state or config.

✅ The vault adds two configs, swapConfig and allowedOracleSlippageBps. Both of these are safe in their default state.

✅ What id no swapper is set? `ISwapper(config.swapper).swap` will revert if `config.swapper` is not a set to a contract, since that behavior is built into solidity.

#### Attack

The two primary enemies of a successful swap are bad orders and MEV sandwich attacks.

✅ The strategists prevent MEV attacks by entering a min resulting amount of the target coin. This is much more accurate than a fixed percentage.

✅ We protect against swap parameter errors by having a check against the asset prices on the swap assets, and a final check against total vault values vs total supply. This prevents looping a bad swap to drain things.

✅ 1inch swaps could be fairly advanced. I'm guessing there could be some kind of reentrancy opportunity in the middle of these, if the path was created maliciously. The vault (and the swap method) at least are reentrancy protected.

✅ This code makes approvals for potentially tens of millions of capital to the 1inch contracts. However the vault itself does not approve to 1inch, but rather directly transfers the correct funds to the swapper contract for the active swap. There shouldn't be anything but dust sitting around for approval theft. We also have the value of vault check that will atomically check the vault value after the swap.

#### Logic

🟧 It's clear that dust is going to accumulate on the swapper contract, and there's no way built in way to move it back to the vault. In theory the dust is going to be ultra small.

However, anyone can come along and take any funds sitting on the swapper contract, simply by calling `swap()` themselves, which will use the funds on the swapper contract to fund the requested swap action.

Even if we decide this is okay, we should leave comments here that we know about it because it's likely going to get reported on Immunefi as a theft of funds.

✅ The swap contract does not have reentrancy protection. In theory there could be multiple swaps happening simultaneously IF someone was able to get execution inside the swap on 1inch. This doesn't seem to be a problem because the `swap` method uses no state, calls only one external call, funds should be transferred out at the beginning of that call, and the swap contract is not responsible for those funds after the call is made. Plus our validation is at the vault level.


#### Tests

	  - [x] Close enough

#### Flavor

🟥 It's kind of sad that we have to split up the post swap validation from the swap code in solidity. That's reusing > 90% of the same variables.

We could combine them back into one method by using the old braces trick. We only need one more stack space, and the toAssetBalBefore is only used once, immediately after the swap. So we can do the braces trick to limit that variable's lifetime,  which will free up stack space for later.

```javacript
SwapConfig memory config = ...;
{
	uint256 toAssetBalBefore = ...;
	IERC20(_fromAsset).safeTransfer(...);
	ISwapper(config.swapper).swap(...);
	toAssetAmount = ...;
}
```

I've tested this locally and it seems to work

💚 I think it's good building the mocks for the swapper contract, so that we can have unit tests for it. It's nice to have unit tests when our own contracts are talking to each other, as much as fork tests are good when we are talking to the outside world.

🟧 I don't think there's any reason for breaking up the swapper contract deploy into a separate file, since the swapper contract is going to be deployed at the same time as the vault swaps, and is a requirement for that deploy file. Won't hurt anything though, just extra bookkeeping.

#### Overflow

#### Proxy
_No new proxies_

#### Black magic

#### Dependencies
_No OZ changes_

#### Deploy

#### Authentication

- [x] Never use tx.origin

#### Cryptographic code

_No Crypto_

#### Gas problems

- [x] For loops correct

#### External calls

	- [x] Validated Enough
- [x] Unsafe external calls
- [x] Oracles, one of:
  - [ ] No oracles
  - [ ] Oracles can't be bent


#### Ethereum

naddison36:
@DanielVF I've pushed the change to move the internal `_postSwapChecks` logic back into `swapCollateral`.

The swapper contract is in a separate deploy script as we have already deployed it to mainnet. [0xcD0fcF8a31Bc78ec07752e9CCD3960E936D18366](https://etherscan.io/address/0xcD0fcF8a31Bc78ec07752e9CCD3960E936D18366)
This was done to make the development of the off-chain swap component easier.

Shall we deploy a new swapper contract with a comment that we know dust assets can be left in the swapper contract?

PR #1571, opened 2023-06-01
fix dapp crash from undefined

HrikB:
The dapp is crashing becuase of an `undefined` error.

I needed to replace 
```
const allowance = parseFloat(
        allowances.coinNeedingApproval.allowanceCheckKey
      )
```
with
```
const allowance = parseFloat(
        get(allowances, 'coinNeedingApproval[allowanceCheckKey]')
      )
```

### DIFF
--- a/dapp-oeth/src/hooks/useCurrencySwapper.js
+++ b/dapp-oeth/src/hooks/useCurrencySwapper.js
@@ -16,7 +16,7 @@ import {
 import { useWeb3React } from '@web3-react/core'
 import addresses from 'constants/contractAddresses'
 import curveRoutes from 'constants/curveRoutes'
-
+import { get } from 'lodash'
 import { calculateSwapAmounts } from 'utils/math'
 
 const useCurrencySwapper = ({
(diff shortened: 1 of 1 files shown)


rafaelugolini (approved):

dapp-oeth/src/hooks/useCurrencySwapper.js
```diff
@@ -140,7 +140,7 @@ const useCurrencySwapper = ({
       }
 
       const allowance = parseFloat(
-        allowances[coinNeedingApproval][allowanceCheckKey]
+        get(allowances, 'coinNeedingApproval[allowanceCheckKey]')
```
sparrowDom, line 143: This shouldn't work:
Screenshot 2023-06-01 at 14 46 11

sparrowDom: I think the correct code should be: 
```
get(allowances, ${coinNeedingApproval}.${allowanceCheckKey}')
```

in case both things in the array are object and you need to go 2 levels deep to get to the property.

sparrowDom (commented):
Was this tested I think it shouldn't work

sparrowDom (commented):
Was this tested I think it shouldn't work

sparrowDom:
not sure why Github posted my comment twice... sorry :D

rafaelugolini:
> Was this tested I think it shouldn't work

it stopped crashing on the preview but maybe worth it for @HrikB to review it

PR #1608, opened 2023-06-07
add runlog for Aave OUSD allocation

sparrowDom:
add runlog for Aave OUSD allocation


sparrowDom:
Merging since the TX has already been executed by the strategist

PR #1611, opened 2023-06-07
Support WETH in FraxETH strategy

shahthepro:
Pending things:
- [x] Add Unit Tests
- [ ] ???
-------

Contract change checklist:
  - [x] Unit tests pass


contracts/contracts/strategies/FraxETHStrategy.sol
```diff
@@ -0,0 +1,115 @@
...
+import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import { IERC20 } from "../utils/InitializableAbstractStrategy.sol";
+import { IWETH9 } from "../interfaces/IWETH9.sol";
+import { IFraxETHMinter } from "../interfaces/IFraxETHMinter.sol";
+import { Generalized4626Strategy } from "./Generalized4626Strategy.sol";
+
+contract FraxETHStrategy is Generalized4626Strategy {
+    using SafeERC20 for IERC20;
+
+    IWETH9 public constant weth =
```
naddison36, line 19: I prefer these public constants to be of type `address` as explorers like Etherscan handle it better.
There are no gas savings in declaring the constant to the interface type.
It just means `weth.withdraw(_amount);` becomes `IWETH9(weth).withdraw(_amount);`
On the flip side, `_asset == address(weth) || _asset == address(assetToken)` becomes `_asset == weth || _asset == address(assetToken)`

contracts/contracts/strategies/FraxETHStrategy.sol
```diff
@@ -0,0 +1,115 @@
...
+        IWETH9(0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2);
+
+    IFraxETHMinter public constant fraxETHMinter =
+        IFraxETHMinter(0xbAFA44EFE7901E04E39Dad13167D089C559c1138);
+
+    function _deposit(address _asset, uint256 _amount) internal override {
+        require(_amount > 0, "Must deposit something");
+        require(
```
naddison36, line 27: I think you can add an `else if (_asset == address(assetToken))` below and then an else to revert if not weth or frxETH.

contracts/contracts/strategies/FraxETHStrategy.sol
```diff
@@ -0,0 +1,115 @@
...
+            _asset == address(weth) || _asset == address(assetToken),
+            "Unexpected asset address"
+        );
+
+        if (_asset == address(weth)) {
+            // Unwrap WETH
+            weth.withdraw(_amount);
+
+            // Deposit ETH for frxETH
```
naddison36, line 36: The comment needs to be clear about what's happening. eg Deposit ETH for frxETH and then stake for sfrETH

contracts/contracts/strategies/FraxETHStrategy.sol
```diff
@@ -0,0 +1,115 @@
...
+        // If it's not WETH, it has to be frxETH
+        require(_asset == address(assetToken), "Unexpected asset address");
+
+        /* We are intentionally not counting the amount of assetToken parked on the
+         * contract toward the checkBalance. The deposit and withdraw functions
+         * should not result in assetToken being unused and owned by this strategy
+         * contract.
+         */
+        return
+            IERC4626(platformAddress).convertToAssets(
```
naddison36, line 89: `convertToAssets` does not take into account fees or slippage. This probably does apply for sfrxETH, but for other vaults it does. I would recommend using `maxWithdraw` instead.

From the 4626 spec "The methods totalAssets, convertToShares and convertToAssets are estimates useful for display purposes, and do not have to confer the exact amount of underlying assets their context suggests."

contracts/contracts/strategies/FraxETHStrategy.sol
```diff
@@ -0,0 +1,115 @@
...
+     * @dev Deposit the entire balance of assetToken to gain shareToken
+     */
+    function depositAll() external virtual override onlyVault nonReentrant {
+        uint256 balance = assetToken.balanceOf(address(this));
+        if (balance > 0) {
+            _deposit(address(assetToken), balance);
+        }
+
+        // ETH + WETH balance
+        uint256 ethBalance = address(this).balance +
```
naddison36, line 104: will this work? If there was ETH in the vault for some reason, wouldn't the `_deposit` call fail as it will try and withdraw too much WETH?

contracts/contracts/strategies/FraxETHStrategy.sol
```diff
@@ -0,0 +1,115 @@
...
+        return
+            IERC4626(platformAddress).convertToAssets(
+                shareToken.balanceOf(address(this))
+            );
+    }
+
+    /**
+     * @dev Deposit the entire balance of assetToken to gain shareToken
+     */
+    function depositAll() external virtual override onlyVault nonReentrant {
```
naddison36, line 97: There is a product decision here as to whether all WETH and frxETH in the vault should be deposited into the strategy. I've raised a design decision for this.

contracts/contracts/strategies/Generalized4626Strategy.sol
```diff
@@ -112,6 +119,7 @@ contract Generalized4626Strategy is InitializableAbstractStrategy {
     function checkBalance(address _asset)
         external
         view
+        virtual
```
naddison36, line 122: I can't comment on line 133 so I'll comment here. `IERC4626(platformAddress).convertToAssets` needs to be replaced with `IERC4626(platformAddress).maxWithdraw`.
`convertToAssets` does not include fees or slippage. It's just an estimate for display purposes.

naddison36 (changes requested):
Looks good. I have a few small changes.
There's also a design decision about what `depositAll` should do.

contracts/contracts/strategies/FraxETHStrategy.sol
shahthepro: Nvm, I guess you're asking about the inverse, yeah, that would fail, will fix it

contracts/contracts/strategies/FraxETHStrategy.sol
sparrowDom: I agree with @naddison36 it would be more accurate for the `checkBalance` to report what is the actual amount extractable from the strategy. 

I've ran both the holdings of our strategy using `convertToAssets` resulting in `2013.0097359133865` and `maxWithdraw`  resulting in `2013.0097359133865`. The difference between the two values is `-0.00010626067205521394`. Interestingly the `maxWithdraw` gives a slightly higher number. 

From the [documentation](https://ethereum.org/en/developers/docs/standards/tokens/erc-4626/#maxwithdraw) it does look like as Nick states that using `maxwithdraw` would be a more accurate approach.

contracts/contracts/strategies/FraxETHStrategy.sol
sparrowDom: I would keep the behavior that our other strategy contracts have. Only the tokens present on the strategy contract (the ones Vault sends over) are being deployed not considering the ones on the Vault.

contracts/contracts/strategies/FraxETHStrategy.sol
```diff
@@ -0,0 +1,118 @@
...
+        uint256 balance = assetToken.balanceOf(address(this));
+        if (balance > 0) {
+            _deposit(address(assetToken), balance);
+        }
+
+        // Wrap any ETH balance to WETH
+        // Wrapping here and unwrapping in `_deposit` isn't gas efficient,
+        // But `depositAll` isn't called frequently; only during reallocation.
+        uint256 ethBalance = address(this).balance;
```
sparrowDom, line 103: I would leave the ethBalance out of it. If strategy works as intended there shouldn't be any ETH lingering around. And there is extra gas cost to doing this check.

contracts/contracts/strategies/FraxETHStrategy.sol
```diff
@@ -0,0 +1,118 @@
...
+        uint256 wethBalance = IWETH9(weth).balanceOf(address(this));
+        if (wethBalance > 0) {
+            _deposit(weth, wethBalance);
+        }
+    }
+
+    /**
+     * @dev Accept ETH
+     */
+    receive() external payable {}
```
sparrowDom, line 117: In case we decide our strategy doesn't want to handle eth we would remove this bit as well.

contracts/contracts/strategies/Generalized4626Strategy.sol
```diff
@@ -35,7 +35,7 @@ contract Generalized4626Strategy is InitializableAbstractStrategy {
      * @param _asset Address of asset to deposit
      * @param _amount Amount of asset to deposit
      */
-    function _deposit(address _asset, uint256 _amount) internal {
+    function _deposit(address _asset, uint256 _amount) internal virtual {
```
sparrowDom, line 38: Since this contract is extended for the first time, lets add some storage slot padding to it. Just in case we ever need to alter it in the future.

contracts/contracts/strategies/FraxETHStrategy.sol
naddison36: Ah, I see. It only deposits what's in the strategy contract, not the vault. My bad.

contracts/contracts/interfaces/IFraxETHMinter.sol
```diff
@@ -0,0 +1,11 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+interface IFraxETHMinter {
+    function submitAndDeposit(address recipient)
+        external
+        payable
+        returns (uint256 shares);
+
+    function submit() external payable;
```
DanielVF, line 10: I don't think this `submit` is being used, and can be removed?

shahthepro: That's true, will remove it

contracts/contracts/strategies/Generalized4626Strategy.sol
DanielVF: For @naddison36, @shahthepro, @sparrowDom, and all who have wondered about the mystery why we seem to be getting slightly different numbers back from `maxWithdraw` and `convertToAssets`, even though the underlying code path appears the same, I think I've solved the issue.

Basically, `maxWithdraw` and `convertToAssets` do actually return the same thing - differences seen are because the number changes block by block, and manually finding and clicking things on etherscan will tend to display numbers pulled from different blocks.

Here's the result of a fork simulation in forge:

```python
  Reported: maxWithdraw     2014259136521724599573
  Reported: convertToAssets 2014259136521724599573
  Actual:   redeemed        2014259136521724599573
  ```

Test code:

```solidity
contract WithdrawTest is Test {
    
    IsfrxETH sfrxETH = IsfrxETH(0xac3E018457B222d93114458476f3E3416Abbe38F);
    address strat = 0x3fF8654D633D4Ea0faE24c52Aec73B4A20D0d0e5;

    function testReportedAmounts() external {
        uint256 shareBalance = sfrxETH.balanceOf(strat);
        console2.log("Share balance", shareBalance);
        console.log("--------");
        console2.log("Reported: maxWithdraw    ", sfrxETH.maxWithdraw(strat));
        console2.log("Reported: convertToAssets", sfrxETH.convertToAssets(shareBalance));
        vm.prank(strat);
        uint256 x = sfrxETH.redeem(shareBalance, strat, strat);
        console2.log("Actual:   redeemed       ", x);
    }
}
```

contracts/contracts/strategies/Generalized4626Strategy.sol
DanielVF: Agree, this should be done.

contracts/contracts/strategies/Generalized4626Strategy.sol
naddison36: For the sfrxETH vault, `maxWithdraw` returns the same as `convertToAssets`. But `Generalized4626Strategy` should work with all ERC-4626 vaults which means it should work with vaults where `convertToAssets` does not match `maxWithdraw`. That is, it should work with vaults that either have slippage or fees.

DanielVF: I'm fine with maxWithdraw on this one.

naddison36 (approved):
Looks good. Well done @shahthepro

contracts/contracts/strategies/FraxETHStrategy.sol
sparrowDom: got it thanks 👍

sparrowDom (approved):
great work @shahthepro

DanielVF (approved):

PR #1618, opened 2023-06-09
OETH Unit Tests Setup

shahthepro:

Contract change checklist:
  - [ ] Unit tests pass


contracts/test/_fixture.js
```diff
@@ -23,10 +23,17 @@ const threepoolSwapAbi = require("./abi/threepoolSwap.json");
...
-  });
+const defaultFixture = deployments.createFixture(async () => {
+  await deployments.fixture(
+    isFork
+      ? undefined
+      : process.env.FORKED_LOCAL_TEST
+      ? ["none"]
+      : ["unit_tests"],
+    {
+      keepExistingDeployments: true,
```
sparrowDom, line 34: that is the thing that does a `snapshot` and `revert` after the fixtures are setup and before each test is ran right?

sparrowDom (commented):
Sorry for being late to the party. Left one minor comment inline

contracts/contracts/oracle/OracleRouter.sol
```diff
@@ -297,3 +297,27 @@ contract OracleRouterDev is OracleRouterBase {
         maxStaleness = fm.maxStaleness;
     }
 }
+
+// @notice Oracle Router that denominates all prices in ETH
+contract OETHOracleRouterDev is OETHOracleRouter {
```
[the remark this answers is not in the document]
shahthepro: Yep, that should work. Will clean it up

contracts/test/_fixture.js
shahthepro: This line tells `hardhat-deploy` to not run deployments again. So, if a deployment hasn't been run before, it'll still run.

Using `deployments.createFixture` with this key always set to true makes test setup/teardown a bit faster. Previously, we only did that for fork tests (though with still incorrect way of using fixtures). I recently changed it to do it for unit tests as well.

The change a few lines above in particular also adds a `unit_tests` tag to some deployments file. Only those are run when setting up or doing a teardown for unit tests, so basically we avoid any of the deployments files that aren't tagged as `unit_tests` (which is almost all deployments other than the first few ones)

PR #1627, opened 2023-06-14
Vault refactoring

naddison36:
Vault changes for next Vault deployment that are not collateral swaps or gas improvements.

This includes:
- https://github.com/OriginProtocol/origin-dollar/issues/1442
- https://github.com/OriginProtocol/origin-dollar/issues/1619
- Renaming OSUD to OToken in the Vault Natspec which covers some of https://github.com/OriginProtocol/origin-dollar/issues/1533
- Upgrade GitHub Actions to remove warnings
- Adding `getAssetConfig` to `VaultCore`

Issue #1533, opened 2023-05-22
Vault generalisation 


sparrowDom:
As part of the OpenZeppelin audit report the OZ team has suggested we could do a better job with Vault generalisation:
<img width="661" alt="Screenshot 2023-05-22 at 15 18 13" src="https://github.com/OriginProtocol/origin-dollar/assets/579910/02a7acb6-bc8c-4f47-9325-b4dbcfc66365">

I have started a draft PR https://github.com/OriginProtocol/origin-dollar/pull/1519 and came pretty quickly to a realisation that this is not going to be a small task and we also need to be quite cognisant of which contract's external functions get renames and what possible contract / dapp / analytics re-deploys that will trigger. 

One thing is sure, that this is not a high priority task and IMHO we should get to it when we have more breathing room.

sparrowDom:
Low priority would love your thoughts on this one @DanielVF @naddison36 @rafaelugolini @micahalcorn

naddison36:
I'm not a fan of changing the contract ABIs once they are deployed. I can do the Vault Natspec changes as part of collateral swaps as that will redeploy the vault contracts.

PR #1629, opened 2023-06-15
Renamed Vault functions and events with Ousd in them

naddison36:
This is an optional addition to the PR https://github.com/OriginProtocol/origin-dollar/pull/1627 that refactors the vaults

It does more more of the items in https://github.com/OriginProtocol/origin-dollar/issues/1533
Specifically, it covers removing `Ousd` from function names and events.
- `setOusdMetaStrategy` -> `setMetaStrategy`
- `setNetOusdMintForStrategyThreshold` -> `setNetMintForStrategyThreshold`
- `OusdMetaStrategyUpdated` -> `MetaStrategyUpdated`
- `NetOusdMintForStrategyThresholdChanged` -> `NetMintForStrategyThresholdChanged`

The two functions and corresponding events are only used by the governor so breaking the ABI is not such a big deal


contracts/contracts/interfaces/IVault.sol
```diff
@@ -165,9 +165,9 @@ interface IVault {
 
     function netOusdMintForStrategyThreshold() external view returns (uint256);
 
-    function setOusdMetaStrategy(address _ousdMetaStrategy) external;
+    function setMetaStrategy(address _strategy) external;
```
shahthepro, line 168: We don't plan to use this for OETH though right? This is exclusive to OUSDMetStrategy contract. the curve AMO uses mintForStrategy I believe

naddison36: It currently is on the OETHVaultAdmin hence the reason for are trying to clean it up.

There may be a better way of doing this. eg
1. Move these OUSD specific functions to an OUSD Vault implementation contract
2. Change the OUSD strategies to use the more generic functions.

Both sound like a bigger change for another day.

sparrowDom:
@naddison36 do we want to pair this PR with deployment of some other Vault functionality? So we don't kick off the Vault deploys just for the naming changes?

sparrowDom:
Maybe it'b be cool to have a list of PRs that are waiting to join some other more impactful functionality to be deployed

naddison36:
For the OETH Vault, I'd say it's best to pair with something else.
For the OUSD Vault, we need to decide if we include with `052_upgrade_ousd_vault_harvester` or do after. I'm not set on either way but I'd lean toward doing after so both OETH and OUSD are on the same codebase.

sparrowDom:
Yes agreed, rolling them out for both protocols simultaneously would be nice - in respect of not complicating our environments and mental load.

naddison36:
I think OUSD has been baked in now

PR #1633, opened 2023-06-16
Collateral swaps total value check

naddison36:
Added post collateral swap check to ensure the vault total has not fallen below the OToken total supply by an allowed percentage.

Contract changes:
- Added new `setSwapAllowedUndervalue` admin function, `allowedSwapUndervalue` view function and `SwapAllowedUndervalueChanged` event
- Moved the `swapCollateral` function from `VaultAdmin` to `VaultCore` as it now needs to read `_totalValue` which is on `VaultCore`.
- Moved the `onlyGovernorOrStrategist` modified up to `VaultStorage` as its now used in both `VaultCore` and `VaultAdmin`
- Removed duplicate `getAssetConfig` function from an incorrect merge
- Moved the `swapper` storage variable into a new `swapConfig` storage variable of type `SwapConfig`. `SwapConfig` includes the new `allowedUndervalueBps` storage data which shares the same slot at the `swapper` address.
- Added new internal `_postSwapChecks` function to workaround stack too deep errors on `swapCollateral`


naddison36:
@DanielVF `nonReentrantView` is no longer used since it has been removed from `VaultCore.totalValue()`. Should I remove the modifier from `Governable`? Or just leave it for possible future use?

At the moment we don't have any code coverage of `nonReentrantView`

contracts/contracts/vault/VaultStorage.sol
```diff
@@ -166,12 +175,4 @@ contract VaultStorage is Initializable, Governable {
             sstore(position, newImpl)
         }
     }
-
-    /**
-     * @dev Verifies that the deposits are not paused.
-     */
-    modifier whenNotCapitalPaused() {
```
DanielVF, line 173: Agree with keeping this at the lowest possible rung of the inheritance tree.

contracts/contracts/vault/VaultAdmin.sol
```diff
@@ -163,17 +164,156 @@ contract VaultAdmin is VaultStorage {
...
+    function swapper() external view returns (address swapper_) {
+        swapper_ = swapConfig.swapper;
+    }
+
+    /**
+     * @notice Set max allowed percentage the vault total value can drop below the OToken total supply in basis points
+     * when executing collateral swaps.
+     * @param _basis Percentage in basis points. eg 100 == 1%
+     */
+    function setSwapAllowedUndervalue(uint16 _basis) external onlyGovernor {
```
DanielVF, line 300: Let's say we have a built in dripper, and have a reserve. This means that the resting vault value would be greater than the total supply by perhaps a few percentage points.

Because this config only works based off undervalue, we would be unable to specify that we want the vault value to remain greater than the total supply. 

We could change this to just be a direct multiplier onto the total supply, for example 10,000 = we need to keep the vault value at at least 100% of total supply after the swap. 10,100 would be 101%

Or we could just keep this as is. I'm okay either way.

contracts/deploy/001_core.js
```diff
@@ -1215,6 +1215,7 @@ const deployOETHSwapper = async () => {
     ]);
 
   await vault.connect(sGovernor).setSwapper(mockSwapper.address);
+  await vault.connect(sGovernor).setSwapAllowedUndervalue(100);
```
DanielVF, line 1218: I like easy math for testing. :)

contracts/contracts/vault/VaultAdmin.sol
```diff
@@ -163,17 +164,156 @@ contract VaultAdmin is VaultStorage {
...
+        swapConfig.allowedUndervalueBps = _basis;
```
DanielVF, line 301: We should have a require that keeps this from going over 100% / 10,000, just for the sanity of the rest of the code.

contracts/contracts/vault/VaultStorage.sol
```diff
@@ -148,8 +149,16 @@ contract VaultStorage is Initializable, Governable {
...
-    /// @notice Contract that swaps the vault's collateral assets
-    address public swapper = address(0);
+    /// @notice Collateral swap configuration.
+    /// @dev is packed into a single storage slot to save gas.
+    struct SwapConfig {
+        // Contract that swaps the vault's collateral assets
+        address swapper;
+        // Max allowed percentage the total value can drop below the total supply in basis points.
+        // For example 100 == 1%
+        uint16 allowedUndervalueBps;
```
DanielVF, line 159: This is going to default to zero percent loss before it gets set, which is a good, safe number.

contracts/test/vault/oeth-swapper.js
```diff
@@ -82,27 +84,44 @@ describe("OETH Vault - Swapper", () => {
...
+        .to.emit(oethVault, "SwapAllowedUndervalueChanged")
+        .withArgs(10);
+
+      expect(await oethVault.allowedSwapUndervalue()).to.equal(10);
+    });
+
+    it("Should not allow anyone else to set allowed swap undervalue", async () => {
+      const { strategist, josh, oethVault } = fixture;
+
+      for (const user of [strategist, josh]) {
```
DanielVF, line 102: I like this style of testing it with multiple user, and including negative  testing the strategist.

contracts/contracts/vault/VaultAdmin.sol
naddison36: good point. I'll add

contracts/contracts/vault/VaultAdmin.sol
naddison36: That's an interesting one. Wouldn't the vault's total value exclude assets that are still to be dripped into the vault? Sure, the assets are physically held in the vault, but I would have expected the total value function to be aware of unrealised dripped assets. So over time the total value slowly increases and any rebases will pick up the newly realised assets.

Regardless, you make a good point on the undervalue being based off the vault value. The term undervalue is referencing the total supply so I'll change the check to 
```Solidity
IVault(address(this)).totalValue() >=
    (oUSD.totalSupply() * ((1e4 - config.allowedUndervalueBps))) /
        1e4
```

I generally prefer multipliers over percentage/basis points as it uses slightly less gas. But I went for basis points here to be consistent with other vault config. eg `redeemFeeBps` and `trusteeFeeBps`

contracts/test/vault/oeth-swapper.js
naddison36: credit to @shahthepro for this

PR #1637, opened 2023-06-19
Vault Natspec

naddison36:
Updated the Vault Natspec 

- Use `@notice` for public and external functions instead of `@dev` as most Natspec parsers ignore `@dev`. eg Etherscan
- Changed `VaultStorage` variables comments to Natspec 
- There are no Solidity changes - only Natspec.

### DIFF
--- a/contracts/contracts/governance/Governable.sol
+++ b/contracts/contracts/governance/Governable.sol
@@ -45,7 +45,7 @@ contract Governable {
     }
 
     /**
-     * @dev Returns the address of the current Governor.
+     * @notice Returns the address of the current Governor.
      */
     function governor() public view returns (address) {
         return _governor();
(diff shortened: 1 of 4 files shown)


shahthepro (approved):
LGTM.

While we are this, we could also use inherited comments at places (especially strategies). I remember doing that in one of the PRs (probably Uniswap), I didn't do it for all contracts though

naddison36:
> LGTM.
> 
> While we are this, we could also use inherited comments at places (especially strategies). I remember doing that in one of the PRs (probably Uniswap), I didn't do it for all contracts though

Sounds good for when we are redeploying strategies. I just did the Vaults in this change as we are about to redeploy them for OETH

sparrowDom (approved):
LGTM

PR #1646, opened 2023-06-21
Add new fuzzing tests for the OUSD/OETH token.

DanielVF:
Add new fuzzing tests for the OUSD/OETH token.

Developed by Rappie.


contracts/contracts/echidna/EchidnaConfig.sol
```diff
@@ -0,0 +1,54 @@
...
+        _;
+    }
+
+    function getAccount(uint8 accountId)
+        internal
+        view
+        returns (address account)
+    {
+        accountId = accountId / 64;
+        if (accountId == 0) return account = ADDRESS_USER0;
```
sparrowDom, line 48: nit: config seems to contain 2 user addresses and 2 contract addresses. Would be great to explain the purpose of this setup in the Readme

contracts/contracts/echidna/EchidnaHelper.sol
```diff
@@ -0,0 +1,94 @@
...
+        ousd.rebaseOptIn();
+    }
+
+    function optOut(uint8 targetAcc) public {
+        address target = getAccount(targetAcc);
+        hevm.prank(target);
+        ousd.rebaseOptOut();
+    }
+
+    function getTotalBalance() public view returns (uint256 total) {
```
sparrowDom, line 61: nit: this function and the one below seem like they could be in a for loop sharing the same fixed size array

contracts/contracts/echidna/EchidnaTestMintBurn.sol
```diff
@@ -0,0 +1,130 @@
...
+
+        uint256 balanceBefore = ousd.balanceOf(target);
+        uint256 amountMinted = mint(targetAcc, amount);
+        uint256 balanceAfter = ousd.balanceOf(target);
+
+        Debugger.log("amountMinted", amountMinted);
+        Debugger.log("balanceBefore", balanceBefore);
+        Debugger.log("balanceAfter", balanceAfter);
+
+        assert(balanceAfter >= balanceBefore + amountMinted);
```
sparrowDom, line 56: should we test that it is within the expected rounding limits? So that it is not too high from expected amount?

contracts/contracts/echidna/EchidnaTestMintBurn.sol
```diff
@@ -0,0 +1,130 @@
...
+        address target = getAccount(targetAcc);
+
+        uint256 balanceBefore = ousd.balanceOf(target);
+        burn(targetAcc, amount);
+        uint256 balanceAfter = ousd.balanceOf(target);
+
+        Debugger.log("balanceBefore", balanceBefore);
+        Debugger.log("balanceAfter", balanceAfter);
+
+        assert(balanceAfter <= balanceBefore - amount);
```
sparrowDom, line 85: Would also test this one to be within the expected range

contracts/contracts/echidna/EchidnaTestMintBurn.sol
```diff
@@ -0,0 +1,130 @@
...
+
+        int256 delta = int256(balanceAfter) - int256(balanceBefore);
+
+        // delta == amount, if no error
+        // delta < amount,  if too little is minted
+        // delta > amount,  if too much is minted
+        int256 error = int256(amountMinted) - delta;
+
+        assert(error >= 0);
+        assert(error <= int256(MINT_ROUNDING_ERROR));
```
sparrowDom, line 104: It feels like 1e18 is being too great of an expected rounding error. Curious what is the reasoning behind picking 1e18?

contracts/contracts/echidna/EchidnaTestSupply.sol
```diff
@@ -0,0 +1,146 @@
...
+        hasKnownIssue
+        hasKnownIssueWithinLimits
+    {
+        uint256 totalSupply = ousd.totalSupply();
+        uint256 totalBalance = getTotalBalance();
+
+        Debugger.log("totalSupply", totalSupply);
+        Debugger.log("totalBalance", totalBalance);
+
+        assert(totalSupply >= totalBalance);
```
sparrowDom, line 60: would be cool to test the limit in the other direction as well. To see we are within some acceptable range

contracts/contracts/echidna/EchidnaTestTransfer.sol
```diff
@@ -0,0 +1,262 @@
...
+
+        uint256 toBalBefore = ousd.balanceOf(to);
+        transfer(fromAcc, toAcc, amount);
+        uint256 toBalAfter = ousd.balanceOf(to);
+
+        Debugger.log("totalSupply", ousd.totalSupply());
+        Debugger.log("toBalBefore", toBalBefore);
+        Debugger.log("toBalAfter", toBalAfter);
+
+        assert(toBalAfter >= toBalBefore + amount);
```
sparrowDom, line 39: should also be within limits in the other direction

sparrowDom (commented):
Great work! The fuzzing tests are awesome 💪 

I've left a similar kind of comment in multiple places (and have stopped commenting at some point to not be repetitive), I guess I am a bit confused why when certain balances change we only test the expected amounts in one direction and not check the other direction as well. Meaning that the amounts are within limits in both directions.

I might be missing something though... Are the fuzzing test runs picking amounts so great that it is hard to pick an expected range for the amounts we are asserting maybe?

PR #1647, opened 2023-06-22
fix auto check snapshot

sparrowDom:
fix auto check snapshot

### DIFF
--- a/brownie/allocations.py
+++ b/brownie/allocations.py
@@ -282,13 +282,14 @@ def auto_check_snapshot():
     print(snapshot)
     vault_change = world.vault_core.totalValue() - snapshot[0]
     supply_change = world.ousd.totalSupply() - snapshot[1]
+    profit = vault_change - supply_change
 
     return [
         world.vault_value_checker.checkDelta(
-            vault_change, # expectedProfit
+            profit, # expectedProfit
             500 * int(1e18), # profitVariance
-            supply_change, # expectedVaultChange
-            500 * int(1e18), # vaultChangeVariance
+            vault_change, # expectedVaultChange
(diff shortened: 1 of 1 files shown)


brownie/allocations.py
```diff
@@ -282,13 +282,14 @@ def auto_check_snapshot():
...
 
     return [
         world.vault_value_checker.checkDelta(
-            vault_change, # expectedProfit
+            profit, # expectedProfit
             500 * int(1e18), # profitVariance
-            supply_change, # expectedVaultChange
-            500 * int(1e18), # vaultChangeVariance
+            vault_change, # expectedVaultChange
+            1000 * int(1e18), # vaultChangeVariance
```
DanielVF, line 292: This is basically the maximum size of the trade in AMO pool for which we don't care about invalidating the transaction. It's probably fine at $250,000

DanielVF (approved):

brownie/allocations.py
sparrowDom: got it changing to that

sparrowDom: just to be clear I am setting it to $250 and not $250,000. Since between the time we create an allocation and the time it is executed on the mainnet the number of expected vault_change shouldn't differ for more than $250 right? The actual `vault_change` can be quite significant, but the variance of the expected value should be quite small

DanielVF: The vault variance is basically the net size of OETH/ETH curve swaps that we want to allow between creating the TX and it running. For example, if the pool was imbalanced against OETH, and someone swaps either way for $20K, then there would be a $20K difference in how much OETH we print, and thus a $20K change in vault value.

Issue #1632, opened 2023-06-15
WalletConnect v1.0 shutdown is happening in 2 weeks!!


sumpg41:
It appears we are using WalletConnect v1.0 - **WalletConnect v1.0 shutdown is happening at 2pm UTC on June 28** - It is essential that we migrate to WalletConnect v2.0 as soon as possible to prevent any disruption when the shutdown finally comes into effect.

Migration paths from v1 to v2 can be found here: https://docs.walletconnect.com/2.0/advanced/migration-from-v1.x/overview

micahalcorn:
> @micahalcorn should this be changed to a P0? There are less than 9 days till WC v1 shutdown

I wouldn't call it a **0** but we'll get it done, thanks.

PR #1651, opened 2023-06-22
WalletConnect v2.0 integration, remove WalletConnect v1.0 from UI

smitch88:
#1632 

**Description:**
* Add support for WalletConnect v2.0 which replaces WalletConnect v1.0 in the UI.
* Change out all `web3React/core` usages in favor of RainbowKit + Wagmi
* Adds Wagmi and RainbowKit providers
* Needed to upgrade NextJS 10 -> 11, to support build, needed to move old babel config (its kept because of `fbt` library), upgraded to webpack 5 build

**Changes:**

The change spans across both dapps, OUSD, and OETH.

* Add in Rainbow kit wallet integration + associated wallet connectors to match previous support
* Kept `Wagmi` on version v0.12 to avoid upgrading to `viem` from `ethers`
* Updates all references to web3react to the equivalent `wagmi` hooks

~* There are package management changes and required dependency upgrades for NextJS `10.3.0 -> 11.1.0` and `webpack 4` -> `webpack 5` to support what we need to include `walletconnect/ethereum-provider@2.8.0` and work within the NextJS build. This required moving to .babelrc, and some other changes. I had to also alias the v8 version of `@web3-react/core` -> `web3-react-v8`, so we could be backward compatible with the existing connectors from 2+ years ago.~

~* Removed "My Ether Wallet" support due to compatibility issues with the upgrades of other packages. Confirmed with product via @micahalcorn that we can revisit this at a later date.~

**Required:**
* Add into `.env`, `NEXT_PUBLIC_WALLETCONNECT_V2_PROJECT_ID`, with walletconnect from `clout.walletconnect.com` project id

**Screenshots**

## OETH
Screenshot 2023-07-03 131921


contracts/test/flipper/flipper.js
```diff
@@ -8,7 +8,7 @@ const {
   loadFixture,
   isFork,
 } = require("../helpers");
-const { parseUnits } = require("ethers/lib/utils");
```
rafaelugolini, line 11: line 106 is broken with this change, the linter got it

dapp-oeth/src/components/TransactionListener.js
```diff
@@ -272,7 +271,9 @@ const TransactionListener = ({ rpcProvider }) => {
         // TODO handle a 404 here. We need to retry. A refresh is required to
         // get this to retry at the moment.
         try {
-          safeData = await connector.sdk.txs.getBySafeTxHash(t.hash)
+          // TODO
+          console.log(connector)
```
dcodes05, line 275: this console.log should be removed, right? it is not needed for users, right?

dapp/src/components/TransactionListener.js
```diff
@@ -272,7 +271,9 @@ const TransactionListener = ({ rpcProvider }) => {
         // TODO handle a 404 here. We need to retry. A refresh is required to
         // get this to retry at the moment.
         try {
-          safeData = await connector.sdk.txs.getBySafeTxHash(t.hash)
+          // TODO
+          console.log(connector)
```
dcodes05, line 275: same thing, do you need console.log in prod?

dcodes05 (commented):
there are a couple of console.logs that should be removed

dapp/src/components/TransactionListener.js
smitch88: Correct, this actually isn't done yet. I need to adjust this equivalent from web3 react. Left a TODO, revisiting it, good call out though.

dapp-oeth/src/components/TransactionListener.js
smitch88: Same comment as the other. Need to find the equivalent of this call from web3-react -> wagmi. The library switch had a gap between them

dapp/src/components/TransactionListener.js
rafaelugolini: @smitch88 does it mean that it's not ready to be merged?

smitch88: This is just an issue with transaction display for gnosis specifically. I think we could merge and unblock most of things while fixing this as a follow up.

micahalcorn (approved):

micahalcorn:
We've tested enough on the OETH side to feel good about this and fix forward. OUSD's dapp will be tested in staging.

PR #451, opened 2020-12-21
OUSD Redeploy

tomlinton:
If you made a contract change, make sure to complete the checklist below before merging it in master.

Refer to our [documentation](https://github.com/OriginProtocol/security) for more details about contract security best practices.

Contract change checklist:
  - [x] Code reviewed by 2 reviewers
  - [x] Unit tests pass
  - [x] Slither tests pass with no warning
  - [ ] Echidna tests pass (not automated, run manually on local)

PR #1653, opened 2023-06-23
last minute collateral swap changes to the Vault

naddison36:
When reviewing the Uniswap V3 Strategy PR, it was clear we need to remove the `_deprecated` property from the `Strategy` struct that is used by the Vault's `strategies` mapping.

Looking at both the OETH and OUSD vault storage, there is no impact in removing this

OUSD `strategies` storage https://evm.storage/eth/17540105/0xe75d77b1865ae93c7eaa3040b038d7aa7bc02f70/strategies#map

OETH `strategies` storage 
https://evm.storage/eth/17540186/0x39254033945aa2e4809cc2977e7087bee48bd7ab/strategies#map

### DIFF
--- a/contracts/contracts/vault/VaultStorage.sol
+++ b/contracts/contracts/vault/VaultStorage.sol
@@ -77,7 +77,6 @@ contract VaultStorage is Initializable, Governable {
     // Strategies approved for use by the Vault
     struct Strategy {
         bool isSupported;
-        uint256 _deprecated; // Deprecated storage slot
     }
     /// @dev mapping of strategy contracts to their configiration
     mapping(address => Strategy) internal strategies;
--- a/contracts/contracts/vault/VaultAdmin.sol
+++ b/contracts/contracts/vault/VaultAdmin.sol
@@ -367,7 +367,7 @@ contract VaultAdmin is VaultStorage {
      */
     function approveStrategy(address _addr) external onlyGovernor {
         require(!strategies[_addr].isSupported, "Strategy already approved");
-        strategies[_addr] = Strategy({ isSupported: true, _deprecated: 0 });
+        strategies[_addr] = Strategy({ isSupported: true });
         allStrategies.push(_addr);
         emit StrategyApproved(_addr);
     }
--- a/contracts/contracts/vault/VaultCore.sol
+++ b/contracts/contracts/vault/VaultCore.sol
@@ -411,7 +411,7 @@ contract VaultCore is VaultStorage {
 
     /**
      * @dev Internal to calculate total value of all assets held in Vault.
-     * @return value Total value in ETH (1e18)
+     * @return value Total value in base currency. eg USD or ETH (1e18)
      */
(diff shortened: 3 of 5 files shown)


contracts/contracts/vault/VaultStorage.sol
```diff
@@ -77,7 +77,6 @@ contract VaultStorage is Initializable, Governable {
     // Strategies approved for use by the Vault
     struct Strategy {
         bool isSupported;
-        uint256 _deprecated; // Deprecated storage slot
```
shahthepro, line 80: This will affect any future changes to the struct (since any new prop added might already have values set for existing strategies). Uniswap strategy already does that. It's probably low impact but I'd prefer not to do it

naddison36: That's why we need to do this now. The Uniswap V3 Strategy will add another bool after this which means we have two bools using three storage slots.

shahthepro: I get your point. But how do we make sure that we don't accidentally use the old value in slot1 (set before deprecation) when we add more vars to this? Right now, having this named `_depreacted` acts as a reminder for us

naddison36: All the old `slot1` values are set to zero which is the default value.

Here's the OETH Vault
- https://evm.storage/eth/17540186/0x39254033945aa2e4809cc2977e7087bee48bd7ab/strategies.0x1827f9ea98e0bf96550b2fc20f7233277fcd7e63#map
- https://evm.storage/eth/17540186/0x39254033945aa2e4809cc2977e7087bee48bd7ab/strategies.0x3ff8654d633d4ea0fae24c52aec73b4a20d0d0e5#map
- https://evm.storage/eth/17540186/0x39254033945aa2e4809cc2977e7087bee48bd7ab/strategies.0xc1fc9e5ec3058921ea5025d703cbe31764756319#map

Here's the OUSD Vault `strategies` mapping which you can drill into each mapping to check the `slot1` values
https://evm.storage/eth/17540105/0xe75d77b1865ae93c7eaa3040b038d7aa7bc02f70/strategies#map

shahthepro: Oh, it makes more sense to remove it then

shahthepro (approved):

sparrowDom (approved):
LGTM. 

Pretty cool demonstration of how the evm storage tool helps answer the question of whether we have any `dirty` data in the deprecated slot.

DanielVF:
Depreciating this variable was done on [December 3, 2020](https://github.com/OriginProtocol/origin-dollar/pull/368). The deploy script file for the relaunch of OUSD was made on [Dec 20th, 2020](https://github.com/OriginProtocol/origin-dollar/pull/451), so it seems likely from the commit history that this depreciated variable has never been live on the current vault. The actual deploy was something like the last day of the year 2020.

PR #1668, opened 2023-06-29
fix unit test

sparrowDom:
fix unit test


shahthepro (approved):

sparrowDom:
Merging it in since only the 1inch test failed with API call rate exceeded //cc: @naddison36

PR #1669, opened 2023-06-29
erase hardhat deployments folder if present

sparrowDom:
If you've ran a node in forkTest environment locally it will copy mainnet deployments to hardhat deployments folder. Also containing the `.chainId` file with value `1`. This makes the command `yarn run deploy:(oeth)` error out with the following error: 
Screenshot 2023-06-29 at 11 40 42

This PR fixes the issue. Also worthy to note this affects only local environments and not "clean" environments like CI.

### DIFF
--- a/contracts/package.json
+++ b/contracts/package.json
@@ -4,9 +4,9 @@
   "description": "Origin Dollar Contracts",
   "main": "index.js",
   "scripts": {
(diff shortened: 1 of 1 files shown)


naddison36 (approved):
LGTM

PR #1670, opened 2023-06-29
Helper, Handle empty arg'd gov actions

DanielVF:
My last python governance action printer would fail when a governance action had no arguments. Fixed.


sparrowDom (approved):
nice lgtm

PR #1676, opened 2023-07-04
OGV to OETH governance deploy part 2

sparrowDom:
- add deployment artifacts from passing the OGV governance proposal
- add tooling so that deployer can be the one submitting the OGV governance proposal


naddison36 (approved):
LGTM

PR #1695, opened 2023-07-12
AMO peg keeping

naddison36:

Contract change checklist:
  - [ ] Unit tests pass

Notion resources

[Solution analysis](https://www.notion.so/originprotocol/Curve-AMO-improvements-9480678f11954fe0bb0aa492214d02f9)
[Design decisions](https://www.notion.so/originprotocol/3defa307080b41f883529c33a88735ad?v=6a32271dfd2940efbb08c4e893750e82&pvs=4)

Significant contract changes

- [x] Added 075 deploy script for upgrade of the `ConvexEthMetaStrategy` contract
- [x] Added three new functions for the strategist to manage the AMO Metapool balance `removeAndBurnOTokens`, `mintAndAddOTokens` and `removeOnlyAssets`


contracts/contracts/strategies/ConvexEthMetaStrategy.sol
```diff
@@ -263,16 +284,134 @@ contract ConvexEthMetaStrategy is InitializableAbstractStrategy {
...
-        weth.deposit{ value: address(this).balance }();
+        // Get the strategy contract's ether balance.
+        // This includes all that was removed from the Metapool and
+        // any ether that was sitting in the strategy contract before the removal.
+        uint256 ethBalance = address(this).balance;
+        // Convert all the strategy contract's ether to WETH and transfer to the vault.
+        weth.deposit{ value: ethBalance }();
         require(
-            weth.transfer(vaultAddress, weth.balanceOf(address(this))),
+            weth.transfer(vaultAddress, ethBalance),
```
DanielVF, line 294: This code changes no longer picks up any WETH if any was left on the contract. However that's okay since "depositAll" would put it into use.

DanielVF:
You are still working on the safety checks that the strategist is not abusing the pool?

What's the design consideration/tradeoffs for having the strategist talk directly to the strategy vs going through the vault?

contracts/contracts/strategies/ConvexEthMetaStrategy.sol
naddison36: It is correct this change no longer picks up any WETH in the strategy contract for `withdraw` and `withdrawAll`. In normal operation there shouldn't be any WETH in the AMO strategy contract. We can see in practice that the AMO strategy contract has always had a zero balance of WETH. https://dune.com/queries/2738518/4557122?account_address_t6c1ea=0x1827F9eA98E0bf96550b2FC20F7233277FcD7E63

If for some reason there was WETH in the AMO strategy contract, for example someone transferred it some WETH, the `depositAll` will pick it up as DVF has pointed out.

naddison36:
> You are still working on the safety checks that the strategist is not abusing the pool?
> 
> What's the design consideration/tradeoffs for having the strategist talk directly to the strategy vs going through the vault?

The thinking was to put the AMO specific functions on the strategy rather than introduce AMO functions on the Vault. Thinking about it some more, we can use the standard `deposit` and `withdraw` functions using OETH as the asset. See PR https://github.com/OriginProtocol/origin-dollar/pull/1703 for how this would look.

As for the safety checks, the thinking is to use the same checks as the existing `deposit` and `withdraw` which basically does a 1% slippage check. The strategist should wrap the `depositToStrategy` and `withdrawFromStrategy` calls around `VaultValueChecker`'s `takeSnapshot` and `checkDelta`. There is currently nothing stopping the strategist calling the `depositToStrategy` and `withdrawFromStrategy` without the `VaultValueChecker` checks. We can change that.

It's also worth noting that depositing or withdrawing just OETH will incur a larger fee that depositing equal amounts of ETH and OETH, which still incurs a fee. As a fee is charged, the vault will make a loss on deposits of both WETH and OETH.

Withdrawing WETH does not incur a fee as it withdraw the OETH and ETH proportionally. Withdrawing OETH will incur a fee as the pool balances will be changed.

It's worth noting that Curve's fee is relative to the pools current proportions. If depositing equal amounts of OETH and ETH and the pool is not exactly balanced, a fee will be charged. The 0.04% (4 bps) fee is charged on the difference between the old and new pool balances. 0.02% of the fee does back to the liquidity providers, the other 0.02% goes to veCRV stakers.

brownie/runlogs/2023_07_strategist.py
```diff
@@ -148,3 +148,110 @@ def main():
       print("Transaction ", idx)
       print("To: ", item.receiver)
       print("Data (Hex encoded): ", item.input, "\n")
+# -------------------------------
```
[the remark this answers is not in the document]
naddison36: I've merged master back and this has now dropped from the changes

naddison36:
The changes since Daniel's last review

- Added modifier that checks if the Metapool balance was improved or overshot
- Moved the calculation of the min amounts into the contract using the virtual price rather than the strategist setting
- Added fork tests for the new repeg functions
- Fixed safeApproveAllTokens which wasn't doing an approve 0 before the reapproval and was approving WETH to the Metapool
- Made the OETH AMO fixture configurable. eg tilt the Metapool in either direction by a configurable amount.
- Beefed out the fork tests of the existing deposit, depositAll, withdraw and withdrawAll functions

contracts/contracts/strategies/ConvexEthMetaStrategy.sol
```diff
@@ -257,19 +337,169 @@ contract ConvexEthMetaStrategy is InitializableAbstractStrategy {
...
+        IVault(vaultAddress).burnForStrategy(oethToBurn);
+
+        emit Withdrawal(address(oeth), address(lpToken), oethToBurn);
+    }
 
-        // Send all ETH and WETH on the contract, including extra
-        weth.deposit{ value: address(this).balance }();
+    /**
+     * @notice Mint OTokens and one-sided add to the Metapool.
+     * This is used when the Metapool has not enough OTokens and too many ETH.
```
sparrowDom, line 403: nit: I might be off here but would it sound better like this maybe: 
```
This is used when the Metapool doesn't have enough OTokens and too much ETH
```

contracts/contracts/strategies/ConvexEthMetaStrategy.sol
```diff
@@ -257,19 +337,169 @@ contract ConvexEthMetaStrategy is InitializableAbstractStrategy {
...
+     * The total supply of OTokens does not change.
+     * The asset value of the strategy reduces.
+     * The asset value of the vault should be close to the same.
+     * @param _lpTokens The amount of Metapool LP tokens to be burned for ETH.
+     * @dev Metapool LP tokens is used rather than WETH assets as Curve does not
+     * have a way to accurately calculate the amount of LP tokens for a required
+     * amount of ETH. Curve's `calc_token_amount` functioun does not include fees.
+     * A 3rd party libary can be used that takes into account the fees, but this
+     * is a gas intensive process. It's easier for the trusted strategist to
+     * caclulate the amount of Metapool LP tokens required off-chain.
```
sparrowDom, line 460: agree, though IMO the strategist should also pass the other side of the trade - min amount of eth expected

contracts/test/strategies/oeth-metapool.fork-test.js
```diff
@@ -1,309 +1,863 @@
...
+      const tx = convexEthMetaStrategy.connect(timelock).withdrawAll();
+      await expect(tx).to.emit(convexEthMetaStrategy, "Withdrawal");
+    });
+  });
 
-    await weth.connect(josh).approve(oethVault.address, unitAmount);
-    await oethVault.connect(josh).mint(weth.address, unitAmount, 0);
-    await oethVault.connect(josh).allocate();
+  describe("with a lot more OETH in the Metapool", () => {
+    const loadFixture = createFixtureLoader(convexOETHMetaVaultFixture, {
```
sparrowDom, line 434: I think these might fail if pool on the mainnet get heavily out of balance. Or at least they won't end up with the required tilt

sparrowDom (commented):
Great job, left a few comments inline

contracts/contracts/strategies/ConvexEthMetaStrategy.sol
```diff
@@ -21,75 +21,140 @@ contract ConvexEthMetaStrategy is InitializableAbstractStrategy {
...
+        // Get the asset and OToken balances in the Curve Metapool
+        uint256[2] memory balancesAfter = curvePool.get_balances();
+        // diff = ETH balance - OETH balance
+        int256 diffAfter = int256(balancesAfter[ethCoinIndex]) -
+            int256(balancesAfter[oethCoinIndex]);
+
+        if (diffBefore <= 0) {
+            // If the pool was originally imbalanced in favor of OETH, then
+            // we want to check that the pool is now more balanced
+            require(diffAfter <= 0, "OTokens overshot peg");
```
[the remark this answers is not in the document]
naddison36: `diffAfter = 0` is ok. The require will revert if `diffAfter > 0`. `diffAfter <= 0` are the required state to continue.

contracts/contracts/strategies/ConvexEthMetaStrategy.sol
```diff
@@ -132,6 +199,8 @@ contract ConvexEthMetaStrategy is InitializableAbstractStrategy {
          */
         IVault(vaultAddress).mintForStrategy(oethToAdd);
 
+        emit Deposit(address(oeth), address(lpToken), oethToAdd);
```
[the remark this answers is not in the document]
naddison36: Yes, I agree.

FYI, I had a design decision on this https://www.notion.so/originprotocol/Should-a-Deposit-event-for-OTokens-be-emitted-for-AMO-strategy-deposits-607484d2db8f4136b14099cb6caf1766

contracts/contracts/strategies/ConvexEthMetaStrategy.sol
```diff
@@ -161,7 +231,7 @@ contract ConvexEthMetaStrategy is InitializableAbstractStrategy {
     }
 
     /**
-     * @dev Deposit the entire balance of any supported asset into the Curve 3pool
+     * @notice Deposit the strategy's entire balance of any ETH or WETH into the Curve Metapool
```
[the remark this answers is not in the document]
naddison36: Good spot. I'll update the comment.

contracts/contracts/strategies/ConvexEthMetaStrategy.sol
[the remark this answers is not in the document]
naddison36: I did spot an edge case not covered. If the pool was exactly balanced with `diffBefore = 0`, the Strategist could tip the pool to have more ETH.
It's a very unlikely scenario but best to have it covered off.

contracts/contracts/strategies/ConvexEthMetaStrategy.sol
naddison36: That's better. Thanks

contracts/test/_fixture.js
```diff
@@ -1590,7 +1680,35 @@ async function replaceContractAt(targetAddress, mockContract) {
...
+ *     depositToStrategy: false,
+ *   });
+ *   beforeEach(async () => {
+ *     fixture = await loadFixture();
+ *   });
+ * @param {*} fixture async function that sets up test data. eg users, contracts and protocols
+ * @param {*} config optional config object passed to the fixture function
+ * @returns loadFixture an async function that loads a fixture
+ */
+function createFixtureLoader(fixture, config) {
```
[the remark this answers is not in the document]
naddison36: The speed hasn't been that noticeable, but it has surfaced a number of false positive tests after switching to the new fixture. These have since been fixed.
And some tests would consistently fail rather than sporadically fail. The missing `await` in the redeem tests comes to mind.

contracts/tasks/amoStrategy.js
```diff
@@ -0,0 +1,395 @@
...
+      netMintedForStrategyDiff
+    )
+  );
+}
+
+/************************************
+          AMO peg keeping
+ ************************************/
+
+async function mintAndAddOTokensTask(taskArguments) {
```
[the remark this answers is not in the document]
naddison36: `removeOnlyAssets` and `removeAndBurnOTokens` have common code. `mintAndAddOTokens` is quite different.

* `removeAndBurnOTokens`
  * Withdraw Metapool LP tokens from Convex pool
  * Convert Metapool LP tokens to ETH value
  * Remove just the OTokens from the Metapool using `remove_liquidity_one_coin`
  * The vault burns the OTokens from this strategy

* `removeOnlyAssets`
  * Withdraw Metapool LP tokens from Convex pool
  * Convert Metapool LP tokens to ETH value
  * Remove just the ETH from the Metapool using `remove_liquidity_one_coin`
  * Convert ETH to WETH
  * Transfer to the vault
 
* `mintAndAddOTokens`
  * Convert OETH to Metapool LP tokens
  * Add the minted OTokens to the Curve Metapool using `add_liquidity`
  * Deposit the Metapool LP tokens to the Convex rewards pool

The gas impact of refactoring out the common code is only adding 50 gas. Given these functions would rarely be used that is not significant.

Gas before change
```
|  Contract               ·  Method                              ·  Min        ·  Max        ·  Avg        ·  # calls      ·  eur (avg)  │
··························|······································|·············|·············|·············|···············|··············
|  ConvexEthMetaStrategy  ·  removeAndBurnOTokens                ·    2399859  ·    2400793  ·    2400326  ·            6  ·          -  │
··························|······································|·············|·············|·············|···············|··············
|  ConvexEthMetaStrategy  ·  removeOnlyAssets                    ·    2040945  ·    2040981  ·    2040965  ·            9  ·          -  │
```

Gas after change
```
|  Contract               ·  Method                              ·  Min        ·  Max        ·  Avg        ·  # calls      ·  eur (avg)  │
··························|······································|·············|·············|·············|···············|··············
|  ConvexEthMetaStrategy  ·  removeAndBurnOTokens                ·    2399912  ·    2400846  ·    2400379  ·            6  ·          -  │
··························|······································|·············|·············|·············|···············|··············
|  ConvexEthMetaStrategy  ·  removeOnlyAssets                    ·    2040998  ·    2041034  ·    2041018  ·            9  ·          -  │
```

I think these are easier to read so I've done the refactoring

```Solidity
    function removeAndBurnOTokens(uint256 _lpTokens)
        external
        onlyStrategist
        improveMetapoolBalance
    {
        // Withdraw Metapool LP tokens from Convex and remove OTokens from the Metapool
        uint256 oethToBurn = _withdrawAndRemoveFromPool(
            _lpTokens,
            oethCoinIndex
        );

        // The vault burns the OTokens from this strategy
        IVault(vaultAddress).burnForStrategy(oethToBurn);

        emit Withdrawal(address(oeth), address(lpToken), oethToBurn);
    }

    function removeOnlyAssets(uint256 _lpTokens)
        external
        onlyStrategist
        improveMetapoolBalance
    {
        // Withdraw Metapool LP tokens from Convex and remove ETH from the Metapool
        uint256 ethAmount = _withdrawAndRemoveFromPool(_lpTokens, ethCoinIndex);

        // Convert ETH to WETH and transfer to the vault
        weth.deposit{ value: ethAmount }();
        require(
            weth.transfer(vaultAddress, ethAmount),
            "Transfer of WETH not successful"
        );

        emit Withdrawal(address(weth), address(lpToken), ethAmount);
    }
```

contracts/test/_fixture.js
```diff
@@ -1308,27 +1334,91 @@ async function convexOETHMetaVaultFixture() {
...
+
+    const ethAmount = parseUnits(config.poolAddEthAmount.toString(), 18);
+    // prettier-ignore
+    await fixture.oethMetaPool
+      .connect(josh)["add_liquidity(uint256[2],uint256)"]([ethAmount, 0], 0, {
+        value: ethAmount,
+      });
+  }
+
+  const oethWhaleAddress = "0xEADB3840596cabF312F2bC88A4Bb0b93A4E1FF5F";
```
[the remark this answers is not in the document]
naddison36: I agree. I've moved it to `addresses.mainnet.oethWhaleAddress`

contracts/contracts/strategies/ConvexEthMetaStrategy.sol
```diff
@@ -257,19 +337,169 @@ contract ConvexEthMetaStrategy is InitializableAbstractStrategy {
...
+
+    /**
+     * @notice One-sided remove of OTokens from the Metapool which are then burned.
+     * This is used when the Metapool has too many OTokens and not enough ETH.
+     * The amount of assets in the vault is unchanged.
+     * The total supply of OTokens is reduced.
+     * The asset value of the strategy and vault is reduced.
+     * @param _lpTokens The amount of Metapool LP tokens to be burned for OTokens.
+     */
+    function removeAndBurnOTokens(uint256 _lpTokens)
```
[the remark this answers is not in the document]
naddison36: This is largely covered in this design decision
https://www.notion.so/originprotocol/Should-the-AMO-repeg-functions-be-on-the-vault-or-AMO-strategy-bf550a1d296546baa1ff6ec2d1150b73

For the people who can't see it in Notion

## Should the AMO repeg functions be on the vault or AMO strategy?
## Options
1. New functions on the strategy. eg removeAndBurnOTokens, mintAndAddOTokens and removeOnlyAssets 
2. Reuse strategy deposit and withdraw using oeth as the asset.
3. Reuse strategy deposit using oeth as the asset and withdraw using Metpool LP as the asset
4. Add new deposit and withdraw functions to the strategy interface that takes an array of assets and amounts like the vault depositToStrategy

## Implications
Option 1 
- removeAndBurnOTokens One-sided remove of OTokens from the Metapool which are then burned.
  Used when the Metapool has not enough ETH and too much OETH.
- mintAndAddOTokens Mint OTokens and one-sided add to the Metapool.
  Used when the Metapool has not enough OETH and too much ETH.
- removeOnlyAssets One-sided remove of ETH from the Metapool, convert to WETH and transfer to the vault.
  Used when the Metapool has not enough OETH and too much ETH.
- Vault does not need upgrading.
- An extra option of removing ETH from the Metapool and converting to WETH in the Vault.
branch nicka/amo-peg-keeping
PR https://github.com/OriginProtocol/origin-dollar/pull/1695 

Option 2
- deposit WETH and OETH or just OETH to the Curve Metapool.
  If just OETH, the asset is OETH. If both, the asset is WETH.
  Used when the Metapool has not enough OETH and too much ETH.
- withdraw ETH and OETH or just OETH from the Curve Metapool.
  If just OETH, the asset is OETH.  If both, the asset is WETH.
  Used when the Metapool has not enough ETH and too much OETH.
- withdraw needs to calculate the amount of Metapool LP token to burn in order to get the remove the required amount of OETH. Curve’s calc_token_amount does not take into account fees. calcWithdraw has been ported from another project that takes into account fees but it’s less than 1 basis point out as the Metapool implementation is slightly different. Getting this accurate could take some days to do. Even if the calcWithdraw and get_D functions can be made accurate, they are gas intensive.
- requires the Vault to be upgraded to allow OETH as an asset in depositToStrategy
branch nicka/amo-peg-keeping-oeth-deposit-withdraw
PR https://github.com/OriginProtocol/origin-dollar/pull/1703

Option 3
- deposit the same as option 2
- withdraw ETH and OETH or just OETH from the Curve Metapool.
  If just OETH, the asset is Metapool LP tokens.  If both, the asset is WETH.
  Used when the Metapool has not enough ETH and too much OETH.
- withdraw doesn’t need to convert the required OETH amount to a Metapool LP token amount.
- requires the Vault to be upgraded to allow OETH as an asset in depositToStrategy
branch nicka/amo-peg-keeping-withdraw-pool-lp
PR https://github.com/OriginProtocol/origin-dollar/pull/1704

Option 4
- requires new deposit and withdraw overrides to be added to the IStrategy interface.
- could be reused by other strategies. eg adding rETH and WETH to Balancer

contracts/test/strategies/oeth-metapool.fork-test.js
naddison36: The tx won't fail if the tilt is improved.

contracts/contracts/strategies/ConvexEthMetaStrategy.sol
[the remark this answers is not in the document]
sparrowDom: I think this design decision mainly decides where to put the interface to the peg-keeping AMO function. What I am proposing is that we pass along the expected amount of OTokens to be burned as a calculation that we perform off-chain. Instead od doing all the math on chain.

contracts/tasks/amoStrategy.js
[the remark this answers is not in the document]
sparrowDom: nice looks better IMHO

contracts/contracts/vault/VaultStorage.sol
```diff
@@ -99,7 +99,7 @@ contract VaultStorage is Initializable, Governable {
     /// @notice OToken mints over this amount automatically allocate funds. 18 decimals.
     uint256 public autoAllocateThreshold;
     /// @notice OToken mints over this amount automatically rebase. 18 decimals.
-    uint256 public rebaseThreshold;
+    uint256 public rebaseThreshold = 0;
```
DanielVF, line 102: This value is actually used from the storage slot on the proxy, while the set here would happen in the constructor on the implementation contract. Because this would be doing something unintuitive, we don't do constructor time variable sets on these storage slots.

contracts/contracts/vault/VaultCore.sol
```diff
@@ -118,12 +118,6 @@ contract VaultCore is VaultStorage {
 
         emit Mint(msg.sender, _amount);
 
-        // Rebase must happen before any transfers occur.
```
DanielVF, line 121: 🎉

contracts/contracts/strategies/ConvexEthMetaStrategy.sol
```diff
@@ -310,7 +542,7 @@ contract ConvexEthMetaStrategy is InitializableAbstractStrategy {
     }
```
DanielVF, line 542: In `checkBalance` the initial `balance += address(this).balance;` doesn't need to be `+=`, but could just be `=`?

contracts/contracts/strategies/ConvexEthMetaStrategy.sol
```diff
@@ -21,75 +21,141 @@ contract ConvexEthMetaStrategy is InitializableAbstractStrategy {
...
+            "Caller is not the Strategist"
+        );
+        _;
+    }
+
+    /**
+     * @dev Checks the Metapools balances have improved and the balances
+     * have not tipped to the other side.
+     */
+    modifier improveMetapoolBalance() {
```
DanielVF, line 78: It is so elegant it brings tears to my eyes.

contracts/contracts/strategies/ConvexEthMetaStrategy.sol
```diff
@@ -21,75 +21,141 @@ contract ConvexEthMetaStrategy is InitializableAbstractStrategy {
...
+     * @param _assets Addresses of supported assets. eg WETH
      */
     function initialize(
         address[] calldata _rewardTokenAddresses, // CRV + CVX
-        address[] calldata _assets,
-        address[] calldata _pTokens,
-        InitializeConfig calldata initConfig
+        address[] calldata _assets // WETH
     ) external onlyGovernor initializer {
         require(_assets.length == 1, "Must have exactly one asset");
```
DanielVF, line 141: Should we require that the one asset be eth? Since that's hardcoded elsewhere in this file.

contracts/contracts/strategies/ConvexEthMetaStrategy.sol
DanielVF, line 78: Can we do these checks on the normal deposit and withdraw methods?

DanielVF, line 78: Are there any possible rounding errors, where the end diffs would be 1 off on these?

contracts/contracts/strategies/ConvexEthMetaStrategy.sol
```diff
@@ -257,19 +338,170 @@ contract ConvexEthMetaStrategy is InitializableAbstractStrategy {
...
+
+    /**
+     * @notice One-sided remove of OTokens from the Metapool which are then burned.
+     * This is used when the Metapool has too many OTokens and not enough ETH.
+     * The amount of assets in the vault is unchanged.
+     * The total supply of OTokens is reduced.
+     * The asset value of the strategy and vault is reduced.
+     * @param _lpTokens The amount of Metapool LP tokens to be burned for OTokens.
+     */
+    function removeAndBurnOTokens(uint256 _lpTokens)
```
DanielVF, line 415: We should probably go ahead and add nonreentrant to these new methods.

contracts/contracts/vault/VaultStorage.sol
naddison36: I put this in to shut slither up. See the error below.

I should have looked deep for the underlying issue which is `VaultCore` was not inheriting from `VaultInitializer` that has the `initialize` function for the Vault. I've changed both `VaultCore` and `VaultAdmin` to inherit from `VaultInitializer`.

This raises the question of whether the `initialize` function should be on `VaultStorage`. The `initialize` function is closely linked to the storage variables so I would vote to just move `initialize` to the `VaultStorage` and get rid of `VaultInitializer`.

So Slither was right to complain!

```
VaultStorage.rebaseThreshold (contracts/vault/VaultStorage.sol#102) is never initialized. It is used in:
        - VaultCore.mint(address,uint256,uint256) (contracts/vault/VaultCore.sol#61-97)
        - VaultCore._redeem(uint256,uint256) (contracts/vault/VaultCore.sol#153-216)
Reference: https://github.com/crytic/slither/wiki/Detector-Documentation#uninitialized-state-variables
```

contracts/contracts/strategies/ConvexEthMetaStrategy.sol
naddison36: yes, and we save 70 gas

Before
```
|  Contract               ·  Method                              ·  Min        ·  Max        ·  Avg        ·  # calls      ·  eur (avg)  │
··························|······································|·············|·············|·············|···············|··············
|  MockVault              ·  checkBalance                        ·          -  ·          -  ·      59858  ·            1  ·          -  │
```
After
```
|  Contract               ·  Method                              ·  Min        ·  Max        ·  Avg        ·  # calls      ·  eur (avg)  │
··························|······································|·············|·············|·············|···············|··············
|  MockVault              ·  checkBalance                        ·          -  ·          -  ·      59788  ·            1  ·          -  │
```

contracts/contracts/strategies/ConvexEthMetaStrategy.sol
naddison36: Yes, the asset must be WETH. I've added this check against the immutable `weth` variable. This works for the unit test using mocks.

contracts/contracts/strategies/ConvexEthMetaStrategy.sol
naddison36: The strategist would have to be compromised for a reentry attack as only the strategist can call these. Adding the reentry check adds another 2.3k gas as it needs to read the reentry flag from storage. This will be significantly cheaper if we implement transient storage (EIP-1153) that is part of the new Cancun network upgrade.

For consistency with `deposit`, `depositAll`, `withdraw` and `withdrawAll` plus the avoidance of spam bug bounties, I'll add them.

Without `nonReentrant`
```
|  Contract               ·  Method                              ·  Min        ·  Max        ·  Avg        ·  # calls      ·  eur (avg)  │
··························|······································|·············|·············|·············|···············|··············
|  ConvexEthMetaStrategy  ·  mintAndAddOTokens                   ·    1555530  ·    1559170  ·    1557350  ·            9  ·          -  │
··························|······································|·············|·············|·············|···············|··············
|  ConvexEthMetaStrategy  ·  removeAndBurnOTokens                ·    1593842  ·    1593854  ·    1593848  ·            6  ·          -  │
··························|······································|·············|·············|·············|···············|··············
|  ConvexEthMetaStrategy  ·  removeOnlyAssets                    ·    1234578  ·    1234614  ·    1234598  ·            9  ·          -  │
```

With `nonReentrant`
```
|  Contract               ·  Method                              ·  Min        ·  Max        ·  Avg        ·  # calls      ·  eur (avg)  │
··························|······································|·············|·············|·············|···············|··············
|  ConvexEthMetaStrategy  ·  mintAndAddOTokens                   ·    1557844  ·    1561484  ·    1559664  ·            9  ·          -  │
··························|······································|·············|·············|·············|···············|··············
|  ConvexEthMetaStrategy  ·  removeAndBurnOTokens                ·    1596159  ·    1596171  ·    1596165  ·            6  ·          -  │
··························|······································|·············|·············|·············|···············|··············
|  ConvexEthMetaStrategy  ·  removeOnlyAssets                    ·    1236965  ·    1237001  ·    1236985  ·            9  ·          -  │
```

contracts/contracts/strategies/ConvexEthMetaStrategy.sol
naddison36: Maybe on the Curve side. We are just subtracting on our side. The strategist shouldn't be aiming for a perfect 18 decimal place balance. That will avoid adding extra on-chain logic for any rounding issues

naddison36: Yes, I think it's a good idea. Its another layer of protection if the vaultValueChecker wasn't used for some reason

The extra gas costs are
deposit 3,659
depositAll 3,662
withdraw  3,660
withdrawAll 3,663

```
|  Contract               ·  Method                              ·  Min        ·  Max        ·  Avg        ·  # calls      ·  eur (avg)  │
··························|······································|·············|·············|·············|···············|··············
|  ConvexEthMetaStrategy  ·  deposit                             ·          -  ·          -  ·    1571158  ·            4  ·      26.35  │
··························|······································|·············|·············|·············|···············|··············
|  ConvexEthMetaStrategy  ·  depositAll                          ·          -  ·          -  ·    1571433  ·            2  ·      26.35  │
··························|······································|·············|·············|·············|···············|··············
|  ConvexEthMetaStrategy  ·  withdraw                            ·          -  ·          -  ·    1570052  ·            4  ·      26.33  │
··························|······································|·············|·············|·············|···············|··············
|  ConvexEthMetaStrategy  ·  withdrawAll                         ·    1548363  ·    1553092  ·    1549939  ·            6  ·      25.99  │
```

```
|  Contract               ·  Method                              ·  Min        ·  Max        ·  Avg        ·  # calls      ·  eur (avg)  │
··························|······································|·············|·············|·············|···············|··············
|  ConvexEthMetaStrategy  ·  deposit                             ·          -  ·          -  ·    1574817  ·            4  ·      26.43  │
··························|······································|·············|·············|·············|···············|··············
|  ConvexEthMetaStrategy  ·  depositAll                          ·          -  ·          -  ·    1575095  ·            2  ·      26.43  │
··························|······································|·············|·············|·············|···············|··············
|  ConvexEthMetaStrategy  ·  withdraw                            ·          -  ·          -  ·    1573712  ·            4  ·      26.41  │
··························|······································|·············|·············|·············|···············|··············
|  ConvexEthMetaStrategy  ·  withdrawAll                         ·    1552026  ·    1556755  ·    1553602  ·            6  ·      26.07  │
```

naddison36: Thinking about is some more, I don't think we need them as we are adding liquidity to both sides. And for withdraws, we are removing proportionally so the balance does not change.

contracts/contracts/strategies/ConvexEthMetaStrategy.sol
```diff
@@ -198,9 +275,13 @@ contract ConvexEthMetaStrategy is InitializableAbstractStrategy {
...
-        // Transfer WETH
+        // Burn all the removed OETH and any that was left in the strategy
+        uint256 oethToBurn = oeth.balanceOf(address(this));
+        IVault(vaultAddress).burnForStrategy(oethToBurn);
+
+        emit Withdrawal(address(oeth), address(lpToken), oethToBurn);
+
+        // Transfer WETH to the recipient
         weth.deposit{ value: _amount }();
         require(
```
naddison36, line 286: this is implementing a safe transfer. Should we use the following instead?
```Solidity
weth.safeTransfer(_recipient, _amount);
```

contracts/contracts/vault/VaultStorage.sol
DanielVF: I would rather we not inherit from VaultInitializer or place VaultInitializer into vault storage. Want to keep the contract space in both contracts, since this has been a long term constraint in the past. Can we just use a slither comment to quite this message?

contracts/contracts/strategies/ConvexEthMetaStrategy.sol
DanielVF: Yes, in theory our math checks out, and we are safe, but I would love to have this guaranteed safeness wrapping this. It's beautiful when no matter how badly you bungle the code, nothing can break the invariant.

The extra gas isn't that bad for something that's used rarely like this.

DanielVF: I was think of what would happen with this wrapping withdrawAll.

DanielVF: I'm wondering if this method actually works for the methods that add or remove from both sides at once. I think this only protects single adds or withdraws.

contracts/contracts/vault/VaultStorage.sol
DanielVF: I can live with it going in just one of the two contracts.

contracts/contracts/strategies/ConvexEthMetaStrategy.sol
DanielVF: We don't need the safe behavior on weth, it's well behaved. Only have to use safeTransfer when we are dealing with coins with incompatible interfaces like USDT. If we've specified a specific coin, and we know it's good, I'm good with just the transfer.

contracts/contracts/strategies/ConvexEthMetaStrategy.sol
DanielVF: I don't think it does. We can leave this off the existing methods, but let's make sure to comment that the check only works for single sided adds or removes.

contracts/contracts/strategies/ConvexEthMetaStrategy.sol
naddison36: ok, thanks

contracts/contracts/vault/VaultStorage.sol
naddison36: I've changed `VaultAdmin` to inherit directly from `VaultStorage`

contracts/contracts/strategies/ConvexEthMetaStrategy.sol
naddison36: I went through the exact same thought process.

I've added a comment to the modifier.

BTW, I took Curve out of the modifier name so its now named `improvePoolBalance`

naddison36: Withdraws are proportional so the pools asset balances doesn't change.

contracts/contracts/strategies/ConvexEthMetaStrategy.sol
```diff
@@ -21,75 +21,146 @@ contract ConvexEthMetaStrategy is InitializableAbstractStrategy {
...
+    // slither-disable-next-line constable-states
+    address private _deprecated_lpToken;
+    // slither-disable-next-line constable-states
+    address private _deprecated_oeth;
+    // slither-disable-next-line constable-states
+    address private _deprecated_weth;
+
+    // Ordered list of pool assets
+    // slither-disable-next-line constable-states
+    uint128 private _deprecated_oethCoinIndex = 1;
```
DanielVF, line 46: These two sets do not do what they look like they do. Even though they are depreciated, we should remove the `=1` and `=0`.

DanielVF:
# AMO Code Review

_Partially complete_

#### Requirements

We want to have finer grain control over responding with the AMO. Primarily we would like to be able to keep a tighter peg by directly controlling adding and removing protocol owned OETH from the pool. This will allow us to respond more cheaply without having to deposit and withdraw large amounts to balance things.

We also want to do this safely, and ensure that we never rug the pool, rug ourselves, or depeg to the coin.

It feels like the verification modifier we came up with here is good - if we only add or remove from one side.

We've also decided to allow the strategist to directly talk to the strategy on these. This is a departure from our usual practice of sending everything through the vault, and is a key area for security examination.

- [ ] Check Direct Call

### Deployment Considerations

#### Internal State

We want any actions that this strategy takes to either move the pool towards peg, or be neutral.

We believe that the algorithms used by our previous deposit and withdraw methods have this property. The new methods act first, according to the strategist, then verify that the action was not harmful. Thus it's very important to carefully verify the verification.

- [ ] Verification Verify

#### Attack

All state changing methods are controlled by the strategist or vault.

Therefor attacks can be broken down into what damage a trusted strategist could do and  damage an external party could do by interfering with the environment this is executing in.

Internal Attacks

We don't think there's any way to execute code in the middle of operations. The AMO deposits and receives ETH from curve, but neither of these movements allow for external execution. The strategy only sends and receives WETH and other trusted, non-execution giving ERC20s.

A malicious strategist is limited in that the AMO can only move toward peg. A very small amount of curve fees can be spent by moving across the flat area of the curve, but since this must always move towards the middle, the amount of damage is limited. 

A malicious strategist could move the peg towards the middle, then externally swap back, then use the AMO to move towards the middle again. However this costs the attacker fees, and we are making some money by bringing the pool back toward peg, and they are loosing some money by imbalancing it.

The only real malicious strategist attack that I've been able to think of would be a self serving repeg then swap of the pool during the depeg of a core coin. This would allow the strategist to alter the mix of OETH and potentially not pay a redeem fee. However this does not make the protocol insolvent.

External attacks 

The only external data we use for calculation is the pool virtual price and pool balances. Note that although the virtual_price on our pool has read only reentrancy protection, the balances do not. However, we should not be vulnerable to read-only reentrancy because we immediately write to curve before/after reading these, and those writes will revert.

We could be vulnerable to a griefing attack where someone looks at our repel efforts and repeatedly slight corrects the peg ahead of us so that our validation fails. This is easily dealt with by either using MEV, or doing about half the size of move, and letting them move to the peg for us.

During a core coin depeg, if we repegged the  pool a MEV bot could arb immediately after we repegged. But someone would do that eventually anyway, and what are we doing reppegging during a core coin depeg.

If an attacker bent the pool more just before a repelling action, they would only increase our profit.

If an attacker added liquidity just before a reppeg, they could make the reppeging effect smaller than we expected (but more profitable) and they would be loosing money.

Removing liquidity just before would make the reppeging effect bigger and less profitable, but would not be bad for us.

#### Logic

#### Tests

#### Flavor

_Could this code be simpler?_

#### Overflow

#### Proxy

#### Black magic

#### Dependencies
No changes to dependancies used.

#### Deploy

#### Authentication

- [x] Never use tx.origin

#### Cryptographic code

_No crypto_

#### Gas problems

    - [ ] A way to remove items

#### External calls

- [x] Unsafe external calls
- [x] Possible malicious behaviors
	- [x] All trusted interactions
- [x] Oracles, one of:
  - [x] No oracles
  - [ ] Oracles can't be bent


#### Ethereum

Contract only sends and receive ETH from a trusted address. Contract wraps/unwraps ETH into WETH for all other uses.

contracts/contracts/strategies/ConvexEthMetaStrategy.sol
naddison36: I agree. I've removed them

DanielVF:
image

image

https://gist.github.com/DanielVF/b785d3f78ce7379f1e86cb65a75529cc

DanielVF:
The verifying checks have also survived the all the thousands of random orders that I've thrown at it, and no constraints have been broken.

DanielVF (approved):
Having completed testing the verification, I'm good with deploying this PR.

contracts/contracts/strategies/ConvexEthMetaStrategy.sol
```diff
@@ -257,19 +343,173 @@ contract ConvexEthMetaStrategy is InitializableAbstractStrategy {
...
+     * This is used when the Curve pool does not have enough OTokens and too many ETH.
+     * The OToken/Asset, eg OETH/ETH, price with increase.
+     * The amount of assets in the vault is unchanged.
+     * The total supply of OTokens is increased.
+     * The asset value of the strategy and vault is increased.
+     * @param _oTokens The amount of OTokens to be minted and added to the pool.
+     */
+    function mintAndAddOTokens(uint256 _oTokens)
+        external
+        onlyStrategist
```
shahthepro, line 379: nit: I feel like Governor should be able to do everything that the Strategist can do

contracts/contracts/vault/VaultCore.sol
```diff
@@ -17,9 +17,9 @@ import { StableMath } from "../utils/StableMath.sol";
 import { IOracle } from "../interfaces/IOracle.sol";
 import { IGetExchangeRateToken } from "../interfaces/IGetExchangeRateToken.sol";
 
-import "./VaultStorage.sol";
+import "./VaultInitializer.sol";
 
-contract VaultCore is VaultStorage {
+contract VaultCore is VaultInitializer {
```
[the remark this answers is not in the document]
naddison36: The main reason is for unit tests. The secondary reason is to keep Slither happy

contracts/contracts/strategies/ConvexEthMetaStrategy.sol
```diff
@@ -302,15 +542,15 @@ contract ConvexEthMetaStrategy is InitializableAbstractStrategy {
         require(_asset == address(weth), "Unsupported asset");
 
         // Eth balance needed here for the balance check that happens from vault during depositing.
-        balance += address(this).balance;
+        balance = address(this).balance;
```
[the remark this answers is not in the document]
naddison36: it wasn't wrong. But its a small gas saving :-)

contracts/contracts/strategies/ConvexEthMetaStrategy.sol
naddison36: I think I've flipped on this a number of times. We can pick this up in the new generalised AMO for frxETH

contracts/deploy/075_oeth_amo_upgrade.js
```diff
@@ -0,0 +1,52 @@
...
+    deployerIsProposer: true,
+    // proposalId: "",
+  },
+  async ({ ethers, deployWithConfirmation }) => {
+    const cConvexEthMetaStrategyProxy = await ethers.getContract(
+      "ConvexEthMetaStrategyProxy"
+    );
+
+    // Deploy and set the immutable variables
+    const dConvexETHMetaStrategy = await deployWithConfirmation(
```
shahthepro, line 20: ℹ️ Checked why there's no `initialize()` method all in the deployment file. Verified that the upgrade will work as expected since the already deployed contract would have the already called the `super._initialize`

shahthepro (approved):

PR #1697, opened 2023-07-12
Balancer MetaStablePool Strategy

sparrowDom:
Adds balancer Meta Stable Pool strategy. Corresponding [Notion design](https://www.notion.so/originprotocol/Support-Balancer-OETH-strategy-9becdea132704e588782a919d7d471eb) document.

Design considerations: since we are planning to potentially also support Composable pool strategy, I've decided to make Meta stable pool contract extend Aura contract. This way Aura implementation can stay the same when/if we add the support for Composable Stable pools.

Things done ✅  : 
- able to add/remove liquidity to/from Balancer meta Stable pool with correct BPT (liquidity tokens) calculation
- able to deploy/withdraw liquidity to/from Aura booster 
- add tests to confirm deposit/withdrawal slippage configuration works as expected
- add test to confirm read-only re-entrancy protection functions


contracts/contracts/strategies/balancer/BaseBalancerStrategy.sol
```diff
@@ -0,0 +1,323 @@
...
+import { IVault } from "../../interfaces/IVault.sol";
+import { IWstETH } from "../../interfaces/IWstETH.sol";
+import { IERC4626 } from "../../../lib/openzeppelin/interfaces/IERC4626.sol";
+import { StableMath } from "../../utils/StableMath.sol";
+import "hardhat/console.sol";
+
+abstract contract BaseBalancerStrategy is InitializableAbstractStrategy {
+    using SafeERC20 for IERC20;
+    using StableMath for uint256;
+    IBalancerVault internal immutable balancerVault =
```
naddison36, line 23: I think this immutable should be set in the constructor so unit tests can mock the balancer vault.

contracts/contracts/strategies/balancer/BaseAuraStrategy.sol
```diff
@@ -0,0 +1,159 @@
...
+        uint256 assetsLength = _assets.length;
+        require(
+            poolAssets.length == assetsLength,
+            "Pool assets and _assets should be the same length."
+        );
+        for (uint256 i = 0; i < assetsLength; ++i) {
+            (address asset, ) = fromPoolAsset(address(poolAssets[i]), 0);
+            require(
+                _assets[i] == asset,
+                "Pool assets and _assets should all have the same numerical order."
```
naddison36, line 70: This doesn't mater that much as it's in the `initialize` function that is only run once, but its better to keep error strings to 31 or less characters so they fit in a 256-bit word. The extra byte stores the string length.

contracts/contracts/strategies/balancer/BaseBalancerStrategy.sol
```diff
@@ -0,0 +1,323 @@
...
+        IBalancerVault(0xBA12222222228d8Ba445958a75a0704d566BF2C8);
+    address internal immutable stEth =
```
naddison36, line 25: My first reaction was to just make these token addresses constants but it's probably better to leave as immutable and set in the constructor so mocks can be used in unit tests

contracts/contracts/strategies/balancer/BaseBalancerStrategy.sol
```diff
@@ -0,0 +1,323 @@
...
+        0xae7ab96520DE3A18E5e111B5EaAb095312D7fE84;
+    address internal immutable wstEth =
+        0x7f39C581F595B53c5cb19bD0b3f8dA6c935E2Ca0;
+    address internal immutable frxEth =
+        0x5E8422345238F34275888049021821E8E08CAa1f;
+    address internal immutable sfrxEth =
+        0xac3E018457B222d93114458476f3E3416Abbe38F;
+
+    address internal pTokenAddress;
```
naddison36, line 34: Any storage variable that can't be changed with an admin function should be immutable in my opinion. This saves over 2k in gas reading the data from storage.

contracts/contracts/strategies/balancer/BaseBalancerStrategy.sol
```diff
@@ -0,0 +1,323 @@
...
+     * Use this modifier with any function that can cause a state change in a pool and is either public itself,
+     * or called by a public function *outside* a Vault operation (e.g., join, exit, or swap).
+     */
+    modifier whenNotInVaultContext() {
+        VaultReentrancyLib.ensureNotInVaultContext(balancerVault);
+        _;
+    }
+
+    /**
+     * @dev Returns bool indicating whether asset is supported by strategy
```
naddison36, line 67: this is an external function so should use @notice so Etherscan will pick up the Natspec.

contracts/contracts/strategies/balancer/BaseBalancerStrategy.sol
```diff
@@ -0,0 +1,323 @@
...
+     * for production usage. Contract allows as low value as 0% for confirming
+     * correct behavior in test suite.
+     */
+    function setMaxWithdrawalSlippage(uint256 _maxWithdrawalSlippage)
+        external
+        onlyVaultOrGovernorOrStrategist
+    {
+        require(
+            _maxWithdrawalSlippage <= 1e18,
+            "Max withdrawal slippage needs to be between 0% - 100%"
```
naddison36, line 286: Error strings should be 31 characters or less to fit in the EVM's 256-bit word

contracts/contracts/strategies/balancer/BaseAuraStrategy.sol
```diff
@@ -0,0 +1,159 @@
...
+import { StableMath } from "../../utils/StableMath.sol";
+import { IRewardStaking } from "../IRewardStaking.sol";
+
+import "hardhat/console.sol";
+
+abstract contract BaseAuraStrategy is BaseBalancerStrategy {
+    using SafeERC20 for IERC20;
+    using StableMath for uint256;
+
+    address internal auraRewardPoolAddress;
```
naddison36, line 21: These should be immutable as they can't be changed with an admin function.

contracts/contracts/strategies/balancer/BaseAuraStrategy.sol
```diff
@@ -0,0 +1,159 @@
...
+ * @author Origin Protocol Inc
+ */
+import { BaseBalancerStrategy } from "./BaseBalancerStrategy.sol";
+import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import { IERC20 } from "../../utils/InitializableAbstractStrategy.sol";
+import { IERC4626 } from "../../../lib/openzeppelin/interfaces/IERC4626.sol";
+import { StableMath } from "../../utils/StableMath.sol";
+import { IRewardStaking } from "../IRewardStaking.sol";
+
+import "hardhat/console.sol";
```
naddison36, line 15: hardhat console should be removed

naddison36 (changes requested):
I'll do the changes on my `nicka/balancer-sfrxETH-stETH-rETH` branch

contracts/contracts/strategies/balancer/BaseBalancerStrategy.sol
```diff
@@ -0,0 +1,323 @@
...
+    function _lpWithdraw(uint256 numBPTTokens) internal virtual;
+
+    function _lpWithdrawAll() internal virtual;
+
+    /**
+     * Balancer returns assets and rateProviders for corresponding assets ordered
+     * by numerical order.
+     */
+    function getPoolAssets() internal view returns (IERC20[] memory assets) {
+        (IERC20[] memory tokens, , ) = balancerVault.getPoolTokens(
```
naddison36, line 162: It's better to set directly to the returned `assets` variable rather than create a new variable `tokens` and then copy to the returned `assets` variable

contracts/contracts/strategies/balancer/BaseAuraStrategy.sol
```diff
@@ -0,0 +1,159 @@
...
+    uint256 internal auraDepositorPTokenId;
+    // renamed from __reserved to not shadow BaseBalancerStrategy.__reserved,
+    int256[50] private __reserved_2;
+
+    struct InitConfig {
+        address platformAddress; // platformAddress Address of the Balancer's pool
+        address vaultAddress; // vaultAddress Address of the vault
+        address auraRewardPoolAddress; // auraRewardPoolAddress Address of the Aura rewards pool
+        address auraRewardStakerAddress; // auraRewardStakerAddress Address of the Aura rewards staker
+        uint256 auraDepositorPTokenId; // auraDepositorPTokenId Address of the Aura rewards staker
```
naddison36, line 32: what's this used for?

naddison36, line 32: What are these used for?

contracts/contracts/strategies/balancer/BalancerMetaPoolStrategy.sol
```diff
@@ -0,0 +1,301 @@
...
+    address internal immutable stETH =
+        0xae7ab96520DE3A18E5e111B5EaAb095312D7fE84;
+    address internal immutable wstETH =
+        0x7f39C581F595B53c5cb19bD0b3f8dA6c935E2Ca0;
+    address internal immutable frxETH =
+        0x5E8422345238F34275888049021821E8E08CAa1f;
+    address internal immutable sfrxETH =
+        0xac3E018457B222d93114458476f3E3416Abbe38F;
+
+    function getRateProviderRate(address _asset)
```
naddison36, line 30: Where is `getRateProviderRate` used?

contracts/contracts/strategies/balancer/BaseAuraStrategy.sol
```diff
@@ -0,0 +1,159 @@
...
+        override
+        onlyHarvester
+        nonReentrant
+    {
+        // Collect CRV and CVX
+        IRewardStaking(auraRewardPoolAddress).getReward();
+        _collectRewardTokens();
+    }
+
+    function checkBalance(address _asset)
```
DanielVF, line 120: 🟥 The way `checkBalance` is currently implemented is the classic way to get rekt calculating funds in an AMM.

As a pool gets more out of balance, the sum of assets in the pool goes up. This allows an attacker to provide us with a falsely inflated view of our assets, allowing an attacker to cause a rebase that is too high. We *must* use something like curve's virtual price, a number that can only go up, and is immune to distortion.

🟥 I don't see read-only reentrancy protection. Do we absolutely know that the pool that we are using does not need this? Even if so, there should dire warnings on this file that it is only compatible with this kind of pool. 

🟧  This allows an attacker to manipulate the mix of coins returned by OETH's redeem function. While not critical like the others, and maybe a good design decision (It means that in a non-manipulating case, we have more accurate numbers.) We should probably doc pro and cons of the different approaches here.

🟧 If I'm reading this right, this does a lot of reads / fairly gas inefficient. Checking the value of a strategy is basically the only strategy function that gas matters, because it affects rebasing, minting, and redeeming.

sparrowDom:
@DanielVF addressing the 2 critical issues you've pointed out: 
- the reentrancy is implemented as a `whenNotInVaultContext` modifier in the `BaseBalancerStrategy`
- will to some fork testing today to confirm the fault with `checkBalance`

contracts/contracts/strategies/balancer/BalancerMetaPoolStrategy.sol
```diff
@@ -0,0 +1,301 @@
...
+import { IERC20 } from "../../utils/InitializableAbstractStrategy.sol";
+import { StableMath } from "../../utils/StableMath.sol";
+
+import "hardhat/console.sol";
+
+contract BalancerMetaPoolStrategy is BaseAuraStrategy {
+    using SafeERC20 for IERC20;
+    using StableMath for uint256;
+
+    address internal immutable stETH =
```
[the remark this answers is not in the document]
sparrowDom: good point agreed

contracts/contracts/strategies/balancer/BalancerMetaPoolStrategy.sol
```diff
@@ -0,0 +1,301 @@
...
+                }
+                return providers[i].getRate();
+            }
+        }
+
+        // should never happen
+        require(false, "Can not find rateProvider");
+    }
+
+    function deposit(address _asset, uint256 _amount)
```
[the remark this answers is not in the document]
sparrowDom: Thanks

contracts/contracts/strategies/balancer/BalancerMetaPoolStrategy.sol
```diff
@@ -0,0 +1,301 @@
...
+                // rate provider doesn't exist, defaults to 1e18
+                if (address(providers[i]) == address(0)) {
+                    return 1e18;
+                }
+                return providers[i].getRate();
+            }
+        }
+
+        // should never happen
+        require(false, "Can not find rateProvider");
```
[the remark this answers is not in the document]
sparrowDom: I see thanks for the info. We have removed this function

contracts/contracts/strategies/balancer/BaseBalancerStrategy.sol
sparrowDom: Yup good point

contracts/contracts/strategies/balancer/BaseAuraStrategy.sol
sparrowDom: got it 👍 useful info

contracts/contracts/strategies/balancer/BaseBalancerStrategy.sol
sparrowDom: 👍

contracts/contracts/strategies/balancer/BaseBalancerStrategy.sol
sparrowDom: 👍

contracts/contracts/strategies/balancer/BaseBalancerStrategy.sol
sparrowDom: 👍

contracts/contracts/strategies/balancer/BaseBalancerStrategy.sol
sparrowDom: good point 👍

contracts/contracts/strategies/balancer/BaseAuraStrategy.sol
sparrowDom: thanks for removing them in your branch

contracts/contracts/strategies/balancer/BalancerMetaPoolStrategy.sol
sparrowDom: Used to be used in `getBPTExpected` but when refactored out of it I hadn't noticed that the function isn't used anymore

contracts/contracts/strategies/balancer/BaseBalancerStrategy.sol
```diff
@@ -0,0 +1,457 @@
...
+            "Max deposit slippage needs to be between 0% - 100%"
+        );
+        emit MaxDepositSlippageUpdated(maxDepositSlippage, _maxDepositSlippage);
+        maxDepositSlippage = _maxDepositSlippage;
+    }
+
+    function _approveBase() internal virtual {
+        IERC20 pToken = IERC20(platformAddress);
+        // Balancer vault for BPT token (required for removing liquidity)
+        pToken.safeApprove(address(balancerVault), 0);
```
DanielVF, line 448: I think we can just use "approve" and skip the double safeApprove()'s. The requirement for the 0 set before a change is actually local code in the local function `safeApprove`.

This will apply throughout the balancer strategy.

contracts/contracts/strategies/balancer/BaseAuraStrategy.sol
```diff
@@ -0,0 +1,140 @@
...
+     */
+    function initialize(
+        address[] calldata _rewardTokenAddresses, // BAL & AURA
+        address[] calldata _assets,
+        address[] calldata _pTokens
+    ) external override onlyGovernor initializer {
+        maxWithdrawalSlippage = 1e15;
+        maxDepositSlippage = 1e15;
+
+        IERC20[] memory poolAssets = getPoolAssets();
```
DanielVF, line 49: Optional - I love the checking concept here, but seems weird that is in the aura abstract contract, because it's not related to aura specific functionality or storage slots. 

Same thing actually with the slippage. That's not aura specific either.

contracts/contracts/strategies/balancer/BaseAuraStrategy.sol
```diff
@@ -0,0 +1,140 @@
...
+
+abstract contract BaseAuraStrategy is BaseBalancerStrategy {
+    using SafeERC20 for IERC20;
+    using StableMath for uint256;
+
+    /// @notice Address of the Aura rewards pool
+    address public immutable auraRewardPoolAddress;
+
+    // renamed from __reserved to not shadow BaseBalancerStrategy.__reserved,
+    int256[50] private __reserved_2;
```
DanielVF, line 25: Open Zeppelin likes the storage slots used + the gap number to equal a nice even number. This is different than starting with a nice even number.

From past audits:

> the gap typically expands the storage size to a round number (like 50 or 100) so all versions of the contract have a predictable size, and the storage layout of descendant contracts starts at a predictable slot number.

DanielVF: Rather than using a number, perhaps we should use something like "__reserved_BaseAuraStrategy"?

contracts/contracts/strategies/balancer/BaseBalancerStrategy.sol
```diff
@@ -0,0 +1,457 @@
...
+    IBalancerVault public immutable balancerVault;
+    /// @notice Balancer pool identifier
+    bytes32 public immutable balancerPoolId;
+
+    // Max withdrawal slippage denominated in 1e18 (1e18 == 100%)
+    uint256 public maxWithdrawalSlippage;
+    // Max deposit slippage denominated in 1e18 (1e18 == 100%)
+    uint256 public maxDepositSlippage;
+
+    int256[50] private __reserved;
```
DanielVF, line 40: Same gap comment as for aura contract.

contracts/contracts/strategies/balancer/BaseBalancerStrategy.sol
```diff
@@ -0,0 +1,457 @@
...
+     * @dev If an asset is rebasing the Balancer pools have a wrapped versions of assets
+     * that the strategy supports. This function converts the pool(wrapped) asset
+     * and corresponding amount to strategy asset.
+     */
+    function toPoolAsset(address asset, uint256 amount)
+        internal
+        view
+        returns (address poolAsset, uint256 poolAmount)
+    {
+        poolAmount = 0;
```
DanielVF, line 279: Options: Is this line needed? Seems like we don't worry about initializing return variables elsewhere in the code.

contracts/contracts/strategies/balancer/BaseBalancerStrategy.sol
```diff
@@ -0,0 +1,457 @@
...
+            wrappedAmount = amount;
+        }
+    }
+
+    /**
+     * @dev Converts wrapped asset to its rebasing counterpart.
+     */
+    function unwrapPoolAsset(address asset, uint256 amount)
+        internal
+        returns (uint256 wrappedAmount)
```
DanielVF, line 342: Isn't this the _unwrapped_ amount, rather than the _wrapped_ amount?

contracts/contracts/strategies/balancer/BaseBalancerStrategy.sol
```diff
@@ -0,0 +1,457 @@
...
+         *  - take in consideration pool's tokens and their exchangeRates. For the queried asset
+         *    deduce what a share of that asset should be in the pool in case of pool being
+         *    completely balanced. To get to this we are using inverted exchange rate accumulator
+         *    and queried asset inverted exchange rate.
+         *  - divide the amount of the previous step with assetRate to convert the ETH
+         *    denominated representation to asset denominated
+         */
+        amount = (bptBalance.mulTruncate(
+            IRateProvider(platformAddress).getRate()
+        ) / tokens.length);
```
DanielVF, line 143: 🟧I don't think using `tokens.length` is correct here.

If `tokens.length` is forever fixed, we are much better off gas wise just reading from our own storage.

If `tokens.length` could change, we would be giving the vault bad totals, since the vault is querying the strategy a number of times, and that number is not by the tokens in the strategy, but by our number of supported assets.

contracts/contracts/strategies/balancer/BaseBalancerStrategy.sol
```diff
@@ -0,0 +1,457 @@
...
+     * and since from_pool_token(asset_amount).amount and pool_asset_rate cancel each-other out
+     * this makes the final equation:
+     *
+     * bptExpected = oracle_price * asset_amount / BPT_token_rate
+     *
+     * more explanation here:
+     * https://www.notion.so/originprotocol/Support-Balancer-OETH-strategy-9becdea132704e588782a919d7d471eb?pvs=4#382834f9815e46a7937f3acca0f637c5
+     */
+    /* solhint-enable max-line-length */
+    function getBPTExpected(address _asset, uint256 _amount)
```
DanielVF, line 216: Just a comment, no change required. Our current oracles are fairly bad in terms of latency and accuracy. The default value set for reverting a deposit/withdraw 1e15, or 0.1% slippage + fee difference from an oracle. That feels like it's going to trigger some times. At a minimum during fork tests.

It may not be much of a production problem, because the strategist can always alter the slippage if needed.

contracts/contracts/strategies/balancer/BalancerMetaPoolStrategy.sol
```diff
@@ -0,0 +1,477 @@
...
+    /**
+     * @notice Deposits an `_amount` of vault collateral assets
+     * from the this strategy contract to the Balancer pool.
+     * @param _asset Address of the Vault collateral asset
+     * @param _amount The amount of Vault collateral assets to deposit
+     */
+    function deposit(address _asset, uint256 _amount)
+        external
+        override
+        whenNotInVaultContext
```
DanielVF, line 39: I don't think we need these read only reentrancy checks on deposits and withdraws, since balancer's primary reentrancy protection should be active since they are writes that interact with the balancer contracts.

Argue with me if I'm wrong. :)

contracts/contracts/strategies/balancer/BalancerMetaPoolStrategy.sol
```diff
@@ -0,0 +1,477 @@
...
+
+            if (amount > 0) {
+                emit Deposit(asset, platformAddress, amount);
+
+                // wrap rebasing assets like stETH and frxETH to wstETH and sfrxETH
+                (, mappedAmounts[i]) = wrapPoolAsset(asset, amount);
+            }
+        }
+
+        // TODO move this loop into the previous loop
```
DanielVF, line 111: TODO comment here. Do we need to do it?

contracts/contracts/strategies/balancer/BalancerMetaPoolStrategy.sol
```diff
@@ -0,0 +1,477 @@
...
+    function _deposit(address[] memory _assets, uint256[] memory _amounts)
+        internal
+    {
+        (IERC20[] memory tokens, , ) = balancerVault.getPoolTokens(
+            balancerPoolId
+        );
+
+        uint256[] memory mappedAmounts = new uint256[](tokens.length);
+        address[] memory mappedAssets = new address[](tokens.length);
+        for (uint256 i = 0; i < _assets.length; ++i) {
```
DanielVF, line 98: 🟧 This method scares me because there's not an obvious check here that the passed in assets are all valid supported one.

contracts/contracts/strategies/balancer/BalancerMetaPoolStrategy.sol
```diff
@@ -0,0 +1,477 @@
...
+        _deposit(assets, amounts);
+    }
+
+    function _deposit(address[] memory _assets, uint256[] memory _amounts)
+        internal
+    {
+        (IERC20[] memory tokens, , ) = balancerVault.getPoolTokens(
+            balancerPoolId
+        );
+
```
DanielVF, line 95: We should require that _assets.length = _amounts.length

DanielVF (changes requested):

contracts/contracts/strategies/balancer/BalancerMetaPoolStrategy.sol
naddison36: that was me. No, it's not needed. It's just a gas optimisation which probably wouldn't save that much relative to the total gas cost of a deposit

naddison36: no, its a gas optimisation that is not needed

contracts/contracts/strategies/balancer/BalancerMetaPoolStrategy.sol
```diff
@@ -0,0 +1,477 @@
...
+            _approveAsset(assetsMapped[i]);
+        }
+        _approveBase();
+    }
+
+    // solhin t-disable-next-line no-unused-vars
+    function _abstractSetPToken(address _asset, address) internal override {
+        address poolAsset = toPoolAsset(_asset);
+        if (_asset == stETH) {
+            IERC20(stETH).safeApprove(wstETH, 1e50);
```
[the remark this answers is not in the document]
naddison36: I agree with this. I hit a similar issue yesterday with the AMO peg-keeping changes. I did two `safeApprove`s when one `approve` would do the job. I'll update the AMO PR

contracts/contracts/strategies/balancer/BaseAuraStrategy.sol
naddison36: wow, I'd be tempted to use 13 just to trigger a response :-)

contracts/contracts/strategies/balancer/BalancerMetaPoolStrategy.sol
DanielVF: Let's remove the comment then.

contracts/contracts/strategies/balancer/BalancerMetaPoolStrategy.sol
```diff
@@ -0,0 +1,477 @@
...
+
+    /**
+     * @notice Returns the rate supplied by the Balancer configured rate
+     * provider. Rate is used to normalize the token to common underlying
+     * pool denominator. (ETH for ETH Liquid staking derivatives)
+     *
+     * @param _asset Address of the Balancer pool asset
+     * @return rate of the corresponding asset
+     */
+    function getRateProviderRate(address _asset)
```
DanielVF, line 450: The rate providers are hard coded, correct? Can we just cache a mapping when creating the contract? Would much rather make one external call than three here, given that balance checking is an area we care about gas on.

contracts/contracts/strategies/balancer/BalancerMetaPoolStrategy.sol
```diff
@@ -0,0 +1,477 @@
...
+        _withdraw(_recipient, _assets, _amounts);
+    }
+
+    /**
+     * @dev Withdraw multiple Vault collateral asset from the Balancer pool.
+     * @param _recipient Address to receive the Vault collateral assets. Typically is the Vault.
+     * @param _assets Addresses of the Vault collateral assets
+     * @param _amounts The amounts of Vault collateral assets to withdraw
+     */
+    function _withdraw(
```
DanielVF, line 197: This technique currently used by this method is fairly ugly. On a ten million dollar withdraw, this could leave $10,000 in limbo. And that's if the oracles are working perfectly and meshed with the pool prices. We might have to use looser oracle numbers.

What we really want to know is the exact number of BPT tokens we need to withdraw before we get them from staking.

When a withdraw happens, the balancer vault asks the balancer pool for the details of what should happen, which ends up including the number of BPT tokens to burn. This is calculated internally to the pool based on the requested tokens.

The math actually isn't that bad:

image

(Though we would need to include the fee adjustments before running this method, which doesn't look nearly as hard as curve.)

While it would be nice if they had an external interface on this, I don't see it in the pool contract. We could just make a little helper contract that runs their version of solidity and has the needed external interface to run this method.

DanielVF: Or, if we are going to withdraw more tokens than needed, we could use _lpDepositAll() at the end to deposit them back to earning.

We could also a simple max cap on the maxBPTtoWithdraw to our current balance - if we estimate more than we have, we just use the amount that we have. This would allow us to withdraw right up to the edge of what we can.

DanielVF: We could also skip all the estimation, and withdraw all but one raw LP token (gas savings), do the withdraw, then put it all back.

contracts/contracts/strategies/balancer/BaseBalancerStrategy.sol
sparrowDom: Good point 👍

contracts/contracts/strategies/balancer/BaseAuraStrategy.sol
```diff
@@ -0,0 +1,140 @@
...
+            IERC20(platformAddress).balanceOf(address(this)) +
+            // maxRedeem is implemented as balanceOf(address) in Aura
+            IERC4626(auraRewardPoolAddress).maxRedeem(address(this));
+    }
+
+    function _approveBase() internal virtual override {
+        super._approveBase();
+
+        IERC20 pToken = IERC20(platformAddress);
+        pToken.safeApprove(auraRewardPoolAddress, 0);
```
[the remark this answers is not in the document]
sparrowDom: Agreed removed all 3 double approves and replaced them with a single `safeApprove`

contracts/contracts/strategies/balancer/BalancerMetaPoolStrategy.sol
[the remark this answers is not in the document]
sparrowDom: agreed thanks

contracts/contracts/strategies/balancer/BalancerMetaPoolStrategy.sol
```diff
@@ -0,0 +1,477 @@
...
+
+    /**
+     * @dev Approves the Balancer Vault to transfer an asset from
+     * this strategy. The assets could be a Vault collateral asset
+     * like WETH or rETH; or a Balancer pool asset that wraps the vault asset
+     * like wstETH or sfrxETH.
+     */
+    function _approveAsset(address _asset) internal {
+        IERC20 asset = IERC20(_asset);
+        asset.safeApprove(address(balancerVault), 0);
```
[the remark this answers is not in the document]
sparrowDom: 👍

contracts/contracts/strategies/balancer/BaseBalancerStrategy.sol
```diff
@@ -0,0 +1,457 @@
...
+
+    /**
+     * @dev Ensure we are not in a Vault context when this function is called, by attempting a no-op internal
+     * balance operation. If we are already in a Vault transaction (e.g., a swap, join, or exit), the Vault's
+     * reentrancy protection will cause this function to revert.
+     *
+     * Use this modifier with any function that can cause a state change in a pool and is either public itself,
+     * or called by a public function *outside* a Vault operation (e.g., join, exit, or swap).
+     */
+    modifier whenNotInVaultContext() {
```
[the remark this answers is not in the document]
sparrowDom: good point have renamed it to `whenNotInBalancerVaultContext` and added link to Notion document explaining the read-only re-entrancy vulnerability

contracts/contracts/strategies/balancer/BalancerMetaPoolStrategy.sol
DanielVF: Exact amounts are probably too much work, on reflection.

contracts/contracts/strategies/balancer/BaseAuraStrategy.sol
```diff
@@ -0,0 +1,140 @@
...
+     *                order as returned by coins on the pool contract, i.e.
+     *                WETH, stETH
+     * @param _pTokens Platform Token corresponding addresses
+     */
+    function initialize(
+        address[] calldata _rewardTokenAddresses, // BAL & AURA
+        address[] calldata _assets,
+        address[] calldata _pTokens
+    ) external override onlyGovernor initializer {
+        maxWithdrawalSlippage = 1e15;
```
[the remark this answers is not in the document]
sparrowDom: agreed 👍

contracts/contracts/strategies/balancer/BaseAuraStrategy.sol
sparrowDom: good catch, initialize fits into BaseBalancerStrategy much better

contracts/contracts/strategies/balancer/BaseAuraStrategy.sol
sparrowDom: @naddison36 haha yes. 

@DanielVF agreed, changed the slots in `BaseBalancerStrategy` to 48 for a total of 50. Leaving the BaseAuraStrategy at 50 since it doesn't have any non inherited storage variables.

contracts/contracts/strategies/balancer/BaseBalancerStrategy.sol
sparrowDom: isn't, removed

contracts/contracts/strategies/balancer/BaseBalancerStrategy.sol
sparrowDom: goo catch thanks, changed

contracts/contracts/strategies/balancer/BaseBalancerStrategy.sol
sparrowDom: that's a very good point, have changed it to `assetsMapped.length`. We could only support 1 asset of 2 in the pool or 2 of 3 (while being exposed to all pool assets while holding BPT of course). We would need to apply other changes to the strategy in order for the contract to support such setup.

contracts/contracts/strategies/balancer/BaseBalancerStrategy.sol
sparrowDom: I have been thinking about this, I think it is problematic in both directions. A lot of ETH pegged oracles have a 24 hour heartbeat and 2% price change trigger. In case e.g. rETH price plunges for 1.5% (rETH -> ETH price that is) and oracles still have not updated, such a low BPTExpected would be set that we could get MEW-ed. 

Check balance wouldn't catch that since that only works off of rate providers. So it wouldn't be shielded with vault value checker.

**Proposed solution:** 
Similar to VaultValue checker we could write an Oracle price checker. This could be an offline brownie module that supplements our allocations scripts and pulls prices from 1inch and passes them to onChain oracle price checker function. If the deviation is too big: 
- in direction of asset price being too high. It could propose to increase the slippage when withdrawing. And prevent deposits since the strategy would be asking for too little BPT
- in direction of asset price being too low. It could propose to increase slippage when depositing.(and undo after deposit) And prevent withdrawing since we'd be leaving too much BPT on the table.

In case we ever make this a default strategy such solutions are of no use of course.

contracts/contracts/strategies/balancer/BalancerMetaPoolStrategy.sol
sparrowDom: correct, I don't know what I was thinking :)

contracts/contracts/strategies/balancer/BalancerMetaPoolStrategy.sol
sparrowDom: 🗑️

contracts/contracts/strategies/balancer/BalancerMetaPoolStrategy.sol
sparrowDom: I don't think VaultCore can be configured to call deposit with the unsupported asset. Still this sort of safety check shouldn't be very gas intensive so adding a: 
```
require(supportsAsset(asset), "Unsupported asset");
```

contracts/contracts/strategies/balancer/BalancerMetaPoolStrategy.sol
sparrowDom: good catch thanks, added

contracts/contracts/strategies/balancer/BalancerMetaPoolStrategy.sol
sparrowDom: It is immutable with the MetaStable pool we are fork testing with. I'd like to understand the composable stable pools better because they have a different implementation of it. We are not supporting those yet and looking at [the implementation ](https://github.com/balancer/balancer-v2-monorepo/blob/master/pkg/pool-stable/contracts/ComposableStablePoolRates.sol)they seem to honor the and keep indexes the same.

I completely agree with you that this change should be made for the reasons already outlined. I would just introduce it more carefully with more time to research that we are completely safe.

contracts/contracts/strategies/balancer/BalancerMetaPoolStrategy.sol
sparrowDom: I've addressed this issue to some extent in another comment. So the problem is that with:
1. exact oracle prices -> as you mention we are still leaving 0.1% on the table
2. oracles lagging, assets more expensive -> we wouldn't calculate sufficient BPT and it will revert
3. oracles lagging, assets cheaper -> we are exposed to being MEVed

Since all the prices are ETH denominated I wouldn't expect big moves to making assets more expensive. On the other hand there could be slashing (point 3.) events that would make assets cheaper and be within oracle tolerances and make us exposed to MEV.

One reservation regarding deriving BPT strictly off of the pool I have is that that number is manipulatable with flash loans tilting the pool.

I need to do some research how rETH/ETH wstETH/ETH oracle prices behave. I would expect them to very slowly move up. In that case our strategy will mostly perform the scenario 1. and leave some extra BPT in the strategy. 

The real danger is the slashing of underlying LSDs. One (temporary) way to mitigate this would be to have it built into our off-chain allocation script, that if Oracle prices and pool balances (price) deviate too much, the allocation doesn't get built (to not lose funds with MEV). Need to think more on this one. 


And you are right the BPT math is pretty straightforward. Tried to cram it into the strategy contract, but it reaches the deployment size limit -> https://github.com/OriginProtocol/origin-dollar/pull/1740 Like you said it would need to be deployed as a separate contract.

contracts/contracts/proxies/Proxies.sol
```diff
@@ -158,6 +158,7 @@ contract OETHMorphoAaveStrategyProxy is InitializeGovernedUpgradeabilityProxy {
 }
 
 /**
+<<<<<<< HEAD
```
DanielVF, line 161: Merge error

contracts/contracts/strategies/balancer/BaseBalancerStrategy.sol
```diff
@@ -0,0 +1,488 @@
...
+    }
+
+    /// @notice Balancer Pool Tokens (BPT) in the Balancer pool.
+    function _getBalancerPoolTokens()
+        internal
+        view
+        virtual
+        returns (uint256 balancerPoolTokens)
+    {
+        balancerPoolTokens = IERC20(platformAddress).balanceOf(address(this));
```
DanielVF, line 223: Noting, I don't think the code in this function is ever used, tested, or included in a deployed contract?

contracts/contracts/strategies/balancer/BalancerMetaPoolStrategy.sol
```diff
@@ -0,0 +1,561 @@
...
+        for (uint256 i = 0; i < assetCount; ++i) {
+            _abstractSetPToken(assetsMapped[i], platformAddress);
+        }
+        _approveBase();
+    }
+
+    // solhint-disable-next-line no-unused-vars
+    function _abstractSetPToken(address _asset, address) internal override {
+        address poolAsset = toPoolAsset(_asset);
+        if (_asset == stETH) {
```
DanielVF, line 504: Could we just use poolAsset and skip all the if's?

DanielVF:
There appears to be a little behavior that's unexpected to me. It looks like an even deposit can make a unit profit.

image

Similarly, it's possible for withdrawAll() to be making a tiny unit loss, which is not what I would expect.

image

Again, when picking different withdraw mixes, it looks like after withdrawing we have a little less money than we expect.

image

All of these point to the the result of check balance being a tiny bit high somehow. Not sure how. Possibly acceptable. But still a bit of a mismatch to what I would we would expect. I wonder where this comes from.

contracts/contracts/proxies/Proxies.sol
sparrowDom: thanks, removed

contracts/contracts/strategies/balancer/BalancerMetaPoolStrategy.sol
sparrowDom: This is used by the `checkBalance` functions [here](https://github.com/OriginProtocol/origin-dollar/pull/1697/files#diff-7d59dfab7a512a013f0696d107a97b7b6b0f1e8d656b8baac6642f61e3d84db6R161) and [here](https://github.com/OriginProtocol/origin-dollar/pull/1697/files#diff-7d59dfab7a512a013f0696d107a97b7b6b0f1e8d656b8baac6642f61e3d84db6R208) and both versions (with asset parameter & without) of checkBalance are used in a few places in the [fork test file](https://github.com/OriginProtocol/origin-dollar/pull/1697/files#diff-4ff4ac51088496da8787ba1d1726b09145dc496c6c9a9f9877bf7c47b3fc2522R665)

contracts/contracts/strategies/balancer/BaseBalancerStrategy.sol
sparrowDom: This is used by the `checkBalance` functions [here](https://github.com/OriginProtocol/origin-dollar/pull/1697/files#diff-7d59dfab7a512a013f0696d107a97b7b6b0f1e8d656b8baac6642f61e3d84db6R161) and [here](https://github.com/OriginProtocol/origin-dollar/pull/1697/files#diff-7d59dfab7a512a013f0696d107a97b7b6b0f1e8d656b8baac6642f61e3d84db6R208) and both versions (with asset parameter & without) of checkBalance are used in a few places in the [fork test file](https://github.com/OriginProtocol/origin-dollar/pull/1697/files#diff-4ff4ac51088496da8787ba1d1726b09145dc496c6c9a9f9877bf7c47b3fc2522R665)

contracts/contracts/strategies/balancer/BalancerMetaPoolStrategy.sol
sparrowDom: We could change it to: 
```
address poolAsset = _toPoolAsset(_asset);
IERC20(_asset).approve(poolAsset, type(uint256).max);
```

Though in case of WETH we would be doing: 
```
IERC20(weth).approve(weth, type(uint256).max);
```
To mitigate that we could then:
```
address poolAsset = _toPoolAsset(_asset);
if (poolAsset != _asset) {
  IERC20(_asset).approve(poolAsset, type(uint256).max);
}
```
But we have the if's back and a bit worse readability

sparrowDom:
@DanielVF you've asked me about caching the decimals yesterday. I forgot that by removing the Oracles (and relying solely on VaultValueChecker) we no longer query the decimals - so no need for caching. 2 birds 1 stone :)

sparrowDom:
@DanielVF awesome approach of testing the behaviour of checkBalance and value in vault functions. I've made a couple of adjustments: 
- by [converting the reth amounts to units](https://github.com/OriginProtocol/origin-dollar/pull/1817/files#diff-4231e9a00092fea5c815707389dd4c247340c9fe5a32c12d74d06318b045b7aeR40-R44) when accounting for tilts
- by [balancing the pool first](https://github.com/OriginProtocol/origin-dollar/pull/1817/files#diff-4231e9a00092fea5c815707389dd4c247340c9fe5a32c12d74d06318b045b7aeR62) before performing any tilts. The original script was probably assuming the pool was balanced to begin with, when in reality it is slightly imbalanced. That is what was causing profitability in some cases. 
 
I've rerun the Jupyter notebook and this is what the deposit profit looks like with adjustments: 
Screenshot 2023-09-13 at 23 26 25

[Link to PR](https://github.com/OriginProtocol/origin-dollar/pull/1817) with all the changes.

naddison36 (approved):
Great work Domen

DanielVF (approved):
#### Requirements

We are adding support for a balancer MetaStablePool strategy.

We are doing the same way of splitting checkBalances into fixed equal percentages that we used for the curve pools.

### Deployment Considerations

- [ ] Monitoring needed for pool
- [ ] Strategy needed in analytics

#### Internal State

_config only_

#### Attack

Biggest concern with our strategy is check balance inflation. We've fork tested this, and an attacker was unable to manipulate our balance downwards via tilting the pool.

Balancer has known read-only reentrancy vectors. We have the balancer approved method on our checkBalance


#### Logic

#### Tests

_I've run my own fork test suite._

#### Flavor

Still rather `whenNotInBalancerVaultContext` was an inline function, rather than a modifier, but can live with this.

#### Overflow

#### Proxy

#### Black magic

#### Dependencies

#### Deploy

#### Authentication

- [x] Never use tx.origin

#### Cryptographic code

_No crypto code_

#### Gas problems

    - [ ] A way to remove items

#### External calls

- [ ] Malicious behaviors
  - [x] Check with vault value checker 
- [x] Oracles, one of:
  - [ ] No oracles
  - [ ] Oracles can't be bent


#### Ethereum

#### Behavior of read-only reentrancy check

PR #1703, opened 2023-07-17
AMO peg keeping using strategy deposit and withdraw functions

naddison36:
This is an alternative to the first AMO rebalance functions.

Changes:
* Uses the standard `deposit` and `withdraw` strategy functions instead of AMO specific functions on the strategy contract that is callable by the strategist. Both `deposit` and `withdraw` asset can either be WETH or OETH. 
* As Curve's `calc_token_amount` does not take into account fees, `calcWithdraw` is introduced that does take into account the Metapool fees. This calculates the amount of LP tokens needed to remove a required amount of ether ETH or OETH.

### DIFF
--- a/contracts/contracts/strategies/ConvexEthMetaStrategy.sol
+++ b/contracts/contracts/strategies/ConvexEthMetaStrategy.sol
@@ -567,19 +567,14 @@ contract ConvexEthMetaStrategy is InitializableAbstractStrategy {
         // Sum the balances
         uint256 S = xp[0] + xp[1];
 
-        // Do these multiplications here rather than in each loop
-        uint256 xp0 = xp[0] * N_COINS;
-        uint256 xp1 = xp[1] * N_COINS;
-
         uint256 Dprev = 0;
         D = S;
         uint256 D_P;
         for (uint256 i = 0; i < 255; ) {
-            // D_P: uint256 = D
-            // for _x in xp:
-            //     D_P = D_P * D / (_x * N_COINS)  # If division by 0, this will be borked: only withdrawal will work.
+            // D_P: uint256 = D * D / _xp[0] * D / _xp[1] / (N_COINS)**2
+            // (N_COINS)**2 = 4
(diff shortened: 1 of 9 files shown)


contracts/contracts/strategies/ConvexEthMetaStrategy.sol
```diff
@@ -21,19 +21,23 @@ contract ConvexEthMetaStrategy is InitializableAbstractStrategy {
...
-    IERC20 internal lpToken;
-    IERC20 internal oeth;
-    IWETH9 internal weth;
+    address public immutable cvxDepositorAddress;
+    IRewardStaking public immutable cvxRewardStaker;
+    uint256 public immutable cvxDepositorPTokenId;
+    ICurveETHPoolV1 public immutable curvePool;
+    IERC20 public immutable lpToken;
+    IERC20 public immutable oeth;
+    IWETH9 public immutable weth;
```
DanielVF, line 33: Is this going to change the storage slots, for future upgrades on top of this? Maybe we should just have a depreciated static sized array?

contracts/contracts/strategies/ConvexEthMetaStrategy.sol
```diff
@@ -89,126 +95,206 @@ contract ConvexEthMetaStrategy is InitializableAbstractStrategy {
...
+                ),
+                "Depositing LP to Convex not successful"
+            );
+        } else if (_asset == address(oeth)) {
+            // If depositing OETH
+
+            uint256 valueInLpTokens = _amount.divPrecisely(
+                curvePool.get_virtual_price()
+            );
+            uint256 minMintAmount = valueInLpTokens.mulTruncate(
```
DanielVF, line 188: Is this essentially a clever way to ensure that we aren't rugging the pool?

naddison36:
Closing as the implementation on the strategy will be used instead https://github.com/OriginProtocol/origin-dollar/pull/1695

PR #1722, opened 2023-07-28
Add multi asset deposit and withdraw to Balancer strategy

naddison36:
* Added optional multi-asset `deposit`, `withdraw` and `checkBalance` to strategies
* Added `isMultiAssets ` to the `Strategy` struct used by the Vault's `strategies` mapping. This config tells the vault whether a strategy supports multi assets.
* Removed `_deprecated` from the `Strategy` struct so only one slot is ready from storage, not two. This is safe as the storage location of the mapped strategies does not overlap.
* Changed Balancer `deposit` and `withdraw` from using a fixed amount of BPT to a fixed amount of pool assets
* Withdraw of Balancer Pool Tokens from Aura adjusted for BPT already held in the strategy contract.
* Balancer strategy implements `checkBalance` that uses the pool's virtual price (`getRate`) to value the Balancer Pool Tokens (BPT) back to ETH


naddison36:
This is not in scope of the Balancer strategy so will close

PR #1727, opened 2023-08-01
Balancer fork tests

naddison36:
* `BaseAuraStrategy` handle withdraws close to the BPT limit in Aura by checking the max BPT is <= the balance of Aura Pool shares.
* Using `maxRedeem` instead of `balanceOf` as that's more technically correct for 4626 vaults. The balance can be greater than what's allowed to be redeemed.
* Fixed bug Balancer in fork tests

### DIFF
--- a/contracts/contracts/strategies/balancer/BaseAuraStrategy.sol
+++ b/contracts/contracts/strategies/balancer/BaseAuraStrategy.sol
@@ -81,6 +81,7 @@ abstract contract BaseAuraStrategy is BaseBalancerStrategy {
      */
     function _lpWithdraw(uint256 numBPTTokens) internal virtual override {
         // Get all the strategy's BPTs in Aura
+        // maxRedeem is implemented as balanceOf(address) in Aura
         uint256 maxBPTTokens = IERC4626(auraRewardPoolAddress).maxRedeem(
             address(this)
         );
@@ -97,6 +98,7 @@ abstract contract BaseAuraStrategy is BaseBalancerStrategy {
      */
     function _lpWithdrawAll() internal virtual override {
         // Get all the strategy's BPTs in Aura
+        // maxRedeem is implemented as balanceOf(address) in Aura
         uint256 bptBalance = IERC4626(auraRewardPoolAddress).maxRedeem(
             address(this)
         );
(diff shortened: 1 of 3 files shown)


contracts/contracts/strategies/balancer/BaseAuraStrategy.sol
```diff
@@ -123,7 +131,7 @@ abstract contract BaseAuraStrategy is BaseBalancerStrategy {
     {
         balancerPoolTokens =
             IERC20(platformAddress).balanceOf(address(this)) +
-            IERC4626(auraRewardPoolAddress).balanceOf(address(this));
+            IERC4626(auraRewardPoolAddress).maxRedeem(address(this));
```
sparrowDom, line 134: good point, also good to know that these things are implementation wise equivalent: 
https://github.com/aurafinance/aura-contracts/blob/45e4507334001a0a8e409be84ced65b73681a480/contracts/compounder/GenericVault.sol#L364-L366

contracts/contracts/strategies/balancer/BaseAuraStrategy.sol
```diff
@@ -78,8 +80,14 @@ abstract contract BaseAuraStrategy is BaseBalancerStrategy {
...
     function _lpWithdraw(uint256 numBPTTokens) internal virtual override {
+        // Get all the strategy's BPTs in Aura
+        // maxRedeem is implemented as balanceOf(address) in Aura
+        uint256 maxBPTTokens = IERC4626(auraRewardPoolAddress).maxRedeem(
+            address(this)
+        );
+
         IRewardStaking(auraRewardPoolAddress).withdrawAndUnwrap(
-            numBPTTokens,
+            Math.min(numBPTTokens, maxBPTTokens),
```
sparrowDom, line 90: I do understand rationale for this, to make it more resilient, but I don't think it is an ok change. I think the caller of this function would expect for AURA to withdraw `numBPTTokens` or fail if it unable to withdraw all.

sparrowDom:
For now I undid the "lenient" implementation that would reduce the amount of BPT withdrawn in case not enough were available. I think that is the more correct way, but we need to merge this in for the audit. 

Feel free to disagree and happy to revisit this issue. There is still one test failing with a very weird stack: 
Screenshot 2023-08-02 at 23 37 59
 
The `withdraw` function doesn't have the ability to call `lpWithdrawAll`. Will merge this in and figure it out in the merged branch.

PR #1740, opened 2023-08-06
add ability to calculate BPT required the way Balancer does it

sparrowDom:
add ability to calculate BPT required the way Balancer does it. Can not be part of Balancer strategy contract since it exceeds max contract size limit. Should be deployed as a separate price providing contract. 

WIP


sparrowDom:
Closing this for now, can revisit if need be

PR #1757, opened 2023-08-10
Fork test env improvements

sparrowDom:
Significantly speeds up fork tests and fixes some other issues. 
Testing flux fork tests file ( 10 fork tests): 
|Node mode|First run time| Second run time|
|-----|-----|-----|
|standalone node, tests in second terminal|303s|45s|
|tests alone|395s|58s|

this branch also: 
- sets fork test block number from the hardhat context. This is important, since the block number is read **after** the deployment scripts have been applied. 
- mints 40 blocks (hardhat requirement) in 1 request instead of 40
- fixes a problem where governor contract delays are more correctly applied


naddison36 (approved):
Thanks for these improvements

PR #1768, opened 2023-08-20
Balancer off chain min BPT check

sparrowDom:
Do a dry-run of how much BPT should depositing liquidity yield, and then pass it along to VaultAdmin's `depositToStrategy`.  This PR also disables withdraw / deposit functions on Balancer strategy since this is never going to be a default strategy. 

Things left to do: 
- make it possible to withdraw using multiple assets with 1 transaction
- make it possible to pass min expected assets on withdrawal

TBD: 
we need a sanity check if strategist is hijacked. We could: 
- still calculate price of BPT on chain and leave ~2% of wiggle room when strategist defines expected BPT on deposit / assets on withdrawal
- limit a number of TX a strategist can execute in a give time interval -> so a circular transactions of depositing / withdrawing to/from strategy can't drain too much funds
- something else?

### DIFF
--- a/contracts/contracts/interfaces/IStrategy.sol
+++ b/contracts/contracts/interfaces/IStrategy.sol
@@ -19,6 +19,12 @@ interface IStrategy {
     function depositAll() external;
 
     /**
+     * @dev Deposit the entire balance of all supported assets in the Strategy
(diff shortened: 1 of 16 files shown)


sparrowDom:
We've decided to use the VaultValue checker and not rely on off chain calculations

PR #1774, opened 2023-08-22
Sparrow dom/balancer check balance check

sparrowDom:
- Check possible manipulations of check balance function. Tilting the pool musn't alter the amounts the function returns


sparrowDom:
This has been implemented in fork tests

PR #1778, opened 2023-08-23
Update collateral swap script

shahthepro:
Have tested the individual functions separately. But still have to test it as a whole after proposal execution. Also, one more thing leftover is the scaling of printed amounts in the `build_swap_tx` function (The actual amounts are scaled and works well, just the ones that get printed to console in human readable format needs to be updated)


naddison36 (approved):
Looks good

PR #1781, opened 2023-08-23
M03 - missing or misleading documentation

sparrowDom:
Todo: 
- [x] explain how reward tokens are collected
- [ ] improve getBptExpected docs
- [x] explain why read-only re-entrency is not required for deposits /withdrawals

**From Audit:**
Throughout the codebase, there are multiple sections which are difficult to understand, and
would benefit from additional or clearer documentation:
It was difficult to assert that reward tokens are actually rewarded to the strategy.
Consider adding more details around how this happens, especially for the AURA tokens
where the process is more complex and not intuitive to follow.
Given the intricacies of how Balancer works, as well as the need to wrap rebasing
tokens, it was difficult to follow the math behind the getBPTExpected functions.
Consider adding a practical example, as well as explanations regarding what each
variable means and what each price is measured in.
Consider documenting why the Balancer read-only reentrancy check is not needed on
deposit and withdrawal functions.
Additionally, while there was documentation around the addition of strategies using Balancer, it
was at times misleading as it was not specifically targeted towards the
BalancerMetaPoolStrategy , and hence would differ from the implementation that was
audited. Consider creating documentation specific to the strategy at hand, explicitly and
concisely explaining the design decisions behind each important piece.

### DIFF
--- a/contracts/contracts/strategies/balancer/BaseAuraStrategy.sol
+++ b/contracts/contracts/strategies/balancer/BaseAuraStrategy.sol
@@ -80,7 +80,9 @@ abstract contract BaseAuraStrategy is BaseBalancerStrategy {
         onlyHarvester
         nonReentrant
     {
-        // Collect BAL and AURA
+        /* Similar to Convex calling this function collects both of the
+         * accrued BAL and AURA tokens.
+         */
         IRewardStaking(auraRewardPoolAddress).getReward();
         _collectRewardTokens();
     }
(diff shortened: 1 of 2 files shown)


naddison36 (approved):

naddison36 (approved):

contracts/contracts/strategies/balancer/BaseAuraStrategy.sol
```diff
@@ -80,7 +80,9 @@ abstract contract BaseAuraStrategy is BaseBalancerStrategy {
         onlyHarvester
         nonReentrant
     {
-        // Collect BAL and AURA
+        /* Similar to Convex calling this function collects both of the
```
DanielVF, line 83: Needs a comma after "Similar to Convex".

DanielVF (approved):

contracts/contracts/strategies/balancer/BaseAuraStrategy.sol
sparrowDom: thanks, added

PR #1786, opened 2023-08-25
N02 gas inefficiencies 

sparrowDom:
notes: 
- I haven't converted the function signature of internal functions (_deposit, _withdrawal_, getBPTExpected) to use calldata. Since other functions can call those with in memory constructed arrays that can not be converted to calldata. 

Have also tested if storing memory array length into a uint256 provides any gas gains. Have ran it twice, with reading length each time inside the loop and storing the length separately. There seemed to be no differences. If the array would be a storage type array, then the gas gains should be more noticeable. In both cases the min/max/average gas usage were the same: 
- depositToStrategy        ·    1294871  ·    1320278  ·    1307819

**From audit report:**
There are several places across the codebase where changes can be made to improve gas
consumption.
The amount = 0; assignment is unnecessary.
There is another, more efficient version of the fromPoolAsset function that can be
used when the amount is not needed. Consider switching to it instead.
The assetAmount variable is unused, consider removing it.
In the withdraw function, the conversion to a pool asset depends only on
poolAssets[i] . Consider doing it only once per pool asset, outside of the for loop
that goes through the strategy assets.
The asset variable can be reused here, to avoid another array access through
_assets[i] .
There are several places where function parameters could be marked as calldata ,
consider changing them for future gas optimizations: deposit , _deposit ,
withdraw , _withdraw , and getBPTExpected .
There are multiple instances of an array's length being accessed on every iteration of a
loop. This can be made more efficient by saving off the length of the array in a local
variable and using that in the for loop instead.
When performing these changes, aim to reach an optimal trade-off between gas optimization
and readability. Having a codebase that is easy to understand reduces the chance of errors in
the future and improves transparency for the community.


naddison36 (approved):
Looks good

Issue #1545, opened 2023-05-25
Wrong calculation of sfrxETH price


rafaelugolini:
sfrxETH shows the same price in USD from frxETH

![Image](https://github.com/OriginProtocol/origin-dollar/assets/2651141/8a20f8de-b1f3-461b-af44-57e494710c4d)
![Image](https://github.com/OriginProtocol/origin-dollar/assets/2651141/866ebb85-ae9c-4189-a852-68edc94c3edf)

smitch88:
@rafaelugolini Just for clarity, the bug here is the USD price conversion of sfrxETH. I would imagine this is an issue across all the LSTs, right now this is just taking eth price and multiplying which would be incorrect.

smitch88:
@micahalcorn Take a look. Basically the USD value is going to be inaccurate for any LST

micahalcorn:
Here's another report that may be the same issue:

![IMG_5845](https://github.com/OriginProtocol/origin-dollar/assets/273937/c228fb3c-1e34-4619-89db-ae5fc635fa93)

PR #1788, opened 2023-08-28
Adds proper USD pricing conversion for token swap

smitch88:
#1545

* Adds in coin gecko token calls to get the correct token price mapping for our LSTs instead of using ETH price

**Examples**
Screenshot 2023-08-27 220203
Screenshot 2023-08-27 220157
Screenshot 2023-08-27 220137

### DIFF
--- a/dapp-oeth/src/hooks/useTokenPrices.js
+++ b/dapp-oeth/src/hooks/useTokenPrices.js
@@ -1,0 +1,115 @@
+import { useEffect, useState } from 'react'
+
+const tokenConfiguration = {
+  ethereum: {
+    id: 'ethereum',
+    symbol: 'eth',
+    name: 'Ethereum',
+  },
+  oeth: {
+    id: 'origin-ether',
+    symbol: 'oeth',
+    name: 'Origin Ether',
+    platforms: {
+      ethereum: '0x856c4efb76c1d1ae02e20ceb03a2a6a08b0b8dc3',
+    },
+  },
+  frxeth: {
+    id: 'frax-ether',
+    symbol: 'frxeth',
+    name: 'Frax Ether',
+    platforms: {
+      ethereum: '0x5e8422345238f34275888049021821e8e08caa1f',
+      'binance-smart-chain': '0x64048a7eecf3a2f1ba9e144aac3d7db6e58f555e',
+      'polygon-pos': '0xee327f889d5947c1dc1934bb208a1e792f953e96',
+      'arbitrum-one': '0x178412e79c25968a32e89b11f63b33f733770c2a',
+      'optimistic-ethereum': '0x6806411765af15bddd26f8f544a34cc40cb9838b',
+      'polygon-zkevm': '0xcf7ecee185f19e2e970a301ee37f93536ed66179',
+      moonbeam: '0x82bbd1b6f6de2b7bb63d3e1546e6b1553508be99',
+      fantom: '0x9e73f99ee061c8807f69f9c6ccc44ea3d8c373ee',
+    },
+  },
+  sfrxeth: {
+    id: 'staked-frax-ether',
+    symbol: 'sfrxeth',
+    name: 'Staked Frax Ether',
+    platforms: {
+      ethereum: '0xac3e018457b222d93114458476f3e3416abbe38f',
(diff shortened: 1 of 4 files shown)


rafaelugolini (dismissed):
I don't think we should coingecko for this, we should use the contracts to get the accurate price:

For stETH, rETH, frxETH: Use the [Oracle contract](https://etherscan.io/address/0x3ccd26e82f7305b12742fbb36708b42f82b61dba#readContract)
For sfrxETH: use [previewRedeem](https://etherscan.io/token/0xac3e018457b222d93114458476f3e3416abbe38f#readContract) and then call the Oracle contract
WETH: 1:1

smitch88:
> I don't think we should coingecko for this, we should use the contracts to get the accurate price:
> 
> For stETH, rETH, frxETH: Use the [Oracle contract](https://etherscan.io/address/0x3ccd26e82f7305b12742fbb36708b42f82b61dba#readContract) For sfrxETH: use [previewRedeem](https://etherscan.io/token/0xac3e018457b222d93114458476f3e3416abbe38f#readContract) and then call the Oracle contract WETH: 1:1

Looks good, will add this and see if works well. For dev environment, will just use coin gecko

dapp-oeth/src/hooks/useTokenPrices.js
```diff
@@ -0,0 +1,189 @@
...
+    },
+  },
+}
+
+const oethOraclePrice = async (contract, tokenAddress) => {
+  try {
+    return await contract.price(tokenAddress)
+  } catch (e) {
+    console.error(e)
+    return utils.parseEther('1')
```
rafaelugolini, line 67: Should we do this? I think I would rather have an error msg

smitch88: I'd rather the dapp still be functional than break on this error. It's just the USD value display and this shouldn't happen unless there is a bad network call

rafaelugolini: ah ok, I thought this was from the exchange rate of the tokens

dapp-oeth/src/hooks/useTokenPrices.js
```diff
@@ -0,0 +1,176 @@
...
+    return {}
+  }
+
+  // Fetch baseline ETH price for conversion
+  const feed = await contracts.chainlinkEthAggregator.latestRoundData()
+  const ethPrice = Number(utils.formatUnits(feed?.answer, 8))
+
+  // Fetch token ratios
+  const tokenToPricingMethod = {
+    frxeth: oethOraclePrice.bind(
```
toniocodo, line 82: long time no see...

dapp-oeth/src/hooks/useTokenPrices.js
```diff
@@ -0,0 +1,176 @@
...
+      setIsLoading(true)
+      setError(null)
+
+      let prices
+
+      if (chainId === 1) {
+        prices = await oraclePrices(queryTokens, contracts)
+      } else {
+        prices = await coingeckoPrices(queryTokens)
+      }
```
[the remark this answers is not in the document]
smitch88: Mainnet is using contract calls, any other would be using coin gecko. Matches our current dapp which fetches price from chain link only in mainnet and uses coin market api for everything else

dapp-oeth/src/hooks/useTokenPrices.js
```diff
@@ -0,0 +1,176 @@
...
+      setIsLoading(false)
+    }
+  }
+
+  useEffect(() => {
+    fetchTokenPrices()
+  }, [chainId, contracts])
+
+  return [{ data: prices, isLoading, error }, { onRefresh: fetchTokenPrices }]
+}
```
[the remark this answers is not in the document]
smitch88: Well this goes into 2 components so it's easiest to do the integration with a hook. Could also expose the fetch and put it in store on the mount after contracts are init.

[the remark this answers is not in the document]
toniocodo: you could also go for react-query to enclose it. That would give you caching, refetching, error handling,... for free

toniocodo (approved):
lgtm

PR #1793, opened 2023-08-29
feat: add dangerjs

rafaelugolini:
Add dangerjs to request 2 reviewers on contracts PRs

### DIFF
--- a/.github/workflows/danger.yml
+++ b/.github/workflows/danger.yml
@@ -1,0 +1,21 @@
+name: Node CI
+
+on:
+  pull_request:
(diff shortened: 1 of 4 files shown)


sparrowDom (approved):
LGTM

at some point if we include for example foundry, we might want to consider only `.sol` files in the contracts source folder. But no need to overengineer this right now IMHO

naddison36 (approved):

DanielVF (approved):

rafaelugolini:
looks like some tests are failing, should I bypass them?

PR #1795, opened 2023-08-29
M04 - minBptFunction robustness

sparrowDom:
Things in this PR: 
- `minBptExpected` function ignores oracle prices and considers only rate of rateProviders. Meaning that strategy is assuming that assets are pegged
- fork tests added to check that with pool manipulation checkBalance almost isn't affected
- fork tests added to check that with pool manipulation depositing is safe when using VaultValueChecker
- deposit functions (that would be used by Vault if this were ever an asset default strategy) disabled
- maxWithdrawal/DepositSlippage renamed to maxWithdrawal/DepositDeviation
 
**From audit report:**
The getBPTExpected functions are used to calculate the minimum amount of BPTs to
receive when depositing liquidity, or the maximum amount of BPTs to provide when
withdrawing liquidity. They calculate the underlying ETH value of the assets to deposit/
withdraw, and divide it by the price of a BPT in terms of the pool's underlying base asset.
There are two issues with this:
Based on which oracles are being used, the strategy's asset price might be stale, which
will result in an over/underestimation of the correct number of BPTs.
The calculation assumes normal market conditions, namely that the pool's underlying
base asset is trading at 1-1 parity with ETH. stETH or frxETH depegging from ETH
would make this untrue, resulting in an overestimation of the amount of BPTs.
Overestimating the BPT amount can result in reverts during deposits while underestimating
them can result in reverts during withdrawals. Consider accurately calculating the exact
number of BPTs required for these operations using Balancer's StableMath.


contracts/contracts/strategies/balancer/BaseBalancerStrategy.sol
```diff
@@ -270,17 +275,15 @@ abstract contract BaseBalancerStrategy is InitializableAbstractStrategy {
         virtual
         returns (uint256 bptExpected)
     {
-        // Get the oracle from the OETH Vault
-        address priceProvider = IVault(vaultAddress).priceProvider();
+        require(
+            _assets.length == _amounts.length,
+            "Assets & amounts length mismatch"
```
naddison36, line 280: Its good to keep strings to 31 or less characters so they fit in an EVM word (32 bytes). The extra byte is for the string length.
`Assets & amounts mismatch` will works

naddison36 (approved):

contracts/contracts/strategies/balancer/BaseBalancerStrategy.sol
sparrowDom: Oh right it is 31 not 32. Thanks corrected

contracts/contracts/strategies/balancer/BaseBalancerStrategy.sol
```diff
@@ -431,50 +431,50 @@ abstract contract BaseBalancerStrategy is InitializableAbstractStrategy {
...
+     * @notice Sets max deposit deviation that is considered when adding
      * liquidity to Balancer pools.
-     * @param _maxDepositSlippage Max deposit slippage denominated in
+     * @param _maxDepositDeviation Max deposit deviation denominated in
      *        wad (number with 18 decimals): 1e18 == 100%, 1e16 == 1%
      *
-     * IMPORTANT Minimum maxDepositSlippage should actually be 0.1% (1e15)
-     * for production usage. Contract allows as low value as 0% for confirming
-     * correct behavior in test suite.
+     * IMPORTANT Minimum maxDepositDeviation will be 1% (1e16) for production
```
DanielVF, line 464: *"will default to" (since we could change it later)

contracts/test/strategies/balancerPoolReentrancy.fork-test.js
```diff
@@ -29,7 +29,7 @@ forkOnlyDescribe(
       fixture = await loadFixture();
     });
 
-    it.only("Should not allow read-only reentrancy", async () => {
```
DanielVF, line 32: Important, lol.

contracts/contracts/strategies/balancer/BaseBalancerStrategy.sol
```diff
@@ -255,13 +264,9 @@ abstract contract BaseBalancerStrategy is InitializableAbstractStrategy {
...
     {
-        address priceProvider = IVault(vaultAddress).priceProvider();
-        uint256 strategyAssetMarketPrice = IOracle(priceProvider).price(_asset);
         uint256 bptRate = IRateProvider(platformAddress).getRate();
-
-        bptExpected = _amount
-            .mulTruncate(strategyAssetMarketPrice)
-            .divPrecisely(bptRate);
+        uint256 poolAssetRate = getRateProviderRate(_asset);
+        bptExpected = _amount.mulTruncate(poolAssetRate).divPrecisely(bptRate);
```
DanielVF, line 269: We don't have to change anything here but,

This: 

```solidity
bptExpected = _amount.mulTruncate(poolAssetRate).divPrecisely(bptRate);
```

is doing this:

```solidity
bptExpected = ((_amount * poolAssetRate / 1e18) * 1e18  / bptRate);
```

Which is the same as

```solidity
bptExpected = _amount * poolAssetRate / bptRate
```

This stable math library we are using is also using the safe math library under the hood, which adds unneeded overhead to the contracts.

Again, no change needed, just something to think about for the future.

contracts/test/strategies/balancerPoolReentrancy.fork-test.js
sparrowDom: yeah :)

contracts/contracts/strategies/balancer/BaseBalancerStrategy.sol
sparrowDom: thanks fixed

contracts/contracts/strategies/balancer/BaseBalancerStrategy.sol
sparrowDom: I haven't though of that, good point. The mulTruncate +  divPrecisely could cancel the 1e18 multiplication & division out.

We could do that change, though adding a comment would probably make sense, since the though process can easily be forgotten.

DanielVF:
I think the oracle portion of this comment may be out of date:

```
  // STEP 5 - Re-deposit any left over BPT tokens back into Aura
  /* When concluding how much of BPT we need to withdraw from Aura we rely on Oracle prices
   * and those can be stale (most ETH based have 24 hour heartbeat & 2% price change trigger)
   * After exiting the pool strategy could have left over BPT tokens that are not earning   
```

contracts/contracts/strategies/balancer/BalancerMetaPoolStrategy.sol
```diff
@@ -214,30 +215,13 @@ contract BalancerMetaPoolStrategy is BaseAuraStrategy {
         );
 
         for (uint256 i = 0; i < _strategyAssets.length; ++i) {
```
DanielVF, line 217: Fun to think about what would happen if we passed in the same asset multiple times in the _strategyAssets array and _strategyAmounts array.

I don't think there would be an issue. The last one would set the amount, but it would try to transfer each of the amounts out, which should fail, since only the last amount was withdrawn from the strategy.

contracts/contracts/strategies/balancer/BaseBalancerStrategy.sol
```diff
@@ -225,28 +225,37 @@ abstract contract BaseBalancerStrategy is InitializableAbstractStrategy {
...
+     * flash loan to manipulate the pool before a deposit/withdrawal since the function ignores
+     * market values of the assets being priced in BPT.
+     *
+     * At the time of writing there is no safe on-chain approach to pricing BPT in a way that it
+     * would make it invulnerable to MEV pool manipulation. See recent Balancer exploit:
+     * https://www.notion.so/originprotocol/Balancer-OETH-strategy-9becdea132704e588782a919d7d471eb?pvs=4#1cf07de12fc64f1888072321e0644348
      *
-     * Since we only have oracle prices for the unwrapped version of the assets the equation
-     * turns into:
+     * To mitigate MEV possibilities VaultValueChecker in conjunction with checkBalance will catch
```
DanielVF, line 242: This sentence could be "To mitigate MEV possibilities during deposits and withdraws, the VaultValueChecker will use checkBalance before and after the move to ensure the expected changes took place."

DanielVF (approved):

contracts/contracts/strategies/balancer/BalancerMetaPoolStrategy.sol
sparrowDom: Yes that is a great thought. Quickly glancing through the code I would expect this inner loop: https://github.com/OriginProtocol/origin-dollar/blob/sparrowDom/balancer-sfrxETH-stETH-rETH/contracts/contracts/strategies/balancer/BalancerMetaPoolStrategy.sol#L265-L293 to only consider the last entry in the `strategyAssets` & `_strategyAmount` of the duplicated assets to be considered. And it would withdraw too little from the balancer pool and pay too much in BPT tokens. 

And as you've said the ERC20 transfer out as the final step of the function should fail.

contracts/contracts/strategies/balancer/BaseBalancerStrategy.sol
sparrowDom: thanks much better!

sparrowDom:
thanks @DanielVF nice catch regarding the oracle comment. Have corrected that one as well.

PR #1797, opened 2023-08-29
N04 more consistant function naming

sparrowDom:
**From audit report:**
In the BaseBalancerStrategy some internal functions start with a leading underscore
while others do not.
The use of standard conventions can quickly convey information to developers and users alike.
Consider standardizing a naming convention for private and internal functions. A leading
underscore is recommended, as mentioned in the Solidity documentation.


naddison36 (approved):

DanielVF (approved):
me gusta

PR #1800, opened 2023-08-30
N05 array length

sparrowDom:
**From audit report:** 
In the _deposit function, the length of mappedAmounts and mappedAssets is initialized
to tokens.length . This is superfluous if only a subset of the Balancer pool's tokens are
supported by the strategy.
While this currently does not present a security risk, consider initializing the array to the correct
length.


DanielVF (approved):

DanielVF:
Should we break out of the inner loop when he hit the correct address?

image

Should we revert after the inner loop if we never hit a correct address?

sparrowDom:
good catch! Yes to both. Have added the break on match found and revert when no asset match can be found

PR #1812, opened 2023-09-07
fix: show error `reason` or `message` in bottom `SwapCurrencyPill`

apexearth:
Proposal to fix the stuck on 'Loading...' issue by showing the error seen in the calculation instead.

Screenshot 2023-09-07 at 1 47 19 PM

### DIFF
--- a/dapp-oeth/src/components/buySell/SwapCurrencyPill.js
+++ b/dapp-oeth/src/components/buySell/SwapCurrencyPill.js
@@ -478,6 +478,7 @@ const SwapCurrencyPill = ({
   selectedCoin,
   selectedSwap,
   swapsLoading,
+  swapsError,
   priceToleranceValue,
   swapMode,
   onErrorChange,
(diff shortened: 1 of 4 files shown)


apexearth:
FYI seems like my user chris@originprotocol.com on Heroku doesn't have access to the pipeline shown here.

toniocodo (approved):
LGTM, let's validate with Micah that he agrees with the error message location and styles, and good to go

rafaelugolini:
> LGTM, let's validate with Micah that he agrees with the error message location and styles, and good to go

@apexearth can we get the price once it's below peg?

rafaelugolini (approved):

apexearth:
> > LGTM, let's validate with Micah that he agrees with the error message location and styles, and good to go
> 
> @apexearth can we get the price once it's below peg?

Yeah this should start working once the price is within the proper bounds.

PR #1815, opened 2023-09-12
OETH Oracle

naddison36:
# Contract Changes

* Added `floorPrice` to the Vault to get the value (USD or ETH) of the collateral assets received from redeeming 1 Origin Token (OUSD or OETH) from the Vault.
* Added Chainlink style `OETHOracle` which is updated by `OETHOracleUpdater` that aggregates on-chain OETH/ETH prices.
* Added `price_oracle` to Curve interface which gets the EMA Oracle price

oethOracles

OETHOracleHierarchy

# Tests

Unit tests
```
yarn test ./test/oracle/oracle-oeth.js 
```

Fork tests
```
yarn test:fork ./test/oracle/oracle.fork-test.js
```

# Review

Contract change checklist:
  - [ ] Unit tests pass


contracts/contracts/vault/VaultCore.sol
```diff
@@ -565,6 +564,27 @@ contract VaultCore is VaultInitializer {
...
+        uint256[] memory redeemAssets = _calculateRedeemOutputs(1e18);
+
+        // For each of the redeemed assets
+        for (uint256 i = 0; i < redeemAssets.length; ++i) {
+            // Sum the value of the asset in ETH = asset amount * oracle price
+            price +=
+                redeemAssets[i] *
+                IOracle(priceProvider).price(allAssets[i]);
+        }
+        price = price / 1e18;
```
DanielVF, line 585: Makes a lot of sense to do the division once. Maybe a comment in the code on it?

contracts/contracts/interfaces/IVault.sol
```diff
@@ -120,6 +120,10 @@ interface IVault {
 
     function priceUnitRedeem(address asset) external view returns (uint256);
 
+    function floorPrice() external view returns (uint256);
+
+    function floorPrice2() external view returns (uint256);
```
DanielVF, line 125: Unused?

contracts/contracts/oracle/BaseOracle.sol
```diff
@@ -0,0 +1,202 @@
...
+        view
+        returns (
+            uint80 roundId,
+            int256 answer,
+            uint256 startedAt,
+            uint256 updatedAt,
+            uint80 answeredInRound
+        )
+    {
+        if (rounds.length <= _roundId) revert NoPriceData();
```
DanielVF, line 131: Is this check built into new solidity versions on storage array access? 

> Accessing an array past its end causes a failing assertion.

DanielVF:
✅ Looks like good design.

contracts/contracts/interfaces/IVault.sol
naddison36: I'm not sure where floorPrice2 came from. I've removed.

contracts/contracts/vault/VaultCore.sol
naddison36: good idea. Done

contracts/contracts/oracle/BaseOracle.sol
naddison36: Solidity will revert. This code is to throw the more usable `NoPriceData` error.

DanielVF:
Second pass review looks good.

DanielVF:
## Review part 1: Requirements

We want to provide an oracle for OETH for use on lending  platforms, both classic lending platforms and CDP style stablecoins, to allow them to determine the value of OETH as collateral.

Oracles for lending platforms and stablecoin CDP collateral have some core considerations:

**There is no safe direction for the oracle to fail to.**

If an oracle reports a price that is too low, then the borrower' collateral will report as insuffienct, and will be liquidated. This make angrey borrowers. If the reported price is extra low, then these liquidations could operate agaist the protocol as well, making the protocol itself insolvant. If prices are too high, then attackers will deposit  OETH, and borrow out more than its value in collateral, making the protocol insolvant.

**Oracles only have to be near the right answer.** 

Because lending platforms / CDP's usually have collateral ratios on ETH of 110%+, 5-8% off in either direction will not result in protocol level insolvancy. This is a pretty big amount of allowed error, and is a good thing. However, even small errors in the direction of low will result in people who were on the edge of liquidation getting liquidated. They would be mad, but at least it's not everybody.

**Transient errors are very bad.**

A single extreme bad value can cause a lot of reckage. Inherint to large lending platforms are a lot of liquidation bots in a constant race with each other. If bad values happen, the consequences can be as fast as the same block that the oracle update happened in.

### Proposed architecture

We will have a chainlink compatibile oracle contract, feed by an on chain updater contract using on chain data, callable by anyone. A chainlink keeper will backstop the calls.

This seems like a good architecure. Great compatibility, cheap reads. No hard requirement that oracle be updated by the origin team, and a good backstop as long as someone keeps feeding the keeper money.

### Proposed pricing strategy

Our current plan for getting new prices to use in updates is:

1. Use the curve price oracle if it is above the redeem fee.
2. If below the redeem fee, take the higher of the curve price and the redeem price.
3. Cap the result at one ETH.

How does this work in practice:

| **Protocol health**       | **Curve price** | **Behavior**                                                           |
|---------------------------|-----------------|------------------------------------------------------------------------|
| Healthy                   | Too high        | 🟢 Correctly capped at 1ETH                                            |
| Healthy                   | Too low         | 🟢 Will do approx 0.995ETH, which should be good.                      |
| Small Depegged Backing    | Too high        | 🟠 Will use curve price. Okay for small depegs.                        |
| Small Depegged Backing    | Too low         | 🟢 Will use vault floor. Accurate price.                               |
| Missing vault collateral  | Too high        | 🔴 Will be too high if curve is manipulated up.                        |
| Missing vault collateral  | Too low         | 🔴 Will assume that the vault has assets to redeem at the target price.|
| Vault reverts floorPrice()| Pegged          | 🟠 Oracle will use curve price. If we would be reverting, things prob bad though|
| Vault reverts floorPrice()| Below threshold | 🔴 Oracle will fail to update.                                         |
| Vault reverts redeems     | Above threshold | 🟠 Oracle will use curve price. If we would be reverting redeems, things prob bad though|
| Vault reverts redeems     | Below threshold | 🔴 We could be reverting becuase the protcool has lost funds or the price oracles are weird.|


Assuming our vault, strategies, and backing coins are healthy, this pricing strategy should work great for the intended use case. The cap prevents curve pool manipulation up, and any manipulation down is quickly limited by the vault's floor price. The 0.5% margin for error here is well within what lending platforms would tolerate with no problems. 

However, the pricing strategy could be broken if something were wrong.

1. If something went wrong, manipulating the curve price up would would bypass all vault level and collateral price checks. Also in case of either collateral level problems or OETH vault problems, the pool size is likely to shrink, making manipulation less costly. How easily and at what price could curve prices be manipulated back to peg?

2. If a backing collateral fell below to 0.70 ETH, then our vault oracles would revert. This both blocks redeems (good), but also reverts our floorPrice() method, causing our oracle to be unable to update at all, and effectively freezing at the last value for the duration of the problem.

3. floorPrice() always assumes the vault can redeem at the current quoted mix of coins. If the vault or an underlying backing strategy lost big funds, then floorPrice() would continue to report at current asset prices, while redeems would be unavailable (since we revert redeems if more than 3%? undercapitalized). For example, in the case that all balancer pools were hacked, and we were down 35% assets, this oracle would still report the price as near peg, even with the curve pool trading at low, accurate levels.

DanielVF:
## Review Part 2: Easy Checks

#### Authentication

- [x] Never use tx.origin

#### Ethereum

#### Cryptographic code

_No crypto_

#### Gas problems

    - [ ] A way to remove items

#### Black magic

#### Overflow

#### Proxy

#### Events

This behavior is okay. We emit events on the updater contract for updates, but not on the oracle contract itself. However, this matches chainlinks behavior, and all updates can be tracked down by getting round info.

contracts/contracts/oracle/OETHOracleUpdater.sol
```diff
@@ -0,0 +1,127 @@
...
+            uint256 answer,
+            uint256 vaultPrice,
+            uint256 marketPrice
+        )
+    {
+        // Get the aggregated market price from on-chain DEXs
+        marketPrice = _getMarketPrice();
+
+        // If market price is above the vault price with the withdraw fee
+        if (marketPrice > MAX_VAULT_PRICE) {
```
DanielVF, line 76: `>=` would be more efficient here, assuming `MAX_VAULT_PRICE` is truly the max the vault can report. If the vault actually did report `MAX_VAULT_PRICE` and the `marketPrice` was also `MAX_VAULT_PRICE`, then we would get the same answer checking the vault as not checking the vault.

contracts/contracts/oracle/OETHOracleUpdater.sol
```diff
@@ -0,0 +1,127 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import { IOracleReceiver } from "./IOracleReceiver.sol";
+import { IVault } from "../interfaces/IVault.sol";
+import { AggregatorV3Interface } from "../interfaces/chainlink/AggregatorV3Interface.sol";
```
DanielVF, line 6: I don't think AggregatorV3Interface is used?

contracts/contracts/oracle/OETHOracleUpdater.sol
```diff
@@ -0,0 +1,127 @@
...
+    /// @dev Callable by anyone as the prices are sourced and aggregated on-chain.
+    /// @param oracle Address of the Oracle that has authorized this contract to add prices.
+    function addPrice(IOracleReceiver oracle) external {
+        (
+            uint256 answer,
+            uint256 vaultPrice,
+            uint256 marketPrice
+        ) = _getPrices();
+
+        emit AddPrice(answer, vaultPrice, marketPrice);
```
DanielVF, line 53: This event fires regardless of which oracle was updated. Someone could call a different success only oracle contract and confuse our analytics about when our oracle has been updated. We should probably include which oracle was updated in the event.

contracts/contracts/oracle/OETHOracleUpdater.sol
naddison36: good point. I'll update

contracts/contracts/oracle/OETHOracleUpdater.sol
naddison36: AggregatorV3Interface  is not used. I'll remove it.

Apparently Slither can detect unused imports. I'll see why it's not detecting ours.

contracts/contracts/oracle/OETHOracleUpdater.sol
naddison36: excellent point. I'll add an indexed oracle address to the AddPrice event.

contracts/contracts/oracle/OETHOracleUpdater.sol
DanielVF: Yeah, I was wondering how this got through the automated tools. I should have chased this down more.

DanielVF:
## Review Part 3: Medium Checks

#### Rounding

#### Dependencies

#### External calls

- [x] Contract addresses passed in are validated 🟫 OracleUpdader#addPrice makes a call to an arbitrary address. This should be okay from a system security point of view, since this is the last thing the code does, and nothing in our system will take any action or depend on any side effects. 🔴 However the event fires regardless of if the correct oracle was updated. This could confuse our analytics. We should probably include the updated oracle in the event. _Added comment in code._
- [ ] Reentrancy guards on all state changing functions. 🟫 OracleUpdader#addPrice only performs external reads before the final write. There should be no room for reentrancy. 🟫 OracleBase#addPrice performs no external calls.
- [x] No Malicious behaviors
- [ ] Oracles, one of: 🟫 See dicussion in requirements. We do use an oracle, and it can't hurt us in normal operations. What happens if something is wrong, and someome is trying to manipulate it requires more analysis.
  - [ ] No oracles
  - [ ] Oracles can't be bent

#### Deploy

DanielVF:
@naddison36, got any thoughts on how we should handle bad vault situations? 

Only thing I've thought of is that we could factor in the difference between the vault total supply and the OUSD supply, if it's less than one. Would add a ton of gas to the oracle update, but in theory we only need to actually call the update if there's been a big price move, and we only the pay the cost if we are off peg.

sparrowDom:
There is some good discussion here. Regarding the bad Vault State: 
- in case the backing collateral's falls below 0.70 ETH/USDT it would make sense to me to still allow our `floorPrice` function to update and not revert (the behaviour for redeems stays the same). This would at least require change in the OracleRouter, to return the price even in cases where there is more than 30% deviation. 
- Thinking what to do in the other case where underlying strategies are being hacked and not able to report the correct value in `checkBalance`. If we can't really rely on the Vault to report the correct value and also not on the Curve pool (easy to manipulate) what is really a trusted source in that case? Can we assume Curve pool being manipulated if latest price and e.g. price X minutes ago diverge too much? We could also take an average of the last X minute price of the Curve pool and consider that as the "closest to truth" with some resistance being manipulated by tilting the pool.

contracts/contracts/oracle/BaseOracle.sol
```diff
@@ -0,0 +1,202 @@
...
+    function setOracleUpdater(address _oracleUpdater) external onlyGovernor {
+        _setOracleUpdater(_oracleUpdater);
+    }
+
+    /***************************************
+                Metadata
+    ****************************************/
+
+    /// @notice The number of decimals in the Oracle price.
+    function decimals()
```
sparrowDom, line 67: This could also be expressed as: 
```
uint8 public constant override decimals = 18;
```

contracts/test/oracle/oracle.fork-test.js
```diff
@@ -54,4 +60,126 @@ describe("ForkTest: Oracle Routers", function () {
...
+      expect(await oethOracleUpdater.MAX_VAULT_PRICE()).to.eq(
+        parseUnits("0.995")
+      );
+      expect(await oethOracleUpdater.curvePool()).to.eq(
+        addresses.mainnet.CurveOETHMetaPool
+      );
+      expect(await oethOracleUpdater.vault()).to.eq(oethVault.address);
+
+      expect(await oethOracle.decimals()).to.eq(18);
+      expect(await oethOracle.version()).to.eq(1);
```
sparrowDom, line 79: would be cool to have a test that maps deployed implementation addresses to versions. And when we attempt to update the Oracle and forget to also bump the version that test would fail.

sparrowDom (commented):
@naddison36 do you mind pinging when you implement the changes we have discussed today and I'll finish my full review?

contracts/contracts/vault/VaultCore.sol
```diff
@@ -565,6 +564,34 @@ contract VaultCore is VaultInitializer {
...
+        uint256[] memory redeemAssets = _calculateRedeemOutputs(1e18);
+
+        // For each of the redeemed assets
+        for (uint256 i = 0; i < redeemAssets.length; ++i) {
+            // Sum the value of the vault asset = asset amount * oracle price
+            // For OUSD's USDC and USDT assets that are to 6 decimals, the oracle
+            // price is to 18 decimals, so we do not need to scale them up to 18 decimals
+            price +=
+                redeemAssets[i] *
+                IOracle(priceProvider).price(allAssets[i]);
```
sparrowDom, line 588: Leaving a note here so it is not forgotten:  we query oracle price per asset in this line above and also in [_toUnitPrice](https://github.com/OriginProtocol/origin-dollar/blob/master/contracts/contracts/vault/VaultCore.sol#L668) function withing the `_calculateRedeemOutputs `

contracts/contracts/oracle/OETHOracleUpdater.sol
```diff
@@ -0,0 +1,131 @@
...
+            // Get the price from the Vault. This includes the withdraw fee
+            // and the vault collateral assets priced using oracles
+            vaultPrice = vault.floorPrice();
+
+            if (marketPrice > vaultPrice) {
+                // Return the market price with the Vault price as the floor price
+                answer = marketPrice;
+            } else {
+                // Return the vault price
+                answer = vaultPrice;
```
sparrowDom, line 95: And I think we've said that if market price is lower than the vaultPrice, then we might go with the market price.

Thinking more about this, I am not so sure this is a good idea, since the market price can be manipulated downwards even though there is a smaller financial incentive to do it. The way it is in the codebase right now seems safer to me.

probably something we should discuss more about.

sparrowDom (commented):
Eyeballed the PR again and left some extra comments

DanielVF:
After some analysis, we decided to not use this oracle approach. We did end up discovering a bug in Curve's oracle implementation though.

PR #1826, opened 2023-09-21
Start of strategy charting tool

DanielVF:
Tool to exercise strategies from the outside and chart the results.

image

The long term idea is to be able to make a small definition class on how to manipulate the strategy, and then have the same reporting code be able to run.

To use:

```
# Have a node:fork already running, then
brownie run strategy_report run_simulations BalancerRethEth
brownie run strategy_report run_report BalancerRethEth
```

The code currently has plenty of room for refactoring, but I figure we'll do that as we add our second strategy to testing.

### DIFF
--- a/brownie/scripts/strategy_report.py
+++ b/brownie/scripts/strategy_report.py
@@ -1,0 +1,434 @@
+import pandas as pd
+import numpy as np
+import matplotlib.pyplot as plt
+import re
+from pathlib import Path
+from brownie import *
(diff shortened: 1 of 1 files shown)


sparrowDom (commented):
This version doesn't balance the pool if it is imbalanced from the start right?

DanielVF:
> This version doesn't balance the pool if it is imbalanced from the start right?

It does not.

PR #1839, opened 2023-09-26
Balancer Composable stable pool strategy

sparrowDom:
Adds support for Balancer's Composable stable pools. 

Connected PR: https://github.com/OriginProtocol/origin-dollar/pull/1849


contracts/contracts/strategies/balancer/BaseBalancerStrategy.sol
```diff
@@ -477,13 +499,43 @@ abstract contract BaseBalancerStrategy is InitializableAbstractStrategy {
...
+     *
+     * @param _asset Address of the Balancer pool asset
+     * @return rate of the corresponding asset
+     */
     function _getRateProviderRate(address _asset)
         internal
         view
         virtual
-        returns (uint256);
+        returns (uint256)
```
DanielVF, line 519: Does a local cache work for this pool type?

contracts/contracts/strategies/balancer/BalancerMetaPoolStrategy.sol
```diff
@@ -448,6 +523,17 @@ contract BalancerMetaPoolStrategy is BaseAuraStrategy {
                 address(this)
             );
 
+            if (strategyAsset == frxETH && poolAssetAmount > 0) {
+                /* _unwrapPoolAsset internally increases the sfrxEth amount by 1 due to
```
DanielVF, line 527: Thanks for the explanation here.

contracts/contracts/strategies/balancer/BalancerMetaPoolStrategy.sol
```diff
@@ -8,14 +8,32 @@ pragma solidity ^0.8.0;
...
+     * 
+     * enum ExitKind { EXACT_BPT_IN_FOR_ONE_TOKEN_OUT, EXACT_BPT_IN_FOR_TOKENS_OUT, BPT_IN_FOR_EXACT_TOKENS_OUT }
+
+     * For Composable stable pools using IBalancerVault.WeightedPoolExitKind is not
+     * ok since the enum values are in different order as they are in MetaStable pools.
+     * From the pool code: 
+     * 
+     * enum ExitKind { EXACT_BPT_IN_FOR_ONE_TOKEN_OUT, BPT_IN_FOR_EXACT_TOKENS_OUT, EXACT_BPT_IN_FOR_ALL_TOKENS_OUT }
+     */
+    uint256 internal balancerBptInExactTokensOutIndex;
```
DanielVF, line 28: Could these be immutable set in a constructor?

contracts/contracts/strategies/balancer/BaseBalancerStrategy.sol
```diff
@@ -477,13 +499,43 @@ abstract contract BaseBalancerStrategy is InitializableAbstractStrategy {
...
+    {
+        IBalancerPool pool = IBalancerPool(platformAddress);
+        IRateProvider[] memory providers = pool.getRateProviders();
```
shahthepro, line 522: Any reason for not doing `IRateProvider[] memory providers = IBalancerPool(pool).getRateProviders();` since `pool` is used only once?

contracts/contracts/strategies/balancer/BaseBalancerStrategy.sol
```diff
@@ -375,11 +390,18 @@ abstract contract BaseBalancerStrategy is InitializableAbstractStrategy {
...
-                amount,
+        } else if (asset == frxETH && amount > 0) {
+            /* redeem takes amount parameter that are the number of
+             * shares (sfrxETH) to be exchanged for assets (frxETH)
+             */
+            unwrappedAmount = IERC4626(sfrxETH).redeem(
+                /* adding + 1 here since a rounding error can 
+                 * return value that is off by 1 WEI
+                 */
+                amount + FRX_ETH_REDEEM_CORRECTION,
```
shahthepro, line 404: nit: I'm just curious, how are other platforms/protocols dealing with this issue?

contracts/contracts/strategies/balancer/BaseBalancerStrategy.sol
```diff
@@ -375,11 +390,18 @@ abstract contract BaseBalancerStrategy is InitializableAbstractStrategy {
         internal
         returns (uint256 unwrappedAmount)
     {
-        if (asset == stETH) {
+        if (asset == stETH && amount > 0) {
```
shahthepro, line 393: nit: I feel like this could changed to an nested `if` structure to not evaluate `amount > 0` multiple times (but the gas cost saved might be very insignificant):
```
if (amount > 0 ) {
    if (asset == stETH) {
     ...
     } else if (asset == frxETH) {
     ...
     }
}

File: contracts/contracts/strategies/balancer/BaseBalancerStrategy.sol (hunk 1 above)
Comment (sparrowDom) [reply contracts/contracts/strategies/balancer/BaseBalancerStrategy.sol:532]: Yes thanks for reminding me. Cached in this commit: [8c8732004095f7c29ea7b569f6a45792f3dce62b](https://github.com/OriginProtocol/origin-dollar/pull/1839/commits/8c8732004095f7c29ea7b569f6a45792f3dce62b)

File: contracts/contracts/strategies/balancer/BaseBalancerStrategy.sol (hunk 5 above)
Comment (sparrowDom) [reply contracts/contracts/strategies/balancer/BaseBalancerStrategy.sol:406]: I have yet to investigate that. Being 1 WEI off is something we've seen with balancer as well. Protocols usually round in the direction that serves the protocol that owns the code. To mitigate flash loan attacks as best as possible.

File: contracts/contracts/strategies/balancer/BalancerMetaPoolStrategy.sol (hunk 3 above)
Comment (sparrowDom) [reply contracts/contracts/strategies/balancer/BalancerMetaPoolStrategy.sol]: Thanks, very good point. I should have though of that... It clears up code nicely and since each implementing contract will have this populated at creation time, there is no storage slot state/collisions to check. 

fix commit: https://github.com/OriginProtocol/origin-dollar/commit/c21d6f44b1342f75b14af5e5bb2504e4363113a9

File: contracts/contracts/strategies/balancer/BaseBalancerStrategy.sol (hunk 4 above)
Comment (sparrowDom) [reply contracts/contracts/strategies/balancer/BaseBalancerStrategy.sol]: Thanks both very good points. Have reduced the code to 1 line. The caching is done in this commit: https://github.com/OriginProtocol/origin-dollar/pull/1839/commits/8c8732004095f7c29ea7b569f6a45792f3dce62b

(the rate providers of a pool are not expected to change)

File: contracts/contracts/strategies/balancer/BaseBalancerStrategy.sol (hunk 6 above)
Comment (sparrowDom) [reply contracts/contracts/strategies/balancer/BaseBalancerStrategy.sol:395]: Yup good point. I've also though about that and decided to go for a little more clarity with an arguably worse gas consumption.

File: contracts/contracts/strategies/balancer/BalancerComposablePoolStrategy.sol [hunk 7]
```diff
@@ -0,0 +1,80 @@
...
+    function _assetConfigVerification(address[] calldata _assets)
+        internal
+        view
+        override
+    {
+        IERC20[] memory poolAssets = _getPoolAssets();
+
+        require(
+            // aside from BPT token all assets must be supported
+            poolAssets.length - 1 == _assets.length,
```
shahthepro, line 40: nit: Minor gas savings by storing `_assets.length` in a local variable. Same with `_amounts.length` in the following functions as well

contracts/deploy/071_balancer_rETH_WETH.js
```diff
@@ -66,13 +66,18 @@ module.exports = deploymentWithGovernanceProposal(
     );
 
     // 3. Encode the init data
-    const initFunction = "initialize(address[],address[],address[])";
+    // TODO change back
```
shahthepro, line 69: nit: Remove all these TODO comments?

shahthepro (dismissed):

contracts/deploy/071_balancer_rETH_WETH.js
sparrowDom: yes thanks

contracts/contracts/strategies/balancer/BaseBalancerStrategy.sol
```diff
@@ -34,8 +41,19 @@ abstract contract BaseBalancerStrategy is InitializableAbstractStrategy {
     uint256 public maxWithdrawalDeviation;
     // Max deposit deviation denominated in 1e18 (1e18 == 100%)
     uint256 public maxDepositDeviation;
+    // Rate providers of the Balancer pool
+    IRateProvider[] public poolRateProvidersCache;
```
DanielVF, line 45: You could check this, but I think a dynamic array reads the length storage slot first to bounds check, then the data slot.

That said, this is clean, and should be a lot less gas than what we were doing. Probably good to keep the clean way.

contracts/contracts/strategies/balancer/BaseBalancerStrategy.sol
```diff
@@ -481,13 +510,62 @@ abstract contract BaseBalancerStrategy is InitializableAbstractStrategy {
...
 
+    /**
+     * @notice Returns the rate supplied by the Balancer configured rate
+     * provider. Rate is used to normalize the token to common underlying
+     * pool denominator. (ETH for ETH Liquid staking derivatives).
+     *
+     * @param _asset Address of the Balancer pool asset
+     * @return rate of the corresponding asset
+     */
     function _getRateProviderRate(address _asset)
```
DanielVF, line 539: Does this method "work" and return a possibly spurious value for assets which are not supported? 

`poolAssetIndex` is mapping, so a bad _asset value should return a zero. This zero would then be used as a valid first index into the rate provider, so everything that's not valid, non-first pool assets would return the rate of the first asset?

We would need to make sure then that only valid addresses are fed into this method.

----

This current method does three storage slot reads.

1. poolAssetIndex mapping lookup
2. poolRateProvidersCache length bounds check
3. poolRateProvidersCache array lookup

But if this method is the only method that uses the `poolRateProvidersCache`, then we could trim this down to one read by making `poolRateProvidersCache` an address => address mapping from the _asset directly to the rate provider.

Also, if we used a special value (say `0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE`) for a fixed exchange rate, then we could revert on unsupported addresses, instead of silently continuing with them.

naddison36 (dismissed):
Looks great. Nice clean code that is well documented. Well done

contracts/contracts/strategies/balancer/BaseBalancerStrategy.sol
```diff
@@ -481,13 +510,62 @@ abstract contract BaseBalancerStrategy is InitializableAbstractStrategy {
 
     function _approveBase() internal virtual {
         IERC20 pToken = IERC20(platformAddress);
+
         // Balancer vault for BPT token (required for removing liquidity)
-        pToken.safeApprove(address(balancerVault), type(uint256).max);
+        // slither-disable-next-line unused-return
+        pToken.approve(address(balancerVault), 0);
```
shahthepro, line 516: nit: Any reason for removing the usage of `safeApprove` but still doing the same thing it was gonna do (i.e. set to 0 before setting it to another value)??

shahthepro (approved):

contracts/contracts/strategies/balancer/BaseBalancerStrategy.sol
sparrowDom: awesome thank you, this is a great find and a good proposal

sparrowDom: So this has moved into a separate PR because it has more profound changes that enable testing of the internal functions: https://github.com/OriginProtocol/origin-dollar/pull/1877

contracts/contracts/strategies/balancer/BalancerMetaPoolStrategy.sol
```diff
@@ -8,24 +8,59 @@ pragma solidity ^0.8.0;
...
+     * 
+     * enum ExitKind { EXACT_BPT_IN_FOR_ONE_TOKEN_OUT, EXACT_BPT_IN_FOR_TOKENS_OUT, BPT_IN_FOR_EXACT_TOKENS_OUT }
+
+     * For Composable stable pools using IBalancerVault.WeightedPoolExitKind is not
+     * ok since the enum values are in different order as they are in MetaStable pools.
+     * From the pool code: 
+     * 
+     * enum ExitKind { EXACT_BPT_IN_FOR_ONE_TOKEN_OUT, BPT_IN_FOR_EXACT_TOKENS_OUT, EXACT_BPT_IN_FOR_ALL_TOKENS_OUT }
+     */
+    uint256 internal immutable balancerBptInExactTokensOutIndex;
```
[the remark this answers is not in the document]
sparrowDom: good point, making the 2 public

contracts/contracts/strategies/balancer/BaseBalancerStrategy.sol
sparrowDom: I've removed it because it is deprecated: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/release-v4.4/contracts/token/ERC20/utils/SafeERC20.sol#L39-L58

Here is some more data on why that is so: https://github.com/OpenZeppelin/openzeppelin-contracts/issues/2219

Also safeApprove didn't actually set the balance to 0 and then to the requested one, rather just checked that the allowance was non zero before setting it to some non zero value.

shahthepro (dismissed):
There're some failing tests, other than that it looks good ot me

### Deployment Considerations

#### Internal State

- [x] Two new constants are added at the top of the existing contract (`BaseBalancerStrategy`), however since they're constant they don't affect the storage layout
- [x] `BaseBalancerStrategy.__reserved` is updated correctly after adding three new state variables

#### Potential Attacks/Edge Cases

- I suspected that doing multiple `withdrawAll` and `depositAll` back to back would cause `withdrawAll` to fail at some point (due to the usage of `FRX_ETH_REDEEM_CORRECTION`) however it turned out to be not true due a few other checks we have that ensures that there would no dust left. Besides we always resort to `faultTolerantWithdrawAllFromStrategies` when/if `withdrawAll` fails


#### Logic

  - [x] _Are there bugs in the logic?_ No


#### Tests
  - [ ] Some fork and unit tests are failing on CI

#### Flavor

_Could this code be simpler?_

#### Overflow

#### Proxy

#### Black magic

- [ ] (If an implementation contract were to call delegatecall under attacker control, it could call selfdestruct the implementation contract, leading to calls through the proxy silently succeeding, even though they were failing.)

#### Dependencies

#### Deploy
- [ ] Check that any deployer permissions are removed after deploy

#### Authentication

- [x] Never use tx.origin

#### Cryptographic code

#### Gas problems

    - [ ] A way to remove items

#### External calls

- [x] No malicious behaviors
- [x] Should not fail from stack depth problems (low level calls must require success)
- [ ] Oracles, one of:
  - [ ] No oracles
  - [ ] Oracles can't be bent


#### Ethereum

contracts/contracts/strategies/balancer/BalancerComposablePoolStrategy.sol
```diff
@@ -0,0 +1,135 @@
...
+        bptTokenPoolPosition = _bptTokenPoolPosition;
+    }
+
+    /*
+     * @dev Value that represents exit encoding where for max BPT given
+     * request exactly specifies the amount of underlying assets
+     * to be returned.
+     */
+    function _btpInExactTokensOutIndex()
+        internal
```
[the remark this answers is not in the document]
sparrowDom: This is a solidity thing where you [can not override constants](https://stackoverflow.com/questions/69931208/solidity-inheritance-override-public-constant). Interestingly (written in the same link) you can override a public function in a child contract using an overridden constant if the signatures and return types match.

contracts/contracts/strategies/balancer/BalancerMetaPoolStrategy.sol
```diff
@@ -227,72 +285,66 @@ contract BalancerMetaPoolStrategy is BaseAuraStrategy {
...
+            );
+
+            if (poolAssetAmount > 0) {
+                strategyAssetsToPoolAssetsAmounts[i] = poolAssetAmount;
+
+                /* This check is triggered when the _withdrawal is called with
+                 * a duplicate asset in the _strategyAssets array
+                 */
+                require(
+                    poolAssetsAmountsOut[poolAssetIndex[poolAsset]] == 0,
```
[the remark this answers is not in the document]
sparrowDom: This is a great catch. Thanks for the gas optimization tip. done here: https://github.com/OriginProtocol/origin-dollar/pull/1839/commits/8fb75ec3d8451937717d5b98670ddbdb368f293f

contracts/contracts/strategies/balancer/BalancerMetaPoolStrategy.sol
```diff
@@ -385,47 +437,80 @@ contract BalancerMetaPoolStrategy is BaseAuraStrategy {
...
+    }
 
+    function _withdrawAll(bool isRecoveryModeWithdrawal) internal {
+        // store poolAssets storage variable into memory for gas optimizations
+        address[] memory _poolAssets = poolAssets;
+
+        // STEP 1 - Withdraw all Balancer Pool Tokens (BPT) from Aura to this strategy contract
         _lpWithdrawAll();
         // Get the BPTs withdrawn from Aura plus any that were already in this strategy contract
         uint256 BPTtoWithdraw = IERC20(platformAddress).balanceOf(
```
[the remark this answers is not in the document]
sparrowDom: yes uncapitalized is a better way to name it thanks, changed. 

Regarding withdrawAll, we can withdraw on an empty strategy and it doesn't revert. Tested [here](https://github.com/OriginProtocol/origin-dollar/pull/1839/files#diff-7a4d77b5a120a2d6ec5e2c2e4fee6cbf48b03fe8803b35e597a1558a2947cc12R571-R572) That said there is no need to spend gas when there is nothing to withdraw.  Test, renames and gas optimisation done here: https://github.com/OriginProtocol/origin-dollar/pull/1839/commits/6dbb9cd4e1360a70d593167dc7d371b686612b93

contracts/contracts/strategies/balancer/BalancerComposablePoolStrategy.sol
```diff
@@ -0,0 +1,135 @@
...
+    }
+
+    function _getUserDataEncodedAmounts(uint256[] memory _amounts)
+        internal
+        view
+        virtual
+        override
+        returns (uint256[] memory amounts)
+    {
+        // remove the position in the array that represents the BPT
```
pandadefi, line 113: I am still learning about this strategy and the parent strategies. I would like to learn more about why the BPT token amount gets removed.

sparrowDom: Hey sure, so the [Composable Stable Pools](https://docs.balancer.fi/concepts/pools/composable-stable.html#what-does-composable-mean)  have this particularity that they have the BPT token positioned next to other pool assets. This unlocks the ability of more easily nesting / batch swapping with other Balancer pools. 

When joining/exiting a Balancer pool (add liquidity / remove liquidity in Curve's terms) [encoded userData](https://github.com/OriginProtocol/origin-dollar/pull/1839/files/a3b31971cfa89c6dc38ab535b7dfb3a021c767d6#diff-87b55ced2ebd9369aef1aeef1ea49285185a8fda30d2d19081a6a3a6b0378b72R372) needs to be passed along the request that encodes expected tokens received after the join/exit. And that function expects the amounts not to have the BPT token amount present in the array (even though it is one of the assets in the Composable Stable Pool). 

Additionally when we are calculating the [amount of BPT expected to be received from withdrawing](https://github.com/OriginProtocol/origin-dollar/pull/1839/files/a3b31971cfa89c6dc38ab535b7dfb3a021c767d6#diff-87b55ced2ebd9369aef1aeef1ea49285185a8fda30d2d19081a6a3a6b0378b72R346) given assets the BPT token needs to be omitted from the calculation. Since we are swapping the BPT token in exchange for the pool assets. 

Meta stable pools (predecessors of Composable Stable Pools) didn't have the BPT token present as one of the pool's assets and thus the exclusion of BPT from the join/exit requests wasn't necessary. 

Does that make sense?

contracts/contracts/strategies/balancer/BalancerComposablePoolStrategy.sol
```diff
@@ -0,0 +1,135 @@
...
+            "Pool assets length mismatch"
+        );
+
+        require(
+            // BPT position should be correctly configured
+            poolAssets[bptTokenPoolPosition] == platformAddress,
+            "BPT token position incorrect"
+        );
+
+        for (uint256 i = 0; i < _assets.length; ++i) {
```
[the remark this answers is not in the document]
sparrowDom: That is a great point thanks. I have just read up on this how; Solidity (after 0.8.0) doesn't need to perform overflow checks in situations where no overflow is guaranteed. And looping through arrays definitely is one of those situations.

The unchecked does provide some gas savings with a little minor impact on code readability. In this case I would rather not make the change, since first loop is done only on initialisation. And the other 2 are done on join/exit pools. We are running those operations relatively rarely - and are paying for the gas ourselves.

If these optimisations would be part of `checkBalance` function ( that gets called in Vault's allocate/rebase that are called in most of mints & redeems) then I'd gladly trade the readability impact for the gas savings.

naddison36:
Closing as we are no longer doing this strategy

PR #1849, opened 2023-10-03
Add generalised pool tilt functionality

sparrowDom:
- add generalized framework to tilt/untilt pools (Balancer supported now, could support Curve in the future)
- add MEV tests for Composable stable pool


contracts/test/fixture/_pool_tilt.js
```diff
@@ -0,0 +1,142 @@
...
+
+  await pool.tiltPool(tiltAmount, attackAsset, sAttacker);
+}
+
+/* Generic tilt pool function
+ *
+ */
+async function unTiltPool({
+  // eslint-disable-next-line no-unused-vars
+  fixture,
```
shahthepro, line 125: `fixture` and `poolContract` seem to be unused, so can remove them right?

contracts/test/fixture/_pool_tilt.js
```diff
@@ -0,0 +1,142 @@
...
+  /* amount of attacking asset in relation to pool TVL. 100 == 100% of TVL meaning
+   * a pool with 1m TVL will get tilted by 1m of attacking asset.
+   */
+  tiltTvlFactor = 100,
+  attackAsset, // asset used to tilt the pool
+  poolContract,
+}) {
+  const poolConfig = await getPoolConfiguration(fixture, poolContract);
+  pool = await factoryCreatePool(poolConfig.config);
```
shahthepro, line 100: Using a global variable works, but we will be having issues when we want to parallelize these tests. Also, I feel like there should be cleaner way to do this.

Perhaps have this `tiltPool()` function return a function that'll untilt the pool?? Something like this:
```
function tiltPool() {
    ... 
    return () => untiltPool(pool)
}
```


On the test file:

```
    const untilt = tiltPool()
    ...
    // Do test thing here
    ...
    untilt()
```

contracts/test/fixture/_pool_tilt.js
```diff
@@ -0,0 +1,142 @@
...
+    };
+  }
+}
+
+async function _fundAttackerOption({
+  asset, // attacking asset
+  assetAmount, // required funded asset amount to perform tilt
+  fixture,
+}) {
+  const attackerEthBalance = await waffle.provider.getBalance(attackerAddress);
```
shahthepro, line 60: `hre.network.provider` or `hre.ethers.provider` should work. `waffle` is an unused dependency we should get rid of 😅

contracts/test/fixture/_pools.js
```diff
@@ -0,0 +1,158 @@
...
+      ]
+    );
+
+    if (this.preTiltTVL) {
+      await this.verifyTiltAmounts(this.preTiltTVL);
+    }
+  }
+
+  async getAssetIndex(asset) {
+    const tokens = (await this.balancerVault.getPoolTokens(this.poolId)).tokens;
```
shahthepro, line 122: nit: Can probably move this to the constructor or make `memoize` this function. But hardhat caches the response anyway, so it shouldn't affect the performance much by doing it this way

contracts/test/fixture/_pools.js
```diff
@@ -0,0 +1,158 @@
...
+  if (config.platform === "balancer") {
+    return new BalancerPool(
+      config.poolId,
+      config.poolType,
+      config.assetAddressArray,
+      config.bptToken,
+      config.balancerVault
+    );
+  }
+  // else if platform === "curve"...
```
shahthepro, line 153: nit: probably a good idea to throw an error here, would be easy to figure out what's wrong when this happens

contracts/test/fixture/_fixture.js
```diff
@@ -1,20 +1,20 @@
-const hre = require("hardhat");
+<const hre = require("hardhat");
```
shahthepro, line 1: Unwanted `<` here causes issues 😅

contracts/test/fixture/_pool_tilt.js
```diff
@@ -0,0 +1,142 @@
...
+const { factoryCreatePool } = require("./_pools");
+
+const attackerAddress = "0xd8dA6BF26964aF9D7eEd9e03E53415D37aA96045";
+let sAttacker;
+let pool;
+
+async function getPoolConfiguration(fixture, pool) {
+  const { rEthBPT, sfrxETHwstETHrEthBPT } = fixture;
+
+  if (!!rEthBPT && pool.address == rEthBPT.address) {
```
shahthepro, line 18: Could use the `?.` operator here like `pool.address == rEthBPT?.address`.

But the only thing that's dynamically fetched here seems to be the `balancerPoolId`, which doesn't change either. All the addresses fetched using `fixture` object are constants as well. So, probably can just make these configuration a global static JSON object?

```
    const poolConfigs = {
        [rEthBPT.address]: { ... },
        [sfrxETHwstETHrEthBPT.address]: { ... }
    }

    function getPoolConfig(pool) {
        return poolConfigs[pool.address];
    }
```

contracts/test/fixture/_pool_tilt.js
```diff
@@ -0,0 +1,142 @@
...
+  assetAmount, // required funded asset amount to perform tilt
+  fixture,
+}) {
+  const attackerEthBalance = await waffle.provider.getBalance(attackerAddress);
+  const assetBalance = await asset.balanceOf(attackerAddress);
+
+  if (
+    attackerEthBalance.gte(ousdUnits("100000")) ||
+    assetBalance.gte(assetAmount)
+  ) {
```
shahthepro, line 66: This is confusing and feels like a bug. If account `a` has 100k ETH but wants 100 frxETH for a test case, this function won't fund them with sfrxETH?

Should this be `&&` instead of `||`? If `&&`, why check ETH balance?

contracts/test/fixture/_pool_tilt.js
sparrowDom: I was thinking of leaving these in since they might be used by curve. On the other hand I see your point, no need to future proof these APIs. Removing them

contracts/test/fixture/_pool_tilt.js
sparrowDom: That's a good point. Have done something similar https://github.com/OriginProtocol/origin-dollar/pull/1849/commits/1c98581fca76f20c039950c7ff070ad730a70595

contracts/test/fixture/_pool_tilt.js
sparrowDom: have changed the way provider is accessed, since it is cleaner. The waffle is actually used in test infrastructure. Removing it failed a test for me: 
Screenshot 2023-10-19 at 18 17 30

It also seems to contain chai matchers and fixtures: https://www.npmjs.com/package/ethereum-waffle

contracts/test/fixture/_pools.js
sparrowDom: good point, will leave it as is for simplicity sake

contracts/test/fixture/_pool_tilt.js
sparrowDom: There is also a pool and BPT token contracts being populated here, example:
```
pool: sfrxETHwstETHrEthBPT,
bptToken: sfrxETHwstETHrEthBPT
```

and while you are right that all these could be re-created outside of the fixture scope from the global addresses / ABIs I think it is more correct to use the ones created by the fixtures. For the following reasons: 
- re-creating contracts seems wasteful if the fixture has already created them
- fixture could potentially add special functionality by replacing code in the contracts, and it seems more correct to use the same object instances as created by the fixture

contracts/test/fixture/_pool_tilt.js
sparrowDom: nice catch thanks yes this is a bug, should definitely be a `&&`.
My thought process is that accounts require ETH & the required asset to function. So either is missed we re-fund it.

sparrowDom:
Thanks a lot for CR @shahthepro

naddison36 (approved):

contracts/test/fixture/_pool_tilt.js
shahthepro: Thanks, I'll deal with the waffle in a different PR

contracts/test/fixture/_pool_tilt.js
shahthepro: Agreed, let's leave it as it is for now

contracts/test/fixture/_pool_tilt.js
shahthepro: nit: I'm just thinking that it's probably better to split this into two conditional statements:

```
if (attackerEthBalance.lte(ousdUnits("100"))) {
// Fund it with ETH
}

if (assetBalance < xxx) {
// Fund with asset
}
```

Also, checking if ETH balance is over `ousdUnits("100000")` seems like an overkill, probably better to just check if it has 1 or 2 ETH as well 😅

shahthepro (approved):

contracts/test/fixture/_pool_tilt.js
sparrowDom: the 1m eth check is there to see if this function has funded it. Later in the function WETH is minted and that requires 500k ETH on the wallet. 

I've made this more efficient like you've suggested and also fund a bug where if accounts would have already been funded a signer wasn't returned.

PR #1846, opened 2023-09-28
Convex frxETH/WETH strategy using Curve abstraction libraries

naddison36:
If you made a contract change, make sure to complete the checklist below before merging it in master.

Refer to our [documentation](https://github.com/OriginProtocol/security) for more details about contract security best practices.

Contract change checklist:
  - [ ] Code reviewed by 2 reviewers. 
  - [ ] Copy & paste code review [security checklist](https://github.com/OriginProtocol/security/blob/master/templates/Contract-Code-Review-Example.md) below this checklist.
  - [ ] Unit tests pass
  - [ ] Slither tests pass with no warning
  - [ ] Echidna tests pass if PR includes changes to OUSD contract (not automated, run manually on local)

PR #1854, opened 2023-10-05
Fix FraxETH strategy tests

shahthepro:
- Only changes Mock contract to fix the unit tests
- Skips one fork tests added in #1838 (since the contract hasn't been upgraded on mainnet yet)
- Moves replacement of WETH strategy for unit tests to mock deployments instead of doing it in fixtures

### DIFF
--- a/contracts/utils/deploy.js
+++ b/contracts/utils/deploy.js
@@ -1248,6 +1248,16 @@ function deploymentWithGuardianGovernor(opts, fn) {
   return main;
 }
 
+async function replaceContractAt(targetAddress, mockContract) {
+  const signer = (await hre.ethers.getSigners())[0];
+  const mockCode = await signer.provider.getCode(mockContract.address);
+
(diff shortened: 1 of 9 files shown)


contracts/utils/deploy.js
```diff
@@ -1248,6 +1248,16 @@ function deploymentWithGuardianGovernor(opts, fn) {
   return main;
 }
 
+async function replaceContractAt(targetAddress, mockContract) {
```
naddison36, line 1251: On my branch, I moved `replaceContractAt` to a new `utils/hardhat.js` file with `hardhatSetBalance`.

https://github.com/OriginProtocol/origin-dollar/pull/1846/files#diff-46bda6023853d155885af601cb43da6cab0ccea9b57fd3bb9b14c89ebcbdf8a6

naddison36 (approved):

PR #1876, opened 2023-10-20
Update Buyback contract for OGV & CVX swaps

shahthepro:

Contract change checklist:
  - [ ] Unit tests pass


DanielVF:
Is there a reason we use one contract with lots of if statements, rather than just deploying one simple contract twice? Seems like it would keep the code simpler, and with that be easier to change in the future?

The different swap routes for each coin could just be an immutable bytestring passed in during the constructor.

DanielVF:
Thanks for cleaning up the now unused swap() method. 👍

contracts/contracts/buyback/OETHBuyback.sol
```diff
@@ -0,0 +1,38 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import { BaseBuyback } from "./BaseBuyback.sol";
+
+contract OETHBuyback is BaseBuyback {
+    // abi.encodePacked(oeth, uint24(500), weth9, uint24(3000), ogv);
+    bytes public constant ogvPath =
+    // solhint-disable-next-line max-line-length
+        hex"856c4efb76c1d1ae02e20ceb03a2a6a08b0b8dc30001f4c02aaa39b223fe8d0a0e5c4f27ead9083c756cc2000bb89c354503c38481a7a7a51629142963f98ecc12d0";
```
sparrowDom, line 10: I think the `OETHBuyback` and `OUSDBuyback` child contracts only differ in the path strings. Maybe those could be the only override differences between them?

sparrowDom (approved):
LGTM

contracts/contracts/buyback/OETHBuyback.sol
shahthepro: I thought of that. But you cannot override constant variables in Solidity. The other way is to define `ogvPath` and `cvxPath` as virtual functions on `BaseBuyback` and then override them with public constant in `OETHBuyback` and `OUSDBuyback` and override it as a function on `MockBuyback` (for unit tests). I can give it a try though

sparrowDom: you could have a mapping of address => bytes where ogv address would map to => bytes path. And have that mapping populated as part of specific buyback constructor body implementation. 

Though having bytes as value might be complicated with mapping and not an elegant solution. 

So if above can't be cleanly implemented your current solution is probably the cleanest

DanielVF: Having the internal method on each final contract that loads these in seems pretty reasonable. The rest of the code cleans up and deduplicates nicely.

contracts/contracts/buyback/BaseBuyback.sol
```diff
@@ -0,0 +1,266 @@
...
+     */
+    function swap(
+        uint256 oTokenAmount,
+        uint256 minOGV,
+        uint256 minCVX
+    ) external onlyGovernorOrStrategist nonReentrant {
+        _swap(oTokenAmount, minOGV, minCVX);
+    }
+
+    function _swap(
```
DanielVF, line 157: Do we need a separate internal function here for swap? Can this code just be in the external swap() function?

contracts/contracts/buyback/BaseBuyback.sol
```diff
@@ -0,0 +1,266 @@
...
+        uint256 ogvBalanceBefore = IERC20(ogv).balanceOf(rewardsSource);
+        uint256 cvxBalanceBefore = IERC20(cvx).balanceOf(address(this));
+
+        // Build swap input
+        bytes[] memory inputs = new bytes[](2);
+
+        inputs[0] = abi.encode(
+            // Send swapped OGV directly to RewardsSource contract
+            rewardsSource,
+            oTokenAmount / 2,
```
DanielVF, line 177: We use this `oTokenAmount / 2` calculation four times, computing it each time. Might be nice to put this into a variable(s) and then use that.

I don't think it matters practically here, but two `oTokenAmount / 2` may not equal same number as `oTokenAmount` because of rounding on each division. Usually if you are calculating some numbers that should make up a total in solidity, you calculate the first numbers, rounding down, and then send the all the rest remaining to the last:

```
// going for x + y == z
x = z / 5
y = z - x
```

DanielVF:
On the whole this looks like a great PR. I'm going to be working though my checklist before marking a final approval, but good job on this.

contracts/contracts/buyback/BaseBuyback.sol
```diff
@@ -0,0 +1,266 @@
...
+    // slither-disable-next-line constable-states
+    uint256 private __deprecated_treasuryBps;
+
+    address public immutable oToken;
+    address public immutable ogv;
+    address public immutable cvx;
+    address public immutable cvxLocker;
+
+    // Ref: https://docs.uniswap.org/contracts/universal-router/technical-reference#command-structure
+    bytes public constant swapCommand = hex"0000";
```
DanielVF, line 56: Probably better internal/private.

contracts/contracts/buyback/BaseBuyback.sol
```diff
@@ -0,0 +1,266 @@
...
+            // behalf of Strategist after the swap
+            address(this),
+            oTokenAmount / 2,
+            minCVX,
+            _getSwapPath(cvx),
+            false
+        );
+
+        // Transfer OToken to UniversalRouter for swapping
+        IERC20(oToken).safeTransfer(universalRouter, oTokenAmount);
```
DanielVF, line 194: We don't have to use `safeTransfer` for our own tokens, since they are ERC20 compatible.

contracts/contracts/buyback/BaseBuyback.sol
DanielVF: Turns out this does actually matter, in the super picky sense, because we send the full amount to router, but then don't use the rounding error to swap back to our coins. The amount is silly and tiny, but I like systems better when they are 100% correct.

DanielVF:
# Buybacks Review


#### Requirements

This PR is shifting to include buying and locking CVX as a part of buyback contract, rather than as a separate strategist action.

We're also now using a mostly shared codebase for both OETH and OUSD buybacks.

### Deployment Considerations

Once we create the governance proposals, we'll need to not do buybacks on the migrated old buyback contracts, since the amounts transferred will be hardcoded into the governance action.

#### Internal State

_Only config_

#### Attack

The consequences of this contract going bad should be small. A week or two of a percentage of yield.

An attacker could try to get the contract to transfer tokens to themselves, however, all transfer destinations are controlled by governance.

The primary threat is MEV action during the swaps. This is defended against via minimum output parameters passed in by the strategist. These minimum output amounts are passed to the Uniswap router for checking.


#### Logic

_Looks correct_


#### Tests

_Not checked_

#### Flavor

_Looking good_

#### Overflow

#### Proxy

#### Black magic

#### Dependencies

#### Deploy

#### Authentication

- [x] Never use tx.origin

#### Cryptographic code

#### Gas problems

    - [x] A way to remove items

#### External calls

- [x] Contract addresses from public in are validated
- [x] Unsafe external calls
- [x] Reentrancy guards on all publicly accessable state changing functions
- [x] Malicious behaviors
- [x] Oracles, one of:
  - [x] No oracles
  - [ ] Oracles can't be bent


#### Ethereum

DanielVF:
Fork tests:

```python
from world import *
OETH_WHALE = '0xEADB3840596cabF312F2bC88A4Bb0b93A4E1FF5F'
NEW_OETH_BUYBACK = '0x85094b52754591A3dE0002AD97F433584389aea0'
buyback_oeth = Contract.from_abi('buyback_oeth', NEW_OETH_BUYBACK, buyback.abi, owner=STRATEGIST)
buyback_ousd = buyback

buyback.swap(100*1e18, 1e18, 1e18, {'from': STRATEGIST})
show_transfers(history[-1])
buyback.swap(100*1e18, 1e18, 20* 1e18, {'from': STRATEGIST})
buyback.swap(100*1e18, 99*1e18, 1* 1e18, {'from': STRATEGIST})
buyback.swap(100*1e18, 999*1e18, 1* 1e18, {'from': STRATEGIST})
buyback.swap(100*1e18, 9999*1e18, 1* 1e18, {'from': STRATEGIST})
buyback.swap(100*1e18, 99999*1e18, 1* 1e18, {'from': STRATEGIST})


oeth.transfer(buyback_oeth, 10*1e18, {'from': OETH_WHALE})
buyback_oeth.swap(1e18, 1e18, 1e18, {'from': STRATEGIST})
buyback_oeth.swap(int(1e18)+1, 1e18, 1e18, {'from': STRATEGIST})
show_transfers(history[-1])
```

contracts/contracts/buyback/BaseBuyback.sol
shahthepro: Yep, good catch, have made the change to fix this

contracts/contracts/buyback/BaseBuyback.sol
shahthepro: Agreed. Changed to `private`

contracts/contracts/buyback/BaseBuyback.sol
shahthepro: Done as well

contracts/contracts/buyback/BaseBuyback.sol
shahthepro: Yep, Have removed the internal function

DanielVF (approved):

PR #1889, opened 2023-10-23
C-01 DoS Due to Unrestricted Growth of poolAssets Array

sparrowDom:
BaseBalancerStrategy's cachePoolAssets [increases the length of ](https://github.com/OriginProtocol/origin-dollar/blob/f0cb237347faa7c650787498e127aba7e0c0488a/contracts/contracts/strategies/balancer/BaseBalancerStrategy.sol#L585)poolAssets[ list](https://github.com/OriginProtocol/origin-dollar/blob/f0cb237347faa7c650787498e127aba7e0c0488a/contracts/contracts/strategies/balancer/BaseBalancerStrategy.sol#L585) with every call, extending the list by the same assets every time. Calling this method more than once will cause DoS via several paths:
- BalancerMetaPoolStrategy's _deposit, _withdraw, and withdrawAll, will send malformed requests to Balancer with duplicated asset lists, and [fail validation](https://github.com/balancer/balancer-v2-monorepo/blob/3f2a99e01338e5258112ced045d074cd0c029616/pkg/vault/contracts/PoolBalances.sol#L294). This will cause all subsequent deposits and withdrawals to fail.
- Calculations based on the duplicated list will result in incorrectly [calculated withdrawal amount of BPT](https://github.com/OriginProtocol/origin-dollar/blob/f0cb237347faa7c650787498e127aba7e0c0488a/contracts/contracts/strategies/balancer/BaseBalancerStrategy.sol#L310-L325) in _withdraw, and create a potential for gas based DoS in the same method.
- BaseBalancerStrategy's cacheRateProviders will fail [its validation check](https://github.com/OriginProtocol/origin-dollar/blob/f0cb237347faa7c650787498e127aba7e0c0488a/contracts/contracts/strategies/balancer/BaseBalancerStrategy.sol#L535-L539) if it will need to be called again.

Crucially, `cachePoolAssets` lacks access control and as a result, is callable by anyone.
Thus, a malicious second call to this function will break the deposit and withdrawal
functionality of all Balancer strategies.

Consider checking that the `poolAssets` array is empty at the beginning of
`cachePoolAssets`


shahthepro (approved):

contracts/test/strategies/balancerMetaStablePool.fork-test.js
```diff
@@ -174,6 +174,15 @@ describe("ForkTest: Balancer MetaStablePool rETH/WETH Strategy", function () {
         auraRewardPool
       );
     });
+
+    // Un-skip once we re-deploy the strategy
+    it.skip("Shouldn't be able to cache assets twice", async function () {
```
DanielVF, line 179: How are we going to make sure we do this?

DanielVF (approved):
Can be merged. Will fully verify when all changes integrated.

contracts/test/strategies/balancerMetaStablePool.fork-test.js
sparrowDom: How about like this :)
https://github.com/OriginProtocol/origin-dollar/pull/1889/commits/74fd116b39c4df5174c427d17eb524d9bfd4c770

PR #1893, opened 2023-10-24
Convex frxETH/WETH Strategy - N-05

naddison36:
The `cvxRewardStaker` state variable of the `ConvexStrategy` contract is provided as a
constructor argument. Alternatively, its value can be retrieved from the `crvRewards` member
of the Convex `poolInfo` , where poolInfo can be retrieved through
`ConvexBooster.poolInfo(poolId)` . In our case, `cvxDepositor` is the address of the
`ConvexBooster` , whereas `cvxDepositorPoolId` is the `poolId`.

Consider removing the `cvxRewardStaker` constructor argument and retrieving its value from
the Convex `Booster` on-chain state. This would make the deployment more encapsulated and
less error-prone.


sparrowDom (commented):
LGTM ✅

PR #1900, opened 2023-10-27
Added signer support for Open Zeppelin's Defender Relayer

naddison36:


contracts/utils/signers.js
```diff
@@ -44,13 +46,35 @@ async function getSigner(address) {
     return await impersonateAndFund(address);
   }
 
+  // If using Defender Relayer
+  if (process.env.DEFENDER_API_KEY && process.env.DEFENDER_API_SECRET) {
```
sparrowDom, line 50: It seems unintuitive to me calling `getSigner(address)` would either return a signer for that address or a signer configured by the environmental variables. What do you think of changing the function signature to: 
`async function getSigner(addressOption)` or `async function getSigner(address=undefined)` to more explicitly signal the optionality of the address?

nit: we might also want to log that a defender signer has been created similar how impersonate above communicates it with a log

sparrowDom (dismissed):
LGTM

contracts/utils/signers.js
naddison36: Good point. I prefer the `async function getSigner(address=undefined)` option.

Yes, I'll also add a log message. Thanks

shahthepro (approved):

PR #1909, opened 2023-11-03
M-01 DoS of Strategy Removal Due to Underflow

sparrowDom:
In emergencies, certain `VaultAdmin` actions relying on `withdrawAll` may be blocked by
an attacker. This can happen in the following cases:
- The strategy does not have any investment.
- The strategy lacks BPT due to an earlier withdrawal performed without removing the
strategy as well.

The check condition and subtraction in the `withdrawAll` function are inconsistent.
Consequently, an amount of `1` wei passes the check but triggers an arithmetic underflow.

This underflow can be maliciously triggered if the Balancer exit returns 0 wei of sfrxETH when
the strategy does not hold any BPT. The required 0 BPT balance may arise from a lack of
allocation or a previous withdrawal.

Thus, an attacker could trigger an underflow by transferring to the strategy either 1 wei of
sfrxETH, a minimal BPT amount (e.g., 4 wei) or a suitable amount of AURA LP tokens.

This would block all functions that depend on `withdrawAll` not reverting. Namely
`removeStrategy` , withdrawFromStrategy and importantly,
`withdrawAllFromStrategies`.

As part of damage mitigation during an attack, the protocol team can transfer more sfrxETH to
the strategy and call `removeStrategy` . However, during emergencies, the time to identify
and resolve this issue could delay addressing the initial emergency, possibly causing severe
consequences.

Consider aligning the checked and subtracted amounts. Additionally, consider using try-
catch in `withdrawAllFromStrategies` , so that one strategy's withdrawal reverting does not revert the entire transaction.

### DIFF
--- a/contracts/contracts/strategies/balancer/BalancerMetaPoolStrategy.sol
+++ b/contracts/contracts/strategies/balancer/BalancerMetaPoolStrategy.sol
@@ -10,6 +10,7 @@ import { BaseAuraStrategy, BaseBalancerStrategy } from "./BaseAuraStrategy.sol";
 import { IBalancerVault } from "../../interfaces/balancer/IBalancerVault.sol";
 import { IERC20, InitializableAbstractStrategy } from "../../utils/InitializableAbstractStrategy.sol";
 import { StableMath } from "../../utils/StableMath.sol";
+import "@openzeppelin/contracts/utils/math/Math.sol";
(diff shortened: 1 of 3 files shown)


DanielVF (approved):
Can be merged. Will fully verify when all changes integrated.

PR #1910, opened 2023-11-03
Hot deploy source code

sparrowDom:
This puts nitro into fork tested development speed. Currently we've had 2 ways of approaching forked development when solidity files needed changing: 
- restart the node and wait for all deployments to finish
- restart node with all the deployments except for the one you are working on. And let test suite run the required deployment. This requires some manual `foceSkip` param configuration in the deployment files

This PR introduces a third option: Let the node run in a separate terminal, never requiring to restart it and then in env config specify which type of contracts you want "hot deployed" - have their source code updated before fork test are ran. This is achieved so that: 
- env var config is being read to understand which files need updating. Options are: 
  - the strategy contract the specific fork test fixture is running against
  - vault admin contract
  - vault core contract
  - harvester (not yet supported)
  - oracleRouter (not yet supported)
- the required contracts are re-deployed by a random address using the latest bytecode (test suite runs contract compilation before executing the tests). 
- at the live/mainnet implementation address of a specified contract the bytecode is replaced with the newly deployed one. 

And this functionality comes with a significant speed increase: 
Screenshot 2023-11-03 at 23 40 27
*Currently the strategies/fixtures in the screenshot above are supported*

**Pros:**
- It is ~+50% faster than the previously known fastest approach to forked development. - it is getting closer to unit test speeds with sub 10 second test execution
- easy configuration -> it is much easier specifying an env var (`HOT_DEPLOY=strategy,vaultCore,vaultAdmin,harvester`) than fiddling with the code
- flexible configuration changes. You can change also to include/exclude Vault(Admin/Core) deployment. Switch to a different fork fixture handled by a different strategy. None of those require the restart of the node or any other performance (speed execution) hit.
- test / debug code changes (say in Vault Admin/Core) without needing to modify deploy scripts

### DIFF
--- a/contracts/README.md
+++ b/contracts/README.md
@@ -105,6 +105,22 @@ yarn test:fork
 
 See [Fork Tests](./fork-test.md) for more information.
 
+### Hot Deploys
+
+You can enable the "hot deploy" mode when doing fork testing development. The mode enables updating the contract code much faster and more conveniently comparing to running deploy scripts. Each time a fork test suite is ran, the configured contracts are updated
+
+To enable Hot Deploys set the HOT_DEPLOY variable in the contracts/.env file. Enable various modes using comma separated flags to direct which contracts need source updated (in the node runtime): 
+- strategy -> strategy contract associated to fixture
+- vaultCore -> vaultCore or oethVaultCore depending on the nature of the fixture
+- vaultAdmin -> vaultAdmin or oethVaultAdmin depending on the nature of the fixture
+- harvester -> harvester or oethHarvester (not yet supported)
+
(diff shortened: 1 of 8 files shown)


naddison36 (commented):
I'm excited to try this.

Can you add a section to the contracts/README.md for hot redeploys. Thanks

sparrowDom:
thanks for pointing it out @naddison36 . Readme added

shahthepro (approved):
Awesome work!

sparrowDom:
the fork test errors seem to be related to Aave issues and rETH Oracle price. Nothing to do with this PR...

PR #1912, opened 2023-11-06
M-02 Incorrect Assumption That Pool’s BPT Is the First Token in Balancer Pool

sparrowDom:
This pull request adds a gas efficient `_bptTokenPoolPosition` immutable parameter to the Composable Stable Pool Strategy that specifies the position of the BPT token in the list of tokens returned from the strategy. That parameter is used for required filtering of token lists. 

**Audit comment:**
`balancerVault.getPoolTokens()` is used to query the tokens supported by the Balancer
pool. The returned array includes the pool's BPT, which is used to track liquidity provisioning.

The `BalancerComposableStrategyContract` assumes that the BPT is at the first
position in the returned array. This assumption is made during `strategy initialization` and when
depositing/withdrawing assets by shifting the array parameters such that they exclude the
BPT.

While this assumption holds for the `wstETH-rETH-sfrxETH` pool, it does so only by
coincidence. The Balancer code does not enforce the BPT to be the first in the array. Instead, it
sorts the tokens by their contract address. Here is an example of a pool where the BPT comes
second in the queried array.

As such, trying to integrate with such a pool will fail during initialization. However, this
initialization check is crucial since the implementation relies on this assumption. Using the
current implementation with a pool that has BPT in another spot will cause unexpected
behavior and potential loss of value because of the faulty results of
_getUserDataEncodedAmounts and _getUserDataEncodedAssets .

Consider explicitly documenting this assumption in case only the pools where this assumption
holds will be used. Alternatively, if other pools are planned to be used, modify the
implementation to remove this assumption from the codebase.

### DIFF
--- a/contracts/contracts/mocks/BalancerComposablePoolTestStrategy.sol
+++ b/contracts/contracts/mocks/BalancerComposablePoolTestStrategy.sol
@@ -15,13 +15,15 @@ contract BalancerComposablePoolTestStrategy is BalancerComposablePoolStrategy {
         BaseStrategyConfig memory _stratConfig,
         BaseBalancerConfig memory _balancerConfig,
         BaseMetaPoolConfig memory _metapoolConfig,
-        address _auraRewardPoolAddress
+        address _auraRewardPoolAddress,
+        uint256 _bptTokenPoolPosition
     )
         BalancerComposablePoolStrategy(
             _stratConfig,
             _balancerConfig,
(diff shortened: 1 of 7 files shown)


DanielVF (commented):
Can be merged. Will review when all changes integrated.

DanielVF (approved):
Can be merged. Will verify when all changes integrated.

PR #1913, opened 2023-11-06
L-01 Duplicate Strategy Assets Can Be Passed to _withdraw

sparrowDom:
This PR adds a check so withdrawal with duplicate assets can not be called. Currently, the deposit with duplicate assets can not be called (contract API doesn't support it) but a future proof check is still added (with a test) to validate duplicate asset deposit calls can also not be made. 

**Audit comments:**
The array arguments passed to `_withdraw` are not validated to ensure that they contain
unique assets. If, in a future extension, it becomes possible to pass duplicates (either due to
user or strategist input), the following impacts may occur:
- The calculated expected BPT from _getBPTExpected may be double-counted.
- Duplicate transfers may be executed, transferring too many tokens to the recipient.

Consider enforcing that poolAssetsAmountsOut[..] is 0 before setting its value.

### DIFF
--- a/contracts/contracts/strategies/balancer/BalancerMetaPoolStrategy.sol
+++ b/contracts/contracts/strategies/balancer/BalancerMetaPoolStrategy.sol
@@ -310,8 +310,8 @@ contract BalancerMetaPoolStrategy is BaseAuraStrategy {
             if (poolAssetAmount > 0) {
                 strategyAssetsToPoolAssetsAmounts[i] = poolAssetAmount;
 
-                /* This check is triggered when the _withdrawal is unexpectedly
-                 * called with a duplicate asset in the _strategyAssets array
+                /* This check is triggered when the _withdrawal is called with
+                 * a duplicate asset in the _strategyAssets array
                  */
                 require(
                     poolAssetsAmountsOut[poolAssetIndex[poolAsset]] == 0,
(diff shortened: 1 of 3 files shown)


contracts/contracts/strategies/balancer/BalancerMetaPoolStrategy.sol
```diff
@@ -157,6 +158,14 @@ contract BalancerMetaPoolStrategy is BaseAuraStrategy {
                     strategyAmount
                 );
 
+                /* This check is triggered when the _deposit is unexpectedly
```
DanielVF, line 161: Probably don't need the word "unexpectedly" here.

contracts/contracts/strategies/balancer/BalancerMetaPoolStrategy.sol
```diff
@@ -300,6 +309,14 @@ contract BalancerMetaPoolStrategy is BaseAuraStrategy {
             if (poolAssetAmount > 0) {
                 strategyAssetsToPoolAssetsAmounts[i] = poolAssetAmount;
 
+                /* This check is triggered when the _withdrawal is unexpectedly
```
DanielVF, line 312: Same, don't need "unexpectedly"

DanielVF (approved):
Can merge, will verify when integrated.

PR #1915, opened 2023-11-07
L-04 Inaccurate IBalancerVault Integration Enums

sparrowDom:
**Audit notes:**
In `IBalancerVault.sol` , there are several enums which are inaccurate:
- The third value of the `IBalancerVault.ComposablePoolExitKind` enum is named
`EXACT_BPT_IN_FOR_TOKENS_OUT` . Consider changing it to
`EXACT_BPT_IN_FOR_ALL_TOKENS_OUT` in accordance with the the Balancer
documentation.
- The `IBalancerVault.WeightedPoolJoinKind` enum does not fully match any of
the `JoinKinds` from the Balancer documentation. While the actual value used is
correct, the above makes the code error-prone and unnecessarily confusing. Additionally,
it is named incorrectly as it is not the right pool type for the strategy.

Consider using the exact values detailed in the documentation. Additionally, consider adding
links to the documentation in inline comments for ease of validation.

### DIFF
--- a/contracts/contracts/interfaces/balancer/IBalancerVault.sol
+++ b/contracts/contracts/interfaces/balancer/IBalancerVault.sol
@@ -4,25 +4,22 @@ pragma solidity ^0.8.0;
 import { IERC20 } from "../../utils/InitializableAbstractStrategy.sol";
 
 interface IBalancerVault {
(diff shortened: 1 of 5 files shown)


DanielVF (approved):
Will verify when integrated.

PR #1916, opened 2023-11-07
L-03 Abstraction of Pool Exit Kinds Is Error-Prone

sparrowDom:
**Audit notes:**
The use of different enum indices, namely `balancerBptInExactTokensOutIndex` and
`balancerExactBptInTokensOutIndex` , to be passed to Balancer for different pool types
is achieved by passing constructor input arguments and storing immutable variables.

However, this approach is error-prone. It reduces readability and makes it difficult for off-chain
tooling and governance to correctly test and pass the arguments during deployment, and verify
the result. As such, any mistake in setting these values could have severe consequences.

Consider creating and overriding internal functions in each class. This will not increase the gas
usage but will ensure that the correct value is always used for each concrete contract. A
simplified example is shown below:

```python
contract A {
    function _someValue() internal pure virtual returns(uint) { return 1; }
}

contract B is A {
    function _someValue() internal pure override returns(uint) { return 2; }
}
```


DanielVF (approved):
Way better.

Can merge, will verify when integrated.

PR #1917, opened 2023-11-07
Frax Convex Strategy with locked Frax Staked Convex LP tokens

naddison36:
This PR extends PR https://github.com/OriginProtocol/origin-dollar/pull/1846 which implemented a Convex Strategy. This PR adds a `FraxConvexStrategy` that supports locking Frax Staked Convex LP tokens.

 There are multiple levels of tokens managed by this strategy
 * Vault assets. eg WETH and frxETH
 * Curve LP tokens. eg frxETH-ng-f
 * Convex LP tokens. eg cvxfrxeth-ng-f
 * Frax Staked Convex LP tokens. eg stkcvxfrxeth-ng-f-frax
 * Locked Frax Staked Convex LP tokens which do not have a LP token. eg FraxUnifiedFarm_ERC20_Convex_frxETH

# Contracts

The new strategy is in green. Modified contracts are in orange.

oethContracts

ConvexStrategiesHierarchy

FraxConvexStrategySquashed

FraxConvexStrategyStorage

# Value Flows

Deposit and locking flows

oethValueFlows-frax-convex-deposit

Harvesting CRV, CVX and FXS token rewards

oethValueFlows-frax-convex-harvest

# Tests

There are no unit tests for the `FraxConvexStrategy` as it has heavy integration to Frax Convex.
There is, however, a unit tests for `ConvexTwoPoolStrategy` which implements `ConvexStrategy` which implements `BaseCurveStrategy`. `FraxConvexStrategy` implements `BaseCurveStrategy` so there is some base unit test coverage.

```
yarn test ./test/strategies/convexFrxEthWeth.js
```

Fork tests

```
yarn test:fork ./test/strategies/frax-convex-weth.fork-test.js
```

# Deploy

The `contracts/deploy/082_frax_convex_strategy.js` script will deploy the new strategy and propose the required governance actions.

# Review

Contract change checklist:
  - [ ] Unit tests pass


DanielVF:
It's possible we could make this far simpler. 

Let's say that we have an internal method `_adjustLockup` that tries to match the pendingWithdraws and the currently available funds. If it has too much, it locks it up, if it has too little, and it can, it pulls back out of the lockup. We set pendingWithdraws from the strategist, directly on the contract before we want to withdraw.


```
_lpDepositAll:
 - _adjustLockup()

_lpWithdrawAll:
  - set pendingWithdraws to all tokens locked balance
  - _adjustLockup()
  - what ever LP tokens we were able to get.
  - pendingWithdraws -= amount withdraw

_lpWithdraw:
  - _adjustLockup()
  -  revert if we don't have enough withdrawn funds now
  - pendingWithdraws -= amount withdraw

setPendingWithdraw(amount) external admins:
  - set pendingWithdraws to amount
```

This means that all our locking/unlocking logic now lives in one function, and work in one place with funds out or in.

naddison36:
> It's possible we could make this far simpler.
> 
> Let's say that we have an internal method `_adjustLockup` that tries to match the pendingWithdraws and the currently available funds. If it has too much, it locks it up, if it has too little, and it can, it pulls back out of the lockup. We set pendingWithdraws from the strategist, directly on the contract before we want to withdraw. When _adjustLockup unlocks funds, it reduces pendingWithdraws.
> 
> ```
> _lpDepositAll:
>  - _adjustLockup()
> 
> _lpWithdrawAll:
>   - set pendingWithdraws to all tokens locked balance
>   - _adjustLockup()
>   - use whatever LP tokens we were able to get.
> 
> _lpWithdraw:
>   - _adjustLockup()
>   -  revert if we don't have enough unlocked funds
> 
> setPendingWithdraw(amount) external admins:
>   - set pendingWithdraws to amount
> ```
> 
> This means that all our locking/unlocking logic now lives in one function, and rather than having four methods that have to do separate logic and calls about funds into and out of locking.

I like this approach. It can also work for the function that rolls the lock.

I got the basic fork tests working today. I'll work on cleaning things up tomorrow and more testing

DanielVF:
Yeah, it should do the lock rolling as well.

One thing I said incorrectly was that unlocking should reduce the pendingWithdraws - it should actually be withdrawing that does that, since otherwise the funds would get locked up again on the next invocation.

naddison36:
I've refactored the `FraxConvexStrategy` contract today. I was originally using `pendingWithdraws` to fit with ERC-7540 but its simpler to have a `targetLockedBalance` set by the Strategist. This allows the strategy to lock a specified amount of Frax Staked Convex LP tokens and hold the rest as unlocked.

There are multiple levels of tokens managed by this strategy:
- Vault assets. eg WETH and frxETH
 - Curve LP tokens. eg frxETH-ng-f
 - Convex LP tokens. eg cvxfrxeth-ng-f
 - Frax Staked Convex LP tokens. eg stkcvxfrxeth-ng-f-frax
 - Locked Frax Staked Convex LP tokens which do not have a LP token

Note there are no partial withdrawals of expired locked tokens so full withdraws of expired tokens are done before locking more.

Below is what the high level design now looks like

```
storage variables:
  - lockKey
  - unlockTimestamp
  - targetLockedBalance

_unlock(lockedBalance):
  - if (there is a lock and
        the lock has expired and
        the lockedBalance > targetLockedBalance)
    - withdraw all tokens from the lock
    - reset lockKey
    - return the unlockedAmount

_lock(unlockedBalance, lockedBalance):
  - If targetLockedBalance > lockedBalance
    - calc lockAmount = min(unlockedBalance, targetLockAmount < unlockedBalance)
    - if lockAmount < MIN_LOCK_AMOUNT return
    - if no lock
      - create new lock for 7 days
      - set lockKey and unlockTimestamp
    - else 
      - add to existing lock 
      - extent existing lock by 7 days

_lpDepositAll:
  - Stake Curve LPs
  - unlock(lockedBalance)
  - lock(unlockedBalance, lockedBalance)

_lpWithdrawAll:
  - set targetLockedBalance = 0
  - unlock(lockedBalance)
  - Unstake all Curve LPs

_lpWithdraw:
  - revert if not enough unlocked tokens or
      not enough unlocked + expired tokens
  - unlock(lockedBalance)
  - Unstake required Curve LPs
  - lock(unlockedBalance, lockedBalance)

- extendLock:
  - unlock(lockedBalance)
  - lock(unlockedBalance, lockedBalance)  

setTargetLockedBalance(balance) external onlyStrategist:
  - set targetLockedBalance
```

I'm now working on the required permutation for the fork tests

naddison36:
Here's an outline of the fork tests to cover the different permutation of states.
This will require a configuration driven testing approach.

```
- no locked amounts == 0
	- extend lock - revert
	- staked amounts == 0
		- target locked balance == 0
			- deposit 
				- stake deposit amount
				- no unlock or lock
			- withdraw - revert
			- withdrawAll
				- no unstake
				- no unlock or lock
		- target locked balance > 0
			- deposit < target locked balance
				- stake deposit amount
				- lock deposit amount
			- deposit > target locked balance
				- stake deposit amount
				- lock target locked balance
			- withdraw - revert
			- withdrawAll
				- no unstake
				- no unlock or lock
	- staked amount > 0
		- target locked balance == 0
			- deposit
				- stake deposit amount
				- no unlock or lock
			- withdraw amount < staked amount
				- unstake withdraw amount
				- no unlock or lock
			- withdraw amount > staked amount
				- revert
			- withdrawAll
				- unstake staked amount
				- no unlock or lock
		- 0 < target locked balance < staked amount
			- deposit
				- stake deposit amount
				- no unlock
				- lock target locked balance
			- withdraw amount < staked amount
				- unstake withdraw amount
				- no unlock or lock
			- withdraw amount > staked amount
				- revert
			- withdrawAll
				- unstake staked amount
				- no unlock or lock
		- 0 < target locked balance > staked amount
			- deposit amount < target locked balance - staked amount
				- stake deposit amount
				- no unlock
				- lock deposit amount
			- deposit amount > target locked balance - staked amount
				- stake deposit amount
				- no unlock
				- lock target locked balance - staked amount
			- withdraw amount < staked amount
				- unstake withdraw amount
				- no unlock or lock
			- withdraw amount > staked amount
				- revert
			- withdrawAll
				- unstake staked amount
				- no unlock or lock
- unexpired locked amount > 0
	- extend lock - revert
	- staked amounts == 0
		- target locked balance == 0
			- deposit 
				- stake deposit amount
				- no unlock or lock
			- withdraw - revert
			- withdrawAll
				- no unstake
				- no unlock or lock
		- target locked balance > 0
			- deposit < target locked balance
				- stake deposit amount
				- lock deposit amount
			- deposit > target locked balance
				- stake deposit amount
				- lock target locked balance
			- withdraw - revert
			- withdrawAll
				- no unstake
				- no unlock or lock
	- staked amount > 0
		- target locked balance == 0
			- deposit
				- stake deposit amount
				- no unlock or lock
			- withdraw amount < staked amount
				- unstake withdraw amount
				- no unlock or lock
			- withdraw amount > staked amount
				- revert
			- withdrawAll
				- unstake staked amount
				- no unlock or lock
	- 0 < target locked balance < staked amount
		- deposit
			- stake deposit amount
			- no unlock
			- lock target locked balance
		- withdraw amount < staked amount
			- unstake withdraw amount
			- no unlock or lock
		- withdraw amount > staked amount
			- revert
		- withdrawAll
			- unstake staked amount
			- no unlock or lock
	- 0 < target locked balance > staked amount
		- deposit amount < target locked balance - staked amount
			- stake deposit amount
			- no unlock
			- lock deposit amount
		- deposit amount > target locked balance - staked amount
			- stake deposit amount
			- no unlock
			- lock target locked balance - staked amount
		- withdraw amount < staked amount
			- unstake withdraw amount
			- no unlock or lock
		- withdraw amount > staked amount
			- revert
		- withdrawAll
			- unstake staked amount
			- no unlock or lock
- expired locked amount
	- staked amounts == 0
		- target locked balance == 0
			- extend lock
				- unlock all expired
				- no lock
			- deposit
				- stake deposit amount
				- unlock all expired 
				- no lock
			- withdraw amount < expired amount
				- unlock all expired
				- unstake withdraw amount 
				- no lock
			- withdraw amount > expired amount
				- revert
			- withdrawAll
				- unlock all expired
				- unstake all expired
				- no lock
		- 0 < target locked balance < expired amount
			- extend lock
				- unlock all expired
				- lock target locked balance
			- deposit
				- stake deposit amount
				- unlock all expired
				- lock target locked balance
			- withdraw amount < expired amount
				- unlock all expired
				- unstake withdraw amount 
				- lock expired amount - withdraw amount 
			- withdraw amount > expired amount
				- revert
			- withdrawAll
				- unlock all expired
				- unstake target locked balance  - expired amount
				- lock target locked balance
		- 0 < target locked balance > expired amount
			- extend lock
				- no unlock
				- add 7 days to lock
			- deposit < target locked balance - expired amount
				- stake deposit amount
				- no unlock
				- add expired amount + deposit to lock
				- add 7 days to lock
			- deposit > target locked balance - expired amount
				- stake deposit amount
				- no unlock
				- add target locked balance to lock
				- add 7 days to lock
			- withdraw amount < expired amount
				- unlock all expired
				- unstake withdraw amount 
				- lock expired amount - withdraw amount 
			- withdraw amount > expired amount
				- revert
			- withdrawAll
				- unlock all expired
				- unstake expired amount
				- no lock
	- staked amount > 0
		- target locked balance == 0
			- extend lock
				- unlock all expired
				- no lock
			- deposit
				- stake deposit amount
				- unlock all expired
				- no lock
			- withdraw amount < expired amount + staked amount
				- unlock all expired
				- unstake withdraw amount 
				- no lock
			- withdraw amount > expired amount + staked amount
				- revert
			- withdrawAll
				- unlock all expired
				- unstake all expired + staked amount
				- no lock
		- 0 < target locked balance < expired amount < staked amount
			- extend lock
				- unlock all expired
				- lock target locked balance
			- deposit
				- stake deposit amount
				- unlock all expired
				- lock target locked balance
			- withdraw amount < expired amount + staked amount
				- unlock all expired
				- unstake withdraw amount 
				- lock expired amount + staked amount - withdraw amount
			- withdraw amount > expired amount + staked amount
				- revert
			- withdrawAll
				- unlock all expired
				- unstake all expired + staked amount
				- no lock
		- 0 < staked amount < target locked balance < expired amount
			- extend lock
				- unlock all expired
				- lock target locked balance
			- deposit
				- stake deposit amount
				- unlock all expired
				- lock target locked balance
			- withdraw amount < staked amount
				- no unlock
				- unstake withdraw amount 
				- add expired amount + staked amount - withdraw amount to lock
				- add 7 days to lock
			- withdraw amount < expired amount + staked amount
				- unlock all expired
				- unstake withdraw amount 
				- lock expired amount + staked amount - withdraw amount
			- withdraw amount > expired amount + staked amount
				- revert
			- withdrawAll
				- unlock all expired
				- unstake all expired + staked amount
				- no lock
		- 0 < expired amount < target locked balance < staked amount
			- extend lock
				- no unlock
				- add target locked balance - expired amount to lock
				- add 7 days to lock
			- deposit
				- stake deposit amount
				- no unlock
				- add target locked balance - expired amount to lock
				- add 7 days to lock
			- withdraw amount < staked amount
				- no unlock
				- unstake withdraw amount 
				- add expired amount + staked amount - withdraw amount to lock
				- add 7 days to lock
			- withdraw amount > staked amount < expired amount + staked amount
				- unlock all expired
				- unstake withdraw amount 
				- lock expired amount + staked amount - withdraw amount
			- withdraw amount > expired amount + staked amount
				- revert
			- withdrawAll
				- unlock all expired
				- unstake all expired + staked amount
				- no lock
		- 0 < expired amount < staked amount < target locked balance < expired amount + staked amount
			- extend lock
				- no unlock
				- add target locked balance - expired amount to lock
				- add 7 days to lock
			- deposit
				- stake deposit amount
				- no unlock
				- add target locked balance - expired amount to lock
				- add 7 days to lock
			- withdraw amount < staked amount
				- no unlock
				- unstake withdraw amount 
				- add expired amount + staked amount - withdraw amount to lock
				- add 7 days to lock
			- withdraw amount > staked amount < expired amount + staked amount
				- unlock all expired
				- unstake withdraw amount 
				- lock expired amount + staked amount - withdraw amount
			- withdrawAll
				- unlock all expired
				- unstake all expired + staked amount
				- no lock
		- 0 < expired amount < staked amount < expired amount + staked amount < target locked balance
			- extend lock
				- no unlock
				- add expired amount + staked amount
				- add 7 days to lock
			- deposit
				- stake deposit amount
				- no unlock
				- add expired amount + staked amount
				- add 7 days to lock
			- withdraw amount < staked amount
				- no unlock
				- unstake withdraw amount 
				- add expired amount + staked amount - withdraw amount to lock
				- add 7 days to lock
			- withdraw amount > staked amount < expired amount + staked amount
				- unlock all expired
				- unstake withdraw amount 
				- lock expired amount + staked amount - withdraw amount
			- withdrawAll
				- unlock all expired
				- unstake all expired + staked amount
				- no lock
```

DanielVF:
It still feels like this code could collapse down even further. Most of the code in the four main methods are doing quite similar things. Couldn't the locking and unlocking code fold down to just one method that each use?

These are all going to admin calls as well, so gas efficient doesn't need to be a priority.

naddison36:
> It still feels like this code could collapse down even further. Most of the code in the four main methods are doing quite similar things. Couldn't the locking and unlocking code fold down to just one method that each use?
> 
> These are all going to admin calls as well, so gas efficient doesn't need to be a priority.

Part of the complication comes from wanting to allow withdraws from expired locks when below the target lock balance.  For exmaple

```
Given
  0 < expired amount < staked amount < target locked balance < expired amount + staked amount
  and withdraw amount > staked amount < expired amount + staked amount
when
  withdraw
then
  unlock all expired
  unstake withdraw amount 
  lock expired amount + staked amount - withdraw amount
```

The code will be a lot easier if we changed the behaviour to

```
Given
  0 < expired amount < staked amount < target locked balance < expired amount + staked amount
  and withdraw amount > staked amount < expired amount + staked amount
when
  withdraw
then
  revert
```

The Strategist can get around the revert by setting the target locked balance to a smaller value so

```
Given
  0 < expired amount < staked amount < target locked balance < expired amount + staked amount
  and withdraw amount > staked amount < expired amount + staked amount
when
  strategist set target locked balance <= expired amount + staked amount - withdraw amount
  strategist withdraw
then
  unlock all expired
  unstake withdraw amount 
  lock target locked balance
```

Note the above is not a problem for `withdrawAll` as it sets the target locked level to zero at the start so will unlock all expired tokens.

```
Given
  0 < expired amount < staked amount < target locked balance < expired amount + staked amount
  and withdraw amount > staked amount < expired amount + staked amount
when
  strategist withdrawAll
then
  unlock all expired
  unstake expired amount + staked amount
  no lock
```

PS I've now realised I should have used Gherkin syntax for the definitions of the fork tests

DanielVF:
We were talking about what would happen if, hypothetically, there weren't enough FXS in the tokens to allow collecting rewards. Since there's a possibility of this happening, we could set the reward collection to `false` on the withdraws. This way we can get our funds. The collect reward tokens method can handle actually getting the reward tokens, and it's since it is called by the harvester, we should still be able to get the reward tokens in the future. This also would not affect normal operation.

```python
from world import *
locking = Contract.from_explorer("0xB4fdD7444E1d86b2035c97124C46b1528802DA35")
fxs = Contract.from_explorer("0x3432b6a60d23ca0dfca7761b7ab56459d9c964d0")
USER = "0x857E629E94D7252C8f71eBAF513548b796641DAc"

with TemporaryFork():
  print(fxs.balanceOf(USER))
  locking.withdrawLocked('0x7106AA15DDDC5E6A21389F9E398B1DD17706D679338C38E3C63883CA56C6F2DA', USER, True, {'from': USER})
  print(fxs.balanceOf(USER))


with TemporaryFork():
  print(fxs.balanceOf(USER))
  locking.withdrawLocked('0x7106AA15DDDC5E6A21389F9E398B1DD17706D679338C38E3C63883CA56C6F2DA', USER, False, {'from': USER})
  print(fxs.balanceOf(USER))
  locking.getReward(USER, {'from': USER})
  print(fxs.balanceOf(USER))


with TemporaryFork():
  print(fxs.balanceOf(USER))
  locking.getReward(USER, {'from': USER})
  print(fxs.balanceOf(USER))
```

naddison36:
> We were talking about what would happen if, hypothetically, there weren't enough FXS in the tokens to allow collecting rewards. Since there's a possibility of this happening, we could set the reward collection to `false` on the withdraws. This way we can get our funds. The collect reward tokens method can handle actually getting the reward tokens, and it's since it is called by the harvester, we should still be able to get the reward tokens in the future. This also would not affect normal operation.
> 
> ```python
> from world import *
> locking = Contract.from_explorer("0xB4fdD7444E1d86b2035c97124C46b1528802DA35")
> fxs = Contract.from_explorer("0x3432b6a60d23ca0dfca7761b7ab56459d9c964d0")
> USER = "0x857E629E94D7252C8f71eBAF513548b796641DAc"
> 
> with TemporaryFork():
>   print(fxs.balanceOf(USER))
>   locking.withdrawLocked('0x7106AA15DDDC5E6A21389F9E398B1DD17706D679338C38E3C63883CA56C6F2DA', USER, True, {'from': USER})
>   print(fxs.balanceOf(USER))
> 
> 
> with TemporaryFork():
>   print(fxs.balanceOf(USER))
>   locking.withdrawLocked('0x7106AA15DDDC5E6A21389F9E398B1DD17706D679338C38E3C63883CA56C6F2DA', USER, False, {'from': USER})
>   print(fxs.balanceOf(USER))
>   locking.getReward(USER, {'from': USER})
>   print(fxs.balanceOf(USER))
> 
> 
> with TemporaryFork():
>   print(fxs.balanceOf(USER))
>   locking.getReward(USER, {'from': USER})
>   print(fxs.balanceOf(USER))
> ```

This is more than a hypothetical. There was not enough FXS in the rewards contract a few weeks ago that was preventing withdrawals of staked Curve LP tokens that have had their locks expire. The error message was “Not enough reward tokens available”. This can be replicated in this simulation https://dashboard.tenderly.co/nicka-origin/project/simulator/6816f018-ffb5-4c03-8d7c-e11a26a4daa5 

The reason for the failure was not enough FXS tokens in the Frax Staking on Convex Pool (FraxUnifiedFarm_ERC20_Convex_frxETH) contract at [0xb4fdd7444e1d86b2035c97124c46b1528802da35](https://etherscan.io/address/0xb4fdd7444e1d86b2035c97124c46b1528802da35).

The Frax team manually transferred 2,500 FXS to resolve this issue in tx [0x60505715c205cf449cf16bc31052542f0c55cb399cdf56cb40f80299d22748b0](https://etherscan.io/tx/0x60505715c205cf449cf16bc31052542f0c55cb399cdf56cb40f80299d22748b0).

Unfortunately, setting `claim_rewards` to `false` does not stop the rewards being collected. This is evident in tx [0xb50cf117802ddbd8e03ddd87c376b5f64cc2472b5dd3a4fa64f0b966fced2bf4](https://etherscan.io/tx/0xb50cf117802ddbd8e03ddd87c376b5f64cc2472b5dd3a4fa64f0b966fced2bf4). 
Rewards are still collected because `collectRewardsOnWithdrawalPaused` in FraxUnifiedFarm_ERC20_Convex_frxETH is set to `false` which means rewards will always be collected on withdraw. The `claim_rewards` parameter is only used if `collectRewardsOnWithdrawalPaused` is set to `true`.

The Frax team added more FXS tokens to the rewards pool within 24 hours of being notified to the problem via Telegram. It was concerning that this problem existed for many days meaning they don't have monitors for this situation.

The worst case is the Origin team purchases and manually transfers FXS to the rewards pool if the Frax team are unresponsive to adding more FXS.

DanielVF:
Confirming what Nick just said, and against what I had previously thought: When I transfer out the FXS tokens from the locking contract, I do see that `withdrawLocked` fails even with the get rewards parameter set to false.

```
fxs.transfer(ousd, fxs.balanceOf(locking), {'from': locking})
# ... Then same test code as before
```

contracts/contracts/oracle/PriceFeedPair.sol
```diff
@@ -0,0 +1,133 @@
...
+        // slither-disable-next-line unused-return
+        (, int256 _price0, , uint256 updatedAt0, ) = AggregatorV3Interface(
+            addressFeed0
+        ).latestRoundData();
+        // slither-disable-next-line unused-return
+        (, int256 _price1, , uint256 updatedAt1, ) = AggregatorV3Interface(
+            addressFeed1
+        ).latestRoundData();
+        updatedAt = Math.min(updatedAt0, updatedAt1);
+        price = _calculatePrice(_price0, _price1);
```
shahthepro, line 110: @sparrowDom @naddison36 Thinking about this, one potential concern I have is that `latestRound` of one feed can be way off from the other feed. Would it make sense to compare the `updatedAt` values and make sure both the updates aren't too far off?

Like if FXS<>USD was last updated 24h ago and ETH<>USD 1h ago, I don't think it'd be very reliable to use this price. Especially if there was too much of mark movement in the past 24h. More so, since this is a generic contract

shahthepro: Another way to do this for FXS<>ETH pair would be just use Uniswap V3's TWAP (by doing something similar to what we do on AuraPriceFeed): <https://blog.uniswap.org/uniswap-v3-oracles>

sparrowDom: We've discussed this on discord and figured out that the errors of multiple oracle price feeds stack. E.g. 2% error and 0.5% error price feeds result in 2.5% error price feed. We are OK keeping it like that for now, but still important to consider.

contracts/contracts/strategies/FraxConvexStrategy.sol
```diff
@@ -0,0 +1,473 @@
...
+
+    /**
+     * @dev Extends an existing lock for 7 days regardless of whether it has expired or not.
+     * Create a new lock if one doesn't exist and there is a positive target locked balance.
+     * Or add to the existing lock if the locked tokens is under the target locked balance.
+     * @param unlockedBalance The balance of the Frax Staked Convex LP tokens that are NOT locked.
+     * @param lockedBalance The balance of the Frax Staked Convex LP tokens that are locked.
+     * This includes any tokens in an expired lock.
+     */
+    function _lock(uint256 unlockedBalance, uint256 lockedBalance) internal {
```
DanielVF, line 198: I think we could simplify the if logic in this somewhat.

Although the overall behavior of when viewed in how this is called works out, inside the function it looks like it could lock funds that we actually wanted to withdraw. Also we duplicate a write.

The first top level if overlaps with another if inside the lower level. Makes it a bit hard to reason about, since some parts of the process are in the earlier if, and some in a later if.

We should extend the time AFTER adding new funds. Otherwise we would be earning less.

Would be nice to emit an event with a zero amount if we extend only, that we we can externally track when our lock ends from events.

Might be cleaner like so

```
//math
// and nopes w/returns
//then
if(lockKey == NO_KEY){
 // new lock
} else {
  // add funds, if needed
  // always extend for max power
}
// event?
```

This way we will always leave this function with a max lock, unless we are waiting to funds to come out, in which case we would do nothing.

Since only the strategist will be moving funds to this, perhaps we can skip the min amount checks.

naddison36: Thanks @DanielVF. Excellent review as always

> Although the overall behavior of when viewed in how this is called works out, inside the function it looks like it could lock funds that we actually wanted to withdraw. Also we duplicate a write.

I originally had logic to withdraw expired locks even if the amount was less than the target locked balance. This made the unlocking logic more complicated as it depended whether it was being called from `withdraw` or `updateLock` as to what it needed to do. If the Strategist wants to withdraw from the expired lock amount, they will need to set the target lock balance lower, call `updateLock` and then call `withdraw`.

> The first top level if overlaps with another if inside the lower level. Makes it a bit hard to reason about, since some parts of the process are in the earlier if, and some in a later if.

Agree. I've gone for a new structure that is hopefully easier to follow.

> We should extend the time AFTER adding new funds. Otherwise we would be earning less.

Good pickup on the rewards being better with extending a lock after adding more funds. I've made this change.

> Would be nice to emit an event with a zero amount if we extend only, that we we can externally track when our lock ends from events.

Good idea on emitting a `Lock` event with zero amount if the lock was extended with no funds added. I've made this change.

> Might be cleaner like so

I've changed to this structure. There was a bug in the old implementation that kept extending the lock even when the locked balance was over the target locked balance.

> Since only the strategist will be moving funds to this, perhaps we can skip the min amount checks.

I've debated with myself a few times whether we need the min amount check. As you point out, its the Strategist who controls the funds in the strategy so I've dropped the min lock amount.

Frax's `lockLonger` will revert if the new unlock time is not extended. That is, it will revert if you lock to the same time. This is why I had a `unlockTimestamp < newUnlockTimestamp` check which would allow `updateLock` to be called twice in a block. But reverting a tx that is doing a second `updateLock` in a block is not a bad thing so I've removed this check.

DanielVF: This function looks really good now!

contracts/contracts/strategies/FraxConvexStrategy.sol
```diff
@@ -0,0 +1,477 @@
...
+        // Do not extend lock time or add funds if already over the target locked balance
+        if (lockedBalance > targetLockedBalance) {
+            return;
+        }
+
+        uint64 newUnlockTimestamp = uint64(block.timestamp + LOCK_DURATION);
+        uint256 lockAmount = 0;
+
+        // Read targetLockedBalance into memory to save a second SLOAD
+        uint256 targetLockedBalanceMem = targetLockedBalance;
```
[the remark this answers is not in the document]
naddison36: Good point. done

contracts/contracts/strategies/FraxConvexStrategy.sol
```diff
@@ -0,0 +1,477 @@
...
+
+        // If the locked balance is under the target lock balance
+        if (lockedBalance < targetLockedBalanceMem) {
```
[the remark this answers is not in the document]
naddison36: yes, we still need this for the `lockedBalance == targetLockedBalanceMem` case

pandadefi:
What is the rationale behind the 7-day lock period? 
Looking at the `FraxUnifiedFarm` contract, a week-long lock will only increase the `_combined_weights` by 0.6%.

contracts/contracts/strategies/FraxConvexStrategy.sol
```diff
@@ -0,0 +1,477 @@
...
+        }
+
+        uint64 newUnlockTimestamp = uint64(block.timestamp + LOCK_DURATION);
+        uint256 lockAmount = 0;
+
+        // If the locked balance is under the target lock balance
+        if (lockedBalance < targetLockedBalanceMem) {
+            // Calculate the amount of Frax Staked Convex LP tokens to lock.
+            // The target lock balance ignoring the unlocked balance
+            uint256 targetLockAmount = targetLockedBalanceMem - lockedBalance;
```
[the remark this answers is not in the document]
DanielVF: The only part of the strategies that is gas critical is the balance checking (since it's used in every rebase). The rest of the strategy code is used only a few times a week, and I'd rather stay out of super gas optimizations in them, and lean towards safety.

contracts/contracts/strategies/FraxConvexStrategy.sol
```diff
@@ -0,0 +1,477 @@
...
+     * Frax Convex Locking contract. See dev note in contract Natspec for more details.
+     */
+    function collectRewardTokens()
+        external
+        override
+        onlyHarvester
+        nonReentrant
+    {
+        // Collect CRV, CVX and FXS rewards from staking contract
+        IFraxConvexStaking(fraxStaking).getReward(address(this));
```
pandadefi, line 452: The `getReward` function can be invoked using the harvester address, allowing for a reduction in one transfer for both the `fraxStaking` and `fraxLocking` contracts.

contracts/contracts/strategies/FraxConvexStrategy.sol
```diff
@@ -0,0 +1,477 @@
...
+     * @param unlockedBalance The balance of the Frax Staked Convex LP tokens that are NOT locked.
+     * @param lockedBalance The balance of the Frax Staked Convex LP tokens that are locked.
+     * This includes any tokens in an expired lock.
+     */
+    function _lock(uint256 unlockedBalance, uint256 lockedBalance) internal {
+        // Read targetLockedBalance into memory to save a second SLOAD
+        uint256 targetLockedBalanceMem = targetLockedBalance;
+
+        // Do not extend lock time or add funds if already over the target locked balance
+        if (lockedBalance > targetLockedBalanceMem) {
```
pandadefi, line 203: Should return if `targetLockedBalanceMem` and `lockedBalance` are both zero.

contracts/contracts/strategies/FraxConvexStrategy.sol
```diff
@@ -0,0 +1,477 @@
...
+     * Any unexpired locked tokens can be withdrawn once the lock has expired. These can not be
+     * relocked after expiry as the target lock balance is zero.
+     *
+     * This will revert with "shutdown" if the Frax Staking contract is shutdown (isShutdown = true).
+     * This will revert with "Withdrawals paused" if the Frax Convex Locking contracts has
+     * expired locked tokens and `withdrawalsPaused` = true.
+     * This will revert if there is not enough FXS rewards in the
+     * Frax Convex Locking contract. See dev note in contract Natspec for more details.
+     */
+    function _lpWithdrawAll() internal override {
```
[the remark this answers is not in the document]
naddison36: this is a good point. I think it makes sense to wrap the `_unlock` in `_lpWithdrawAll` in a try catch so the unlocked LP tokens can still be withdrawn.

contracts/contracts/strategies/FraxConvexStrategy.sol
```diff
@@ -0,0 +1,477 @@
...
+     * This assume the locked or unlocked Frax Staked Convex LP tokens equals the Curve LP tokens.
+     * The average is taken prevent the asset balances being manipulated by tilting the Curve pool.
+     *
+     * @dev This will NOT revert if the Frax Convex Staking contract has been shutdown or the
+     * Frax Convex Locking contract has been paused for staking or withdraws.
+     * @param _asset      Address of the asset
+     * @return balance    Total value of the asset in the platform
+     */
+    function checkBalance(address _asset)
+        public
```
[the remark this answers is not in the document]
naddison36: Is it still the case you get a gas saving for declaring a function `external`? Especially for a view function.

When I changed `checkBalance` on `FraxConvexStrategy` to external I didn't see a gas saving from when it was public.

Doing some quick testing in Remix with the following also shows no gas difference between the functions being declared `external` or `public`. What am I missing?

```Solidity
// SPDX-License-Identifier: MIT
pragma solidity ^0.8.0;

contract ExternalFuncs {
    address asset = 0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2;
    uint256 balance = 1e18;

    function checkBalance(address _asset) public view returns (uint256) {
        require(_asset == asset, "invalid asset");

        return balance;
    }

    function updateAsset(address _asset) public {
        asset = _asset;
    }
}
```

naddison36:
> What is the rationale behind the 7-day lock period? Looking at the `FraxUnifiedFarm` contract, a week-long lock will only increase the `_combined_weights` by 0.6%.
> 
> Edit: Locking funds is necessary to earn rewards. However, even a maximum lock of 3 years only enhances the combined_weight by 100%, indicating a relatively modest benefit for such a prolonged lock-in period.

More analysis is needed on how long the lock should be for.

contracts/contracts/strategies/FraxConvexStrategy.sol
naddison36: This is a valid case - especially after `withdrawAll`.

Since there are no locks and the lock amount is zero, `_lock` will exit on line 226. But I like your suggestion of returning earlier by checking if the target locked balance is zero.

I'll add to a PR

contracts/contracts/strategies/FraxConvexStrategy.sol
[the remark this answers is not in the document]
pandadefi: Gas savings are no longer available; the purpose is now solely for clarity.

contracts/contracts/strategies/FraxConvexStrategy.sol
naddison36: Raised PR https://github.com/OriginProtocol/origin-dollar/pull/1975 for this

pandadefi:
> > What is the rationale behind the 7-day lock period? Looking at the `FraxUnifiedFarm` contract, a week-long lock will only increase the `_combined_weights` by 0.6%.
> > Edit: Locking funds is necessary to earn rewards. However, even a maximum lock of 3 years only enhances the combined_weight by 100%, indicating a relatively modest benefit for such a prolonged lock-in period.
> 
> More analysis is needed on how long the lock should be for.

You could probably create multiple strategies with different durations.

contracts/contracts/strategies/FraxConvexStrategy.sol
naddison36: interesting, so Frax allow rewards to be delegated. In that case, the strategy can delegate rewards to the Harvester.

This requires a change to the Harvester which would be out of scope for this deployment. I'll add this Harvester feature request to the backlog

contracts/contracts/strategies/FraxConvexStrategy.sol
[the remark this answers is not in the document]
naddison36: I've put this in PR https://github.com/OriginProtocol/origin-dollar/pull/1976

contracts/contracts/strategies/ThreePoolStrategy.sol
```diff
@@ -143,7 +143,7 @@ contract ThreePoolStrategy is BaseCurveStrategy, CurveThreeCoinFunctions {
     }
 
     /**
-     * @dev Collect accumulated CRV and send to Vault.
+     * @dev Collect accumulated CRV rewards and send to the Harvester.
```
sparrowDom, line 146: thanks for catching these mistakes 👍

contracts/contracts/strategies/FraxConvexStrategy.sol
```diff
@@ -0,0 +1,493 @@
...
+     * @param _assets Addresses of supported assets. MUST be passed in the same
+     *                order as returned by coins on the pool contract, i.e. WETH, frxETH.
+     * @param _pTokens Address of the Curve pool for each asset.
+     */
+    function initialize(
+        address[] calldata _rewardTokenAddresses, // CRV + CVX + FXS
+        address[] calldata _assets,
+        address[] calldata _pTokens
+    ) external onlyGovernor initializer {
+        require(
```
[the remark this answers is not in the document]
sparrowDom: also maybe add a comment: 
currently all the assets in the curve pool need to be supported. Otherwise `checkBalance` would return incorrect results

contracts/contracts/strategies/FraxConvexStrategy.sol
```diff
@@ -0,0 +1,493 @@
...
+     *   lock any excess unlocked tokens.
+     *   If no lock exists, create a new lock.
+     *   If a lock exists, add tokens to the existing lock.
+     *
+     * This will revert with "Staking paused" if the Frax Convex Locking contract has
+     * paused staking (`stakingPaused` = true).
+     * This will revert if there is not enough FXS rewards in the
+     * Frax Convex Locking contract. See dev note in contract Natspec for more details.
+     */
+    function updateLock() external {
```
sparrowDom, line 365: maybe it sounds a bit paranoid, but I would add `nonReentrant` here

sparrowDom (commented):
**Internal Code Review**
[This is not yet a complete review - still need to check the tests]

**Deployment Considerations**
- creates a new price feed
- updates the OracleRouter with the new price feed
- sets new OracleRouter as new OETHVault price provider
- deploys a new strategy proxy
- deploys a new strategy implementation
- adds Uniswap config to sell FXS to the Harvester. Correct 1% FXS / ETH pool is configured
🟢  all above steps are correct and expected

**Attack**
This will never be a default strategy so the attacks to look for would be pool manipulating ones. And as long as we use vaultValue checker with strategist moves that check the expected values (inflated checkBalance would be noticed) manipulations shouldn't be possible

**Logic**
🟢 Are there bugs in the logic?
🟢 Correct usage of global & local variables. -> they might differentiate only by an underscore that can be overlooked (e.g. address vs _address).
🟢 Do public/external functions having asset parameter verify the asset is supported (where applicable)

**Tests**
⚪ Each publicly callable method has a test
⚪ Each logical branch has a test
⚪ Each require() has a test
⚪ Edge conditions are tested
⚪ If tests interact with AMM make sure enough edge cases (pool tilts) are tested.  Ideally with fuzzing.

**Flavor**
🟢 Could this code be simpler?
🟢 Could this code be less vulnerable to other code behaving weirdly?

**Overflow**
🟢 Never use "+" or "-", always use safe math or have contract compile in solidity version > 0.8
~~Check that all for loops use uint256~~

**Proxy**
🟢 Make sure proxy implementation contracts don't initialize variable state on variable declaration and do it rather in initialize function.

**Black magic**
🟢 Does not contain selfdestruct
🟢 Does not use delegatecall outside of proxying
 (If an implementation contract were to call delegatecall under attacker control, it could call selfdestruct the implementation contract, leading to calls through the proxy silently succeeding, even though they were failing.)
 ~~Address.isContract should be treated as if could return anything at any time, because that's reality.~~
 
**Dependencies**
🟢 Review any new contract dependencies thoroughly (e.g. OpenZeppelin imports) when new dependencies are added or version of dependencies changes.
 ~~If OpenZeppelin ACL roles are use review & enumerate all of them.~~
 Check OpenZeppelin [security vulnerabilities](https://github.com/OpenZeppelin/openzeppelin-contracts/security/advisories) and see if any apply to current PR considering the version of OpenZeppelin contract used.
 
**Deploy**
🟢 Check that any deployer permissions are removed after deploy
Authentication
🟢 Never use tx.origin
🟢 Check that every external/public function should actually be external
🟢 Check that every external/public function has the correct authentication

**Cryptographic code**
 ~~Contracts that roll their own crypto are terrifying~~
 ~~Note that a failed signature check will result in a 0x00 result. Make sure that the result throws if it returns this.~~
~~ Beware of signed data being used in a replay attack to other contracts.~~

**Gas problems**
 Contracts with for loops must have either:
 ~~A way to remove items~~
 🟢Can be upgraded to get unstuck
 ~~Size can only controlled by admins~~
🟢 Contracts with for loops must not allow end users to add unlimited items to a loop that is used by others or admins.

**External calls**
 🟢 Contract addresses passed in are validated
 ⚪  No unsafe external calls
🟠  Reentrancy guards on all state changing functions
 🟢 Still doesn't protect against external contracts changing the state of the world if they are called.
 
**Malicious behaviors**
 🟢 Could fail from stack depth problems (low level calls must require success)
 🟢 No slippage attacks (we need to validate expected tokens received)
 
 **Oracles, one of:**
 No oracles
⚪    Oracles can't be bent
⚪   If oracle can be bent, it won't hurt us.
 🟢 Don't call balanceOf for external contracts to determine what they will do, when they instead use internal accounting?
Ethereum
 🟢 Contract does not send or receive Ethereum.
 🟢 Contract has no payable methods.

sparrowDom:
This PR https://github.com/OriginProtocol/origin-dollar/pull/1981 fixes a test that had to be skipped

pandadefi (approved):

sparrowDom (commented):
Really amazing test coverage of the convex frax strategy 💪  

nit: There are so many different cases of locked, unlocked, staked, target locked amount variations it might be helpful to add a legend in the comments of the various cases covered.

DanielVF:
Closing this PR because we are moving to Simplified OETH.

PR #1866, opened 2023-10-17
Behaviour unit tests and other test infra changes

naddison36:
# Change Summary

## Contracts
- [Removed unused imports from Harvester contracts](https://github.com/OriginProtocol/origin-dollar/pull/1866/commits/491db5cd73bfd0310e943849b7752c7af08a55b7)
- [Added Withdrawal event to Aave and Compound withdrawAll](https://github.com/OriginProtocol/origin-dollar/pull/1866/commits/95e95292edd30afe4afee1c3a5e000a4511c908e)
- [Generalized4626Strategy to not emit Withdrawal event when no assets are withdrawn](https://github.com/OriginProtocol/origin-dollar/pull/1866/commits/a7ef9da3b42bcc925c901054639331486527b3d4)
- [Fix HarvesterAddressesUpdated event in setHarvesterAddress](https://github.com/OriginProtocol/origin-dollar/pull/1866/commits/6e62bef6ed7903f3869e34332ecab594f675590e)

## Contract Mocks
- [Fixes to Convex mock contracts](https://github.com/OriginProtocol/origin-dollar/pull/1866/commits/3170b2520240670ee6d0f14f3dd71281f85835de)
- [Fixes to Curve mock contracts](https://github.com/OriginProtocol/origin-dollar/pull/1866/commits/1a8e0bc23308372b29d9e821dd374c443845af38)
- [Fix withdraw on sfrxETH mock contract](https://github.com/OriginProtocol/origin-dollar/pull/1866/commits/a24477d3bd143288c4d0bb92d532323b1d1b9021)

# Test infrastructure
- [Added behaviour tests](https://github.com/OriginProtocol/origin-dollar/pull/1866/commits/0d1d545775d7fdc9ee1c180e7e4a83d4427f8c29)
- [Added IERC20Metadata which is used by the resolveAsset function](https://github.com/OriginProtocol/origin-dollar/pull/1866/commits/c2874bbe7777ba9f2cb5af26a26f0785ceeea76e)
- [Moved replacing of WETH contract to 000_mock](https://github.com/OriginProtocol/origin-dollar/pull/1866/commits/d6391ca8b4908ea1d5d4453f8a4bed593d234523)
- [Added support deploying contracts with linked libraries](https://github.com/OriginProtocol/origin-dollar/pull/1866/commits/d6391ca8b4908ea1d5d4453f8a4bed593d234523)
[...]

PR #1918, opened 2023-11-07
L-06 Test Coverage Gaps

sparrowDom:
Tests added: 
- [x] single asset deposit call should revert
- [x] multi asset deposit call should revert
- [x] setMaxDepositDeviation should work as expected
- [x] deposit contains duplicates -> [Done in L-01 PR](https://github.com/OriginProtocol/origin-dollar/pull/1913)
- [x] withdrawal contains duplicates -> [Done in L-01 PR ](https://github.com/OriginProtocol/origin-dollar/pull/1913)
- [x] deposit should fail with an unsupported asset
- [x] withdrawal should fail with an unsupported asset
- [x] deposit should work with any order of the assets
- [x] withdrawal should fail when incorrect order
- [x] deposit should work with small asset amounts
- [x] deposit should fail with too large amounts

Comments on "Suggested Invariants"

These are great points. We are working on this effort separately to achieve a generalized way of testing `checkBalance` and protocol profitability in cases of strategies utilizing pools (Balancer, Curve or Uniswap). We are primarily focusing on flash loan attacks which could heavily tilt the pool towards one asset, let the transaction (deposit or withdraw) execute and then tilt the pool back (sandwich attack). For now I think we should leave it out of scope for these audit fixes. 

Comments on "Absence of Contract-Level Tests (Unit Tests)"

Thanks for these points. There is also [an ongoing separate effort](https://github.com/OriginProtocol/origin-dollar/pull/1866) to create unit test suite where strategies are tested in a generalized way - testing common strategy behaviours. E.g. approving assets, governance actions, being able to deposit/withdraw only supported assets... For that reason I think this is out of scope for these fixes as well

**Audit notes:**
Several types of test coverage gaps were identified.

**Specific Line Coverage Gaps**
- In Balancer strategies, lines `deposit 1` and `deposit 2` are disabled.
- `setMaxDepositDeviation`

**Edge Case Gaps**
- Ensure that both `_deposit` and `_withdraw` function as expected or revert when
arguments are in an incorrect order or contain duplicates, unsupported assets, and/or
incorrect amounts. Although the external deposit and withdrawal functions are not
exposed in a way that will enable user inputs, a future change may expose or introduce a
bug with input validation.

**Suggested Invariants**
-Validate that running `checkBalance(address)` across all assets, converting the
results to units and summing them up arrives at the same value as `checkBalance()`
before and after various test scenarios.
- Ensure `totalValue()` remains unchanged by deposits/withdrawals with various
parameters. If it does change, these actions will leak value through rebasing. Although
this is validated in some tests, a more comprehensive validation of the invariant can
catch uncovered edge cases.

**Absence of Contract-Level Tests (Unit Tests)**
The codebase predominantly depends on fork testing which assesses contract integration,
dependencies, and on-chain state. However, it lacks contract-level (local "unit") tests.

While fork tests are of supreme importance, they do have a number of weaknesses that are
typically addressed via even larger local test suites:

- Fork tests require more maintenance due to their reliance on on-chain values (e.g., price
oracles). This makes large and thorough test suites more difficult to maintain.
- Fork tests are more difficult to write for edge cases and scenarios that deviate from the
current on-chain state. Thus, they tend to cover edge cases less thoroughly compared to
local test suites.
- Fork tests make testing internal logic in an isolated way much more difficult, resulting in
many edge cases in the implementation of base and abstract classes, internal functions,
or internally used libraries being insufficiently covered.
- Fork tests are usually slower to run and are thus typically run on CI instead of locally,
extending the length of the development cycle.

Consider creating a comprehensive suite of local tests that are quick to write and run, focusing
on edge cases and internal logic.

### DIFF
--- a/contracts/contracts/interfaces/balancer/IBalancerPool.sol
+++ b/contracts/contracts/interfaces/balancer/IBalancerPool.sol
@@ -8,4 +8,6 @@ interface IBalancerPool {
         external
         view
         returns (IRateProvider[] memory providers);
+
+    function inRecoveryMode() external view returns (bool);
 }
--- a/contracts/contracts/mocks/BalancerComposablePoolTestStrategy.sol
+++ b/contracts/contracts/mocks/BalancerComposablePoolTestStrategy.sol
@@ -15,13 +15,15 @@ contract BalancerComposablePoolTestStrategy is BalancerComposablePoolStrategy {
         BaseStrategyConfig memory _stratConfig,
         BaseBalancerConfig memory _balancerConfig,
         BaseMetaPoolConfig memory _metapoolConfig,
-        address _auraRewardPoolAddress
+        address _auraRewardPoolAddress,
+        uint256 _bptTokenPoolPosition
     )
         BalancerComposablePoolStrategy(
             _stratConfig,
             _balancerConfig,
             _metapoolConfig,
-            _auraRewardPoolAddress
+            _auraRewardPoolAddress,
+            _bptTokenPoolPosition
         )
     {}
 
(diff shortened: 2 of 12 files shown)


sparrowDom:
Merging this in @DanielVF will apply any fixes in separate branches if required

PR #1919, opened 2023-11-07
N-01 Gas Inefficiencies

sparrowDom:
**Audit notes:**
- Throughout the _withdraw and `withdrawAll`, functions the storage array `poolAssets`
and its length are loaded from the storage multiple times in each function. Consider
loading it into a memory array and then reading from there.
- Throughout the strategy, ++i is used in all for loops. Consider changing this i++
occurrence to ++i as well in order to maintain consistency and save some gas.

When performing these changes, aim to reach an optimal tradeoff between gas optimization
and readability. Having a codebase that is easy to understand reduces the chance of errors in
the future and improves transparency for the community.

### DIFF
--- a/contracts/contracts/strategies/balancer/BalancerMetaPoolStrategy.sol
+++ b/contracts/contracts/strategies/balancer/BalancerMetaPoolStrategy.sol
@@ -123,6 +123,8 @@ contract BalancerMetaPoolStrategy is BaseAuraStrategy {
             _strategyAssets.length == _strategyAmounts.length,
             "Array length missmatch"
         );
+        // store poolAssets storage variable into memory for gas optimizations
+        address[] memory poolAssets = poolAssets;
 
         uint256[] memory amountsIn = new uint256[](poolAssets.length);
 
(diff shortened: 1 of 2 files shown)


contracts/contracts/strategies/balancer/BalancerMetaPoolStrategy.sol
```diff
@@ -123,6 +123,8 @@ contract BalancerMetaPoolStrategy is BaseAuraStrategy {
             _strategyAssets.length == _strategyAmounts.length,
             "Array length missmatch"
         );
+        // store poolAssets storage variable into memory for gas optimizations
+        address[] memory poolAssets = poolAssets;
```
DanielVF, line 127: 🔴 I don't thinking we should shadow and use the same variable name. The local one should get an "_"

sparrowDom: Great point, fixed as suggested

DanielVF (commented):
Somehow I can't mark this as approved - no UI for it.

(Now I understand, it's merged)

sparrowDom:
Yup I've merged it to submit it to OZ team. Though I probably wouldn't need to be trigger happy with merging

PR #1923, opened 2023-11-08
L-02 Balancer Pool Recovery Mode Is Not Supported

sparrowDom:
Balancer pools may enter a recovery mode, allowing for a simplified exit code path in case the
usual exit path malfunctions. In order to be able to exit when this mode is enabled, userData
must be specifically encoded to signal the exit path type.

According to the recent Balancer post-mortem, the recovery mode is an important feature that
has already been used several times:
`It has saved us more than once, including during the present incident.`

Consider supporting this exit path via an additional withdrawal function.


DanielVF (approved):
Good to merge, will full verify when integrated.

PR #1926, opened 2023-11-13
Hot deploy fix

sparrowDom:
This fixed hot deploying vault-admin


sparrowDom:
Merging this in since it has only a developer environment consequences

PR #1927, opened 2023-11-13
Sparrow dom/balancer composable fixes

sparrowDom:
prettier


sparrowDom:
Merging this in, since these are just prettier changes.

PR #1929, opened 2023-11-13
L-01, L-02 & L-03 addendum

sparrowDom:
- changes `_exactBptInTokensOutIndex` & `_btpInExactTokensOutIndex` to `uint8` data type
- changes `RECOVERY_MODE_EXIT_KIND` type to `uint8`
- adds comment & test regarding duplicate withdrawal with 0 / near 0 amount


DanielVF (approved):
Nice to see that the cast to a uint is safe. From the solidity docs:

> The explicit conversion from integer checks at runtime that the value lies inside the range of the enum and causes a Panic error otherwise.

Can merge. Will fully verify when integrated. 👍

PR #1930, opened 2023-11-14
M-02 addendum

sparrowDom:
This PR adds tests for wstETH/WETH composable pool to confirm BPT position works as expected even when BPT token is in the non first position of the Balancer pool (returned by the `getPoolTokens` function on the Balancer Vault)


sparrowDom:
Merging this in, since these are only test file changes.

PR #1931, opened 2023-11-14
Add support for Balancer, Curve and Uniswap V3 to Harvester

shahthepro:
### Harvester:
- Updates Harvester to support Uniswap V3, Balancer and Curve exchange routes
- Adds new state variables to store additional swap data (`uniswapV2Path`, `uniswapV3Path`, `balancerPoolId` and `curvePoolData`)  
- Updates `RewardTokenConfig` struct to add `platform` enum (Doesn't affect storage slot) and rename `uniswapV2CompatibleAddr` to just `swapRouterAddr`
- Deduplicates the code in `Harvester` and `OETHHarvester`
- Removes unused event `UniswapUpdated`
- Adds events for swaps (`RewardTokenSwapped`) and reward distribution (`RewardProceedsTransferred`)
- Fixes a bug with `liquidationLimit` being set to 0
- Changes modifier on `setSupportedStrategy` to just `onlyGovernor` and removed the unused modifier `onlyVaultAndGovernor`
- Renames `setRewardsProceedsAddress` to `setRewardProceedsAddress` for consistency (the variable is named `rewardProceedsAddress`)

### Oracle Router
- OETHOracleRouter now inherits from `OracleRouterBase` instead of `OracleRouter`
- Moved `OETHOracleRouter` and `OracleRouterBase` to their own files
- Adds feed address for `BAL/ETH` and `AURA/ETH` pairs to `OETHOracleRouter`
- Changes the constructor of `OETHOracleRouter` to add AuraWETHPriceFeed as an arg

### Aura Price Feed
- A new contract that's compatible with `AggregatorV3Interface`
- Queries TWAP of two different time frames (recent 1h and the last 5m) and makes sure the price deviation isn't much before returning the price or reverting
- Strategist can pause the contract (causing it to always revert) but only Governor can unpause it
- Only governor can change the tolerance level
- Doesn't use a Proxy

### Swap Route changes
`081` deployment file will do the following changes to the swap routes (for better exchange rates)
- CRV & CVX on OETH will be harvested using Curve pools
- BAL & AURA on OETH will be harvested using Balancer pools
- All tokens on OUSD will be harvested using Uniswap V3

### Tests
- Adds behaviour test for Vault to check for any unconfigured reward tokens on all strategies
- Cleans up `MockUniswapRouter` to be a bit more simple and straight forward
- Adds for fork tests for OETH Harvester as well 
- Adds for tests for swaps as well
-------

Contract change checklist:
  - [ ] Unit tests pass


contracts/contracts/harvest/BaseHarvester.sol
```diff
@@ -58,38 +83,46 @@ abstract contract BaseHarvester is Governable {
...
+     **/
+    address public immutable baseTokenAddress;
+    uint256 public immutable baseTokenDecimals;
+
+    mapping(address => address[]) public uniswapV2Path;
+    mapping(address => bytes) public uniswapV3Path;
+    mapping(address => bytes32) public balancerPoolId;
+
+    struct CurvePoolData {
+        // TODO: Check if storing as single slot would help with gas here
```
DanielVF, line 101: It would help with gas.

contracts/contracts/harvest/BaseHarvester.sol
```diff
@@ -58,38 +83,46 @@ abstract contract BaseHarvester is Governable {
...
+        uint256 rewardTokenIndex;
+        uint256 baseTokenIndex;
+    }
+    mapping(address => CurvePoolData) public curvePoolData;
+
+    // "_usdtAddress" is set to Vault's address, but is really not used
```
DanielVF, line 107: This comment is outdated?

contracts/contracts/harvest/BaseHarvester.sol
```diff
@@ -221,15 +341,16 @@ abstract contract BaseHarvester is Governable {
     }
 
     /**
-     * @dev Swap all supported swap tokens for stablecoins via Uniswap.
+     * @dev Swap all supported reward tokens for
+     *      base token on the configured swap platforms
      */
     function swap() external onlyGovernor nonReentrant {
```
DanielVF, line 347: I think we could get rid of all the governor only harvest and swap methods.

We've never used them since adding incentivized harvesting, and I can't see any reason we would need to use them in the future.

I think this would be:

- `harvest()`
- `swap()`
- `harvestAndSwap()`
- `harvest(address)`
- `swapRewardToken(address)`
- `_harvest() internal`
- `_swap(address)`

contracts/contracts/harvest/BaseHarvester.sol
```diff
@@ -342,17 +463,253 @@ abstract contract BaseHarvester is Governable {
...
+        );
+        emit RewardTokenSwapped(
+            _swapToken,
+            baseTokenAddress,
+            tokenConfig.platform,
+            balance,
+            amountReceived
+        );
+
+        IERC20 desiredToken = IERC20(baseTokenAddress);
```
DanielVF, line 559: I think this would be much clearer as `baseToken` rather than `desiredToken`. Particularly `desiredTokenBalance` is confusing because it sounds like it's how many tokens we want, instead of how many we have.

contracts/contracts/harvest/BaseHarvester.sol
```diff
@@ -342,17 +463,253 @@ abstract contract BaseHarvester is Governable {
...
+
+        uint256 vaultBps = 1e4 - tokenConfig.harvestRewardBps;
+        uint256 rewardsProceedsShare = (desiredTokenBalance * vaultBps) / 1e4;
+
+        require(
+            vaultBps > tokenConfig.harvestRewardBps,
+            "Address receiving harvest incentive is receiving more rewards than the rewards proceeds address"
+        );
+
+        uint256 farmerShare = desiredTokenBalance - rewardsProceedsShare;
```
DanielVF, line 570: This share calculation is a bit convoluted, and is actually rounding against the vault, in favor of the farmer. 

I think this can much simpler if we just calculate the farmer share first, then the vault share by subtraction, then compare those two totals.

contracts/contracts/harvest/BaseHarvester.sol
```diff
@@ -342,17 +463,253 @@ abstract contract BaseHarvester is Governable {
...
+
+        desiredToken.safeTransfer(rewardProceedsAddress, rewardsProceedsShare);
+        desiredToken.safeTransfer(
+            _rewardTo,
+            farmerShare // Remaining share of the rewards
```
DanielVF, line 575: We might not need this comment.

contracts/contracts/harvest/BaseHarvester.sol
```diff
@@ -104,82 +137,169 @@ abstract contract BaseHarvester is Governable {
...
+        // Do some validation
+        require(
+            path.length >= 2 &&
+                path[0] == token &&
+                path[path.length - 1] == baseTokenAddress,
+            "Invalid Uniswap V2 path"
+        );
+    }
+
+    function _decodeUniswapV3Path(bytes calldata data, address token)
```
DanielVF, line 250: I think you might be able to uses slices on this calldata data + abi.decode to make this be a bit cleaner and simpler.

Off the top of my head: `abi.decode(data[0:20], (address))` or something.

contracts/contracts/harvest/BaseHarvester.sol
```diff
@@ -284,19 +406,8 @@ abstract contract BaseHarvester is Governable {
     }
```
DanielVF, line 406: We can also get rid of swapRewardToken, since it's governor only, and unlikely to be used.

DanielVF (changes requested):
# Harvester Code Review

## Requirements

Our new harvester will allow us to sell our tokens from our balancer strategies. 

Secondly, we want to revamp our token sales to allow more than just Uniswap v2 compatible swap paths. This means that we are switching to using balancer, Uniswap v3, Uniswap v2, and curve as harvesting strategies.

We need to make sure that we're actually using all these and that they are configured correctly and fork tested. 

## Easy Checks

#### Authentication

- [x] Never use tx.origin

#### Ethereum

#### Cryptographic code

#### Gas problems

    - [ ] A way to remove items

#### Black magic

#### Overflow

#### Proxy

#### Events
  - [ ] 🟥 setRewardsProceeds does not emit an event. We should set for completeness
  - [x] 🟤 TransferToken does not, but that's okay. Nothing in the contract changes.

## Medium Checks

#### Rounding
   - [x] 🟤 Min expected rounds down against the protocol, but that's okay. Not going to be worth pool maniputlating for a free 0.000001

#### Dependencies

#### External calls

- [x] Oracles, one of:
  - [ ] No oracles
  - [x] Oracles can't be bent

#### Tests

__skipped__

#### Deploy

## Thinking

#### Logic

  - [x] _Are there bugs in the logic?_
  - [] Correct usage of global & local variables. -> they might differentiate only by an underscore that can be overlooked (e.g. address vs _address).

#### Internal State

No internal state beyond config

#### Attack Scenarios

1. The worst case scenario for a harvester failure would be allow yield to be stolen by an attacker. This could happen for a few days or a week before we noticed. It's not that epic in the scope of things. 

2. The harvester could be misconfigured (particularly a mismatch betwen min harvest amounts, slippage, and incentives), which will block yield from being harvested until such time as it's reconfigured. Since these configurations are governance requiring, that means that we block yield for about a week.

3. We could forget to setup tokens to be harvested.

In order for an attacker to steal tokens:

1. There would either need to be a bug in our reward tokens calculation for the rewards.
2. They would need to have admin powers to change the rewards, 
3. An attacker could conduct some kind of swapping attack
4. An attacker could fool our oracle? Perhaps they could max out the differnce between the five minute price and the old price?

One disadvantage of using these minimum amounts on the swap protocols rather than checking our own balances is that instead of reverting ourselves afterwards is that this tells MEVBOTS exactly how much they can manipulate the pool in order to profit from us. I think we've seen in the past that it may be better to run these at zero and do our own check on the return balances. However this is much more imporant when swapping huge protocol principle sized amounts, and we can leave this here, and see what happens.

#### Flavor

I've already done a simplicity pass in a previous review. The one thing I see, though it may not be worth changing, is that we have a huge amount of seperate variables to configure our swaps with. It might be better to just have a single opaque set of bytes to configure the swap, which is decoded on demand when the swap happens. We don't have to do this though.

contracts/contracts/harvest/BaseHarvester.sol
shahthepro: Have changed it to pack it into a single slot

contracts/contracts/harvest/BaseHarvester.sol
shahthepro: Removed all of those now

contracts/contracts/harvest/BaseHarvester.sol
```diff
@@ -342,17 +463,253 @@ abstract contract BaseHarvester is Governable {
...
-     *       a registered price feed with the price provider.
-     * @param _swapToken Address of the token to swap.
+     * @dev Swap a reward token for the base token on the configured
+     *      swap platform. The token must have a registered price feed
+     *      with the price provider
+     * @param _swapToken Address of the token to swap
      * @param _rewardTo Address where to send the share of harvest rewards to
      */
-    function _swap(address _swapToken, address _rewardTo) internal virtual;
+    function _swap(address _swapToken, address _rewardTo) internal virtual {
```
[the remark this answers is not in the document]
shahthepro: Rather than return the swap function, I just changed it to do the actual swap to keep it a bit simple. But let me know if you feel there's a better way

contracts/contracts/harvest/BaseHarvester.sol
shahthepro: Apparently you couldn't use `abi.decode` with sliced bytes (it's treated as type `bytes memory slice`). So I had to do explicit conversions after slicing them. Should be much shorter and cleaner now, thanks for the tip

shahthepro:
@DanielVF Have added the event for `setRewardProceedsAddress` function as well

contracts/contracts/harvest/BaseHarvester.sol
```diff
@@ -12,37 +12,63 @@ import { IVault } from "../interfaces/IVault.sol";
...
     // Configuration properties for harvesting logic of reward tokens
     struct RewardTokenConfig {
         // Max allowed slippage when swapping reward token for a stablecoin denominated in basis points.
         uint16 allowedSlippageBps;
         // Reward when calling a harvest function denominated in basis points.
         uint16 harvestRewardBps;
-        /* Address of Uniswap V2 compatible exchange (Uniswap V2, SushiSwap).
-         */
-        address uniswapV2CompatibleAddr;
+        // Address of compatible exchange router (Uniswap V2, SushiSwap).
```
sparrowDom, line 64: nit: also Curve and Balancer in the comment

contracts/contracts/harvest/BaseHarvester.sol
```diff
@@ -58,38 +84,46 @@ abstract contract BaseHarvester is Governable {
...
+    // Cached decimals for `baseTokenAddress`
+    uint256 public immutable baseTokenDecimals;
+
+    // Uniswap V2 path for reward tokens using Uniswap V2 Router
+    mapping(address => address[]) public uniswapV2Path;
+    // Uniswap V3 path for reward tokens using Uniswap V3 Router
+    mapping(address => bytes) public uniswapV3Path;
+    // Pool ID to use for reward tokens on Balancer
+    mapping(address => bytes32) public balancerPoolId;
+    // Packed indices of assets on the Curve pool
```
sparrowDom, line 104: mental note: we should have a test that confirms these 4 mappings are correctly configured. Not only by checking values in fork tests but also fork tests doing queries to e.g. Curve and check that `curvePoolIndices` return correct values for queried assets. 

In short if we only change the reward token address to use a different Curve pool to swap out the reward token,(and forget to update `curvePoolindices`) a test should fail.

contracts/contracts/harvest/BaseHarvester.sol
```diff
@@ -12,37 +12,63 @@ import { IVault } from "../interfaces/IVault.sol";
...
     struct RewardTokenConfig {
         // Max allowed slippage when swapping reward token for a stablecoin denominated in basis points.
         uint16 allowedSlippageBps;
         // Reward when calling a harvest function denominated in basis points.
         uint16 harvestRewardBps;
-        /* Address of Uniswap V2 compatible exchange (Uniswap V2, SushiSwap).
-         */
-        address uniswapV2CompatibleAddr;
+        // Address of compatible exchange router (Uniswap V2, SushiSwap).
+        address swapRouterAddr;
```
sparrowDom, line 65: with Balancer this is a BalancerVault and with Curve this is a Curve pool. Would a `swapPlatform` or `swapProtocol` or `swapProtocolEntryPoint` be more accurate?

contracts/contracts/harvest/BaseHarvester.sol
```diff
@@ -12,37 +12,63 @@ import { IVault } from "../interfaces/IVault.sol";
...
         /* When true the reward token is being swapped. In a need of (temporarily) disabling the swapping of
          * a reward token this needs to be set to false.
          */
         bool doSwapRewardToken;
+        // Platform to use for Swapping
+        SwapPlatform platform;
```
sparrowDom, line 71: good call adding SwapPlatform next to a bool. Since a bool and uint8 (underlying enum implementation) will be packed together

sparrowDom: nit: we could do gas optimisation and pack `doSwapRewardToken` & `platform` next to the 2 `uint16` at the top of the struct. To save on storage reads. But that would require updating reward token config of all of the tokens. Which already might be the case will all these additions?

contracts/contracts/harvest/BaseHarvester.sol
```diff
@@ -99,87 +133,208 @@ abstract contract BaseHarvester is Governable {
...
+        } else if (_platform == SwapPlatform.Balancer) {
+            balancerPoolId[_tokenAddress] = _decodeBalancerPoolId(swapData);
+        } else if (_platform == SwapPlatform.Curve) {
+            curvePoolIndices[_tokenAddress] = _decodeCurvePoolIndices(
+                swapData,
+                newRouterAddress,
+                _tokenAddress
+            );
+        } else {
+            revert("Unknown SwapPlatfrom");
```
sparrowDom, line 221: We could replace this with a solidity [custom error](https://ethereum.stackexchange.com/questions/101782/requirecondition-message-vs-revert-with-a-custom-error-which-is-better-a) that reduces the gas cost. @naddison36 has started doing this in [OETH Oracle PR](https://github.com/OriginProtocol/origin-dollar/pull/1815/files#diff-42dbcd9f1d9adfaa494e307193605f6ddf3ed0c17dbf0dfd03f5fc97514dc603R97) and I think it is a better approach. 

also Codecov is complaining this line isn't tested

contracts/contracts/harvest/BaseHarvester.sol
```diff
@@ -99,87 +133,208 @@ abstract contract BaseHarvester is Governable {
...
+        returns (address[] memory path)
+    {
+        (path) = abi.decode(data, (address[]));
+
+        // Do some validation
+        require(
+            path.length >= 2 &&
+                path[0] == token &&
+                path[path.length - 1] == baseTokenAddress,
+            "Invalid Uniswap V2 path"
```
sparrowDom, line 256: you could do custom errors here for each of the 3 conditions and pass the incorrect value to an error as a parameter

contracts/contracts/harvest/BaseHarvester.sol
```diff
@@ -99,87 +133,208 @@ abstract contract BaseHarvester is Governable {
...
+        // Do some validation
+        require(
+            address(uint160(bytes20(data[0:20]))) == token,
+            "Invalid Reward Token in swap path"
+        );
+        require(
+            address(uint160(bytes20(data[path.length - 20:]))) ==
+                baseTokenAddress,
+            "Invalid Base Token in swap path"
+        );
```
sparrowDom, line 284: same possibility for custom errors here

contracts/contracts/harvest/BaseHarvester.sol
```diff
@@ -99,87 +133,208 @@ abstract contract BaseHarvester is Governable {
...
+     */
+    function _decodeBalancerPoolId(bytes calldata data)
+        internal
+        pure
+        returns (bytes32 poolId)
+    {
+        (poolId) = abi.decode(data, (bytes32));
+
+        // Do some validation
+        require(poolId != bytes32(0), "Invalid Balancer Pool ID");
```
sparrowDom, line 301: and here (custom errors)

contracts/contracts/harvest/BaseHarvester.sol
```diff
@@ -99,87 +133,208 @@ abstract contract BaseHarvester is Governable {
...
+            (uint256, uint256)
+        );
+
+        // Pack it into one slot for gas savings
+        // Shifts left rewardTokenIndex by 128 bits and then
+        // adds baseTokenIndex to avoid overlaps
+        indices = (rewardTokenIndex << 128) + baseTokenIndex;
+
+        ICurvePool pool = ICurvePool(poolAddress);
+        require(
```
sparrowDom, line 331: this check is awesome

contracts/contracts/harvest/BaseHarvester.sol
```diff
@@ -99,87 +133,208 @@ abstract contract BaseHarvester is Governable {
...
+            token == pool.coins(rewardTokenIndex),
+            "Invalid Reward Token Index"
+        );
+        require(
```
sparrowDom, line 335: same custom errors comment as above. (there are more in the file, I won't write for each to not flood with the comments too much)

contracts/contracts/harvest/BaseHarvester.sol
```diff
@@ -214,40 +369,9 @@ abstract contract BaseHarvester is Governable {
...
-    function harvestAndSwap() external onlyGovernor nonReentrant {
-        _harvest();
-        _swap(rewardProceedsAddress);
-    }
-
-    /**
-     * @dev Collect reward tokens for a specific strategy.
-     * @param _strategyAddr Address of the strategy to collect rewards from
-     */
-    function harvest(address _strategyAddr) external onlyGovernor nonReentrant {
```
sparrowDom, line 243: good riddance for all of these :)

contracts/contracts/harvest/BaseHarvester.sol
```diff
@@ -328,31 +428,255 @@ abstract contract BaseHarvester is Governable {
...
+    function _swapWithCurve(
+        address poolAddress,
+        address swapToken,
+        uint256 amountIn,
+        uint256 minAmountOut
+    ) internal returns (uint256 amountOut) {
+        uint256 packedIndices = curvePoolIndices[swapToken];
+
+        // Indices are stored in a single uint256 slot, unpack them
+        // The first 128-bit contains rewardTokenIndex
```
sparrowDom, line 670: ha this is nice :)

contracts/contracts/harvest/BaseHarvester.sol
```diff
@@ -328,31 +428,255 @@ abstract contract BaseHarvester is Governable {
...
-            }
+    function _swap(address _swapToken, address _rewardTo) internal virtual {
+        RewardTokenConfig memory tokenConfig = rewardTokenConfigs[_swapToken];
+
+        /* This will trigger a return when reward token configuration has not yet been set
+         * or we have temporarily disabled swapping of specific reward token via setting
+         * doSwapRewardToken to false.
+         */
+        if (!tokenConfig.doSwapRewardToken) {
+            return;
```
sparrowDom, line 445: would be so nice to bubble this information up somehow. An error might be wrong, if multiple tokens are being harvested. Maybe an event saying token swap disabled for a specific token?

contracts/contracts/harvest/BaseHarvester.sol
```diff
@@ -328,31 +428,255 @@ abstract contract BaseHarvester is Governable {
...
+        }
+
+        address priceProvider = IVault(vaultAddress).priceProvider();
+        uint256 balance = IERC20(_swapToken).balanceOf(address(this));
+
+        if (balance == 0) {
```
sparrowDom, line 451: nit: we could move this above fetching the `priceProvider` for an earlier more gas efficient exit when balance is 0.

contracts/contracts/harvest/BaseHarvester.sol
```diff
@@ -328,31 +428,255 @@ abstract contract BaseHarvester is Governable {
...
+
+        /* This will trigger a return when reward token configuration has not yet been set
+         * or we have temporarily disabled swapping of specific reward token via setting
+         * doSwapRewardToken to false.
+         */
+        if (!tokenConfig.doSwapRewardToken) {
+            return;
+        }
+
+        address priceProvider = IVault(vaultAddress).priceProvider();
```
sparrowDom, line 448: idea: we could query for the priceProvider in a higher up function e.g. `_harvestAndSwap` and pass it down, so it is not fetched multiple times (for some gas savings). It will dirty up the code to some extent though

contracts/contracts/harvest/BaseHarvester.sol
```diff
@@ -328,31 +428,255 @@ abstract contract BaseHarvester is Governable {
...
+        uint256 oraclePrice = IOracle(priceProvider).price(_swapToken);
+
+        // Oracle price is 1e18
+        uint256 minExpected = (balance *
+            (1e4 - tokenConfig.allowedSlippageBps) * // max allowed slippage
+            oraclePrice).scaleBy(
+                baseTokenDecimals,
+                Helpers.getDecimals(_swapToken)
+            ) /
+            1e4 / // fix the max slippage decimal position
```
sparrowDom, line 469: You can use `StableMath`'s `mulTruncate` && `mulTruncateScale` that will make this operation a bit more readable:
```
balance
     .mulTruncateScale(1e4 - tokenConfig.allowedSlippageBps, 1e4)
     .mulTruncate(oraclePrice)
    .scaleBy(
        baseTokenDecimals,     
        Helpers.getDecimals(_swapToken)
    )
```

contracts/contracts/harvest/BaseHarvester.sol
```diff
@@ -328,31 +428,255 @@ abstract contract BaseHarvester is Governable {
...
+            baseTokenAddress,
+            tokenConfig.platform,
+            balance,
+            amountReceived
+        );
+
+        IERC20 baseToken = IERC20(baseTokenAddress);
+        uint256 baseTokenBalance = baseToken.balanceOf(address(this));
+        uint256 farmerShare = (baseTokenBalance *
+            tokenConfig.harvestRewardBps) / 1e4;
```
sparrowDom, line 491: nit: `mulTruncateScale` here as well

sparrowDom (commented):
Amazing work @shahthepro . A lot of the code really cleans up nicely while you still managed to capture great complexity. 

I am submitting a partial review. Will continue

contracts/contracts/harvest/BaseHarvester.sol
sparrowDom: nvm, you do check for this one

contracts/contracts/harvest/BaseHarvester.sol
```diff
@@ -12,37 +12,63 @@ import { IVault } from "../interfaces/IVault.sol";
...
+        address indexed rewardToken,
+        address indexed swappedInto,
+        SwapPlatform platform,
+        uint256 amountIn,
+        uint256 amountOut
+    );
+    event RewardProceedsTransferred(
+        address indexed token,
+        address farmer,
+        uint256 protocolShare,
```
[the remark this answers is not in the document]
shahthepro: > one could expect this to represent a percentage of the total

Isn't that the case here? 😅 

Harvester swaps `x` token for `y` token and send a percentage of that (dictated by `harvestRewardBps`) to the farmer and the protocol? 

May be something like `farmerFee` (or `farmerReward`) and `protocolYield` would work?

contracts/contracts/harvest/BaseHarvester.sol
shahthepro: Yeah, the check is there for Curve. For Uniswap V2 and V3, I just parse the path and make sure the source and destination token are as expected. For Balancer, I couldn't do much there, other than null check. I could probably query the pool and check if the pool tokens match. Will make that change

contracts/contracts/harvest/BaseHarvester.sol
shahthepro: Since we already use `platform` for the `SwapPlatform` enum, something like `swapPlatformAddr` would be consistent ig, will make this change. Also, changing the `platform` to `swapPlatform` to avoid any confusion

contracts/contracts/harvest/BaseHarvester.sol
shahthepro: Isn't it already in a single slot? `uint16,uint16,address,bool` would use `16 + 16 + 160 + 8` bits, so 200 bits (or 25 bytes), which is still under one slot.

I thought of changing the order of the elements at first as well, but we check the old router address and remove any allowance when using `setRewardTokenConfig` method. Changing this struct would mean that we might potentially cause it to fail on mainnet after the upgrade since the value of the old router address might not be `address(0)` or the right one due to change in slot

contracts/contracts/harvest/BaseHarvester.sol
shahthepro: Good point about Custom Errors, Will change to use them. 

As for CodeCov, this line is unreachable since we revert in `setRewardTokenConfig` method if the `swapPlatform` is something else. I just added it here as a defence

contracts/contracts/harvest/BaseHarvester.sol
shahthepro: I'm not entirely sure if we should do that. I don't think we would have any use of that event on analytics side either?

But it won't hurt to add it, can do it if you think there would be a need for it

contracts/contracts/harvest/BaseHarvester.sol
shahthepro: Good point, I just realized, I could as well get rid of the variable `priceProvider` and make it inline

contracts/contracts/harvest/BaseHarvester.sol
shahthepro: It's only fetched once though

`harvestAndSwap` can no longer be called for multiple strategies. And `_harvest` doesn't use priceProvider. Only `_swap` does. 

The only other place we fetch and use the priceProvider is under `setRewardTokenConfig` method

contracts/contracts/harvest/BaseHarvester.sol
shahthepro: I'm not sure if chaining these calls will work as expected due to rounding. But I can probably nest them

shahthepro: Okay, we do the `balance * slippage * price` multiplication first and then scale it by base token's decimals and then adjust the decimals. 

Changing the order gives us a less accurate value due to rounding. But let me test this a bit more. For now, I've not committed this change

shahthepro: So, one potential bug with your suggested approach is that it won't work for tokens with decimals of <18 and a really low price. Let's say `NotARugToken` uses 6 decimals, if the Oracle price is less than 1e6, it's gonna be rounded down to 0. But this is probably an edge case.

Right now none of our reward tokens use <18 decimals, but I don't wanna make that assumption for future as well. And I guess I'd prefer a bit higher accuracy than readability for this line. But let me know if you have any other suggestions on this one

contracts/contracts/harvest/BaseHarvester.sol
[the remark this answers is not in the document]
sparrowDom: `farmerFee` and `protocolYield` sound great to me.

Sorry for not being clear enough with the previous message. I meant that when something is called `share` it could also be interpreted as being expressed in percentage points rather than part of the absolute number.

contracts/contracts/harvest/BaseHarvester.sol
sparrowDom: Yup that sounds like a great suggestion you could call `getPoolTokens` on the Balancer Vault with the given poolId to see if all the tokens there are expected.

I think that ETH (that can be converted to WETH) will be shown as a `0xEeeeEe....`

contracts/contracts/harvest/BaseHarvester.sol
sparrowDom: `swapPlatformAddr` is great

contracts/contracts/harvest/BaseHarvester.sol
```diff
@@ -58,129 +97,277 @@ abstract contract BaseHarvester is Governable {
...
+
+        if (poolId == bytes32(0)) {
+            revert EmptyBalancerPoolId();
+        }
+
+        IBalancerVault bVault = IBalancerVault(balancerVault);
+
+        // Note: this reverts if token is not a pool asset
+        // slither-disable-next-line unused-return
+        bVault.getPoolTokenInfo(poolId, token);
```
sparrowDom, line 332: This is a great way to perform a check. Does it work for WETH as well?

contracts/contracts/harvest/BaseHarvester.sol
shahthepro: Ah, I get it now. Will make that change

contracts/contracts/harvest/BaseHarvester.sol
```diff
@@ -58,139 +97,287 @@ abstract contract BaseHarvester is Governable {
...
+    uint256 public immutable baseTokenDecimals;
+
+    // Uniswap V2 path for reward tokens using Uniswap V2 Router
+    mapping(address => address[]) public uniswapV2Path;
+    // Uniswap V3 path for reward tokens using Uniswap V3 Router
+    mapping(address => bytes) public uniswapV3Path;
+    // Pool ID to use for reward tokens on Balancer
+    mapping(address => bytes32) public balancerPoolId;
+    // Packed indices of assets on the Curve pool
+    mapping(address => uint256) public curvePoolIndices;
```
DanielVF, line 118: We could avoid packing, unpacking, and casting these in the code, by just using a struct with two 128 sized members? The rest would then be handled by the complier.

Would also document what's happening.

contracts/contracts/harvest/BaseHarvester.sol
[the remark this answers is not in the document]
DanielVF: This works.

contracts/contracts/harvest/BaseHarvester.sol
```diff
@@ -12,37 +12,76 @@ import { IVault } from "../interfaces/IVault.sol";
...
-         */
-        address uniswapV2CompatibleAddr;
+        // Address of compatible exchange protocol (Uniswap V2/V3, SushiSwap, Balancer and Curve).
+        address swapPlatformAddr;
         /* When true the reward token is being swapped. In a need of (temporarily) disabling the swapping of
          * a reward token this needs to be set to false.
          */
         bool doSwapRewardToken;
+        // Platform to use for Swapping
+        SwapPlatform swapPlatform;
```
DanielVF, line 84: 👍 Looks like this should upgrade okay, since it won't be clobbering anything, even though added into the middle of the struct.
image

DanielVF (dismissed):
Post review fixes look good. Just need to change that packing to use a struct.

contracts/contracts/harvest/BaseHarvester.sol
shahthepro: Have changed the code to use a struct now

contracts/contracts/harvest/BaseHarvester.sol
```diff
@@ -58,139 +97,285 @@ abstract contract BaseHarvester is Governable {
...
      */
     function setRewardTokenConfig(
         address _tokenAddress,
-        uint16 _allowedSlippageBps,
-        uint16 _harvestRewardBps,
-        address _uniswapV2CompatibleAddr,
-        uint256 _liquidationLimit,
-        bool _doSwapRewardToken
+        RewardTokenConfig calldata tokenConfig,
+        bytes calldata swapData
```
sparrowDom, line 176: these 2 local variables should be prepended with an underscore.

contracts/contracts/harvest/BaseHarvester.sol
```diff
@@ -58,139 +97,285 @@ abstract contract BaseHarvester is Governable {
...
 
+    /**
+     * @dev Decodes the data passed into Uniswap V2 path and validates
+     *      it to make sure the path is for `token` to `baseToken`
+     *
+     * @param data Ecnoded data passed to the `setRewardTokenConfig`
+     * @param token The address of the reward token
+     * @return path The validated Uniswap V2 path
+     */
+    function _decodeUniswapV2Path(bytes calldata data, address token)
```
sparrowDom, line 265: underscore for local vars..

contracts/contracts/harvest/BaseHarvester.sol
```diff
@@ -58,139 +97,285 @@ abstract contract BaseHarvester is Governable {
...
+
+    /**
+     * @dev Decodes the data passed into Uniswap V3 path and validates
+     *      it to make sure the path is for `token` to `baseToken`
+     *
+     * @param data Ecnoded data passed to the `setRewardTokenConfig`
+     * @param token The address of the reward token
+     * @return path The validated Uniswap V3 path
+     */
+    function _decodeUniswapV3Path(bytes calldata data, address token)
```
sparrowDom, line 296: also underscore here

contracts/contracts/harvest/BaseHarvester.sol
```diff
@@ -58,139 +97,285 @@ abstract contract BaseHarvester is Governable {
...
+    /**
+     * @dev Decodes the data passed to Balancer Pool ID
+     *
+     * @param data Ecnoded data passed to the `setRewardTokenConfig`
+     * @return poolId The pool ID
+     */
+    function _decodeBalancerPoolId(
+        bytes calldata data,
+        address balancerVault,
+        address token
```
sparrowDom, line 326: underscore these as well

contracts/test/vault/harvester.fork-test.js
```diff
@@ -14,22 +18,122 @@ describe("ForkTest: Harvester", function () {
...
+      await expect(tx).to.emit(balancerREthStrategy, "RewardTokenCollected");
+      await expect(tx).to.emit(oethHarvester, "RewardTokenSwapped");
+      await expect(tx).to.emit(oethHarvester, "RewardProceedsTransferred");
+
+      // Should've transferred everything to Harvester
+      expect(await bal.balanceOf(balancerREthStrategy.address)).to.equal("0");
+      expect(await aura.balanceOf(balancerREthStrategy.address)).to.equal("0");
+
+      // Should've transferred swapped WETH to Dripper
+      expect(await weth.balanceOf(oethDripper.address)).to.be.gte(wethBefore);
```
sparrowDom, line 105: This line should pass even when no WETH has been collected (because of gt**e**  -> making it greater or equal) 

Lets check for balance being greater than some margin: 
```
expect(await weth.balanceOf(oethDripper.address)).to.be.gte(wethBefore.add(parseUnits("0.2", 18)));
```

contracts/deploy/081_upgrade_harvester.js
```diff
@@ -0,0 +1,315 @@
...
+        {
+          contract: cOETHHarvester,
+          signature: setRewardTokenConfigSig,
+          args: [
+            addresses.mainnet.BAL,
+            {
+              allowedSlippageBps: 300,
+              harvestRewardBps: 200,
+              swapPlatform: 2, // Balancer
+              swapPlatformAddr: "0xBA12222222228d8Ba445958a75a0704d566BF2C8",
```
sparrowDom, line 157: also this

contracts/contracts/harvest/BaseHarvester.sol
```diff
@@ -307,52 +437,279 @@ abstract contract BaseHarvester is Governable {
...
+        emit RewardTokenSwapped(
+            _swapToken,
+            baseTokenAddress,
+            tokenConfig.swapPlatform,
+            balance,
+            amountReceived
+        );
+
+        IERC20 baseToken = IERC20(baseTokenAddress);
+        uint256 baseTokenBalance = baseToken.balanceOf(address(this));
```
sparrowDom, line 524: We could do a sanity check here: 
```
if (baseTokenBalance < amountReceived) {
    revert ...
}
if (minExpected > amountReceived) {
    revert ...
}
```
It might seem like paranoid defensive programming, but I would argue that it is ok to check this most critical part of the transaction.

sparrowDom:
# Full Code Review
#### Attack
The amount of funds stolen from the harvester at any point can only be equal to the amount of rewards our strategies have earned. Which makes the size of the honey pot relatively small and probably not interesting for the attackers. Still if harvester were to be attacked, my main concerns are governance miss-configuration of reward token configs or slippage check fails. 

We have the former well covered in the fork tests. And for the latter I have added [a comment](https://github.com/OriginProtocol/origin-dollar/pull/1931/files#r1406607650) here. 

#### Logic

  - [x] _Are there bugs in the logic?_
  - 🟠  Correct usage of global & local variables. -> they might differentiate only by an underscore that can be overlooked (e.g. address vs _address).
    - there are a few occurrences where local vars don't have underscores


#### Tests

  - [x]  Each logical branch has a test
  - 🟠  Each require() has a test
    - There is no test for InvalidSwapPlatform in the `setRewardTokenConfig`
  - [] If tests interact with AMM make sure enough edge cases (pool tilts) are tested. Ideally with fuzzing.

#### Flavor

I think the code is simple considering the amount of complexity it is handling. 

#### Overflow

#### Proxy

#### Black magic

#### Dependencies

#### Deploy

#### Authentication

- [x] Never use tx.origin

#### Cryptographic code

#### Gas problems

    - [x] A way to remove items

#### External calls

- [x] Malicious behaviors
- [x] Oracles, one of:
  - [ ] No oracles
  - [x] Oracles can't be bent


#### Ethereum

contracts/contracts/harvest/BaseHarvester.sol
shahthepro: Since all of the swap protocols we intend to use accept `minExpected` value this seems a bit redundant to me. Besides `baseTokenBalance < amountReceived` wouldn't be the right way to do this, since someone can send USDT/WETH directly to Harvester to bypass this check. If we were do a balance diff it's gonna increase the gas cost for farmers.

But I get your point about being extra safe. @naddison36 @DanielVF any thoughts here?

shahthepro: I just checked the Collateral Swap code and we have a similar check there: https://github.com/OriginProtocol/origin-dollar/blob/master/contracts/contracts/vault/VaultAdmin.sol#L219-L222

But then Collateral Swap is not as frequently used as the Harvester. Anyway, I'm adding this check as a precaution but gonna do a `baseTokenBalance > minExpected` check, since if we don't trust the swap protocol to do slippage protection, we shouldn't trust the value it returns either.

As for someone sending tokens directly to Harvester, there's no incentive for someone to do that from what I can tell. So, I'm just gonna leave a comment about it

naddison36: It depends on whether we trust the protocol that is doing the swap. eg Uniswap and SushiSwap.
If we trust them we can use the `amountReceived` and trust that they will revert if `amountReceived < minExpected`.

If we don't trust them, we have to get the balance before and after the swap to check the received tokens are > `minExpected`. This adds significant gas costs.

contracts/contracts/harvest/BaseHarvester.sol
```diff
@@ -307,52 +437,279 @@ abstract contract BaseHarvester is Governable {
...
+        uint256 farmerFee = baseTokenBalance.mulTruncateScale(
```
naddison36, line 525: we should only reward the farmer on the amount swapped. ie  `amountReceived`
The farmer should not be rewarded if there is already USDT or WETH in the Harvester.

contracts/contracts/harvest/BaseHarvester.sol
```diff
@@ -307,52 +437,279 @@ abstract contract BaseHarvester is Governable {
...
+     */
+    function _swapWithCurve(
+        address poolAddress,
+        address swapToken,
+        uint256 amountIn,
+        uint256 minAmountOut
+    ) internal returns (uint256 amountOut) {
+        CurvePoolIndices memory indices = curvePoolIndices[swapToken];
+
+        amountOut = ICurvePool(poolAddress).exchange(
```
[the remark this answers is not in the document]
shahthepro: Will add it as a comment. In those cases, the call to `exchange` itself should fail since there won't be a function matching the signature we expect

contracts/contracts/harvest/BaseHarvester.sol
shahthepro: Thanks, that's good point, will incorporate that. cc @DanielVF @sparrowDom

contracts/contracts/harvest/BaseHarvester.sol
sparrowDom: The harvester after each harvest shouldn't hold any USDT / WETH. If someone were to send USDT / WETH to bypass the `baseTokenBalance < amountReceived` check I think it is ok. Since it doesn't exploit the protocol in any way. And profit made in the favorable swap would be un-done by sending funds to the harvester. 

I think the 2 new checks are a good compromise between security and gas usage: not doing another `balanceOf` of USDT / WETH before the swap.

contracts/contracts/harvest/BaseHarvester.sol
sparrowDom: Excellent point. There shouldn't be any USDT / WETH lingering on the harvester, but that doesn't make the point any less true.

PR #1935, opened 2023-11-17
Discovery day: check for ".only" and prevent commit

sparrowDom:
This PR adds a husky enabled git pre-commit hook that scans all test (unit & fork) files for the presence of ".only" modifier that is used for local testing. If found the commit fails. This is an example of a failure: 

Screenshot 2023-11-17 at 23 20 43

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,0 +1,25 @@
+#!/bin/sh
+
+RED='\033[0;31m'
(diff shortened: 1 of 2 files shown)


shahthepro (approved):

thanhvietmkb123456789 (commented):
tuyệt vời quá mọi người

PR #1936, opened 2023-11-19
add gas inspection plugin

sparrowDom:
To measure the gas usage of any transaction wrap it into `GetGas` globally defined function. Example: 
```
	const gas = await GetGas(await cEvilContract.connect(daniel).approveAllTokens());
	console.log('Gas used: ' + gas);
```

example output: 
Screenshot 2023-11-20 at 00 22 16

### DIFF
--- a/contracts/hardhat.config.js
+++ b/contracts/hardhat.config.js
@@ -12,6 +12,7 @@ require("hardhat-deploy-ethers");
 require("hardhat-gas-reporter");
 require("solidity-coverage");
 require("@openzeppelin/hardhat-upgrades");
+require("hardhat-gas-trackooor");
 
 require("./tasks/tasks");
 const { accounts } = require("./tasks/account");
--- a/contracts/README.md
+++ b/contracts/README.md
@@ -265,3 +265,15 @@ yarn test:fork:coverage
 ```
 
 The CI will upload the coverage reports to Codecov if they complete successfully.
+
+# Utils
+
+## Transaction gas inspector
+
+To measure the gas usage of any transaction wrap it into `GetGas` globally defined function. Example: 
+```
+	const gas = await GetGas(await cEvilContract.connect(daniel).approveAllTokens());
+	console.log('Gas used: ' + gas);
+```
+
+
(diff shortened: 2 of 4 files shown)


contracts/hardhat.config.js
```diff
@@ -12,6 +12,7 @@ require("hardhat-deploy-ethers");
 require("hardhat-gas-reporter");
 require("solidity-coverage");
 require("@openzeppelin/hardhat-upgrades");
+require("hardhat-gas-trackooor");
```
shahthepro, line 15: It looks like this plugin hasn't been updated in about 18 months. The GitHub repo has no activity either. Would be better if we can use something that's being actively maintained

naddison36:
How is this better than `hardhat-gas-reporter` which can be enabled with
`export REPORT_GAS=true`

Example output from running the OETH Oracle Router fork tests
```
·----------------------------------------------------|---------------------------|-------------|-----------------------------·
|                Solc version: 0.8.7                 ·  Optimizer enabled: true  ·  Runs: 200  ·  Block limit: 30000000 gas  │
·····················································|···························|·············|······························
|  Methods                                                                                                                   │
································|····················|·············|·············|·············|·············|················
|  Contract                     ·  Method            ·  Min        ·  Max        ·  Avg        ·  # calls    ·  eur (avg)    │
································|····················|·············|·············|·············|·············|················
|  MockOETHOracleRouterNoStale  ·  price             ·      22044  ·      41364  ·      34511  ·          4  ·            -  │
································|····················|·············|·············|·············|·············|················
|  OETHOracle                   ·  latestRoundData   ·          -  ·          -  ·      28373  ·          1  ·            -  │
································|····················|·············|·············|·············|·············|················
|  OETHOracle                   ·  setOracleUpdater  ·          -  ·          -  ·      30157  ·          2  ·            -  │
································|····················|·············|·············|·············|·············|················
|  OETHOracleUpdater            ·  addPrice          ·      77520  ·      89434  ·      87051  ·          5  ·            -  │
·-------------------------------|--------------------|-------------|-------------|-------------|-------------|---------------·
```

PR #1954, opened 2023-12-05
Enable FXS harvesting

sparrowDom:
- add a PriceFeedPair contract that is able to combine 2 chainlink price feeds.
- add that price feed to OETH Oracle
- configure harvester to support harvesting of FXS token


contracts/contracts/oracle/PriceFeedPair.sol
```diff
@@ -0,0 +1,129 @@
...
+
+    error PriceFeedAddressError(address _address);
+
+    constructor(
+        address _addressFeed0,
+        address _addressFeed1,
+        bool _reverseFeed0,
+        bool _reverseFeed1
+    ) {
+        if (_addressFeed0 == address(0)) {
```
shahthepro, line 45: A check to ensure `_addressFeed0 != _addressFeed1` would be nice to have here

contracts/deploy/082_frax_convex_strategy.js
```diff
@@ -108,7 +154,33 @@ module.exports = deploymentWithGovernanceProposal(
           signature: "setHarvesterAddress(address)",
           args: [cHarvester.address],
         },
-        // TODO add Harvest of FXS
+        // 4. Set OETH Oracle Router on Vault
+        {
+          contract: cVault,
+          signature: "setPriceProvider(address)",
+          args: [dOETHRouter.address],
```
shahthepro, line 161: Makes me wonder if we should start using a proxy for Oracle Router contracts as well

contracts/test/abi/AggregatorInterfaceAbi.json
```diff
@@ -0,0 +1,113 @@
+[
```
shahthepro, line 1: I don't think this is needed here, since we already have the interface file under `/contracts/interfaces` directory. You can just do `ethers.getContractAt("AggregatorV3Interface", address)` and it'll work

contracts/utils/addresses.js
```diff
@@ -61,6 +61,9 @@ addresses.mainnet.CVXRewardsPool = "0x7D536a737C13561e0D2Decf1152a653B4e615158";
 addresses.mainnet.CVXLocker = "0x72a19342e8F1838460eBFCCEf09F6585e32db86E";
 // Frax
 addresses.mainnet.FXS = "0x3432B6A60D23Ca0dFCa7761B7ab56459D9C964D0";
+addresses.mainnet.Frax = "0x853d955acef822db058eb8505911ed77f175b99e";
+addresses.mainnet.FraxCommunityBags =
```
shahthepro, line 65: Is this used somewhere?

contracts/contracts/oracle/PriceFeedPair.sol
sparrowDom: Sure I can add that in.

I think we are going to need to fork test each feed configuration anyway, since combining 2 feeds and making it possible to reverse any of them leaves much room for error.

naddison36 (approved):
Looks good.,

I've raised PR https://github.com/OriginProtocol/origin-dollar/pull/1955 with documentation updates

contracts/deploy/082_frax_convex_strategy.js
sparrowDom: hmm actually thinking more about it, we would need a governor role on the proxy, which is another small addition to complexity. 

As long as the router doesn't have any complex state to maintain where a proxy contract really pays off, I'd vote not to have it as a proxy

contracts/test/abi/AggregatorInterfaceAbi.json
sparrowDom: that is a great point thanks! removed

contracts/deploy/082_frax_convex_strategy.js
shahthepro: Good point, We can just let it be simple for now

contracts/utils/addresses.js
sparrowDom: Yes in the _fixture to fund the FXS tokens. Those can unfortunately not be gained by either calling `mint` on the FXS contract or by altering the storage slots. Since right after the slot manipulation / minting the transfer fails because vote count check triggers.

PR #1959, opened 2023-12-18
Strategist run Dec 20

sparrowDom:
Strategist run december 20th allocating 2026 eth to AMO

### DIFF
--- a/brownie/README.md
+++ b/brownie/README.md
@@ -82,7 +82,7 @@ brownie run spell_apy --network hardhat
 ```
 
 ### For strategist runs (IMPORTANT!)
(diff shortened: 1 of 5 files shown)


brownie/README.md
```diff
@@ -59,6 +59,9 @@ Or run brownie scripts:
 brownie run spell_apy --network hardhat
 ```
 
+### For strategist runs (IMPORTANT!)
```
sparrowDom, line 62: @shahthepro I like your suggestion how we should find a better more system wide solution to enforce this behaviour. From what I gathered the networks can only be configured outside this project scope (in `~/.brownie/network-config.yaml` file) and there isn't an elegant solution that could be applied within this project folder. 

For now I have only added this documentation, though we should find a better solution in the long run.

Maybe as a part of brownie installation we should make everyone install anvil and configure brownie network in their home folder to use the anvil node.

shahthepro: Agreed, we can probably add this as a task on Notion, so that we don't lose track of it. I'm merging this in

sparrowDom: good thinking, added a taks here: https://www.notion.so/originprotocol/Brownie-dev-environment-configuration-458d73f324f84871863d5f0243a64196

PR #1968, opened 2024-01-03
Deploy 83

sparrowDom:
Deploy 83


shahthepro (dismissed):

sparrowDom:
Merging this in to unblock any fork tests on master.

Explanation: if proposal Id is not present a fork environment will try to re-create the proposal with the same parameters as one already exists and the governor will throw an error: "Proposal already exists"

PR #1975, opened 2024-01-15
Exist `_lock` earlier if target lock balance is zero

naddison36:
Changes to Frax Convex Strategy from yAudit comments.

* Exit `_lock` function earlier if target and locked balances are both zero
* Added a fork test to measure the gas usage of `checkBalance`


naddison36:
This PR has been replaced with https://github.com/OriginProtocol/origin-dollar/pull/1976

PR #1981, opened 2024-01-29
make fxs tokens usable for tests

sparrowDom:
This fixes a test. 

FXS when performing a transfer also creates voting snapshots. This uses FXS owner tokens instead of setting ERC20 balance via snapshots.


naddison36:
closing as we are no longer doing the Convex locking strategy

PR #1995, opened 2024-03-11
Simplify Mint & Redeem

shahthepro:

Contract change checklist:
  - [ ] Unit tests pass


contracts/contracts/vault/OETHVaultCore.sol
```diff
@@ -1,26 +1,137 @@
...
+        }
+
+        // Burn OETH from user
+        oUSD.burn(msg.sender, _amount);
+
+        // Until we can prove that we won't affect the prices of our assets
+        // by withdrawing them, this should be here.
+        // It's possible that a strategy was off on its asset total, perhaps
+        // a reward token sold for more or for less than anticipated.
+        uint256 totalUnits = 0;
```
DanielVF, line 119: We can move the totalUnits math inside the `maxSupplyDiff > 0` since it is not used otherwise.

DanielVF:
We should add a gap[] storage array to VaultStorage, so that we have future room to add variables there. Otherwise, any future variable adds there will clobber the inherited OETHValueCore variables.

contracts/contracts/vault/OETHVaultCore.sol
```diff
@@ -1,26 +1,137 @@
...
+                wethAssetIndex = i;
+                break;
+            }
+        }
+
+        require(allAssets[wethAssetIndex] == weth, "Invalid WETH Asset Index");
+    }
+
     // @inheritdoc VaultCore
     function mint(
```
DanielVF, line 43: This should match the internal `_mint()` and use those modifiers, the same way we override the internal methods on redeem and on calculate values. This way we don't have an internal `_mint()` hanging around that does something completely different than the public mint.

contracts/contracts/vault/OETHVaultCore.sol
```diff
@@ -1,26 +1,137 @@
...
         address _asset,
         uint256 _amount,
-        uint256 _minimumOusdAmount
+        uint256
```
DanielVF, line 46: Let's actually check these minimum amounts on both mints and redeems. This keeps integrations from using bogus high values, which would prevent us from using this for real at some point in the future.

contracts/contracts/vault/OETHVaultCore.sol
```diff
@@ -1,26 +1,137 @@
...
+        // Ensure that the WETH index is cached
+        uint256 wethIndex = wethAssetIndex;
+        require(allAssets[wethIndex] == weth, "WETH Asset index not cached");
+
+        outputs = new uint256[](allAssets.length);
+        outputs[wethIndex] = _amount;
+    }
+
+    // @inheritdoc VaultCore
+    function _redeem(uint256 _amount, uint256) internal virtual override {
```
DanielVF, line 91: Let's add back in the redeem fee.

It looks like roadmap wise, going zero fee on these instant redeems is going to quite far out, and it would be nice to go ahead and have this simplified code in use before that.

contracts/contracts/vault/OETHVaultCore.sol
```diff
@@ -88,50 +98,48 @@ contract OETHVaultCore is VaultCore {
...
+        override
+    {
         // Override `VaultCore._redeem` to simplify it. Gets rid of oracle
         // usage and looping through all assets for LST-mix redeem. Instead
-        // does a simple WETH-only redeem with zero fee.
+        // does a simple WETH-only redeem.
         emit Redeem(msg.sender, _amount);
 
-        if (IERC20(weth).balanceOf(address(this)) >= _amount) {
+        require(_amount > 0, "Amount must be greater than 0");
```
DanielVF, line 111: We don't want to change behavior here from the past code and revert where we didn't revert before - can you replace this require with an `if(_amount==0){ return; }` ?

contracts/contracts/vault/OETHVaultCore.sol
```diff
@@ -1,26 +1,147 @@
...
+        // for backward compatibility
+
+        // Calculate redeem fee
+        if (redeemFeeBps > 0) {
+            uint256 redeemFee = _amount.mulTruncateScale(redeemFeeBps, 1e4);
+            _amount = _amount - redeemFee;
+        }
+
+        // Ensure that the WETH index is cached
+        uint256 wethIndex = wethAssetIndex;
```
DanielVF, line 93: Lets do `_wethAssetIndex` so it's clear it's a more local version of wethAssetIndex, rather than having a name which similar, but scoped differently.

contracts/contracts/vault/OETHVaultCore.sol
```diff
@@ -1,26 +1,147 @@
...
+
     // @inheritdoc VaultCore
-    function mint(
+    function _mint(
         address _asset,
         uint256 _amount,
         uint256 _minimumOusdAmount
-    ) external virtual override whenNotCapitalPaused nonReentrant {
+    ) internal virtual override {
         require(_asset == weth, "Unsupported asset for minting");
```
DanielVF, line 48: We've removed the check that this is an asset marked as supported. This check is indeed not needed for both minting and redeeming with hardcoded WETH. asset.isSupported is also ignored in balance checks as well.

I think we should just add a note to the asset.isSupported struct parameter that it is ignored in minting, redeeming, and balance checking on OETH.

contracts/contracts/vault/OETHVaultCore.sol
```diff
@@ -1,26 +1,147 @@
...
+        if (_amount == 0) {
+            return;
+        }
+
+        // Amount excluding fees
+        uint256 amountMinusFee = _calculateRedeemOutputs(_amount)[
+            wethAssetIndex
+        ];
+
+        if (_minimumUnitAmount > 0) {
```
DanielVF, line 120: We can remove this if statement. It's now just an extra gas costs, since the computation inside costs the same as it does in the worst case, and in the normal case we have two if's getting called. A zero _minimumUnitAmount should still pass all checks.

DanielVF:
# Simple Mints and Redeems - Code Review

## Requirements

We want to move OETH to mint and redeem only to from WETH, rather than a basket of assets, and to use no oracles for assets.

## Easy Checks

#### Authentication

- [x] Never use tx.origin

#### Ethereum

#### Cryptographic code

_No crypto_

#### Gas problems

    - [ ] A way to remove items

#### Black magic

#### Overflow

#### Proxy

#### Events
🟣 We don't emit in an event on `cacheWETHAssetIndex` but that's not even a configuration, more like an internal implementation detail. Should be okay.

## Medium Checks

#### Rounding
🟣 We round fees down, in favor of the user. This means that at small amounts, there is no fee. This is probably okay since the amounts (less than a billionth of an eth) mean that gas factors primarily matter. This is not a change from how it was before.

#### Dependencies

_All dependencies are currently used.

#### External calls

- [x] No Malicious behaviors
- [x] Oracles, one of:
  - [x] No oracles
  - [ ] Oracles can't be bent

#### Tests

_Not reviewed_

#### Deploy

#### Downstream

_No external protocol dependancies added_

## Thinking

#### Logic

🟠 Some changes needed. Comments already posted inline with the code.

  - [x] _Are there bugs in the logic?_

We need to either have moved to WETH backed only, or be good with only redeeming to WETH before going live.

#### Internal State

🟢 wethAssetIndex needs to be set correctly or redeems / redeem calls will revert. Code does a job of not just checking that it is set, but also that it is set correctly. 

🟢 cacheWETHAssetIndex will revert if there is no WETH asset. Correct.
#### Attack

This code covers mints and redeems, so its correct operation is critical. However it is a major simplification and much easier to reason about.

Mints should never mint more than 1 OETH per 1 WETH. Given that both the mint and the asset transferFrom use the same `amount` variable these should be 1:1 from the vault's POV.

Redeems should never burn less than 1 ETH per 1 WETH. Given that the post redeem fee amount is calculated using subtraction of uints, the amount should always be equal or less. 

Exact behavior of the OETH token itself is out of the scope of this PR (and into the fuzzing campaign), but we know that redeems will always have fees for now, until we've made other changes.

#### Flavor

Good!

```

File: contracts/contracts/vault/OETHVaultCore.sol (hunk 7 above)
Comment (shahthepro) [reply contracts/contracts/vault/OETHVaultCore.sol:48]: Have added the comment in `VaultStorage` where the `Asset` struct is defined

File: contracts/contracts/vault/OETHVaultCore.sol [hunk 9]
```diff
@@ -1,26 +1,148 @@
...
+
+        if (IERC20(weth).balanceOf(address(this)) >= amountMinusFee) {
+            // Use Vault funds first if sufficient
+            IERC20(weth).safeTransfer(msg.sender, amountMinusFee);
+        } else {
+            address strategyAddr = assetDefaultStrategies[weth];
+            if (strategyAddr != address(0)) {
+                // Nothing in Vault, but something in Strategy, send from there
+                IStrategy strategy = IStrategy(strategyAddr);
+                strategy.withdraw(msg.sender, weth, amountMinusFee);
```
pandadefi, line 136: It is probably necessary to first withdraw from the vault and then withdraw what is missing from the strategy. Otherwise `revert("Liquidity error");` might trigger even if there is enough weth on vault and strategy combined.

contracts/contracts/vault/OETHVaultCore.sol
```diff
@@ -1,26 +1,148 @@
...
+        }
+
+        // Mint oTokens
+        oUSD.mint(msg.sender, _amount);
+
+        // Transfer the deposited coins to the vault
+        IERC20(_asset).safeTransferFrom(msg.sender, address(this), _amount);
+
+        // Auto-allocate if necessary
+        if (_amount >= autoAllocateThreshold) {
```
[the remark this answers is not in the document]
DanielVF: The goal is to allow small sized mints to not spend the very large gas to allocate to strategies. `_allocate()` can be fairly expensive currently.

pandadefi:
@shahthepro You can have a look at my comments.

contracts/contracts/vault/OETHVaultCore.sol
DanielVF: Agree that we might want to move to a more sophisticated system here eventually (if we are spending the gas to pull funds out, we might want to exactly rebalance our vault buffer for example.) For the time being I think we'll keep this PR small and focused and continue with the code previous code that the vault has been using.

DanielVF (approved):

contracts/contracts/vault/OETHVaultCore.sol
pandadefi: Codewise it's not a lot more sophisticated to pull vault balance if there is any and then take what is missing from the strategy. You are saving gas by not having to `_allocate` the pending weth if any.

contracts/contracts/vault/OETHVaultCore.sol
```diff
@@ -1,26 +1,148 @@
...
+
     // @inheritdoc VaultCore
-    function mint(
+    function _mint(
         address _asset,
         uint256 _amount,
         uint256 _minimumOusdAmount
-    ) external virtual override whenNotCapitalPaused nonReentrant {
+    ) internal virtual override {
         require(_asset == weth, "Unsupported asset for minting");
```
naddison36, line 48: I would have preferred the old error message `Asset is not supported` was kept for backward compatibility

contracts/contracts/vault/OETHVaultCore.sol
```diff
@@ -1,26 +1,148 @@
...
 
     constructor(address _weth) {
         weth = _weth;
     }
 
+    /**
+     * @dev Caches WETH's index in `allAssets` variable.
+     *      Reduces gas usage by redeem by caching that.
+     */
+    function cacheWETHAssetIndex() external onlyGovernor {
```
naddison36, line 30: If we adding `wethAssetIndex` for gas improvements, it'd be much better if `wethAssetIndex` was an immutable. The problem with that, though, is this the immutable has to be set in the constructor where the storage slots from the proxy contract are unavailable. `wethAssetIndex` would have to passed in as a constructor param and then checked in an `initialize` or `upgrade` function that is called as part of the proxy upgrade.

contracts/contracts/vault/OETHVaultCore.sol
```diff
@@ -1,26 +1,148 @@
...
 /**
  * @title OETH VaultCore Contract
  * @author Origin Protocol Inc
  */
 contract OETHVaultCore is VaultCore {
+    using SafeERC20 for IERC20;
+    using StableMath for uint256;
+
     address public immutable weth;
+    uint256 public wethAssetIndex;
```
naddison36, line 20: I assumed `wethAssetIndex` is added to `OETHVaultCore` and not `VaultStorage` as `wethAssetIndex` is specific to OETH. It's not relevant for OUSD which also uses  `VaultStorage`. I think this is fair enough.

contracts/contracts/vault/OETHVaultCore.sol
```diff
@@ -1,26 +1,148 @@
...
+
+        // Calculate redeem fee
+        if (redeemFeeBps > 0) {
+            uint256 redeemFee = _amount.mulTruncateScale(redeemFeeBps, 1e4);
+            _amount = _amount - redeemFee;
+        }
+
+        // Ensure that the WETH index is cached
+        uint256 _wethAssetIndex = wethAssetIndex;
+        require(
```
naddison36, line 94: this is another read from storage (2,100 gas). Is it really needed when the same check is in `cacheWETHAssetIndex` which sets the `wethAssetIndex` storage variable?

contracts/contracts/vault/OETHVaultCore.sol
```diff
@@ -1,26 +1,148 @@
...
+            allAssets[_wethAssetIndex] == weth,
+            "WETH Asset index not cached"
+        );
+
+        outputs = new uint256[](allAssets.length);
```
naddison36, line 99: `allAssets.length` is another storage read. This value could be packed into the same slot as `wethAssetIndex` so there is only one slot read for both `wethAssetIndex` and the number of vaults assets.
I don't think it's worth doing this change at this late stage, though.

contracts/contracts/vault/OETHVaultCore.sol
```diff
@@ -1,26 +1,148 @@
...
+        if (_amount >= autoAllocateThreshold) {
+            _allocate();
+        }
+    }
+
+    // @inheritdoc VaultCore
+    function _calculateRedeemOutputs(uint256 _amount)
+        internal
+        view
+        virtual
```
naddison36, line 78: nit `virtual` is not needed here as `OETHVaultCore` is not inherited. Only the `override` is required

contracts/contracts/vault/OETHVaultCore.sol
```diff
@@ -1,26 +1,148 @@
...
+        );
+
+        outputs = new uint256[](allAssets.length);
+        outputs[_wethAssetIndex] = _amount;
+    }
+
+    // @inheritdoc VaultCore
+    function _redeem(uint256 _amount, uint256 _minimumUnitAmount)
+        internal
+        virtual
```
naddison36, line 106: nit `virtual` is not needed here as `OETHVaultCore` is not inherited. Only the `override` is required

contracts/contracts/vault/OETHVaultCore.sol
```diff
@@ -1,26 +1,148 @@
...
+    }
+
     // @inheritdoc VaultCore
-    function mint(
+    function _mint(
         address _asset,
         uint256 _amount,
         uint256 _minimumOusdAmount
-    ) external virtual override whenNotCapitalPaused nonReentrant {
+    ) internal virtual override {
```
naddison36, line 47: nit `virtual` is not needed here as `OETHVaultCore` is not inherited. Only the `override` is required

naddison36 (commented):
Given this PR has already been merged to master and contract deployed, there's no need to make any changes at this late stage.

contracts/contracts/vault/OETHVaultCore.sol
shahthepro: I initially thought of making it an immutable, yeah. However, we have #2012 PR that's gonna go live in the near future. When we remove other assets from the Vault, it can change the asset index. I assumed we do not want to redeploy the Vault everytime we remove an asset

contracts/contracts/vault/OETHVaultCore.sol
shahthepro: Yep, just defensive programming keeping the #2012 PR in mind

contracts/contracts/vault/OETHVaultCore.sol
naddison36: Good point, we don't want to redeploy the contract every time we remove an asset

PR #1996, opened 2024-03-12
FrxEth Redeem Strategy

DanielVF:
This will allow us to convert amounts of FrxEth to WETH via the Frax ETH redemption queue.

### DIFF
--- a/contracts/contracts/strategies/FrxEthRedeemStrategy.sol
+++ b/contracts/contracts/strategies/FrxEthRedeemStrategy.sol
@@ -1,11 +1,6 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.0;
 
-/**
- * @title Compound Strategy
- * @notice Investment strategy for Compound like lending platforms. eg Compound and Flux
- * @author Origin Protocol Inc
- */
 import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
 import { IERC721Receiver } from "@openzeppelin/contracts/token/ERC721/IERC721Receiver.sol";
 
@@ -27,6 +22,7 @@ interface IFraxEtherRedemptionQueue {
 /**
  * @title Frax ETH Redeem Strategy
  * @notice This strategy redeems Frax ETH for ETH via the Frax Eth Redemption Queue contract
+ * @author Origin Protocol Inc
  */
 contract FrxEthRedeemStrategy is InitializableAbstractStrategy {
     IWETH9 private constant weth =
--- a/contracts/contracts/strategies/FrxEthRedeemStrategy.sol
+++ b/contracts/contracts/strategies/FrxEthRedeemStrategy.sol
@@ -116,11 +116,14 @@ contract FrxEthRedeemStrategy is InitializableAbstractStrategy {
      * @param _amount Amount of assets to withdraw
      */
     function withdraw(
+        // solhint-disable-next-line no-unused-vars
         address _recipient,
+        // solhint-disable-next-line no-unused-vars
         address _asset,
+        // solhint-disable-next-line no-unused-vars
         uint256 _amount
     ) external override onlyVault nonReentrant {
-        // Does nothing - all redeems need to be called manually by the 
+        // Does nothing - all redeems need to be called manually by the
         // strategist via redeemTickets
         require(false, "use redeemTickets() instead");
     }
@@ -213,9 +216,13 @@ contract FrxEthRedeemStrategy is InitializableAbstractStrategy {
     }
 
     function onERC721Received(
+        // solhint-disable-next-line no-unused-vars
         address operator,
+        // solhint-disable-next-line no-unused-vars
         address from,
+        // solhint-disable-next-line no-unused-vars
         uint256 tokenId,
+        // solhint-disable-next-line no-unused-vars
         bytes calldata data
     ) external returns (bytes4) {
         return IERC721Receiver.onERC721Received.selector;
--- a/contracts/contracts/strategies/FrxEthRedeemStrategy.sol
+++ b/contracts/contracts/strategies/FrxEthRedeemStrategy.sol
@@ -207,8 +207,10 @@ contract FrxEthRedeemStrategy is InitializableAbstractStrategy {
     {
         if (_asset == address(weth)) {
             return outstandingRedeems;
-        } else {
+        } else if (_asset == address(frxETH)) {
             return 0;
+        } else {
+            revert("Unexpected asset address");
         }
     }
 
--- a/contracts/contracts/strategies/FrxEthRedeemStrategy.sol
+++ b/contracts/contracts/strategies/FrxEthRedeemStrategy.sol
@@ -165,7 +165,7 @@ contract FrxEthRedeemStrategy is InitializableAbstractStrategy {
         );
     }
 
-    function _abstractSetPToken(address _asset, address _pToken)
+    function _abstractSetPToken(address, address)
         internal
         override
     {
--- a/contracts/contracts/strategies/FrxEthRedeemStrategy.sol
+++ b/contracts/contracts/strategies/FrxEthRedeemStrategy.sol
@@ -1,12 +1,9 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.0;
 
-import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
 import { IERC721Receiver } from "@openzeppelin/contracts/token/ERC721/IERC721Receiver.sol";
 
-import { BaseCompoundStrategy, InitializableAbstractStrategy } from "./BaseCompoundStrategy.sol";
-import { IComptroller } from "../interfaces/IComptroller.sol";
-import { IERC20 } from "../utils/InitializableAbstractStrategy.sol";
+import { IERC20, InitializableAbstractStrategy } from "../utils/InitializableAbstractStrategy.sol";
 import { IWETH9 } from "../interfaces/IWETH9.sol";
 import { IVault } from "../interfaces/IVault.sol";
 
--- a/contracts/contracts/strategies/FrxEthRedeemStrategy.sol
+++ b/contracts/contracts/strategies/FrxEthRedeemStrategy.sol
@@ -168,7 +168,9 @@ contract FrxEthRedeemStrategy is InitializableAbstractStrategy {
     function _abstractSetPToken(address _asset, address _pToken)
         internal
         override
-    {}
+    {
+        revert("No pTokens are used");
+    }
 
     /**
      * @notice Withdraw all assets from this strategy, and transfer to the Vault.
--- a/contracts/contracts/strategies/FrxEthRedeemStrategy.sol
+++ b/contracts/contracts/strategies/FrxEthRedeemStrategy.sol
@@ -59,6 +59,9 @@ contract FrxEthRedeemStrategy is InitializableAbstractStrategy {
         safeApproveAllTokens();
     }
 
+    /**
+     * @notice deposit() function not used for this strategy. Use depositAll() instead.
+     */
     function deposit(address, uint256) public override onlyVault nonReentrant {
         // This method no longer used by the VaultAdmin, and we don't want it
         // to be used by VaultCore.
@@ -66,7 +69,7 @@ contract FrxEthRedeemStrategy is InitializableAbstractStrategy {
     }
 
     /**
-     * @notice Takes all frxETH and creates new redeem tickets
+     * @notice Takes all given frxETH and creates new redeem tickets
      */
     function depositAll() external override onlyVault nonReentrant {
         uint256 frxETHStart = frxETH.balanceOf(address(this));
@@ -211,7 +214,13 @@ contract FrxEthRedeemStrategy is InitializableAbstractStrategy {
         frxETH.approve(address(redemptionQueue), type(uint256).max);
     }
 
+     /**
+     * @notice Check if an asset is supported.
+     * @param _asset    Address of the asset
+     * @return bool     Whether asset is supported
+     */
     function supportsAsset(address _asset) public view override returns (bool) {
+        // frxETH can be deposited by the vault and balances are reported in weth
         return _asset == address(frxETH) || _asset == address(weth);
     }
 
--- a/contracts/contracts/strategies/FrxEthRedeemStrategy.sol
+++ b/contracts/contracts/strategies/FrxEthRedeemStrategy.sol
@@ -135,6 +135,7 @@ contract FrxEthRedeemStrategy is InitializableAbstractStrategy {
             msg.sender == IVault(vaultAddress).strategistAddr(),
             "Caller is not the Strategist"
         );
+        uint256 startingBalance = payable(address(this)).balance;
         for (uint256 i = 0; i < _nftIds.length; i++) {
             uint256 nftId = _nftIds[i];
             redemptionQueue.burnRedemptionTicketNft(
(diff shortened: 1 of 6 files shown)


DanielVF:
```python
from world import *
FRXETH_REDEEM_STRAT = '0xB262b69d1dB3dc092C73384B8553DA3d00e93682'
frxeth_redeem_strat = load_contract('frxeth_redeem_strat', FRXETH_REDEEM_STRAT)
oeth_vault_admin.withdrawAllFromStrategy(FRAX_ETH_STRATEGY, {'from': STRATEGIST})
print(frxeth.balanceOf(vault_oeth_core) / 1e18, weth.balanceOf(vault_oeth_core) / 1e18)
vault_oeth_admin.depositToStrategy(FRXETH_REDEEM_STRAT, [frxeth], [1003e18], {'from': STRATEGIST})
tickets = [int(x['data'][2:66],16) for x in history[-1].logs if x['address'] == FRXETH_REDEEM_STRAT and len(x['topics'])==1]
print(tickets)
chain.sleep(16*24*3600)
frxeth_redeem_strat.redeemTickets(tickets, 1003e18, {'from': STRATEGIST})
```

naddison36 (approved):
This looks good.

I've pushed some changes to fix the prettier, linter, slither and the unit tests.
The fork tests are currently failing but that's to do with OUSD's Flux strategy.

contracts/contracts/strategies/FrxEthRedeemStrategy.sol
```diff
@@ -0,0 +1,232 @@
...
+
+/**
+ * @title Frax ETH Redeem Strategy
+ * @notice This strategy redeems Frax ETH for ETH via the Frax Eth Redemption Queue contract
+ */
+contract FrxEthRedeemStrategy is InitializableAbstractStrategy {
+    IWETH9 private constant weth =
+        IWETH9(0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2);
+    IERC20 private constant frxETH =
+        IERC20(0x5E8422345238F34275888049021821E8E08CAa1f);
```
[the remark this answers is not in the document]
DanielVF: Yes, for most things I would use immutable as well, but for this it is purely going to be fork tested, and makes no sense out of a mainnet context.

DanielVF:
Thanks for the cleanup, Nick.

contracts/contracts/strategies/FrxEthRedeemStrategy.sol
```diff
@@ -0,0 +1,233 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import { IERC721Receiver } from "@openzeppelin/contracts/token/ERC721/IERC721Receiver.sol";
+
+import { BaseCompoundStrategy, InitializableAbstractStrategy } from "./BaseCompoundStrategy.sol";
```
sparrowDom, line 7: the `BaseCompoundStrategy` seems not to be used

contracts/contracts/strategies/FrxEthRedeemStrategy.sol
```diff
@@ -0,0 +1,233 @@
...
+import { IComptroller } from "../interfaces/IComptroller.sol";
```
sparrowDom, line 8: also an unused import

contracts/contracts/strategies/FrxEthRedeemStrategy.sol
```diff
@@ -0,0 +1,233 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
```
sparrowDom, line 4: unused import

sparrowDom (dismissed):
good job lgtm

contracts/contracts/strategies/FrxEthRedeemStrategy.sol
naddison36: I'll fix these are I've got the branch checked out

contracts/contracts/strategies/FrxEthRedeemStrategy.sol
naddison36: fixed

contracts/contracts/strategies/FrxEthRedeemStrategy.sol
naddison36: fixed

contracts/contracts/strategies/FrxEthRedeemStrategy.sol
```diff
@@ -0,0 +1,241 @@
...
+        // slither-disable-next-line unused-return
+        frxETH.approve(address(redemptionQueue), type(uint256).max);
+    }
+
+     /**
+     * @notice Check if an asset is supported.
+     * @param _asset    Address of the asset
+     * @return bool     Whether asset is supported
+     */
+    function supportsAsset(address _asset) public view override returns (bool) {
```
shahthepro, line 222: Visibility can be set to `pure`

contracts/contracts/strategies/FrxEthRedeemStrategy.sol
```diff
@@ -0,0 +1,241 @@
...
+    }
+
+    /**
+     * @notice Returns the amount of queued FraxEth that will be returned as WETH.
+     * We return this as a WETH asset, since that is what it will eventually be returned as.
+     * We only return the outstandingRedeems, because the contract itself should never hold any funds.
+     * @param _asset      Address of the asset
+     * @return balance    Total value of the asset in the platform
+     */
+    function checkBalance(address _asset)
```
shahthepro, line 195: nit: For all other strategies, `checkBalance` reverts if asset isn't supported. This is a temporary strategy, so it shouldn't matter though

contracts/contracts/strategies/FrxEthRedeemStrategy.sol
```diff
@@ -0,0 +1,241 @@
...
+        uint256 currentBalance = payable(address(this)).balance;
+        uint256 redeemedAmount = currentBalance - startingBalance;
+        require(
+            expectedAmount == redeemedAmount,
+            "Redeemed amount does not match expected amount"
+        );
+        outstandingRedeems -= redeemedAmount;
+        weth.deposit{ value: currentBalance }();
+        // slither-disable-next-line unchecked-transfer
+        weth.transfer(vaultAddress, currentBalance);
```
shahthepro, line 160: Can we emit a `Withdraw` event here? Would be easier to track it with analytics that way

DanielVF: Added

contracts/contracts/strategies/FrxEthRedeemStrategy.sol
DanielVF: Changed to match

contracts/contracts/strategies/FrxEthRedeemStrategy.sol
DanielVF: Done

shahthepro (commented):
A couple more comments related to gas savings, but nothing major

contracts/contracts/strategies/FrxEthRedeemStrategy.sol
```diff
@@ -0,0 +1,250 @@
...
+     * @notice Redeem specific tickets from the Queue.
+     * Called by the strategist.
+     * @param _nftIds Array of NFT IDs to redeem
+     */
+    function redeemTickets(uint256[] memory _nftIds, uint256 expectedAmount)
+        external
+        nonReentrant
+    {
+        require(
+            msg.sender == IVault(vaultAddress).strategistAddr(),
```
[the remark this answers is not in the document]
DanielVF: I'd rather have it just update and always follow the latest value on the vault. Plus it's only used in an admin action, so we don't have to worry about gas.

For our non-default strategies, the only gas important methods are supportsAsset and checkBalance.

shahthepro (dismissed):
#### Requirements
- Adds a one-off strategy to redeem frxETH into ETH using Frax's redemption queue

### Deployment Considerations

#### Internal State

Does this code do that?

#### Attack

- ⚪ An malicious actor could send their redeemable NFT to the strategy. And if the strategist accidentally uses that NFT while calling `redeemTickets` method, it could cause underflow of `outstandingRedeems`. However this will not result in loss of funds in any way. Worst case, we will have to upgrade the contract in future. Could be totally avoided if strategist takes necessary precaution to only use NFTs that the strategy minted. **No need to fix**


#### Logic

  - [x] _Are there bugs in the logic?_: **No logical bugs**
  - [x] _Correct usage of global & local variables. -> they might differentiate only by an underscore that can be overlooked (e.g. address vs _address)._: The contract initialises some of the storage variables to `address(0)`, however it has zero impact on overall functioning of the contract and doesn't result in any bugs. So, it's fine
  - [x] _Do public/external functions having asset parameter verify the asset is supported (where applicable)_ **Yes**


#### Tests

  - [x] If tests interact with AMM make sure enough edge cases (pool tilts) are tested. Ideally with fuzzing.

#### Flavor

_Could this code be simpler?_ All good

_Could this code be less vulnerable to other code behaving weirdly?_ All good

#### Overflow

#### Proxy

#### Black magic

#### Dependencies

#### Deploy
- [ ] Check that any deployer permissions are removed after deploy

#### Authentication

- [x] Never use tx.origin
- [ ] ⚪ _Check that every external/public function should actually be external._ **`deposit()` could be `external`. But not really a blocker**

#### Cryptographic code

#### Gas problems

    - [x] A way to remove items

#### External calls

- [x] Malicious behaviors
- [x] Oracles, one of:
  - [x] No oracles
  - [ ] Oracles can't be bent


#### Ethereum

- [x] Contract wraps ETH and can also receive ETH from the redemption queue
- [x] Contract has no payable methods (other than fallback)

contracts/contracts/strategies/FrxEthRedeemStrategy.sol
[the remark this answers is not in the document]
shahthepro: Sounds good

contracts/contracts/strategies/FrxEthRedeemStrategy.sol
```diff
@@ -0,0 +1,250 @@
...
+        // strategist via redeemTickets
+        require(false, "use redeemTickets() instead");
+    }
+
+    /**
+     * @notice Redeem specific tickets from the Queue.
+     * Called by the strategist.
+     * @param _nftIds Array of NFT IDs to redeem
+     */
+    function redeemTickets(uint256[] memory _nftIds, uint256 expectedAmount)
```
shahthepro, line 133: nit: Could probably add a `expectedAmount > 0` check

contracts/contracts/strategies/FrxEthRedeemStrategy.sol
```diff
@@ -0,0 +1,250 @@
...
+        // slither-disable-next-line unchecked-transfer
+        weth.transfer(vaultAddress, currentBalance);
+        emit Withdrawal(
+            address(weth),
+            address(redemptionQueue),
+            currentBalance
+        );
+    }
+
+    function _abstractSetPToken(address _asset, address _pToken)
```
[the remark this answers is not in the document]
DanielVF: I've added a revert on _abstractSetPToken which will prevent a pToken set from happening. I've not adjusted remove, since if there are no adds a remove should not work.

contracts/contracts/strategies/FrxEthRedeemStrategy.sol
```diff
@@ -0,0 +1,250 @@
...
+    function redeemTickets(uint256[] memory _nftIds, uint256 expectedAmount)
+        external
+        nonReentrant
+    {
+        require(
+            msg.sender == IVault(vaultAddress).strategistAddr(),
+            "Caller is not the Strategist"
+        );
+        uint256 startingBalance = payable(address(this)).balance;
+        for (uint256 i = 0; i < _nftIds.length; i++) {
```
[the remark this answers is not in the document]
DanielVF: Not really gas important here, and future solidity compilers will optimize this to be the same thing.

contracts/contracts/strategies/FrxEthRedeemStrategy.sol
```diff
@@ -0,0 +1,250 @@
...
+        IWETH9(0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2);
+    IERC20 private constant frxETH =
+        IERC20(0x5E8422345238F34275888049021821E8E08CAa1f);
+    IFraxEtherRedemptionQueue private constant redemptionQueue =
+        IFraxEtherRedemptionQueue(0x82bA8da44Cd5261762e629dd5c605b17715727bd);
+    uint256 public constant maxRedeemTicket = 250e18;
+    uint256 public outstandingRedeems;
+
+    event RedeemNFTMinted(uint256 _nftId, uint256 _amount);
+    event RedeemNFTBurned(uint256 _nftId);
```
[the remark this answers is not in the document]
DanielVF: For which reason? To know the state of the contract and what you could redeem you would need to fetch all RedeemNFTMinted and RedeemNFTBurned events.

contracts/contracts/strategies/FrxEthRedeemStrategy.sol
DanielVF: Will keep as is on this one.

shahthepro (dismissed):

DanielVF:
# FrxEth Redeem Code Review

#### Authentication

- [x] Never use tx.origin
- [x] Every external/public function should be intended to be externally accessible

🟣 `deposit` could be external, rather than public. Fine as is though.

🟣 `oneERC721Received` could be view/pure. Fine as is though.


#### Ethereum

_Contract does have to receive Ethereum, however in normal operation the contract immediately wraps it and sends it to the vault as WETH.

#### Cryptographic code

#### Gas problems

    - [ ] A way to remove items

#### Black magic

#### Overflow

#### Proxy

#### Events

## Medium Checks

#### Rounding

#### Dependencies
_No new dependancies_

#### External calls

- [ ] No Malicious behaviors
- [x] Oracles, one of:
  - [x] No oracles
  - [ ] Oracles can't be bent

#### Deploy

## Strategy Specific

#### Smart contract

- [x] **Check balance cannot be manipulated up AND down by an attacker**
- [x] No funds fall into the cracks and are not reported in check balance as a result of depositing or withdrawing
- [ ] WithdrawAll() can always withdraw an amount equal to or larger than checkBalances report, even in spite of attacker manipulation. (🟣 We inherently can't, since these are asynchronous redeems)

#### Downstream

_Even without a backend system, we can just look at events fired from mint transactions to know what we can redeem.

## Thinking

#### Logic

  - [x] Logic free of bugs

#### Internal State

`outstandingRedeems` must match the sum of all minted NFTs minus the sum of redeemed NFTs. It appears to do so.

#### Attack

Possible Impact. This contract may be moving and accounting for a large portion of OETH's funds. Permanently losing funds would be bad.

Attacker theft - The contract should hold no funds at rest between strategist operations. This limits theft operations to inside deposit and withdraw operations. Both of these can only be initiated by the strategist or governance. All funds are transferred to only the vault or allowed to be transferred to the frxeth redeem queue.

Attacker balance manipulation - Donations to the strategy do not alter the reported funds. NFT donations _that are used by the strategist_ would result in a profit to the protocol, though would results in other NFT's not being able to be redeeming until accounting could be fixed with a contract upgrade.

Given that strategy funds are empty outside of actions, and actions can only kicked off by trusted users, the most interesting attack angles would be either directly outside the actions (sandwiches, frontrunning, backrunning) or inside the action itself. 

It's possible for someone to redeem just in front of us in the frxETH pool, and thus bring the redeemable total down under what we are trying to redeem in a transaction. But we can just resubmit with a private transaction.

How about inside attacks? redeemTickets does an adjacent decrement of outstandingRedeems with sending funds to the vault. There shouldn't be any time when the vault balances are incorrect from an external point of view.

depositAll does get frxEth sent to it before the deposit happens and before outstandingRedeems is updated. This would result from the outside like a temporary drop in vault total supply. However there is no way for an attacker to get execution during this time, the vault is reentrancy locked during depositToStrategy, and a temporary down in funds should not be exploitable like a temporary up in funds would be.

NFT's provide a reentrancy danger, in that transferring them can give execution to the destination address. However we never transfer the NFTs, so this is not an issue.

Deposit process failure - It is possible to send too much funds to the deposit function that you could run out of gas creating NFT's. If this was a strategist action, then the revert would put the funds back in the vault and a smaller amount could be chosen. If someone donated a huge amount of frxETH to the strategy to cause this, we could withdraw the frxETH to vault via withdrawAll() and be very happy.

Redeem process failure - Too many NFT's could cause out of gas, but strategist could just adjust number of NFTs in redeemTickets call.

#### Flavor

There was a bit of tradeoff between writing just enough code to do the job, and handling cases like unexpected donations of eth, or incorrect governance behavior. Code seems reasonable.

DanielVF:
I had an unsigned commit, and needed to force push to remove it as a merge blocker. Also added a lint fix.

Needless to say, re-review the last set of commits carefully before merging.

shahthepro (approved):

PR #2012, opened 2024-03-28
Add support for removing assets from Vault

shahthepro:
# Changes

* Added `removeAsset` to `VaultAdmin`

## Gas impact

Both the following are after the Frax strategy has been removed.
The before is when frxETH is still an asset of the Vault.
After has the frxETH asset removed.

| Test |  Before | After | Saving |
| --- | --- | --- | --- |
mint using 1 WETH | 531,368 | 512,991 | 18,377 |
mint using 11 WETH | 745,615 |  717,792 | 27,823 |
redeem 10 OETH | 574,376 |  555,992 | 18,377 |
redeem 100 OETH | 574,400 | 556,016 | 18,377 |
rebase | 467,203 | 448,826 | 18,377 |

# Security

Contract change checklist:
  - [x] Copy & paste code review [security checklist](https://github.com/OriginProtocol/security/blob/master/templates/Contract-Code-Review-Example.md) below this checklist.
  - [x] Unit tests pass


contracts/contracts/vault/VaultAdmin.sol
```diff
@@ -343,6 +343,59 @@ contract VaultAdmin is VaultStorage {
...
+            IVault(address(this)).checkBalance(_asset) == 0,
+            "Vault still holds asset"
+        );
+
+        uint256 assetsCount = allAssets.length;
+        uint256 assetIndex = assetsCount;
+        for (uint256 i = 0; i < assetsCount; ++i) {
+            if (allAssets[i] == _asset) {
+                assetIndex = i;
+                break;
```
DanielVF, line 358: We should really revert after this loop if the asset is not found. In theory anything with isSupported will be the allAssets, but this could break stuff badly if it were not.

contracts/contracts/vault/VaultAdmin.sol
```diff
@@ -343,6 +343,59 @@ contract VaultAdmin is VaultStorage {
...
+        assets[_asset].isSupported = false;
+
+        emit AssetRemoved(_asset);
+    }
+
+    /**
+     * @notice Remove a supported asset from the Vault
+     * @param _asset Address of asset
+     */
+    function removeAsset(address _asset) external onlyGovernor {
```
DanielVF, line 385: Let's just have the one `removeAsset`, and not have `removeAssets`. We're not likely to call this often enough to justify the extra code size in the contract.

shahthepro: Will do it

contracts/contracts/vault/VaultAdmin.sol
shahthepro: Actually, `assetIndex` is set to `assetsCount` initially. So it'll revert if the asset doesn't exist `allAssets` with an out-of-bound index error. Have also added a comment about this (a few lines below this)

contracts/contracts/vault/VaultAdmin.sol
```diff
@@ -343,6 +343,45 @@ contract VaultAdmin is VaultStorage {
...
+        // Update allAssets array
+        allAssets[assetIndex] = allAssets[assetsCount - 1];
+        allAssets.pop();
+
+        // Reset default strategy
+        assetDefaultStrategies[_asset] = address(0);
+        emit AssetDefaultStrategyUpdated(_asset, address(0));
+
+        // Mark as unsupported
+        assets[_asset].isSupported = false;
```
DanielVF, line 380: Let's actually clear everything in the asset struct with `delete`, rather than just isSupported. Feels cleaner and will keep us from having weird data around lurking for the future if we ever decided to turn the asset back on.

shahthepro: Sounds good. Will make that change

contracts/contracts/vault/VaultAdmin.sol
DanielVF: Clever. I've pushed up one additional code comment on this.

contracts/contracts/vault/VaultAdmin.sol
```diff
@@ -343,6 +343,45 @@ contract VaultAdmin is VaultStorage {
...
     }
 
+    /**
+     * @notice Remove a supported asset from the Vault
+     * @param _asset Address of asset
+     */
+    function removeAsset(address _asset) external onlyGovernor {
+        require(assets[_asset].isSupported, "Asset not supported");
+        require(
+            IVault(address(this)).checkBalance(_asset) == 0,
```
DanielVF, line 353: I'm wondering if we are actually going to be able to hit zero on some assets that are weird like stETH. Maybe this should be < 1e6 to allow for dust.

DanielVF (commented):
# Code review remove assets

## Requirements

We will be removing some assets from OETH. We want the data structures in the contract to be like these assets were never there.

## Easy Checks

#### Authentication

- [x] Never use tx.origin

#### Ethereum

#### Cryptographic code

_no crypto_

#### Gas problems

    - [x] A way to remove items

#### Black magic

#### Overflow

#### Proxy

#### Events

## Medium Checks

#### Rounding
_No rounding_

#### Dependencies
_No dependency changes_

#### External calls

- [x] No Malicious behaviors
- [x] Oracles, one of:
  - [x] No oracles
  - [ ] Oracles can't be bent

#### Tests

#### Deploy

#### Downstream

_No downstream_

## Thinking

#### Logic

The code needs to remove the asset from all storage that holds the asset config.

- assets ✅
- allAssets ✅
- assetDefaultStrategies ✅

We also have sanity code to ensure that the vault already has been emptied of the asset before removing it.

_No deployment considerations_

#### Internal State

See above under "Logic"

#### Attack

Removing the wrong asset could result in an apparent drop in the vaults assets, thus scaring users and stopping all redeems and rebases. Also would stop mints in that asset. 

However, we do check that the asset is actually unused before allowing the removal of an asset.

We also check that the asset is a supported asset, and is in the assets array before removing.

This action is governance only, and so should not be available for external users to attack.

This method does not make external, non-static calls. It should be safe from any kind of reentrancy attack happening inside it.

It is not marked non-reentrant, so it could be triggered externally. Even though it is a governance action that can only be called from the timelock, the governance actions that have come out of queue in the timelock can be triggered by anyone, thus allowing an attacker to trigger a governance queued remove.

Removing an asset results in that asset not being available for withdraw and a lower total vault value. Neither of these are good for an attacker. Removing an asset during an OUSD redeem would cause a revert since the asset for loop would cause a revert. For both mints and redeems, we do all calculations first, before calling external assets.

🟣 An attacker could break the strategy remove by front running a donation to the vault. I think we'll just deal with this if it happens with an upgrade.



#### Flavor

Flavor great. Code is simple!

contracts/contracts/vault/VaultAdmin.sol
```diff
@@ -343,6 +343,45 @@ contract VaultAdmin is VaultStorage {
...
     }
 
+    /**
+     * @notice Remove a supported asset from the Vault
+     * @param _asset Address of asset
+     */
+    function removeAsset(address _asset) external onlyGovernor {
+        require(assets[_asset].isSupported, "Asset not supported");
+        require(
+            IVault(address(this)).checkBalance(_asset) <= 1e6,
```
pandadefi, line 353: I saw the original version had a check for zero balance, considering you might have dust on multiple strategies as well 1e6 might be a little low, and would advise for 10e12 which is even for a coin like ETH not worth more than a cent.

pandadefi (approved):

contracts/contracts/vault/VaultAdmin.sol
shahthepro: Sounds good to me, making that change

shahthepro: Done

contracts/contracts/vault/VaultAdmin.sol
shahthepro: Panda suggested to increase it to 1e13 (since the dust can exist across multiple strategies and it'd still be insignificant), so used that limit

contracts/contracts/vault/VaultAdmin.sol
shahthepro: Done

DanielVF (dismissed):

contracts/contracts/vault/VaultAdmin.sol
```diff
@@ -343,6 +343,45 @@ contract VaultAdmin is VaultStorage {
...
     }
 
+    /**
+     * @notice Remove a supported asset from the Vault
+     * @param _asset Address of asset
+     */
+    function removeAsset(address _asset) external onlyGovernor {
+        require(assets[_asset].isSupported, "Asset not supported");
+        require(
+            IVault(address(this)).checkBalance(_asset) <= 1e13,
```
naddison36, line 353: why an external call and not an internal `_checkBalance(_asset)`?

contracts/deploy/092_vault_admin_upgrade.js
```diff
@@ -0,0 +1,40 @@
...
+    Part of simplified OETH proposal. This adds the ability to remove a supported asset from the Vault. \
+    ",
+      actions: [
+        {
+          contract: cVaultAdmin,
+          signature: "setAdminImpl(address)",
+          args: [dVaultAdminImpl.address],
+        },
+      ],
+    };
```
naddison36, line 38: can we add the removal of frxETH to this deployment?

naddison36 (commented):
It'd be interesting to see the gas impacts of mint, redeem and rebase when frxETH is removed.

contracts/contracts/vault/VaultAdmin.sol
shahthepro: `checkBalance` is in `VaultCore`, not `VaultAdmin`

contracts/deploy/092_vault_admin_upgrade.js
shahthepro: We can do it, but we need strategist to do swap the frxETH it holds. Right now, this PR allows dust upto `e12`. But the Vault right now has a little over `1e16` frxETH

contracts/test/vault/oeth-vault.fork-test.js
```diff
@@ -148,7 +148,36 @@ describe("ForkTest: OETH Vault", function () {
       await expect(tx).to.revertedWith("Liquidity error");
     });
     it("OETH whale can redeem after withdraw from all strategies", async () => {
-      const { oeth, oethVault, timelock } = fixture;
+      const { oeth, oethVault, timelock, domen, strategist, frxETH } = fixture;
```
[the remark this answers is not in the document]
shahthepro: frxETH strategy is already removed. It's just the asset that we have to remove right now

contracts/contracts/vault/VaultAdmin.sol
naddison36: Ah, yes, that makes sense.

naddison36 (dismissed):

contracts/deploy/092_vault_admin_upgrade.js
naddison36: I've added the remove of frxETH to the deploy script as the remaining frxETH in the Vault has been swapped out for WETH

PR #2015, opened 2024-04-04
SSV Native staking 

sparrowDom:
# Contracts

## Pull Requests to this branch

* https://github.com/OriginProtocol/origin-dollar/pull/2060
* https://github.com/OriginProtocol/origin-dollar/pull/2066
* https://github.com/OriginProtocol/origin-dollar/pull/2072
* https://github.com/OriginProtocol/origin-dollar/pull/2073
* https://github.com/OriginProtocol/origin-dollar/pull/2074
* https://github.com/OriginProtocol/origin-dollar/pull/2076
* https://github.com/OriginProtocol/origin-dollar/pull/2081
* https://github.com/OriginProtocol/origin-dollar/pull/2082
* https://github.com/OriginProtocol/origin-dollar/pull/2083
* https://github.com/OriginProtocol/origin-dollar/pull/2084
* https://github.com/OriginProtocol/origin-dollar/pull/2085
* https://github.com/OriginProtocol/origin-dollar/pull/2087 
* https://github.com/OriginProtocol/origin-dollar/pull/2088
* https://github.com/OriginProtocol/origin-dollar/pull/2091
* https://github.com/OriginProtocol/origin-dollar/pull/2092

## Related pull requests

* https://github.com/OriginProtocol/origin-dollar/pull/2080

## Changes

* Added new `NativeStakingSSVStrategy` that takes WETH from the Vault and stakes ETH into SSV validators.
* Added `FeeAccumulator` to receive ETH execution rewards.
* Changes `BaseHarvester` so it can collect WETH and send it straight to the `Dripper`

oethContracts

### NativeStakingSSVStrategy

NativeStakingSSVStrategyHierarchy

NativeStakingSSVStrategySquashed

## FeeAccumulator

FeeAccumulatorHierarchy

FeeAccumulatorSquashed

# Processes

## Value Flows

The flow of native currency (ETH) and tokens between the wallets and contracts.

oethValueFlows-native-staking

## Validator Registration and Staking

oethProcesses-register

## Validator withdrawals

oethProcesses-withdraw

## Validator rewards

oethProcesses-rewards

## Native Staking Admin

oethProcesses-admin

# Testing

## Unit Tests

Unit tests are in `contracts/test/strategies/nativeSSVStaking.js`

```bash
yarn run test
```

## Holesky fork tests

Holesky fork tests are in `contracts/test/strategies/nativeSsvStaking.holesky.fork-test.js`

```bash
yarn run node:hol
yarn run test:hol-fork
```

## Mainnet fork tests

Mainnet fork tests are in `contracts/test/strategies/nativeSsvStaking.fork-test.js`

```bash
yarn run node
yarn run test:fork
```

## Holesky end-to-end testing using Hardhat tasks

```
export DEBUG=origin*

# uncomment DEFENDER_API_KEY and DEFENDER_API_SECRET in the .env file
npx hardhat depositWETH --amount 64 --network holesky
npx hardhat mint --asset WETH --amount 64 --network holesky

# Uncomment DEPLOYER_PK in .env file to private key of 0x1b94CA50D3Ad9f8368851F8526132272d1a5028C
npx hardhat depositToStrategy --assets WETH --amounts 64 --strategy NativeStakingSSVStrategyProxy --network holesky

# uncomment DEFENDER_API_KEY and DEFENDER_API_SECRET in the .env file
npx hardhat resetStakeETHTally --network holesky
npx hardhat registerValidators --days 2 --network holesky

# give it 10 minutes before running
npx hardhat stakeValidators --network holesky

# Claim SSV from faucet https://faucet.ssv.network/
# transfer claimed SSV to Native Staking Strategy proxy
# Uncomment DEPLOYER_PK in .env file to private key of 0x1b94CA50D3Ad9f8368851F8526132272d1a5028C
npx hardhat depositSSV --amount 2 --operatorids 563,564,565,566 --network holesky

# uncomment DEFENDER_API_KEY and DEFENDER_API_SECRET in the .env file
npx hardhat exitValidator --operatorids 563,564,565,566 --pubkey 0x95883A44A9629017441D318E51D80F89B5A13E98CAC6EC2481167319753C14F4C9C694462B057D98B7AD1CEACA9441E3 --network holesky

npx hardhat removeValidator --operatorids 563,564,565,566 --pubkey 0x95883A44A9629017441D318E51D80F89B5A13E98CAC6EC2481167319753C14F4C9C694462B057D98B7AD1CEACA9441E3 --network holesky

npx hardhat doAccounting --network holesky

# transfer some ETH to the FeeAccumulator 0x65a289f4BF934c964C942eFF6E6F83b6481BE550
npx hardhat harvest --strategy NativeStakingSSVStrategyProxy --network holesky

# Uncomment DEPLOYER_PK in .env file to private key of 0x1b94CA50D3Ad9f8368851F8526132272d1a5028C
npx hardhat pauseStaking --network holesky

npx hardhat fixAccounting --network holesky
```

# Deployment

The deploy script for the Native Staking Strategy is `contracts/deploy/mainnet/097_native_ssv_staking.js`.

# Operations

Hardhat tasks


contracts/contracts/strategies/NativeStaking/NativeStakingSSVStrategy.sol
```diff
@@ -0,0 +1,287 @@
...
+        override
+        returns (uint256 balance)
+    {
+        if (_asset != WETH_TOKEN_ADDRESS) {
+            revert NotWeth();
+        }
+
+        balance = activeDepositedValidators * 32 ether;
+        balance += beaconChainRewardWETH;
+        balance += FEE_ACCUMULATOR_ADDRESS.balance;
```
naddison36, line 250: I don't think we want to include the MEV rewards in the FeeAccumulator if those are going to be harvested and sent to through the dripper.

contracts/contracts/strategies/NativeStaking/ValidatorAccountant.sol
```diff
@@ -0,0 +1,257 @@
...
+    ///
+    /// We need to be able to keep a separate accounting of the WETH so we understand how much we can pass oh to
+    /// the harvester as a consequence of rewards harvesting and how much registrator can pick up as a result of WETH
+    /// deposit into the strategy contract.
+    /// To achieve this the beacon chain rewards are accounted for using below variable, all other WETH is assumed to be
+    /// present as a result of a deposit.
+    uint256 public beaconChainRewardWETH = 0;
+
+    /// @dev start of fuse interval
+    uint256 public fuseIntervalStart = 0;
```
naddison36, line 27: I've squash `fuseIntervalStart` and `fuseIntervalEnd` into a single slot.

contracts/contracts/strategies/NativeStaking/ValidatorAccountant.sol
```diff
@@ -0,0 +1,257 @@
...
+    /// @dev end of fuse interval
+    uint256 public fuseIntervalEnd = 0;
+    /// @dev Governor that can manually correct the accounting
+    address public accountingGovernor;
+    /// @dev Strategist that can pause the accounting
+    address public strategist;
```
naddison36, line 33: Best to read the `strategist` from the Vault rather than duplicating the maintenance in the strategy. This is what OETH AMO strategy does

```Solidity
    modifier onlyStrategist() {
        require(
            msg.sender == IVault(vaultAddress).strategistAddr(),
            "Caller is not the Strategist"
        );
        _;
    }
```

contracts/contracts/strategies/NativeStaking/NativeStakingSSVStrategy.sol
```diff
@@ -0,0 +1,287 @@
...
+        beaconChainRewardWETH += wethCollected;
+        _collectRewardTokens();
+    }
+
+    /// @dev Need to override this function since the strategy doesn't allow for all the WETH
+    /// to be collected. Some might be there as a result of deposit and is waiting for the Registrar
+    /// to be deposited to the validators.
+    function _collectRewardTokens() internal override {
+        uint256 rewardTokenCount = rewardTokenAddresses.length;
+        for (uint256 i = 0; i < rewardTokenCount; ++i) {
```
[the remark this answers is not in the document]
sparrowDom: yeah that is a good point this is done because of the SSV tokens. I was thinking that the Relayer would just send the SSV tokens to the strategy. Though given that SSV doesn't have a Chainlink oracle, we'd probably need to use Uniswap's TWAP.... we might skip some complexity and launch without the support for Harvester to swap the SSV for ETH. 

And add that functionality after the launch.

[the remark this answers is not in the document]
naddison36: Are we going to sell the SSV tokens? Given we need to pay the SSV Operators using SSV tokens, I was thinking the SSV rewards was more like a rebate. We can reuse the SSV rewards to pay for future SSV Operations.

contracts/contracts/strategies/NativeStaking/FeeAccumulator.sol
```diff
@@ -0,0 +1,39 @@
...
+import { Governable } from "../../governance/Governable.sol";
+
+/**
+ * @title Fee Accumulator for Native Staking SSV Strategy
+ * @notice Receives execution rewards which includes tx fees and
+ * MEV rewards like tx priority and tx ordering.
+ * It does NOT include swept ETH from beacon chain consensus rewards or full validator withdrawals.
+ * @author Origin Protocol Inc
+ */
+contract FeeAccumulator is Governable {
```
pandadefi, line 15: It seems this contract missing a receive `payable` function?

contracts/contracts/strategies/NativeStaking/NativeStakingSSVStrategy.sol
```diff
@@ -0,0 +1,264 @@
...
+    /// Will NOT revert if the strategy is paused from an accounting failure.
+    /// @param _asset Address of asset to deposit. Has to be WETH.
+    /// @param _amount Amount of assets that were transferred to the strategy by the vault.
+    function deposit(address _asset, uint256 _amount)
+        external
+        override
+        onlyVault
+        nonReentrant
+    {
+        require(_asset == WETH_TOKEN_ADDRESS, "Unsupported asset");
```
pandadefi, line 128: I see you have the same `require` twice, you could use a modifier.

contracts/contracts/strategies/NativeStaking/NativeStakingSSVStrategy.sol
```diff
@@ -0,0 +1,264 @@
...
+            IERC20(WETH_TOKEN_ADDRESS).balanceOf(address(this));
+    }
+
+    function pause() external onlyStrategist {
+        _pause();
+    }
+
+    /// @notice Returns bool indicating whether asset is supported by strategy.
+    /// @param _asset The address of the asset token.
+    function supportsAsset(address _asset) public view override returns (bool) {
```
pandadefi, line 238: This function can be set to external.

contracts/contracts/strategies/NativeStaking/ValidatorAccountant.sol
```diff
@@ -0,0 +1,232 @@
...
+/// as either full or partial withdrawals. Partial withdrawals being consensus rewards.
+/// Full withdrawals are from exited validators.
+/// @author Origin Protocol Inc
+abstract contract ValidatorAccountant is ValidatorRegistrator {
+    /// @notice The maximum amount of ETH that can be staked by a validator
+    /// @dev this can change in the future with EIP-7251, Increase the MAX_EFFECTIVE_BALANCE
+    uint256 public constant MAX_STAKE = 32 ether;
+
+    /// @notice Keeps track of the total consensus rewards swept from the beacon chain
+    uint256 public consensusRewards = 0;
```
pandadefi, line 18: In Solidity, variables are automatically initialized to zero. Therefore, explicitly initializing variables to zero is not mandatory and can be skipped.

contracts/contracts/strategies/NativeStaking/ValidatorRegistrator.sol
```diff
@@ -0,0 +1,239 @@
...
+            IDepositContract(BEACON_CHAIN_DEPOSIT_CONTRACT).deposit{
+                value: 32 ether
+            }(
+                validators[i].pubkey,
+                withdrawal_credentials,
+                validators[i].signature,
+                validators[i].depositDataRoot
+            );
+
+            activeDepositedValidators += 1;
```
pandadefi, line 146: You could save on some gas by adding the `validators.length` once instead of doing a +1 on every interaction..

contracts/contracts/strategies/NativeStaking/ValidatorAccountant.sol
```diff
@@ -0,0 +1,232 @@
...
+            return _failAccounting(pauseOnFail);
+        }
+
+        // Calculate all the new ETH that has been swept to the contract since the last accounting
+        uint256 newSweptETH = address(this).balance - consensusRewards;
+        accountingValid = true;
+
+        // send the ETH that is from fully withdrawn validators to the Vault
+        if (newSweptETH >= MAX_STAKE) {
+            uint256 fullyWithdrawnValidators = newSweptETH / MAX_STAKE;
```
pandadefi, line 119: This safe division could use `unchecked`

contracts/contracts/strategies/NativeStaking/FeeAccumulator.sol
```diff
@@ -0,0 +1,39 @@
...
+
+    /**
+     * @notice sends all ETH in this FeeAccumulator contract to the Native Staking Strategy.
+     * @return eth The amount of execution rewards that were sent to the Native Staking Strategy
+     */
+    function collect() external returns (uint256 eth) {
+        require(msg.sender == STRATEGY, "Caller is not the Strategy");
+
+        eth = address(this).balance;
+        if (eth > 0) {
```
pandadefi, line 34: You have a couple of superior to zero checks, it's more gas-efficient to use a not zero check.

contracts/contracts/strategies/NativeStaking/ValidatorRegistrator.sol
```diff
@@ -0,0 +1,239 @@
...
+        require(
+            requiredETH <= IWETH9(WETH_TOKEN_ADDRESS).balanceOf(address(this)),
+            "insufficient WETH"
+        );
+
+        // Convert required ETH from WETH
+        IWETH9(WETH_TOKEN_ADDRESS).withdraw(requiredETH);
+
+        // For each validator
+        for (uint256 i = 0; i < validators.length; ) {
```
pandadefi, line 118: `validators.length` should be cached outside of the loop.

contracts/contracts/strategies/NativeStaking/FeeAccumulator.sol
sparrowDom: Thanks this is a great find. Fixed in this commit: https://github.com/OriginProtocol/origin-dollar/pull/2053/commits/3fb038a61f7275665e689df8e812e590c7b56394

contracts/contracts/strategies/NativeStaking/NativeStakingSSVStrategy.sol
sparrowDom: this one overrides the function from `InitializableAbstractStrategy.sol` and can not have a different visibility

contracts/contracts/strategies/NativeStaking/ValidatorAccountant.sol
sparrowDom: Good point, fixed here: https://github.com/OriginProtocol/origin-dollar/pull/2053/commits/8bf9a268ffc433b95f79ef5b6c15cb9d19a9cc6d

contracts/contracts/strategies/NativeStaking/ValidatorRegistrator.sol
sparrowDom: great gas savings thanks: https://github.com/OriginProtocol/origin-dollar/pull/2053/commits/fd4f2d24d5db8723e7eef9ed8630a4f221684749

contracts/contracts/strategies/NativeStaking/ValidatorRegistrator.sol
sparrowDom: thanks, fixed here: https://github.com/OriginProtocol/origin-dollar/pull/2053/commits/fd4f2d24d5db8723e7eef9ed8630a4f221684749

contracts/contracts/strategies/NativeStaking/ValidatorAccountant.sol
sparrowDom: agreed, thanks

contracts/contracts/strategies/NativeStaking/FeeAccumulator.sol
sparrowDom: I've tested this with the [hardhat-gas-reporter](https://www.npmjs.com/package/hardhat-gas-reporter) and Remix and didn't see a noticeable difference in gas usage.

contracts/contracts/strategies/NativeStaking/ValidatorAccountant.sol
```diff
@@ -0,0 +1,234 @@
...
+        // pause if not already
+        if (pauseOnFail) {
+            _pause();
+        }
+        // fail the accounting
+        accountingValid = false;
+    }
+
+    /// @notice Allow the Strategist to fix the accounting of this strategy and unpause.
+    /// @param _validatorsDelta adjust the active validators by plus one, minus one or unchanged with zero
```
DanielVF, line 184: Comment here says +1, but code seems to be +3?

contracts/contracts/strategies/NativeStaking/ValidatorAccountant.sol
```diff
@@ -0,0 +1,234 @@
...
+        require(ethRemaining < 32 ether, "unexpected accounting");
+
+        // If no Beacon chain consensus rewards swept
+        if (ethRemaining == 0) {
+            // do nothing
+            return accountingValid;
+        }
+        // Beacon chain consensus rewards swept (partial validator withdrawals)
+        else if (ethRemaining < fuseIntervalStart) {
+            // solhint-disable-next-line reentrancy
```
DanielVF, line 149: Wonder why `consensusRewards += ethRemaining;` needs a `solhint-disable-next-line reentrancy` but `activeDepositedValidators -= 1;` in the next if statement does not - that's very odd.

contracts/contracts/strategies/NativeStaking/ValidatorAccountant.sol
```diff
@@ -0,0 +1,234 @@
...
+        }
+        // fail the accounting
+        accountingValid = false;
+    }
+
+    /// @notice Allow the Strategist to fix the accounting of this strategy and unpause.
+    /// @param _validatorsDelta adjust the active validators by plus one, minus one or unchanged with zero
+    /// @param _wethToVaultAmount the amount of WETH to be sent to the Vault
+    /// @param _consensusRewardsDelta adjust the accounted for consensus rewards up or down
+    function manuallyFixAccounting(
```
DanielVF, line 187: I think we had discussed some kind of rate limiting on this? Otherwise it can be called over and over again to add 1.3 million dollars to total supply each time?

DanielVF: Requiring whenPaused as a modifier to start the function and reverting unless the function is fuse blown can be bypassed by self destruct donations to the contract, to allow for another call.

contracts/contracts/strategies/NativeStaking/ValidatorAccountant.sol
sparrowDom: yup thanks, fixed here: https://github.com/OriginProtocol/origin-dollar/pull/2015/commits/fc48427a5acb235499913c7705af009a7b4a961c

contracts/contracts/strategies/NativeStaking/ValidatorAccountant.sol
sparrowDom: @naddison36 was battling with the slither maybe best for him to respond

contracts/contracts/strategies/NativeStaking/ValidatorAccountant.sol
sparrowDom: This is a great point. I will prepare a PR for this over the weekend

contracts/contracts/strategies/NativeStaking/NativeStakingSSVStrategy.sol
```diff
@@ -0,0 +1,264 @@
...
+         */
+        emit Deposit(_asset, address(0), _amount);
+    }
+
+    /// @notice Unlike other strategies, this does not deposit assets into the underlying platform.
+    /// It just emits the Deposit event.
+    /// To deposit WETH into validators `registerSsvValidator` and `stakeEth` must be used.
+    /// Will NOT revert if the strategy is paused from an accounting failure.
+    function depositAll() external override onlyVault nonReentrant {
+        uint256 wethBalance = IERC20(WETH_TOKEN_ADDRESS).balanceOf(
```
sparrowDom, line 154: I just realized that calling `depositAll` will communicate incorrect values since one can deposit WETH and that WETH will stay on the contract until the defender action stakes it. 

If `depositAll` is called again before all WETH is unwrapped and staked this value will be too high, since it will include the WETH that was on the contract before the Vault sent additional WETH to it. This might be misleading for the monitoring tools if they rely on `Deposit` events instead of calling `checkBalance`.

contracts/contracts/strategies/NativeStaking/ValidatorAccountant.sol
```diff
@@ -0,0 +1,220 @@
...
+            consensusRewards += ethRemaining;
+            emit AccountingConsensusRewards(ethRemaining);
+        }
+        // Beacon chain consensus rewards swept but also a slashed validator fully exited
+        else if (ethRemaining > fuseIntervalEnd) {
+            IWETH9(WETH_TOKEN_ADDRESS).deposit{ value: ethRemaining }();
+            IWETH9(WETH_TOKEN_ADDRESS).transfer(VAULT_ADDRESS, ethRemaining);
+            activeDepositedValidators -= 1;
+
+            emit AccountingValidatorSlashed(
```
sparrowDom, line 158: Since we basically withdraw WETH to the Vault should we emit a `Withdrawal` event here?

contracts/contracts/strategies/NativeStaking/ValidatorAccountant.sol
```diff
@@ -0,0 +1,220 @@
...
+                // explicitly cast to uint256 as we want to round to a whole number of validators
+                fullyWithdrawnValidators = uint256(newSweptETH / MAX_STAKE);
+            }
+            activeDepositedValidators -= fullyWithdrawnValidators;
+
+            uint256 wethToVault = MAX_STAKE * fullyWithdrawnValidators;
+            IWETH9(WETH_TOKEN_ADDRESS).deposit{ value: wethToVault }();
+            IWETH9(WETH_TOKEN_ADDRESS).transfer(VAULT_ADDRESS, wethToVault);
+
+            emit AccountingFullyWithdrawnValidator(
```
sparrowDom, line 130: Since we basically withdraw WETH to the Vault should we emit a `Withdrawal` event here?

naddison36: yes, pls add a `Withdrawal` event

contracts/contracts/strategies/NativeStaking/NativeStakingSSVStrategy.sol
naddison36: That's true, but what's the alternative? Store the deposited WETH balance in storage so the actual deposit amount can be calculated?

sparrowDom: Yeah I can't think of a better alternative than that. And adding so much complexity doesn't seem to justify the correct values in the Deposit events.

contracts/contracts/strategies/NativeStaking/ValidatorAccountant.sol
```diff
@@ -0,0 +1,266 @@
...
+        if (address(this).balance < consensusRewards) {
+            return _failAccounting(pauseOnFail);
+        }
+
+        // Calculate all the new ETH that has been swept to the contract since the last accounting
+        uint256 newSweptETH = address(this).balance - consensusRewards;
+        accountingValid = true;
+
+        // send the ETH that is from fully withdrawn validators to the Vault
+        if (newSweptETH >= MAX_STAKE) {
```
pandadefi, line 122: Depending on how often and how many validators you plan to run, this condition could return true even if no validator has fully withdrawn.

Lido operates approximately 300,000 validators. If you manage 1% of that, or 3,000 validators, your base daily rewards would be around 7 ETH, excluding execution rewards, which could quickly add up to 32 in just a few days.

naddison36: The idea is to cap the number a validators a strategy can manage to 600 (19,200 ETH)

naddison36: We need to decide if we are going to enforce the 600 validator (19,200 ETH) limit in the strategy contract or not. If all stETH and rETH currently in the OETH Vault was convert to WETH and staked, that would be 20,188 ETH.

contracts/contracts/strategies/NativeStaking/ValidatorRegistrator.sol
```diff
@@ -0,0 +1,341 @@
...
+        );
+
+        require(
+            stakeETHTally + requiredETH <= stakeETHThreshold,
+            "Staking ETH over threshold"
+        );
+        stakeETHTally += requiredETH;
+
+        // Convert required ETH from WETH
+        IWETH9(WETH_TOKEN_ADDRESS).withdraw(requiredETH);
```
pandadefi, line 174: A payable receive function seems to be missing to get the ETH out of the WETH withdraw.

naddison36: The `receive` is on the `NativeStakingSSVStrategy` contract as it also need to accept ETH from the FeeAccumulator.

I could change to use inheritance so it's clearer. That is, implement a `receive` on `ValidatorRegistrator` to accept payments from WETH and then override with a `super.receive` call on `NativeStakingSSVStrategy` to also accept payments from the `FeeAccumulator`. Thoughts?

naddison36: The inheritance idea with super will not work as that will execute the two require statements separately. Payment needs to be allowed from WETH or the FeeAccumulator.
The `receive` on `NativeStakingSSVStrategy` would have to override the `receive` on `ValidatorRegistrator` with no `super` call.

contracts/contracts/strategies/NativeStaking/FeeAccumulator.sol
```diff
@@ -0,0 +1,44 @@
...
+import { Governable } from "../../governance/Governable.sol";
+
+/**
+ * @title Fee Accumulator for Native Staking SSV Strategy
+ * @notice Receives execution rewards which includes tx fees and
+ * MEV rewards like tx priority and tx ordering.
+ * It does NOT include swept ETH from beacon chain consensus rewards or full validator withdrawals.
+ * @author Origin Protocol Inc
+ */
+contract FeeAccumulator is Governable {
```
DanielVF, line 15: I'm wondering if this contract should have a transferToken method that's only callable by the governor and does a safeTransfer of the coin.

Otherwise, we don't use Governable, and could remove it from this contract.

contracts/contracts/strategies/NativeStaking/ValidatorAccountant.sol
```diff
@@ -0,0 +1,269 @@
...
+    /// @dev this can change in the future with EIP-7251, Increase the MAX_EFFECTIVE_BALANCE
+    uint256 public constant MAX_STAKE = 32 ether;
+
+    /// @notice Keeps track of the total consensus rewards swept from the beacon chain
+    uint256 public consensusRewards;
+
+    /// @notice start of fuse interval
+    uint256 public fuseIntervalStart;
+    /// @notice end of fuse interval
+    uint256 public fuseIntervalEnd;
```
DanielVF, line 25: These will aways be relatively small numbers and will be read together. Perhaps we could pack them together as `uint128`. This will save a cold storage read during our accounting, which the protocol will be paying for to happen across multiple instances of this strategy.

contracts/contracts/strategies/NativeStaking/ValidatorAccountant.sol
```diff
@@ -0,0 +1,269 @@
...
+    /// @notice set fuse interval values
+    function setFuseInterval(
+        uint256 _fuseIntervalStart,
+        uint256 _fuseIntervalEnd
+    ) external onlyGovernor {
+        require(
+            _fuseIntervalStart < _fuseIntervalEnd &&
+                _fuseIntervalEnd < 32 ether &&
+                _fuseIntervalEnd - _fuseIntervalStart >= 4 ether,
+            "incorrect fuse interval"
```
DanielVF, line 79: In our existing code base, we use an initial uppercase letter in errors (unless talking about a function name).

For example:


image

contracts/contracts/strategies/NativeStaking/ValidatorAccountant.sol
```diff
@@ -0,0 +1,269 @@
...
+            IWETH9(WETH_TOKEN_ADDRESS).transfer(
+                VAULT_ADDRESS,
+                _ethToVaultAmount
+            );
+            _wethWithdrawnToVault(_ethToVaultAmount);
+        }
+
+        // rerun the accounting to see if it has now been fixed.
+        // Do not pause the accounting on failure as it is already paused
+        require(_doAccounting(false), "fuse still blown");
```
DanielVF, line 257: Start error with upper case

contracts/contracts/strategies/NativeStaking/ValidatorAccountant.sol
```diff
@@ -0,0 +1,269 @@
...
+            lastFixAccountingBlockNumber + MIN_FIX_ACCOUNTING_CADENCE <
+                block.number,
+            "manuallyFixAccounting called too soon"
+        );
+        require(
+            _validatorsDelta >= -3 &&
+                _validatorsDelta <= 3 &&
+                // new value must be positive
+                int256(activeDepositedValidators) + _validatorsDelta >= 0,
+            "invalid validatorsDelta"
```
DanielVF, line 221: Start error with upper case

contracts/contracts/strategies/NativeStaking/ValidatorAccountant.sol
```diff
@@ -0,0 +1,269 @@
...
+    /// to the vault, so it doesn't interfere with further accounting.
+    function manuallyFixAccounting(
+        int256 _validatorsDelta,
+        int256 _consensusRewardsDelta,
+        uint256 _ethToVaultAmount
+    ) external onlyStrategist whenPaused {
+        require(
+            lastFixAccountingBlockNumber + MIN_FIX_ACCOUNTING_CADENCE <
+                block.number,
+            "manuallyFixAccounting called too soon"
```
DanielVF, line 214: Start error with upper case

contracts/contracts/strategies/NativeStaking/ValidatorAccountant.sol
```diff
@@ -0,0 +1,269 @@
...
+    // slither-disable-end reentrancy-eth
+
+    /// @dev pause any further accounting if required and return false
+    function _failAccounting(bool pauseOnFail)
+        internal
+        returns (bool accountingValid)
+    {
+        // pause if not already
+        if (pauseOnFail) {
+            _pause();
```
DanielVF, line 191: If we are pausing the contract for accounting reasons, perhaps we should emit an event, beyond just the paused event?

contracts/contracts/strategies/NativeStaking/ValidatorAccountant.sol
```diff
@@ -0,0 +1,269 @@
...
+        uint256 _fuseIntervalEnd
+    ) external onlyGovernor {
+        require(
+            _fuseIntervalStart < _fuseIntervalEnd &&
+                _fuseIntervalEnd < 32 ether &&
+                _fuseIntervalEnd - _fuseIntervalStart >= 4 ether,
+            "incorrect fuse interval"
+        );
+
+        emit FuseIntervalUpdated(_fuseIntervalStart, _fuseIntervalEnd);
```
DanielVF, line 82: In our codebase, we almost always emit events _after_ setting the relevant values. 

image

The emit in this function, and the emit in `manuallyFixAccounting`, should move down below the variable sets. The intermediate emission in `_doAccounting` are fine, and happen right after the relevant variable setting.

contracts/contracts/strategies/NativeStaking/ValidatorAccountant.sol
```diff
@@ -0,0 +1,269 @@
...
+
+        // Calculate all the new ETH that has been swept to the contract since the last accounting
+        uint256 newSweptETH = address(this).balance - consensusRewards;
+        accountingValid = true;
+
+        // send the ETH that is from fully withdrawn validators to the Vault
+        if (newSweptETH >= MAX_STAKE) {
+            uint256 fullyWithdrawnValidators;
+            // safe since MAX_STAKE is hardcoded to 32ETH
+            unchecked {
```
DanielVF, line 128: Why do we need an unchecked around this? In general we avoid unchecked unless we really really really care. (and the rest of this function isn't using it in places that it could)

contracts/contracts/strategies/NativeStaking/ValidatorAccountant.sol
```diff
@@ -0,0 +1,269 @@
...
+        }
+        // Beacon chain consensus rewards swept (partial validator withdrawals)
+        else if (ethRemaining < fuseIntervalStart) {
+            // solhint-disable-next-line reentrancy
+            consensusRewards += ethRemaining;
+            emit AccountingConsensusRewards(ethRemaining);
+        }
+        // Beacon chain consensus rewards swept but also a slashed validator fully exited
+        else if (ethRemaining > fuseIntervalEnd) {
+            IWETH9(WETH_TOKEN_ADDRESS).deposit{ value: ethRemaining }();
```
DanielVF, line 164: Is it possible to disable slither reentrancy checking on just this single line. Then we don't need to disable it in other places in this function.

contracts/contracts/strategies/NativeStaking/ValidatorRegistrator.sol
```diff
@@ -0,0 +1,350 @@
...
+    ) {
+        WETH_TOKEN_ADDRESS = _wethAddress;
+        BEACON_CHAIN_DEPOSIT_CONTRACT = _beaconChainDepositContract;
+        SSV_NETWORK_ADDRESS = _ssvNetwork;
+        VAULT_ADDRESS = _vaultAddress;
+        MAX_VALIDATORS = _maxValidators;
+    }
+
+    /// @notice Set the address of the registrator which can register, exit and remove validators
+    function setRegistrator(address _address) external onlyGovernor {
```
DanielVF, line 130: In the next several functions, the emits should be after the variable sets.

contracts/contracts/strategies/NativeStaking/ValidatorRegistrator.sol
```diff
@@ -0,0 +1,350 @@
...
+        external
+        onlyRegistrator
+        whenNotPaused
+    {
+        uint256 requiredETH = validators.length * 32 ether;
+
+        // Check there is enough WETH from the deposits sitting in this strategy contract
+        require(
+            requiredETH <= IWETH9(WETH_TOKEN_ADDRESS).balanceOf(address(this)),
+            "insufficient WETH"
```
DanielVF, line 169: Capital first letter of error for consistency. (immediately following errors are capitalized, for example.)

image

contracts/contracts/strategies/NativeStaking/ValidatorRegistrator.sol
```diff
@@ -0,0 +1,350 @@
...
+         * bytes11(0) to fill up the required zeros
+         * remaining bytes20 are for the address
+         */
+        bytes memory withdrawal_credentials = abi.encodePacked(
+            bytes1(0x01),
+            bytes11(0),
+            address(this)
+        );
+
+        uint256 validatorsLength = validators.length;
```
DanielVF, line 197: We actually use `validators.length` earlier in the function. Perhaps this `validatorsLength` could move up right before `requiredETH` and then `requiredETH` could use it in calculations.

contracts/contracts/strategies/NativeStaking/ValidatorRegistrator.sol
```diff
@@ -0,0 +1,350 @@
...
+                pubKeyHash,
+                validators[i].pubkey,
+                32 ether,
+                withdrawal_credentials
+            );
+
+            validatorsStates[pubKeyHash] = VALIDATOR_STATE.STAKED;
+
+            unchecked {
+                ++i;
```
DanielVF, line 227: Let's just put this up in the for loop, and have the code be clearer.

contracts/contracts/strategies/NativeStaking/ValidatorRegistrator.sol
```diff
@@ -0,0 +1,350 @@
...
+    function registerSsvValidator(
+        bytes calldata publicKey,
+        uint64[] calldata operatorIds,
+        bytes calldata sharesData,
+        uint256 ssvAmount,
+        Cluster calldata cluster
+    ) external onlyRegistrator whenNotPaused {
+        bytes32 pubKeyHash = keccak256(publicKey);
+        require(
+            validatorsStates[pubKeyHash] == VALIDATOR_STATE.NON_REGISTERED,
```
DanielVF, line 253: The other methods do ` VALIDATOR_STATE currentState = validatorsStates[pubKeyHash];` then compare `currentState` in the require. Either way is fine, but let's be consistent. I lean towards the temp variable since we aren't running into stack space issues in these.

contracts/contracts/strategies/NativeStaking/NativeStakingSSVStrategy.sol
```diff
@@ -0,0 +1,322 @@
...
+        // collect ETH from execution rewards from the fee accumulator
+        uint256 executionRewards = FeeAccumulator(FEE_ACCUMULATOR_ADDRESS)
+            .collect();
+
+        // total ETH rewards to be harvested = execution rewards + consensus rewards
+        uint256 ethRewards = executionRewards + consensusRewards;
+
+        require(
+            address(this).balance >= ethRewards,
+            "insufficient eth balance"
```
DanielVF, line 130: Should start with a capital letter

contracts/contracts/strategies/NativeStaking/NativeStakingSSVStrategy.sol
```diff
@@ -0,0 +1,322 @@
...
+     * @notice Only accept ETH from the FeeAccumulator and the WETH contract - required when
+     * unwrapping WETH just before staking it to the validator
+     * @dev don't want to receive donations from anyone else as this will
+     * mess with the accounting of the consensus rewards and validator full withdrawals
+     */
+    receive() external payable {
+        require(
+            msg.sender == FEE_ACCUMULATOR_ADDRESS ||
+                msg.sender == WETH_TOKEN_ADDRESS,
+            "eth not from allowed contracts"
```
DanielVF, line 302: Capitalization on error message.

contracts/contracts/strategies/NativeStaking/NativeStakingSSVStrategy.sol
```diff
@@ -0,0 +1,322 @@
...
+        }
+    }
+
+    /// @notice Unlike other strategies, this does not deposit assets into the underlying platform.
+    /// It just checks the asset is WETH and emits the Deposit event.
+    /// To deposit WETH into validators `registerSsvValidator` and `stakeEth` must be used.
+    /// Will NOT revert if the strategy is paused from an accounting failure.
+    /// @param _asset Address of asset to deposit. Has to be WETH.
+    /// @param _amount Amount of assets that were transferred to the strategy by the vault.
+    function deposit(address _asset, uint256 _amount)
```
DanielVF, line 158: Since the vault no longer calls `deposit`, could we make this function revert, and then we can remove the code around `depositedWethAccountedFor` here and in `depositAll`?

contracts/contracts/strategies/NativeStaking/ValidatorAccountant.sol
DanielVF, line 164: 🔵 It's possible there's no clean way to do this, but it bugs me a little that we could be making two different transfers of the same coin to the same address in the same function. Feels like it's often better to do computation first, then transfers after.

contracts/contracts/strategies/NativeStaking/ValidatorAccountant.sol
```diff
@@ -0,0 +1,269 @@
...
+
+        uint256 ethRemaining = address(this).balance - consensusRewards;
+        // should be less than a whole validator stake
+        require(ethRemaining < 32 ether, "unexpected accounting");
+
+        // If no Beacon chain consensus rewards swept
+        if (ethRemaining == 0) {
+            // do nothing
+            return accountingValid;
+        }
```
DanielVF, line 155: It's odd to have a newline in between the end of if statement brace and the start of the else if. It makes the if statement look standalone, rather than part of a block of options. Our usual practice in the code base is to do `} else if(...) {

image

DanielVF: Is this something to do with the comments on the else if statement? Perhaps we could move them inside the else if?

contracts/contracts/strategies/NativeStaking/ValidatorAccountant.sol
```diff
@@ -0,0 +1,269 @@
...
+            emit AccountingFullyWithdrawnValidator(
+                fullyWithdrawnValidators,
+                activeDepositedValidators,
+                wethToVault
+            );
+        }
+
+        uint256 ethRemaining = address(this).balance - consensusRewards;
+        // should be less than a whole validator stake
+        require(ethRemaining < 32 ether, "unexpected accounting");
```
DanielVF, line 149: We seem to alternate between writing `32 ether` inline, or using MAX_STAKE, sometimes even inside the same function. We should probably use one consistently?

contracts/contracts/strategies/NativeStaking/ValidatorAccountant.sol
naddison36: Not sure. I'll remove

contracts/contracts/strategies/NativeStaking/ValidatorAccountant.sol
```diff
@@ -0,0 +1,269 @@
...
+/// @notice Attributes the ETH swept from beacon chain validators to this strategy contract
+/// as either full or partial withdrawals. Partial withdrawals being consensus rewards.
+/// Full withdrawals are from exited validators.
+/// @author Origin Protocol Inc
+abstract contract ValidatorAccountant is ValidatorRegistrator {
+    /// @notice The minimum amount of blocks that need to pass between two calls to manuallyFixAccounting
+    uint256 public constant MIN_FIX_ACCOUNTING_CADENCE = 7200; // 1 day
+    /// @notice The maximum amount of ETH that can be staked by a validator
+    /// @dev this can change in the future with EIP-7251, Increase the MAX_EFFECTIVE_BALANCE
+    uint256 public constant MAX_STAKE = 32 ether;
```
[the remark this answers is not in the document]
naddison36: changed

contracts/contracts/strategies/NativeStaking/ValidatorRegistrator.sol
```diff
@@ -0,0 +1,350 @@
...
+        // Convert required ETH from WETH
+        IWETH9(WETH_TOKEN_ADDRESS).withdraw(requiredETH);
+        _wethWithdrawnAndStaked(requiredETH);
+
+        /* 0x01 to indicate that withdrawal credentials will contain an EOA address that the sweeping function
+         * can sweep funds to.
+         * bytes11(0) to fill up the required zeros
+         * remaining bytes20 are for the address
+         */
+        bytes memory withdrawal_credentials = abi.encodePacked(
```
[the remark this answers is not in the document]
naddison36: no. I'll change

contracts/contracts/strategies/NativeStaking/NativeStakingSSVStrategy.sol
naddison36: fixed

contracts/contracts/strategies/NativeStaking/NativeStakingSSVStrategy.sol
naddison36: There's a design decision on whether the Native Staking Strategy should be the default strategy
https://www.notion.so/originprotocol/Should-the-Native-Staking-strategy-be-the-default-for-WETH-718c1b9a43154b13a05caee97e8a75d6

contracts/contracts/strategies/NativeStaking/ValidatorRegistrator.sol
```diff
@@ -0,0 +1,350 @@
...
+}
+
+/**
+ * @title Registrator of the validators
+ * @notice This contract implements all the required functionality to register, exit and remove validators.
+ * @author Origin Protocol Inc
+ */
+abstract contract ValidatorRegistrator is Governable, Pausable {
+    /// @notice The address of the Wrapped ETH (WETH) token contract
+    address public immutable WETH_TOKEN_ADDRESS;
```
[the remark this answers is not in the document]
naddison36: I've renamed `WETH_TOKEN_ADDRESS` to `WETH`

I've also renamed `SSV_TOKEN_ADDRESS` to `SSV_TOKEN` and `SSV_NETWORK_ADDRESS` to `SSV_NETWORK`

contracts/contracts/strategies/NativeStaking/ValidatorAccountant.sol
naddison36: I changed to use `FULL_STAKE`

contracts/contracts/strategies/NativeStaking/ValidatorAccountant.sol
naddison36: the weird format was from the comments. I've moved them inside the if and prettier has fixed

contracts/contracts/strategies/NativeStaking/ValidatorRegistrator.sol
naddison36: fixed

contracts/contracts/strategies/NativeStaking/ValidatorRegistrator.sol
naddison36: putting the array length in memory only saves gas when dealing with storage variable arrays. In this case, its actually a little less gas to use `validators.length`. 139,421 compared to 139,424 using `validatorsLength`.

I'll remove `validatorsLength`

contracts/contracts/strategies/NativeStaking/ValidatorRegistrator.sol
naddison36: changed

contracts/contracts/strategies/NativeStaking/ValidatorRegistrator.sol
naddison36: I've added the temp variable `currentState`

contracts/contracts/strategies/NativeStaking/ValidatorAccountant.sol
naddison36: I thought the same but it actually costs more gas. Doing this change the average. For the Holesky fork tests, `doAccounting` average gas costs increased by 642 gas.

The reason is reading the second storage slot with `fuseIntervalEnd` is conditional on `ethRemaining` being greater than `fuseIntervalStart`. Most of the time the `ethRemaining` will be less than `fuseIntervalStart` hence the `ethRemaining > fuseIntervalEnd` condition is not executed.

contracts/contracts/strategies/NativeStaking/FeeAccumulator.sol
naddison36: I've got rid of `Governable`.

The contract is proxied so if a large amount of ERC20 tokens were accidentally transferred to this contract we could upgrade and recover the tokens. Otherwise, `Governable` doesn't do anything.

contracts/contracts/strategies/NativeStaking/ValidatorAccountant.sol
naddison36: I'm not convinced adding a second `AccountingPaused` event will add much value over the existing `Paused` event.

contracts/contracts/strategies/NativeStaking/NativeStakingSSVStrategy.sol
```diff
@@ -0,0 +1,312 @@
...
+    /// can happen when:
+    ///   - the deposit was not a multiple of 32 WETH
+    ///   - someone sent WETH directly to this contract
+    /// Will NOT revert if the strategy is paused from an accounting failure.
+    /// @param _recipient Address to receive withdrawn assets
+    /// @param _asset WETH to withdraw
+    /// @param _amount Amount of WETH to withdraw
+    function withdraw(
+        address _recipient,
+        address _asset,
```
DanielVF, line 177: We never validate this asset. This function should have a `require(_asset == WETH, "Unsupported asset");`.

contracts/contracts/strategies/NativeStaking/NativeStakingSSVStrategy.sol
```diff
@@ -0,0 +1,312 @@
...
+    /// @param _asset Address of asset to deposit. Has to be WETH.
+    /// @param _amount Amount of assets that were transferred to the strategy by the vault.
+    function deposit(address _asset, uint256 _amount)
+        external
+        override
+        onlyVault
+        nonReentrant
+    {
+        require(_asset == WETH, "Unsupported asset");
+        depositedWethAccountedFor += _amount;
```
DanielVF, line 131: Withdrawing does not adjust `depositedWethAccountedFor`. If the strategist were to

- deposit(32 weth)
- withdraw(32 weth)
- deposit(32 weth)

then the `depositedWethAccountedFor` would be doubled. This would also cause a subsequent `depositAll()` to revert when doing `wethBalance - depositedWethAccountedFor;`

contracts/contracts/strategies/NativeStaking/NativeStakingSSVStrategy.sol
```diff
@@ -0,0 +1,312 @@
...
+
+    function _withdraw(
+        address _recipient,
+        address _asset,
+        uint256 _amount
+    ) internal {
+        require(_amount > 0, "Must withdraw something");
+        require(_recipient != address(0), "Must specify recipient");
+
+        emit Withdrawal(_asset, address(0), _amount);
```
DanielVF, line 191: bump emit of event down

contracts/contracts/strategies/NativeStaking/NativeStakingSSVStrategy.sol
```diff
@@ -0,0 +1,312 @@
...
+        );
+
+        if (ethRewards > 0) {
+            // reset the counter keeping track of beacon chain consensus rewards
+            consensusRewards = 0;
+
+            // Convert ETH rewards to WETH
+            IWETH9(WETH).deposit{ value: ethRewards }();
+
+            emit RewardTokenCollected(harvesterAddress, WETH, ethRewards);
```
DanielVF, line 290: bump emit of event down and transfer up right under the deposit?

contracts/contracts/strategies/NativeStaking/NativeStakingSSVStrategy.sol
```diff
@@ -0,0 +1,312 @@
...
+        view
+        override
+        returns (uint256 balance)
+    {
+        require(_asset == WETH, "Unsupported asset");
+
+        balance =
+            // add the ETH that has been staked in validators
+            activeDepositedValidators *
+            32 ether +
```
DanielVF, line 228: We use `32 eth` several times in this file. Perhaps this should change to FULL_STAKE?

contracts/contracts/strategies/NativeStaking/ValidatorRegistrator.sol
```diff
@@ -0,0 +1,349 @@
...
+    /// @param validators A list of validator data needed to stake.
+    /// The `ValidatorStakeData` struct contains the pubkey, signature and depositDataRoot.
+    /// Only the registrator can call this function.
+    // slither-disable-start reentrancy-eth
+    function stakeEth(ValidatorStakeData[] calldata validators)
+        external
+        onlyRegistrator
+        whenNotPaused
+    {
+        uint256 requiredETH = validators.length * 32 ether;
```
DanielVF, line 164: Should this function switch to using FULL_STAKE instead of 32 ether? Would mean we move the definition of FULL_STAKE down into this contract

contracts/contracts/strategies/NativeStaking/ValidatorRegistrator.sol
```diff
@@ -0,0 +1,349 @@
...
+                validators[i].depositDataRoot
+            );
+
+            validatorsStates[pubKeyHash] = VALIDATOR_STATE.STAKED;
+
+            emit ETHStaked(
+                pubKeyHash,
+                validators[i].pubkey,
+                32 ether,
+                withdrawalCredentials
```
DanielVF, line 222: These `withdrawalCredentials` are essentially hardcoded and will never change without some fairly major changes to the ethereum. Is there any reason for emitting the withdrawalCredentials?

contracts/contracts/strategies/NativeStaking/ValidatorRegistrator.sol
```diff
@@ -0,0 +1,349 @@
...
+
+    /// @notice Registers a new validator in the SSV Cluster.
+    /// Only the registrator can call this function.
+    /// @param publicKey The public key of the validator
+    /// @param operatorIds The operator IDs of the SSV Cluster
+    /// @param sharesData The validator shares data
+    /// @param ssvAmount The amount of SSV tokens to be deposited to the SSV cluster
+    /// @param cluster The SSV cluster details including the validator count and SSV balance
+    // slither-disable-start reentrancy-no-eth
+    function registerSsvValidator(
```
DanielVF, line 239: No change needed, just a note. I see that each of these SSV methods can only operate on one validator at a time. That may be somewhat less gas efficient, since I think we could do multiple submits on one cluster data.

However, I'm good keeping this as is for launch and we can tweak as needed later if active use shows it would matter.

contracts/contracts/strategies/NativeStaking/ValidatorAccountant.sol
```diff
@@ -0,0 +1,261 @@
...
+    /// the accounting function will treat that ETH as Beacon chain consensus rewards.
+    /// On the contrary, if after dividing by 32, the ETH remaining on the contract falls between fuseIntervalEnd and 32,
+    /// the accounting function will treat that as a validator slashing.
+    /// @notice Perform the accounting attributing beacon chain ETH to either full or partial withdrawals. Returns true when
+    /// accounting is valid and fuse isn't "blown". Returns false when fuse is blown.
+    /// @dev This function could in theory be permission-less but lets allow only the Registrator (Defender Action) to call it
+    /// for now.
+    /// @return accountingValid true if accounting was successful, false if fuse is blown
+    /* solhint-enable max-line-length */
+    function doAccounting()
```
DanielVF, line 101: Is there any need for the strategist to call this, or should the strategist just call manuallyFixAccounting with zero values to force an accounting?

DanielVF (changes requested):

contracts/contracts/strategies/NativeStaking/NativeStakingSSVStrategy.sol
naddison36: added

contracts/contracts/strategies/NativeStaking/NativeStakingSSVStrategy.sol
naddison36: good idea as EIP-7251, Increase the MAX_EFFECTIVE_BALANCE, will allow us to stake more than 32 ether.

contracts/contracts/strategies/NativeStaking/NativeStakingSSVStrategy.sol
naddison36: done

contracts/contracts/strategies/NativeStaking/NativeStakingSSVStrategy.sol
naddison36: done

contracts/contracts/strategies/NativeStaking/NativeStakingSSVStrategy.sol
naddison36: Good pickup. We should have caught this.

I've put this change in a separate PR back to this branch as its non-trivial https://github.com/OriginProtocol/origin-dollar/pull/2091

contracts/contracts/strategies/NativeStaking/ValidatorRegistrator.sol
naddison36: Yes, I've switched it and a few other spots in `ValidatorRegistrator` to use FULL_STAKE instead of 32 ether. It puts us in a better place for EIP-7251, Increase the MAX_EFFECTIVE_BALANCE

contracts/contracts/strategies/NativeStaking/ValidatorRegistrator.sol
naddison36: with the recent change to the P2P API introducing a delay between registering the SSV validator and getting the deposit data, which incudes the signature, I think we need to change `registerSsvValidator` to register multiple validators. I'll do this in a separate PR

contracts/contracts/strategies/NativeStaking/ValidatorRegistrator.sol
naddison36: Good point. It's just the strategy proxy address left padded. I'll take it out as it saves 1k in gas

contracts/contracts/strategies/NativeStaking/ValidatorAccountant.sol
naddison36: Good question. `manuallyFixAccounting` requires the strategy to be paused, but the Strategist can `pause` the contract, call `manuallyFixAccounting` which then calls `doAccounting`. I think that is enough.

contracts/contracts/strategies/NativeStaking/NativeStakingSSVStrategy.sol
```diff
@@ -188,8 +191,10 @@ contract NativeStakingSSVStrategy is
         require(_amount > 0, "Must withdraw something");
         require(_recipient != address(0), "Must specify recipient");
 
-        emit Withdrawal(_asset, address(0), _amount);
+        _wethWithdrawn(_amount);
```
DanielVF, line 194: This `_wethWithdrawn` is really just an emit? It should go after the transfer?

contracts/contracts/strategies/NativeStaking/ValidatorAccountant.sol
```diff
@@ -0,0 +1,258 @@
...
+    uint256 public consensusRewards;
+
+    /// @notice start of fuse interval
+    uint256 public fuseIntervalStart;
+    /// @notice end of fuse interval
+    uint256 public fuseIntervalEnd;
+    /// @notice last block number manuallyFixAccounting has been called
+    uint256 public lastFixAccountingBlockNumber;
+
+    uint256[49] private __gap;
```
DanielVF, line 26: I think the usual best practice is to make the initial size of the gap variable such that the total storage slots used fit into some nice round number. If a contract inheritance layer had one storage slot used, it would make sense to gap 49, which we do elsewhere. But here with 4 storage slots used, a gap of 46 might make more sense.

contracts/contracts/strategies/NativeStaking/ValidatorRegistrator.sol
```diff
@@ -0,0 +1,345 @@
...
+    mapping(bytes32 => VALIDATOR_STATE) public validatorsStates;
+    /// @notice The account that is allowed to modify stakeETHThreshold and reset stakeETHTally
+    address public stakingMonitor;
+    /// @notice Amount of ETH that can be staked before staking on the contract is suspended
+    /// and the governor needs to approve further staking
+    uint256 public stakeETHThreshold;
+    /// @notice Amount of ETH that can has been staked since the last governor approval.
+    uint256 public stakeETHTally;
+    // For future use
+    uint256[47] private __gap;
```
DanielVF, line 54: Similar gap comment - we have six storage slot variables used at this layer, so our gap should be 44 slots?

contracts/deploy/mainnet/097_native_ssv_staking.js
```diff
@@ -0,0 +1,224 @@
...
+          contract: cStrategy,
+          signature: "setRegistrator(address)",
+          args: [addresses.mainnet.validatorRegistrator],
+        },
+        // 6. set staking threshold
+        {
+          contract: cStrategy,
+          signature: "setStakeETHThreshold(uint256)",
+          // TODO: confirm this number makes sense
+          args: [ethers.utils.parseEther("1024")], // 32ETH * 32
```
DanielVF, line 206: This is roughly 4 million dollars. May want to start with a lower amount since this essentialy max loss per strategy to a badly behaving strategist. 512 ETH?

contracts/contracts/strategies/NativeStaking/ValidatorAccountant.sol
naddison36: the problem is we have been deploying to Holesky whilst making changes. That's why its now 49 and not 50 which it started at.

contracts/contracts/strategies/NativeStaking/ValidatorRegistrator.sol
naddison36: same as the other one. It started at 50 when we first deployed to Holesky but with changes it's now 47

contracts/deploy/mainnet/097_native_ssv_staking.js
naddison36: We can start at 512. The Strategist will have to call `resetStakeETHTally` after 16 validators have been registered. Once we get a roll on the migration we could up the `stakeETHThreshold` to something higher using governance and then drop it back down after the migration.

naddison36: actually, its the 5/8 multisig that needs to call `resetStakeETHTally`

contracts/contracts/strategies/NativeStaking/ValidatorAccountant.sol
DanielVF: If we already have contracts live that we are upgrading, it is what is. We'll keep.

DanielVF:
_Posting my mostly complete review_

# Code Review: Native Staking

## Requirements

OETH should be able to deposit and withdraw ETH from SSV controlled native staking as a yield earning strategy. It is planned that this will be the primary yield for OETH and that OETH will become an LST.

This is a difficult strategy to build because validators operate cross chain, and the execution chain only has very poor information about the actual status of validators. What information can be gathered far to gas intensive to be effective.

Because Ethereum has horrible support for LSTs, there are essentially two approaches to building an LST. The first is YOLO trust of an address that updates the execution layer about what is going on the validator layer. The second is building a massive offchain oracle infrastructure and operating it.

We've chosen a third route:
1. If everything goes smoothly, the accounting operates off onchain data.
3. A limited-trust address interacts with P2P/SSV for validator provisioning. The amount of damage this address can do is limited to a maximum amount.
4. In the event of things getting out of sync enough to cause accounting errors, the strategist can make limited corrections at a certain rate over time.
5. In the event of things being very out of sync, governance can act to correct.

The overall concept is contract enforced damage mitigation.

We want to transition to fully onchain accounting once that is possible from future Ethereum updates.

#### Authentication

- [x] Never use tx.origin

#### Ethereum

_Unlike almost all of contracts, we do send and receive ETH from this contract. However, we only send ETH to two trusted addresss.

🟠 There are three methods to prevent reentrancy: only calling trusted contracts, reentrancy guards, and writing functions in order such that reentrancy is not an issue. Usually we try to hit all three. This code only does the first. 



#### Cryptographic code

The only crypto used is hashing the validator public keys in order to look them up in a mapping. I don't think that can be abused to get different public keys to map to the same hash.

Otherwise the contract passes through signature data, but does not use it itself.


#### Gas problems

    - [ ] A way to remove items

#### Black magic

- [x] Does not use `delegatecall` outside of proxying. 

#### Overflow

#### Proxy

#### Events

🟣 Deposit SSV does not emit directly, but that's okay. We could monitor the SSV contract for this if we wished.

#### Rounding

#### Dependencies

#### External calls

	- [x] 🔴 No validation in withdraw (_made inline note, fixed_).
- [x] No malicious behaviors
- [x] Oracles, one of:
	- [x] It's complicated
  - [x] No oracles
  - [ ] Oracles can't be bent

#### Tests

_not checked yet_

#### Deploy

## Strategy Specific

#### Strategy checks

- [x] **Check balance cannot be manipulated up AND down by an attacker**
	- [x] Way more complicated, but we can get funds out.
	- [x] Limited amounts though 



#### Accounting Invariants

- [x] depWeth should go up whenever WETH is added via the vault
- [x] depWeth should go down whenever WETH is staked
- [x] depWeth should go down whenever WETH is manually withdrawn to the vault
- [ ] depWETH should never be greater than the standing WETH balance (_not checked yet_)
- [x] depWETH should never be less than the standing WETH balance unless someone donates
- [x] ETH only arrives from off chain + mev rewards (outside donations)
- [x] Sitting ETH is only ever sent to WETH
- [x] All funds sent upwards to the vault and harvester are done in WETH
- [x] ActVal * 32 should always be less than the sum of WETH sent to the contract
	- [x] not counting manual adjustments
- [x] ETH balance at rest should always be greater than or equal to consensus rewards
	- [x] manual adjustments can set this below, but it will revert at the end of the manual set when it calls do accounting
- [x] After a successful doAccounting, eth balance should equal consensus rewards
	- [x] unless accounting fails and the contract is paused

#### What happens if:
- _We exit all validators?_ - should be fairly sane. No more rewards than we would normally get.
- _Lots of validators get slashed?_ - accounting will undercount how many validators have withdrawn. We must pause() the strategy accounting when this happens on the beacon chain, before the slashed validators eventually get withdrawn.
- _All validators go down for a week_ - accounting thinks we have an extra validator, and would need manual adjustment.
- _We miss a rewards beat_ - we blow the fuse and stop updating.
- _gas prices spike for a week_ - no real effect
- _Validator max sizes go up, the number of global active validators goes down by 20x, resulting in a 20x withdraw cycle  speedup_ - okay if we call do accounting 20x more often. Otherwise we need to migrate to larger stakes and/or switch to using the better accounting which is probably gas possible now.

#### Logic

  _Still not specifically reviewed yet_

- [ ] SSV strategy proxy must be deployed by the relayer account
- [ ] Before non testing amounts of funds are in, we must have monitoring for slashings and missed validations

#### Internal State

See above in "Accounting Invariants" section


#### Attack

This code is inherently dangerous. We focus on damage mitigation rather than the usual perfection.

First, there is no way for a smart contract to ensure that a staking deposit will end up with correct withdrawal address in the beacon chain. We are essentially shipping out gold bars via packing labels created on an EOA.

To mitigate this risk, we limit how many deposits can be made before requiring a 5 of 8 multisig to verify that the previous deposits were correctly made. This caps our max losses at about two million per strategy contract.

We could reduce the deposit risk by a magnitude and remove all humans from the loop by doing the following:

- Deposit 1 ETH into a validator, increment, mark validator as pending
- When this validator is active, post back to the contract with a proof that the validator withdraw address is recorded correctly. This deposit 31 ETH and marks the validator as registered.

By keeping a limited number of deposits in the pending state, we reduce our risk. 1 ETH instead of 32 ETH to test out a validator  deposit is also a fairly big win.

Secondly, it would take approximately 50% of all yield to be spent in gas to be mostly sure about what each validator is doin on the beacon chain. Instead we take a guess based on the amount of funds we see coming in. This can be wrong and need manual correction. We also throttle the amount of manual correction to avoid malicious actions.

Outside the scope of this contract, failures or vulnerabilities with SSV or P2P could result in mass slashings of our validators.

Withdrawing funds is also ultimately outside the contract's control. We can ask the SSV contract to emit an event which hopefully is picked up. There are other off-chain ways to kick off withdraws.

SSV rewards are also outside our contracts control, since they go only to the EOA that deployed the contracts.

ETH rewards are not distinguished from withdraws, both go to the same place, via the same mechanism. We use heuristics to discriminate between them.

Our strategy contracts have four jobs:

1) Deposit money
2) Withdraw money
3) Collect yield + rewards
4) Know the balance

Four out of four of these cannot be guaranteed by the code.

Monitoring is critical here.

Other code:

We might be in trouble if the SSV deposit made callbacks to somewhere that an attacker could hook into. A quick glance at the SSV code shows that not only is it dynamic, but extremely dynamic in its dynamicness.

🟠 Let's add `nonReentrant` to `stakeETH`, `doAccounting` and `manuallyFixAccounting` so something weird happening with SSV can't mess anything up on our side.

#### Flavor

The strategy is broken into three different files in an inheritance hierarchy. In general I'm not a fan of this, as it makes it more difficult to see the big picture, but the cutpoints here are very good, and each has clear ownership of difference concerns, and small interaction with other layers. Works here.

Otherwise, the flavor is now good.

contracts/contracts/strategies/NativeStaking/ValidatorRegistrator.sol
```diff
@@ -0,0 +1,346 @@
...
+        stakeETHTally = 0;
+        emit StakeETHTallyReset();
+    }
+
+    /// @notice Stakes WETH to the node validators
+    /// @param validators A list of validator data needed to stake.
+    /// The `ValidatorStakeData` struct contains the pubkey, signature and depositDataRoot.
+    /// Only the registrator can call this function.
+    // slither-disable-start reentrancy-eth
+    function stakeEth(ValidatorStakeData[] calldata validators)
```
DanielVF, line 159: Let's add `nonReentrant` to stakeETH, doAccounting and manuallyFixAccounting so something weird happening with SSV can't mess anything up on our side.

naddison36: I've added `nonReentrant` to `stakeETH`, `doAccounting` and `manuallyFixAccounting`

DanielVF (commented):
-

DanielVF (dismissed):
The NativeSSVStrategy contract is approved for use with test amounts of funds.

I have not yer reviewed the rest of this PR.

naddison36 (approved):
All pre deployment checks have bene done

PR #2019, opened 2024-04-15
Aerodrome OETH AMO

PraneshASP:

Contract change checklist:
  - [ ] Unit tests pass

### Notion resources:
- [Aero AMO RFC](https://notion.so/originprotocol/Pranesh-AMO-RFC-57f22709b6634c408d8f7f3572961bb7)


contracts/contracts/strategies/AerodromeEthStrategy.sol
```diff
@@ -0,0 +1,592 @@
...
+    }
+
+    /**
+     * @dev Calculates the constant K for the sAMM pool based on the reserves r1 and r2.
+     * The formula for K is: K = (r1^3 * r2) + (r2^3 * r1)
+     * @param r1 The reserve of wETH.
+     * @param r2 The reserve of OETH.
+     * @return The calculated constant K.
+     */
+    function getK(uint256 r1, uint256 r2) internal pure returns (uint256) {
```
DanielVF, line 584: Since this is an internal function, it should get an underscore.

contracts/contracts/strategies/AerodromeEthStrategy.sol
```diff
@@ -0,0 +1,592 @@
...
+        nonReentrant
+    {
+        // Collect AERO
+        aeroGaugeAddress.getReward(address(this));
+        _collectRewardTokens();
+    }
+
+    function _lpWithdraw(uint256 _lpTokensAmount) internal {
+        // Claim remaining rwards before withdrawing LP Tokens
+        aeroGaugeAddress.getReward(address(this));
```
DanielVF, line 454: Do we lose these reward tokens if we don't withdraw them? If not, it might just make things easier to not withdraw them until they are ready to be passed on. Less gas, and less things that could go wrong during a withdraw.

contracts/contracts/strategies/AerodromeEthStrategy.sol
```diff
@@ -0,0 +1,592 @@
...
+    }
+
+    /**
+     * @dev Checks the Aero sAMM pool's balances have improved and the balances
+     * have not gone beyond the threshold.
+     * The standard deposit function adds to both sides of the pool in a way that
+     * the pool's balance is not worsened.
+     * Withdrawals are proportional so doesn't change the pools asset balance.
+     */
+    modifier improvePoolBalance() {
```
DanielVF, line 73: This method should check the pool's balances before the action that the modifier wraps, then compare them again after. 

This way we check that things got better during the action.

If I'm reading this correctly, this only compares the balances/ratios after the action has taken place, which means it will not be able to compare the before and after to see if the difference is better?

What is this modifier securing as written?

contracts/contracts/strategies/AerodromeEthStrategy.sol
```diff
@@ -0,0 +1,592 @@
...
+        uint256 poolWETHBalance = wethCoinIndex == 0
+            ? lpTokenAddress.reserve0()
+            : lpTokenAddress.reserve1();
+
+        /* K is multiplied by 1e36 which is used for higher precision calculation of required
+         * pool LP tokens. Without it the end value can have rounding errors up to precision of
+         * 10 digits. This way we move the decimal point by 36 places when doing the calculation
+         * and again by 36 places when we are done with it.
+         */
+        uint256 k = (1e36 * IERC20(address(lpTokenAddress)).totalSupply()) /
```
DanielVF, line 324: Might want to avoid calling this 'k' because it's a very different thing than the AMM's k, which is also calculated elsewhere in the code.

contracts/contracts/strategies/AerodromeEthStrategy.sol
```diff
@@ -0,0 +1,592 @@
...
+            poolWETHBalance;
+        // prettier-ignore
+        // slither-disable-next-line divide-before-multiply
+        uint256 diff = (_wethAmount + 1) * k;
```
DanielVF, line 328: This calculation currently calculates a ratio, then uses that ratio to calculate the end result. This is slightly lossy in doing two calculations with rounding errors, which requires the extra high intermediate precision.

Would this be much simpler and less  if we just move the (_wethAmount+1) to be directly multiplied by the totalSupply in the first step?

Something like:

`(_wethAmount+1) * totalSupply  / poolWETHBalance` ?

contracts/contracts/strategies/AerodromeEthStrategy.sol
```diff
@@ -0,0 +1,592 @@
...
+     * @param _amountIn Amount of tokens to swap for.
+     * @param _minAmountOut Minimum amount of expected output amount (slippage adjusted)
+     * @param _tokenIn Address of the input token.
+     * @param _recipient Address of the recipient of weth tokens if we swap oeth for weth.
+     */
+    function swapAndRebalancePool(
+        uint256 _amountIn,
+        uint256 _minAmountOut,
+        address _tokenIn,
+        address _recipient
```
DanielVF, line 384: The strategist should not be able to steal tokens from the strategy / vault. We should probably remove the `_recipient` option here.

contracts/contracts/strategies/AerodromeEthStrategy.sol
PraneshASP: > What is this modifier currently trying to do?
This modifier is used to make sure after rebalancing, the pool ratio doesn't skew more than the allowed threshold. As long as the ratio remains within threshold, we should be good. It helps avoid errors from the strategist while rebalancing the pool. This also gives the strategist the flexibility to rebalance the pool as per the market conditions. 

> This way we check that things got better during the action.
To validate that we should define what we mean by "better" for the pool. Does it mean the pool should be closer to 50:50 ratio or should it be unbalanced till the safest point.

contracts/contracts/strategies/AerodromeEthStrategy.sol
PraneshASP: If we remove the _recipient, where should the swapped `weth` sent? Will it be held in the AMO contract?

contracts/contracts/strategies/AerodromeEthStrategy.sol
DanielVF: "Better" is towards 50/50.

Going towards 50/50 is safer because the strategist can't trade away from 50/50 which is an action that guaranteed loses money.

Going towards 50/50 is also more flexible, in that the strategist may want to move only a little towards repegging things. This can keep the price still off peg, but be profitable for the strategy, and slow down redeems.

PraneshASP: Fixed here: [7933c0e](https://github.com/OriginProtocol/origin-dollar/pull/2019/commits/7933c0e137c7efc9b826ae1bf7074337aa84163b)

contracts/contracts/oracle/BaseOETHOracleRouter.sol
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    /**
+     * @notice Returns the total price in 18 digit units for a given asset.
+     *         This implementation does not (!) do range checks as the
+     *         parent OracleRouter does.
+     * @param asset address of the asset
+     * @return uint256 unit price for 1 asset unit, in 18 decimal fixed
+     */
+    function price(address asset)
```
sparrowDom, line 25: this function is still identical to the one in `OETHOracleRouter`. You can create a `OETHOracleRouterBase` and implement the `price` function in it. And then have `OETHOracleRouter` and `BaseOETHOracleRouter` inherit from it

PraneshASP: Refactored here: [826361f](https://github.com/OriginProtocol/origin-dollar/pull/2061/commits/826361ffd619e8c2f87bb7883d7fafb84f86c165)

contracts/contracts/strategies/AerodromeEthStrategy.sol
PraneshASP: I tired using multiple values, it doesn't make any difference. so yea we can use the one you suggested, might save some gas too. 

| Action   | Value                         | Formula 1               | Formula 2               |
|----------|-------------------------------|-------------------------|-------------------------|
| withdraw | 220                           | 82862338377910162699    | 82862338377910162699    |
| withdraw | 1                             | 1035779229723877034     | 1035779229723877034     |
| withdraw | 0.0000000001                  | 103577924               | 103577924               |
| withdrawAll | 30000000000000000001       | 30000000000000000001    | 30000000000000000001    |

contracts/contracts/strategies/AerodromeEthStrategy.sol
PraneshASP: Good point, looks like the gauge tracks the rewards during withdrawals. So no need to harvest during withdrawals. Removed.

contracts/contracts/strategies/AerodromeEthStrategy.sol
PraneshASP: Renamed it.

PR #2031, opened 2024-04-25
Fix WOETH Zapper: Include feeAmount in `ccipSend()` call

PraneshASP:
Fees were not being accounted for. This fixes that.

Contract change checklist:
  - [x] Unit tests pass


contracts/contracts/zapper/WOETHCCIPZapper.sol
```diff
@@ -146,4 +172,30 @@ contract WOETHCCIPZapper {
         // Return the message ID
         return messageId;
     }
+
+    function _estimateFee(uint256 amount, address receiver)
```
DanielVF, line 176: This method is called estimateFee. Is this calculating the exact fee, or could it be wrong? If it's not exact, what happens when high or low?

contracts/contracts/zapper/WOETHCCIPZapper.sol
```diff
@@ -104,6 +122,13 @@ contract WOETHCCIPZapper {
         internal
         returns (bytes32 messageId)
     {
+        // Estimate fee for zapping.
+        uint256 feeAmount = estimateFee(amount, receiver);
+        if (feeAmount >= amount) revert AmountLessThanFee();
```
DanielVF, line 127: I think stylistically we just always use braces on ifs.

contracts/contracts/zapper/WOETHCCIPZapper.sol
DanielVF: > returns guaranteed execution fee for the specified message delivery to the destination chain

If this is true from the chainlink docs, perhaps we should remove the "estimate" from the name.

contracts/contracts/zapper/WOETHCCIPZapper.sol
```diff
@@ -92,6 +95,21 @@ contract WOETHCCIPZapper {
...
 
+    /**
+     * @notice Used to estimate fee for CCIP transaction
+     * @param amount The amount of ETH to be zapped
+     * @param receiver The address of the EOA on the destination chain
+     * @return feeAmount The CCIP tx fee in ETH.
+     */
+
+    function estimateFee(uint256 amount, address receiver)
+        public
```
DanielVF, line 106: Can be external, or can just be public and move the actual logic up into it.

DanielVF (approved):

PR #2034, opened 2024-04-26
manuallyFixAccounting changes

naddison36:
# Contract changes

* `manuallyFixAccounting` now uses delta values
* `manuallyFixAccounting` calls `doAccounting` to check the fuse is still not blown
* Removed `accountingGovernor` as the Strategist now calls `manuallyFixAccounting`


contracts/contracts/strategies/NativeStaking/ValidatorAccountant.sol
```diff
@@ -170,66 +158,65 @@ abstract contract ValidatorAccountant is ValidatorRegistrator {
...
-        uint256 ethBalance = address(this).balance;
-        uint256 wethBalance = IWETH9(WETH_TOKEN_ADDRESS).balanceOf(
-            address(this)
+        int256 _validatorsDelta,
+        int256 _consensusRewardsDelta,
+        uint256 _wethToVaultAmount
+    ) external onlyStrategist whenPaused {
+        require(
+            _validatorsDelta >= -1 &&
+                _validatorsDelta <= 1 &&
```
sparrowDom, line 183: the -1 -> 1 accepted validator delta and -32 -> 32 accepted consensus delta could be constants so that it is easier to configure these values if/when we want. 

Also if these ranges are not able to fix the accounting we need to re-deploy the contract via OUSD governance (slow process) to introduce larger deltas. I think we should be comfortable changing up to 3 or even 5 validators (and 3*32 or 5*32 ETH in consensus rewards) without being worried that such transaction could rug anyone or produce so much rewards for the attacker (controlling the strategist) to somehow damage the protocol.

contracts/contracts/strategies/NativeStaking/ValidatorAccountant.sol
```diff
@@ -111,16 +91,24 @@ abstract contract ValidatorAccountant is ValidatorRegistrator {
...
     function doAccounting()
         external
         onlyRegistrator
         whenNotPaused
         returns (bool accountingValid)
     {
+        accountingValid = _doAccounting();
+    }
+
+    function _doAccounting() internal returns (bool accountingValid) {
```
sparrowDom, line 105: I know this works for now, I am slightly worried that we will in future forget that internal _doAccounting needs to be in `whenNotPaused` context when doing the accounting and in `whenPaused` context when manually correcting the accounting.

What if we introduce a bool property to the internal `_doAccounting` function and be explicit whether a `_pause()` call should be made depending on who the caller of the transaction is. Smth like: 

```
function _doAccounting(bool callPause) internal...
...
  // pause and fail the accounting
  if (callPause) {
      _pause();
  }
  // fail the accounting
  return false;
....
```

sparrowDom (commented):
Great work. Looks good, the one thing that I'd like to increase is the deltas that the strategist can manipulate.

contracts/contracts/strategies/NativeStaking/ValidatorAccountant.sol
naddison36: Thanks, I'll update to 2 and 332

contracts/contracts/strategies/NativeStaking/ValidatorAccountant.sol
naddison36: using a `pauseOnFail` param also saves as storage slot read. I'll make the change

sparrowDom (approved):
LGTM

PR #2038, opened 2024-04-29
add deploy and test tools for multi chain deployments

sparrowDom:
add deploy and test tools for multi chain deployments


sparrowDom:
Rerunning tests. 6 fork tests were failing. 5 of them ran perfectly fine for me locally. The Frx failing one will be resolved when we remove frx strategies from the Vault. The vault has no more frxETH

naddison36 (approved):
Looks good

contracts/hardhat.config.js
```diff
@@ -46,63 +60,20 @@ task("accounts", "Prints the list of accounts", async (taskArguments, hre) => {
...
-  }).json();
-} else if (isForkTest) {
-  console.log(`Starting a fresh node on block: ${forkBlockNumber}`);
+const paths = {};
+if (isHolesky || isHoleskyForkTest || isHoleskyFork) {
+  // holesky deployment files are in contracts/deploy/holesky
+  paths.deploy = "deploy/holesky";
+} else {
+  // holesky deployment files are in contracts/deploy/mainnet
+  paths.deploy = "deploy/mainnet";
```
shahthepro, line 71: If we are going to do this, should we also split it based on the chain? Like `mainnet` for Ethereum, `arbitrum` for Arb One and `base` for Base?

One thing about splitting is that we won't be able to test all contracts at once in unit tests. However, it'll help trimming down the fixture (we already have a different fixture file for each network though)

sparrowDom: good point, how do you feel about it? What would you rather have?

My thinking is that we should separate and compartmentalise into separate modules things that aren't strictly connected. This should enable us (like you mention) to separate the fixture files. Also when different people work on functionalities / deploys of different chains, there should be less code overlap and thus less merge conflicts. Also tests for the chain 1 person is working on should ran faster if things are separated.

sparrowDom:
I will merge this in and apply any changes in later PRs if so requested

PR #2042, opened 2024-05-02
add `woethReceived` as `woethAmount` in zap result

apexearth:

Contract change checklist:
  - [ ] Unit tests pass

### DIFF
--- a/contracts/contracts/zapper/WOETHCCIPZapper.sol
+++ b/contracts/contracts/zapper/WOETHCCIPZapper.sol
@@ -90,7 +90,7 @@ contract WOETHCCIPZapper {
     function zap(address receiver)
         external
         payable
-        returns (bytes32 messageId)
+        returns (bytes32 messageId, uint256 woethAmount)
     {
         return _zap(receiver, msg.value);
     }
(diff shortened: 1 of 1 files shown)


contracts/contracts/zapper/WOETHCCIPZapper.sol
```diff
@@ -190,6 +190,6 @@ contract WOETHCCIPZapper {
         emit Zap(messageId, msg.sender, receiver, amount);
 
         // Return the message ID
-        return messageId;
+        return (messageId, woethReceived);
```
naddison36, line 193: no need to explicitly return. Any return variables assigned in the function body like `messageId` and `woethReceived` will be returned even if the return is not called.

FYI, in older Solidity compilers it used to be less gas to not explicitly return, but that's no longer the case with newer Solidity compilers

apexearth: Should uint256 be removed from line 156?

naddison36: yes it should

PR #2043, opened 2024-05-02
Add AERO/WETH Oracle

PraneshASP:
### Description:
Added a new contract `AeroWethPriceFeed`. This contract helps fetching the price of the Aerodrome token denominated in ETH. 

_**Note**: Since chainlink doesn't support AERO/ETH price feed yet, we fetch the AERO/USD price and ETH/USD price then divide both to get the `AERO/ETH` rate._ 

---

Contract change checklist:
  - [ ] Unit tests pass
  
#### Notes:
This is required for the Vault and harvester contracts in Base to support the Aerodrome AMO.


contracts/contracts/oracle/AeroWethPriceFeed.sol
```diff
@@ -0,0 +1,76 @@
...
+
+            ) = wethUsdPriceFeed.latestRoundData();
+
+            require(
+                ethPriceUpdatedAt + 1 hours >= block.timestamp, // staleness check
+                "ETH_STALE_PRICE"
+            );
+            // There is no AERO/ETH feed.
+            // Get AERO/USD price and divide it with ETH/USD price to fetch the AERO/ETH price.
+            uint256 _price = uint256(_aeroUsdPrice).divPrecisely(
```
shahthepro, line 67: The contract seems to be using different staleness tolerance for both the pairs. It might cause some issues especially when the market is volatile.

Shouldn't be much of an issue though since heartbeat for AERO/USD on Base is at 24h anyway

shahthepro: I'd say the current numbers are okay

contracts/contracts/oracle/AeroWethPriceFeed.sol
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    /**
+     * @notice Returns the total price in 18 digit units for the given asset.
+     *         This implementation does not (!) do range checks as the
+     *         parent OracleRouter does.
+     * @param asset address of the asset
+     * @return uint256 unit price for 1 asset unit, in 18 decimal fixed
+     */
+    function price(address asset) external view returns (uint256) {
```
sparrowDom, line 34: A price feed usually communicates a price rate between 2 different assets. There is no notion of a base asset (e.g. ETH or USD). Similar to how it is done [here](https://github.com/OriginProtocol/origin-dollar/blob/master/contracts/contracts/oracle/AuraWETHPriceFeed.sol#L45).

For that reason the price shouldn't be accepting the asset parameter

sparrowDom (changes requested):
I would change a couple of things: 
- We use [OracleRouters](https://github.com/OriginProtocol/origin-dollar/blob/master/contracts/contracts/oracle/OETHOracleRouter.sol#L59) as a collection of price feeds. OETH and OUSD oracle routers also make sure that the prices are denominated in the correct underlying units. This price feed should be available there. If a new OracleRouter is required for Base, a new one should be deployed.
- rather than calling the price feed class `AeroWethPriceFeed` I'd call it `PriceFeedPair` since what it is effectively doing is combining 2 price feeds into 1. The  USD/WETH & WETH/AERO price feeds become a USD/AERO price feed. And this could work with any 2 chainlink price feeds (that share a token as in the above example obviously). This way we have a general contract price feed, that we can reuse in the future.

PraneshASP:
Thanks for the suggestions. If I understand correctly, we need to make a couple of changes:
- Remove the asset parameter since the pricefeed always returns the price for a given asset.
- Create a new oracle router for base, which should return the prices for WETH (1:1) and AERO. We can add more assets to the router if required. 

Currently both the above actions are coupled in the same contract. 

I'll refactor and split them into two contracts: 
- `AeroWethPriceFeed`
- `BaseOracleRouter`

let me know if that makes sense. @sparrowDom

sparrowDom:
Yes if you check the codebase or the API of all the chainlink price feeds we are using, you'll see that they communicate a price rate between 2 assets. See [RETH/ETH](https://data.chain.link/feeds/ethereum/mainnet/reth-eth), [AAVE/ETH](https://data.chain.link/feeds/ethereum/mainnet/aave-eth) & [AAVE/USD](https://data.chain.link/feeds/ethereum/mainnet/aave-usd). There is no notion of `base asset`, meaning if you return only the price for AAVE what is it priced in? Usd, ETH, Eur? For that reason the price feed returns the price rate between the 2 assets.  Actually I remembered we've already implemented this in another PR. You can probably reuse [all the code from it](https://github.com/OriginProtocol/origin-dollar/pull/1917/files/230d15e43577f434ce2965e2576b4dc0c9ed7f4c#diff-997d733d4c49ff4d71b3cb7e33e33f55681bb6ba3029ab46905283cd646b56aeR2).

Regarding oracle router. Yes just check out how things are setup currently for OUSD & OETH and follow that pattern. Feel free to ask questions if somethings don't make sense to you.

sparrowDom (commented):
Left a couple of comments PTAL

contracts/contracts/oracle/BaseOETHOracleRouter.sol
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    /**
+     * @notice Returns the total price in 18 digit units for a given asset.
+     *         This implementation does not (!) do range checks as the
+     *         parent OracleRouter does.
+     * @param asset address of the asset
+     * @return uint256 unit price for 1 asset unit, in 18 decimal fixed
+     */
+    function price(address asset)
```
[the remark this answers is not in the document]
PraneshASP: If we inherit from OETHOracleRouter, we might wanna initialize the parent contract by passing the Aura price feed address. Since it is designed to work with Base, it made more sense to inherit from the `OracleRouterBase` rather than `OETHOracleRouter.sol`. let me know wdyt.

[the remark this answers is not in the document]
sparrowDom: Ah yes you are right I haven't noticed that. What I'd do is create then another abstract contract called `OracleRouterNoPriceRange` that inherits from `OracleRouterBase` and only implements the `price` the way it is implemented in OETHOracleRouter. 

Then alter the code so that `OETHOracleRouter` and `BaseOETHOracleRouter` both inherit from that contract and not re-implement the `price` function.

PraneshASP:
Merging this with the AMO branch. Will address the comments in the AMO PR

PR #2051, opened 2024-05-08
Sparrow dom/native staking defender action

sparrowDom:
- deploy script for updating the registrator on the native staking contract to defender relayer
- fixes to operate validators script


naddison36 (approved):
LGTM

PR #2057, opened 2024-05-12
add rate limit to calling manually fix accounting

sparrowDom:
add rate limit to calling manually fix accounting

### DIFF
--- a/contracts/contracts/strategies/NativeStaking/ValidatorAccountant.sol
+++ b/contracts/contracts/strategies/NativeStaking/ValidatorAccountant.sol
@@ -10,6 +10,8 @@ import { IWETH9 } from "../../interfaces/IWETH9.sol";
 /// Full withdrawals are from exited validators.
 /// @author Origin Protocol Inc
 abstract contract ValidatorAccountant is ValidatorRegistrator {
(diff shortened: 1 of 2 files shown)


naddison36 (approved):
Happy to add this but its worth noting that `manuallyFixAccounting` can only be run when the fuse has blown. That is, the accounting has been paused. `manuallyFixAccounting` will unpause if successful so can not be run against until the fuse is next blown.

sparrowDom:
The workaround to continually trigger this would be:
- create a contract and send 23 ETH to it
- call [selfDestruct](https://solidity-by-example.org/hacks/self-destruct/) on the contract and set the recipient address to native staking contract
- the `payable receive` on native staking contract [is not triggered](https://ethereum.stackexchange.com/questions/124571/how-do-i-activate-selfdestruct-once-my-smart-contract-is-deployed)
- attacker calls `doAccounting`, fuse is blown
- `manuallyFixAccounting` is called (given the strategist is compromised). Contract balance can be inflated by 332ETH

Above steps can be repeated continually if there is no rate limit on `manuallyFixAccounting` function and strategist account has been compromised.

PR #2059, opened 2024-05-13
Add withdrawal event

sparrowDom:


naddison36 (approved):
This is better than changing the contract inheritance. Thanks

PR #2060, opened 2024-05-13
Fix event values

sparrowDom:
- adds tests to detect the new Withdrawal events
- allows for the `manuallyFixAccounting` function to wrap ETH into WETH and send it to the vault to correct a possible state where the accounting function needs to classify a blown fuse as a full (slashed) withdrawal of the validator and send its assets as WETH to the Vault
- add better comments
- improve accounting for depositAll to communicate correct values


contracts/contracts/strategies/NativeStaking/NativeStakingSSVStrategy.sol
```diff
@@ -33,8 +54,20 @@ contract NativeStakingSSVStrategy is
...
+    /// deposit it to an underlying platform. Rather a special privilege account stakes it to the validators.
+    /// For that reason calling WETH.balanceOf(this) in a deposit function can contain WETH that has just been
+    /// deposited and also WETH that has previously been deposited. To keep a correct count we need to keep track
+    /// of WETH that has already been accounted for.
+    /// This value represents the amount of WETH balance of this contract that has already been accounted for by the
+    /// deposit events.
+    /// It is important to note that this variable is not concerned with WETH that is a result of full/partial
+    /// withdrawal of the validators. It is strictly concerned with WETH that has been deposited and is waiting to
+    /// be staked.
+    uint256 depositedWethAccountedFor;
```
naddison36, line 67: nit storage variables should explicitly be public

PR #2062, opened 2024-05-16
OETH withdrawal queue

naddison36:
## Contract Changes

* Added a withdrawal queue to the OETH Vault which includes `requestWithdrawal`, `claimWithdrawal` and `addWithdrawalQueueLiquidity`.
* Any function that transfers WETH to or from the vault modified so the withdrawal queue is prioritised. This includes `allocate`, `mint`, `redeem`, `depositToStrategy`, `withdrawFromStrategy`, `withdrawAllFromStrategy`, `withdrawAllFromStrategies` and `swapCollateral`.
* `redeem` will only work if there is enough WETH liquidity in the Vault to satisfy all outstanding withdrawal requests.
* There needs to be a minimum of 10 minute between request and claim. Once the liquidity is available in the vault after the 10 minutes, the user can claim.
* OETH Vault's `allocate` changed to use the OETH total supply rather than calculating the vault's total assets when calculating the vault buffer.
* To prevent yield theft, mints over the `rebaseThreshold` (1 OETH) will call the `Dripper` to collect any streamed rewards to that point. The minter's OETH is calculated after the rewards are collected. This prevents someone minting with a large amount and then withdrawing 10 minutes later collecting most of the yield from the Dripper.

## Dependencies

* https://github.com/OriginProtocol/origin-dollar/pull/2015

## Interfaces

```Solidity
    struct WithdrawalQueueMetadata {
        // cumulative total of all withdrawal requests included the ones that have already been claimed
        uint128 queued;
        // cumulative total of all the requests that can be claimed including the ones that have already been claimed
        uint128 claimable;
        // total of all the requests that have been claimed
        uint128 claimed;
        // index of the next withdrawal request starting at 0
        uint128 nextWithdrawalIndex;
    }

    struct WithdrawalRequest {
        address withdrawer;
        bool claimed;
        // Amount of oTokens to redeem
        uint128 amount;
        // cumulative total of all withdrawal requests including this one.
        // this request can be claimed when this queued amount is less than or equal to the queue's claimable amount.
        uint128 queued;
    }

    event WithdrawalRequested(
        address indexed _withdrawer,
        uint256 indexed _requestId,
        uint256 _amount,
        uint256 _queued
    );
    event WithdrawalClaimed(
        address indexed _withdrawer,
        uint256 indexed _requestId,
        uint256 _amount
    );
    event WithdrawalClaimable(uint256 _claimable, uint256 _newClaimable);

    /**
     * @notice Request an asynchronous withdrawal of WETH in exchange for OETH.
     * The OETH is burned on request and the WETH is transferred to the withdrawer on claim.
     * This request can be claimed once the withdrawal queue's `claimable` amount
     * is greater than or equal this request's `queued` amount.
     * There is no minimum time or block number before a request can be claimed. It just needs
     * enough WETH liquidity in the Vault to satisfy all the outstanding requests to that point in the queue.
     * OETH is converted to WETH at 1:1.
     * @param _amount Amount of OETH to burn.
     * @param requestId Unique ID for the withdrawal request
     * @param queued Cumulative total of all WETH queued including already claimed requests.
     */
    function requestWithdrawal(uint256 _amount)
        external
        returns (uint256 requestId, uint256 queued);

    /**
     * @notice Claim a previously requested withdrawal once it is claimable.
     * This request can be claimed once the withdrawal queue's `claimable` amount
     * is greater than or equal this request's `queued` amount and 30 minutes has passed.
     * If the requests is not claimable, the transaction will revert with `Queue pending liquidity`.
     * If the request is not older than 30 minutes, the transaction will revert with `Claim delay not met`.
     * OETH is converted to WETH at 1:1.
     * @param _requestId Unique ID for the withdrawal request
     * @return amount Amount of WETH transferred to the withdrawer
     */
    function claimWithdrawal(uint256 requestId)
        external
        returns (uint256 amount);

    /**
     * @notice Claim a previously requested withdrawals once they are claimable.
     * This requests can be claimed once the withdrawal queue's `claimable` amount
     * is greater than or equal each request's `queued` amount and 30 minutes has passed.
     * If one of the requests is not claimable, the whole transaction will revert with `Queue pending liquidity`.
     * If one of the requests is not older than 30 minutes,
     * the whole transaction will revert with `Claim delay not met`.
     * @param _requestIds Unique ID of each withdrawal request
     * @return amounts Amount of WETH received for each request
     * @return totalAmount Total amount of WETH transferred to the withdrawer
     */
    function claimWithdrawals(uint256[] memory requestIds)
        external
        returns (uint256[] memory amounts, uint256 totalAmount);

    /// @notice Global metadata for the withdrawal queue including:
    /// queued - cumulative total of all withdrawal requests included the ones that have already been claimed
    /// claimable - cumulative total of all the requests that can be claimed including the ones already claimed
    /// claimed - total of all the requests that have been claimed
    /// nextWithdrawalIndex - index of the next withdrawal request starting at 0
    function withdrawalQueueMetadata()
        external
        view
        returns (VaultStorage.WithdrawalQueueMetadata memory);

    /// @notice Mapping of withdrawal request indices to the user withdrawal request data
    function withdrawalRequests(uint256 requestId)
        external
        view
        returns (VaultStorage.WithdrawalRequest memory);

    /// @notice Collects harvested rewards from the Dripper as WETH then
    /// adds WETH to the withdrawal queue if there is a funding shortfall.
    /// @dev is called from the Native Staking strategy when validator withdrawals are processed.
    /// It also called before any WETH is allocated to a strategy.
    function addWithdrawalQueueLiquidity() external;
```

## Test Chains

The OETH Vault on Holesky has been upgraded to include the withdrawal queue [0x19d2bAaBA949eFfa163bFB9efB53ed8701aA5dD9](https://holesky.etherscan.io/address/0x19d2bAaBA949eFfa163bFB9efB53ed8701aA5dD9)

Tenderly testnet [OETH ARM](https://dashboard.tenderly.co/origin-protocol/origin/testnet/614d920e-92bd-46bd-9f56-e933d3453f6f) is an older fork of mainnet.

Tenderly testnet [OETH ARM 2](https://dashboard.tenderly.co/origin-protocol/origin/testnet/040f6eee-991c-4c93-98b5-dcd68607bec1) is a newer fork of mainnet with the last Vault code.

# Processes

## requestWithdrawal

62488c55

## claimWithdrawal

d11c27ca

# Deployment

## Holesky

The Holesky deploy script to upgrade the OETHVault is `contracts/deploy/holesky/017_upgrade_vault.js`

```
# uncomment DEPLOYER_PK and GOVERNOR_PK in the .env file for the Holesky deployer PK

yarn run deploy:holesky

echo "module.exports = [
\"0x94373a4919B3240D86eA41593D5eBa789FEF3848\", // WETH
]" > vault-args.js

npx hardhat --network holesky verify --contract contracts/vault/OETHVaultCore.sol:OETHVaultCore --constructor-args vault-args.js 0xB716b6a92B77cDF2c5d0fd46eba05838E30f77c1

npx hardhat --network holesky verify --contract contracts/vault/OETHVaultAdmin.sol:OETHVaultAdmin --constructor-args vault-args.js 0xdF96E53406F6E6873B20E996d7fcB8e1f61ecc5c
```

## Mainnet

The deploy script for the OETH Withdrawal Queue is `contracts/deploy/mainnet/103_oeth_withdraw_queue.js`.

# Testing

## Unit Tests

The OETH Vault unit tests are in `contracts/test/vault/oeth-vault.js`.

```
yarn run test
```

## Fork Tests

The mainnet OETH Vault fork tests are in `contracts/test/vault/oeth-vault.mainnet.fork-test.js`

```
# In one terminal
yarn run node

# In another terminal
yarn run test:fork
```

## Holesky using Hardhat

```
npx hardhat snapVault --network holesky
npx hardhat rebase --network holesky
npx hardhat snapStaking --index 2 --network holesky
npx hardhat queueLiquidity --network holesky

npx hardhat depositWETH --amount 64 --network holesky
npx hardhat approve --symbol WETH --amount 10000 --spender 0x39254033945AA2E4809Cc2977E7087BEE48bd7Ab --network holesky
npx hardhat mint --asset WETH --amount 64 --network holesky
npx hardhat snapVault --network holesky

npx hardhat requestWithdrawal --amount 30 --network holesky
npx hardhat claimWithdrawal --request-id 1 --network holesky

npx hardhat balance --symbol WETH --network holesky
npx hardhat transfer --symbol WETH --amount 4 --to 0x19d2bAaBA949eFfa163bFB9efB53ed8701aA5dD9 --network holesky
```

## Security

Contract change checklist:
  - [ ] Unit tests pass


contracts/deploy/mainnet/097_withdraw_queue.js
```diff
@@ -0,0 +1,64 @@
...
+          signature: "upgradeTo(address)",
+          args: [dVaultCore.address],
+        },
+        // 2. set OETH Vault proxy to the new admin vault implementation
+        {
+          contract: cVault,
+          signature: "setAdminImpl(address)",
+          args: [dVaultAdmin.address],
+        },
+        // 3. Set the Dripper contract
```
DanielVF, line 55: Did I miss something? This PR shouldn't have to mess with the dripper? And doesn't make any contract changes to the dripper?

contracts/contracts/vault/VaultStorage.sol
```diff
@@ -165,8 +178,41 @@ contract VaultStorage is Initializable, Governable {
...
+        // this request can be claimed when this queued amount is less than or equal to the queue's claimable amount.
+        uint128 queued;
+    }
+
+    // Mapping of withdrawal requests indexes to the user withdrawal request data
+    mapping(uint256 => WithdrawalRequest) public withdrawalRequests;
+
     // For future use
-    uint256[50] private __gap;
+    uint256[46] private __gap;
```
DanielVF, line 215: I see three slots added? `dripper`, `withdrawalQueueMetadata`, and  `withdrawalRequests`. Shouldn't the gap change by 3, not four?

This particular gap is one that keeps me up at night. We should make sure our fork tests will catch an incorrect number here.

contracts/deploy/mainnet/097_withdraw_queue.js
DanielVF: Nevermind!

contracts/contracts/vault/VaultStorage.sol
naddison36: There's three storage variables added but `withdrawalQueueMetadata` is a struct that takes up 2 slots hence the gap was reduced by 4 slots.

contracts/deploy/mainnet/097_withdraw_queue.js
naddison36: This PR includes the change to the Vault to call `collect` on the `Dripper` on mints to prevent yield theft.

contracts/contracts/vault/OETHVaultCore.sol
```diff
@@ -145,4 +156,212 @@ contract OETHVaultCore is VaultCore {
...
+        WithdrawalQueueMetadata memory queue = withdrawalQueueMetadata;
+        WithdrawalRequest memory request = withdrawalRequests[requestId];
+
+        require(request.claimed == false, "already claimed");
+        require(request.withdrawer == msg.sender, "not requester");
+
+        // Try and get more liquidity in the withdrawal queue if there is not enough
+        if (request.queued > queue.claimable) {
+            // Stream any harvested rewards (WETH) that are available to the Vault
+            IDripper(dripper).collect();
```
pandadefi, line 249: This will get called on every loop from line 225.

contracts/contracts/vault/OETHVaultCore.sol
```diff
@@ -145,4 +156,212 @@ contract OETHVaultCore is VaultCore {
...
+    {
+        // Check if there's enough liquidity to cover the withdrawal request
+        WithdrawalQueueMetadata memory queue = withdrawalQueueMetadata;
+        WithdrawalRequest memory request = withdrawalRequests[requestId];
+
+        require(request.claimed == false, "already claimed");
+        require(request.withdrawer == msg.sender, "not requester");
+
+        // Try and get more liquidity in the withdrawal queue if there is not enough
+        if (request.queued > queue.claimable) {
```
pandadefi, line 247: This branch only needs to be executed once on the first iteration of the loop line 225.

contracts/contracts/vault/OETHVaultCore.sol
```diff
@@ -145,4 +156,212 @@ contract OETHVaultCore is VaultCore {
...
+            // Stream any harvested rewards (WETH) that are available to the Vault
+            IDripper(dripper).collect();
+
+            // Add any WETH from the Dripper to the withdrawal queue
+            uint256 addedClaimable = _addWithdrawalQueueLiquidity();
+
+            // If there still isn't enough liquidity in the queue to claim, revert
+            require(
+                request.queued <= queue.claimable + addedClaimable,
```
pandadefi, line 256: Have you considered returning zero and letting the previous withdrawal execute in the case of a call to claimWithdrawals with one not being claimable in the array?

jayjaysecurity (approved):
LGTM!

contracts/contracts/vault/VaultStorage.sol
```diff
@@ -166,8 +179,44 @@ contract VaultStorage is Initializable, Governable {
...
+    /// @dev The vault is proxied so needs to be set with setDripper against the proxy contract.
+    // slither-disable-start constable-states
+    // slither-disable-next-line uninitialized-state
+    address public dripper;
+    // slither-disable-end constable-states
+
+    /// Withdrawal Queue Storage /////
+
+    struct WithdrawalQueueMetadata {
+        // cumulative total of all withdrawal requests included the ones that have already bene claimed
```
DanielVF, line 192: `bene` -> `been`

contracts/contracts/vault/VaultStorage.sol
```diff
@@ -166,8 +179,44 @@ contract VaultStorage is Initializable, Governable {
...
+        uint128 queued;
+        // cumulative total of all the requests that can be claimed included the ones that have already bene claimed
```
DanielVF, line 194: `bene` -> `been`

DanielVF:
The counters are beautiful in this PR. Nice way to solve the fairness issues.

Thinking in invariants:

`queue.queued >= queue.claimable => queue.claimed`

DanielVF:
This does basically add instant 1:1 redeems in certain situations. Fortunately our vault side mint math is now very simple, and vault side burn math is simple too.

The real question then is if there are any possible problems in the OETH token itself that would make mint->burn or burn->mint a profitable cycle.

contracts/contracts/vault/VaultStorage.sol
naddison36: thanks. Fixed

contracts/contracts/vault/VaultStorage.sol
naddison36: thanks. Fixed

contracts/contracts/vault/OETHVaultAdmin.sol
```diff
@@ -1,12 +1,112 @@
...
+        address _strategyToAddress,
+        address[] calldata _assets,
+        uint256[] calldata _amounts
+    ) internal override {
+        require(
+            strategies[_strategyToAddress].isSupported,
+            "Invalid to Strategy"
+        );
+        require(
+            _assets.length == 1 && _amounts.length == 1 && _assets[0] == weth,
```
shahthepro, line 35: Would be great if we can move `_amounts.length` to its own `require` statement so that we can know why it reverts when it does

contracts/contracts/vault/OETHVaultAdmin.sol
```diff
@@ -1,12 +1,112 @@
...
+            "Only WETH is supported"
+        );
+
+        // Check the there is enough WETH to transfer once the WETH reserved for the withdrawal queue is accounted for
+        require(_amounts[0] <= _wethAvailable(), "Not enough WETH available");
```
shahthepro, line 40: nit: Short-circuiting for `_amounts[0] == 0` would be nice

contracts/contracts/vault/OETHVaultCore.sol
```diff
@@ -145,4 +157,230 @@ contract OETHVaultCore is VaultCore {
...
+            amount: uint128(_amount),
+            queued: uint128(queued)
+        });
+    }
+
+    /**
+     * @notice Claim a previously requested withdrawal once it is claimable.
+     * This request can be claimed once the withdrawal queue's `claimable` amount
+     * is greater than or equal this request's `queued` amount.
+     * There is no minimum time or block number before a request can be claimed. It just needs
```
shahthepro, line 205: The comment seems invalid since it doesn't account for `CLAIM_DELAY`

contracts/contracts/vault/OETHVaultCore.sol
```diff
@@ -145,4 +157,230 @@ contract OETHVaultCore is VaultCore {
...
+    function claimWithdrawals(uint256[] memory _requestIds)
+        external
+        whenNotCapitalPaused
+        nonReentrant
+        returns (uint256[] memory amounts, uint256 totalAmount)
+    {
+        amounts = new uint256[](_requestIds.length);
+        for (uint256 i = 0; i < _requestIds.length; ++i) {
+            amounts[i] = _claimWithdrawal(_requestIds[i]);
+            totalAmount += amounts[i];
```
shahthepro, line 241: nit: `withdrawalQueueMetadata` is updated multiple times when claiming several withdrawal requests, Could probably save some gas by having a separate method for multiple claims instead of just  looping through the same method several times

contracts/contracts/vault/OETHVaultAdmin.sol
naddison36: _wethAvailable() will be > 0 for the deposit to work. It's basically all the WETH that is not reserved for the withdrawal queue.

shahthepro: Not about `_wethAvailable` but I was just wondering if `_amounts[0]` will ever be called with `0`. We have faced this issue while removing a few strategies (where we call withdrawFromAllStrategies and it failed because we didn't have a zero-amount check and we had to resort to workarounds)

naddison36: I see, you want to revert the `depositToStrategy` if the asset amount is zero. It's not something we have previously done.

The issue with `withdrawFromAllStrategies` was some strategies didn't support withdrawing a zero amount. We had to add a tiny amount so we could do a withdrawal. Checking `depositToStrategy` amounts are not zero would not have prevented this.

contracts/contracts/vault/OETHVaultCore.sol
```diff
@@ -145,4 +157,230 @@ contract OETHVaultCore is VaultCore {
...
+
+            // If there still isn't enough liquidity in the queue to claim, revert
+            require(
+                request.queued <= queue.claimable + addedClaimable,
+                "Queue pending liquidity"
+            );
+        }
+
+        // Store the updated claimed amount
+        withdrawalQueueMetadata.claimed = queue.claimed + request.amount;
```
shahthepro, line 281: nit: Any reason for not doing a `queue.claimable -= request.amount` here? I feel like that'll simplify the calculations in the other places like `wethAvailable`

contracts/contracts/vault/OETHVaultAdmin.sol
shahthepro: Not revert, but I meant to say just a `return` for zero deposits. 

> Checking depositToStrategy amounts are not zero would not have prevented this.

That's true

contracts/contracts/vault/OETHVaultCore.sol
naddison36: you mean `queue.claimable += request.amount`?

`queue` is an in memory variable. We need to update the storage variable `withdrawalQueueMetadata`.
`withdrawalQueueMetadata.claimed += request.amount` would work but that would do an extra read from storage. So I've done with adding the two amounts using in memory variables and then storing back to storage.

shahthepro: Nope, I meant that we can subtract claimed amount from `claimable` (I see that we do a `queue.claimable - queue.claimed` in a lot of places that could be avoided. Besides `claimable` still including claimed amount seems a bit off since I am thinking `claimable` == `unclaimed`)

naddison36: `queue.claimable` is important as it drives when a request can be claimed. It's a global variable that goes up whenever liquidity is reserved for the queue. A request can be claimed when this invariant is true `request.queued <= queue.claimable`. 

`claimable != unclaimed` as `uint256 unclaimed = queue.claimable - queue.claimed;`
`queue.claimed` is also a global storage variable of all requests that have every been claimed.

The advantage of using the global `queued` and `claimable` is we don't need to iterate over user requests to work out if the request can be claimed or not.

contracts/contracts/vault/OETHVaultCore.sol
```diff
@@ -145,4 +157,230 @@ contract OETHVaultCore is VaultCore {
...
+        WithdrawalQueueMetadata memory queue = withdrawalQueueMetadata;
+
+        // Check if the claimable WETH is less than the queued amount
+        uint256 queueShortfall = queue.queued - queue.claimable;
+        uint256 wethBalance = IERC20(weth).balanceOf(address(this));
+        // Of the claimable withdrawal requests, how much is unclaimed?
+        uint256 unclaimed = queue.claimable - queue.claimed;
+
+        if (wethBalance > queueShortfall + unclaimed) {
+            wethAvailable = wethBalance - queueShortfall - unclaimed;
```
clement-ux, line 352: I think that this can be simplified in: `wethAvailable = wethBalance  + queue.claimed - queue.queued`, this can avoid 2 unnecessary operations.

contracts/contracts/vault/OETHVaultCore.sol
```diff
@@ -145,4 +157,230 @@ contract OETHVaultCore is VaultCore {
...
+    function _wethAvailable() internal view returns (uint256 wethAvailable) {
+        WithdrawalQueueMetadata memory queue = withdrawalQueueMetadata;
+
+        // Check if the claimable WETH is less than the queued amount
+        uint256 queueShortfall = queue.queued - queue.claimable;
+        uint256 wethBalance = IERC20(weth).balanceOf(address(this));
+        // Of the claimable withdrawal requests, how much is unclaimed?
+        uint256 unclaimed = queue.claimable - queue.claimed;
+
+        if (wethBalance > queueShortfall + unclaimed) {
```
clement-ux, line 351: Maybe, this can be simplified in: `wethBalance  > queue.queued - queue.claimed`, this can avoid 2 unnecessary operations.

contracts/contracts/vault/OETHVaultCore.sol
clement-ux, line 352: Same here with `wethAvailable = wethBalance  + queue.claimed - queue.queued`.

contracts/contracts/vault/OETHVaultCore.sol
```diff
@@ -145,4 +157,230 @@ contract OETHVaultCore is VaultCore {
...
+     * @return totalAmount Total amount of WETH transferred to the withdrawer
+     */
+    function claimWithdrawals(uint256[] memory _requestIds)
+        external
+        whenNotCapitalPaused
+        nonReentrant
+        returns (uint256[] memory amounts, uint256 totalAmount)
+    {
+        amounts = new uint256[](_requestIds.length);
+        for (uint256 i = 0; i < _requestIds.length; ++i) {
```
clement-ux, line 239: Maybe we can save a bit gas by caching len of the array, instead on calling it multiple time. Something like:
 ```solidity
 uint256 len = _requestIds.length.
 amounts = new uint256[](len);
 for (uint256 i = 0; i < len; ++i) {}
```

Comment (shahthepro) [reply contracts/contracts/vault/OETHVaultCore.sol:239]: I thought caching length only saves gas when dealing with global storage. I don't think there'd be any difference, if at all, when accessing the length of `calldata` or `memory` parameters. I might be wrong though

File: contracts/contracts/vault/OETHVaultCore.sol [hunk 16]
```diff
@@ -145,4 +157,230 @@ contract OETHVaultCore is VaultCore {
...
+        view
+        override
+        returns (uint256 balance)
+    {
+        balance = super._checkBalance(_asset);
+
+        if (_asset == weth) {
+            WithdrawalQueueMetadata memory queue = withdrawalQueueMetadata;
+            // Need to remove WETH that is reserved for the withdrawal queue
+            return balance + queue.claimed - queue.queued;
```
shahthepro, line 367: 🔴 This can cause the Vault to break due to underflow issues.

The code assumes that `balance + queue.claimed` will be always be greater than `queue.queued`. However, that's not the case all the time.

Someone can come and do a `requestWithdrawal` for an amount that's larger than whatever WETH Vault holds and essentially break all of our contracts

shahthepro (changes requested):

contracts/contracts/vault/OETHVaultCore.sol
naddison36: Good spot. Fixed. Thanks

contracts/contracts/vault/OETHVaultCore.sol
clement-ux: You're right, I tested it with optimization enabled and it is better to keep it like this!

contracts/contracts/vault/OETHVaultCore.sol
```diff
@@ -345,23 +384,79 @@ contract OETHVaultCore is VaultCore {
...
         _addWithdrawalQueueLiquidity();
 
-        super._allocate();
+        _allocate();
     }
 
+    /// @dev Allocate WETH to the default WETH strategy if there is excess to the Vault buffer.
+    /// This is called from either `mint` or `allocate` and assumes `_addWithdrawalQueueLiquidity`
+    /// has been called before this function.
+    function _allocate() internal override {
```
DanielVF, line 406: Gotta say I love how linear this function is. ❤️

contracts/contracts/vault/OETHVaultCore.sol
```diff
@@ -345,23 +384,79 @@ contract OETHVaultCore is VaultCore {
...
+        if (value + queue.claimed >= queue.queued) {
+            return value + queue.claimed - queue.queued;
+        }
+    }
+
+    /// @dev Only WETH is supported in the OETH Vault so return the WETH balance only
+    /// Any ETH balances in the Vault will be ignored.
+    /// Amounts from previously supported vault assets will also be ignored.
+    /// For example, there is 1 wei left of stETH in the OETH Vault but is will be ignored.
+    function _totalValueInVault()
```
DanielVF, line 454: Great perf optimization.

contracts/contracts/vault/OETHVaultCore.sol
```diff
@@ -345,23 +384,79 @@ contract OETHVaultCore is VaultCore {
...
+    // withdrawal requests then return a total value of 0.
     function _totalValue() internal view override returns (uint256 value) {
-        value = super._totalValue();
+        value = _totalValueInVault() + _totalValueInStrategies();
 
         // Need to remove WETH that is reserved for the withdrawal queue.
         // reserved for the withdrawal queue = cumulative queued total - total claimed
         WithdrawalQueueMetadata memory queue = withdrawalQueueMetadata;
-        value = value + queue.claimed - queue.queued;
+        if (value + queue.claimed >= queue.queued) {
```
DanielVF, line 445: 🟥 This appears to be extremely wrong.

As `queued` increases, the total returned goes down. However when this if statement flips, the total instantly jumps up to the balance, ignoring `claimed` and `queued`

image

Example:

> Vault controls 100 million WETH.
>
> Someone requests a 50 million withdraw. Vault now reports 50 million total value. Correct.
> Someone requests a 49 million withdraw next. Vault now reports 1 million total value. Correct.
> Someone requests a 1 million + 1 raw withdraw next. Vault now reports 100 million total value instead of zero. Not good.

This could result in an insane rebase, considering at this point there would be a low totalSupply.


This also applies to `_checkBalance`, which uses a similar if statement.

contracts/contracts/vault/VaultCore.sol
```diff
@@ -224,7 +224,7 @@ contract VaultCore is VaultInitializer {
         }
 
         // Check that the OTokens are backed by enough assets
-        if (maxSupplyDiff > 0) {
+        if (maxSupplyDiff > 0 && totalUnits > 0) {
```
DanielVF, line 227: 🟥 This skips the solvency check if an attacker has managed to steal all the money. That's the wrong time to skip the solvency check.

Now in practice, we of course don't think that an attacker could steal all the money, but still, this function should catch it if they did.

contracts/contracts/vault/OETHVaultCore.sol
```diff
@@ -143,6 +155,308 @@ contract OETHVaultCore is VaultCore {
...
+    {
+        // Burn the user's OETH
+        // This also checks the requester has enough OETH to burn
+        oUSD.burn(msg.sender, _amount);
+
+        requestId = withdrawalQueueMetadata.nextWithdrawalIndex;
+        queued = withdrawalQueueMetadata.queued + _amount;
+
+        // store the next withdrawal request
+        withdrawalQueueMetadata.nextWithdrawalIndex = uint128(requestId + 1);
```
DanielVF, line 188: These casts will silently clip values. While I think uint128's should be safe for these values, I think we should default to using some kind of safe cast that reverts on overflows.

contracts/contracts/vault/OETHVaultCore.sol
```diff
@@ -125,7 +136,8 @@ contract OETHVaultCore is VaultCore {
             "Redeem amount lower than minimum"
         );
 
-        if (IERC20(weth).balanceOf(address(this)) >= amountMinusFee) {
+        // If there is any WETH in the Vault available after accounting for the withdrawal queue
+        if (_wethAvailable() >= amountMinusFee) {
```
DanielVF, line 140: 🟥 This appears to bypass our check of only giving out available (unreserved) funds.

If we don't have funds available, the redeem should fail, not collect funds to give the user from another source.

contracts/contracts/vault/OETHVaultCore.sol
DanielVF: I love reverts rather than silent ignores when something weird happens.

contracts/contracts/vault/OETHVaultCore.sol
```diff
@@ -143,6 +155,308 @@ contract OETHVaultCore is VaultCore {
...
+            "Claim delay not met"
+        );
+
+        // Try and get more liquidity in the withdrawal queue if there is not enough
+        if (request.queued > queue.claimable) {
+            // Stream any harvested rewards (WETH) that are available to the Vault
+            IDripper(dripper).collect();
+
+            // Add any WETH from the Dripper to the withdrawal queue
+            uint256 addedClaimable = _addWithdrawalQueueLiquidity();
```
DanielVF, line 283: 🟧 Strong personal preference here against this code.

We have a dangerous pattern here in this method, where we check that the request is claimed, then make an external state changing call, and then set that the request has been claimed.

This is a classic setup for a reentrancy vulnerability. If the external call would allow an attacker to call back in, we could claim over and over again. This is why we had to add slither exception to this method. (Yes, they would have to setup the accounting just right to make this work) (yes, dripper should be safe)

1) As a general rule of thumb, whenever we check some kind of one-time only gate value, like `request.claimed`, we should immediately set it right in the next line of code. In this way it's not vulnerable to reentrancy issues, or missing getting set because some if statement returned somewhere.

For example, from primeETH, this is a good way to handle this:

image

2) It's awkward having an external call and global accounting update right in the middle of all this internal request queue updating. Also makes the math ugly. Can we do this after loading the structs from storage, before we do the require checks and then updates on the request? If it needs to run, we can reload the queue struct after. 

3) I have a small preference for not having a multi `claimWithdrawals`. It's one more code path that could go wrong. I'm open though if it's useful.

contracts/contracts/vault/OETHVaultCore.sol
```diff
@@ -143,6 +155,308 @@ contract OETHVaultCore is VaultCore {
...
+
+    /***************************************
+                View Functions
+    ****************************************/
+
+    function _wethAvailable() internal view returns (uint256 wethAvailable) {
+        WithdrawalQueueMetadata memory queue = withdrawalQueueMetadata;
+
+        // Check if the claimable WETH is less than the queued amount
+        uint256 queueShortfall = queue.queued - queue.claimable;
```
DanielVF, line 358: 🟠 As clement mentioned, we use the equivalent of  `(queue.queued - queue.claimable) + (queue.claimable - queue.claimed)` twice. Instead of two variables, these can be combined into one set with `queue.queued - queue.claimed`

contracts/contracts/vault/OETHVaultAdmin.sol
DanielVF: I'm good with the current line. If someone is only using WETH, everything on this line will be true.

contracts/contracts/vault/OETHVaultCore.sol
naddison36: it went through a number of iterations. Each time getting simpler

contracts/contracts/vault/OETHVaultCore.sol
naddison36: Good pickup and an embarrassing bug by me.

This bug can triggered with this test scenario

1. mint 100 WETH
WETH in vault 100
queued 0
claimed 0
vault + claimed < queue (100 + 0 <= 0) is false
Total value = 100

2. Request withdraw for 90
WETH in vault 100
queue 90
claimed 0
vault + claimed <= queue (100 + 0 <= 90) is false
total value 100 when it should be 10

naddison36: I've pushed a fix that handles the scenario where the vault becomes insolvent enough that the total value in the vault and all strategies is less than the outstanding withdrawal requests.

For example, there was a mass slashing event and most users request to withdraw.

contracts/contracts/vault/OETHVaultCore.sol
naddison36: at a rate of one billion requests per second, it would still take 1.08 × 10^22 years to process 2^128  requests.

I've added a `SafeCast.toUint128` so we don't have to deal with this in bug bounties

contracts/contracts/vault/VaultCore.sol
naddison36: It also skips the solvency check if `totalUnits` is zero because the vault value is less than the outstanding withdrawal requests. This can happen when there is a mass slashing and most users have requested a withdrawal.

I need to consider how to handle this.

naddison36: I've changed `_postRedeem` in `VaultCore` to

```Solidity
// Check that the OTokens are backed by enough assets
if (maxSupplyDiff > 0) {
    // If there are more outstanding withdrawal requests than assets in the vault and strategies
    // then the available assets will be negative and totalUnits will be rounded up to zero.
    // As we don't know the exact shortfall amount, we will reject all redeem and withdrawals
    require(totalUnits > 0, "Too many outstanding requests");


```

contracts/contracts/vault/OETHVaultCore.sol
naddison36: I think we simplify redeem and don't try and pull funds from the default strategy. If there's not enough available funds in the vault after accounting for the withdrawal queue, we will revert with `Liquidity error`.
If the user really wants to, they can request a withdrawal.

contracts/contracts/vault/OETHVaultCore.sol
naddison36: this is an easy one. Not sure why I didn't pick it up when Clement mentioned it. Changed

DanielVF: @naddison36 I think we have this function code twice in the code? Another one in OETHVaultAdmin. Need to update both of them?

contracts/contracts/vault/OETHVaultCore.sol
naddison36: 1. This makes sense. I've marked the request at claimed immediately after it's checked. I also changed the order of the requires as `Claim delay not met` is the more likely revert reason.

naddison36: 2. I've done a number of interactions refactoring the claim functions. 

`claimWithdrawal` will only call the Dripper and do the global accounting if there is not enough liquidity to claim. These are now done before calling `_claimWithdrawal`

`_claimWithdrawals` doesn't try to work out what the last request is in the parameter list. It calls the Dripper and does the global accounting regardless.

`_claimWithdrawal` is an internal function with no external calls

naddison36: 3. The frontend is using  `claimWithdrawal` and `claimWithdrawals` so I'll leave them

contracts/contracts/vault/OETHVaultCore.sol
```diff
@@ -125,24 +137,338 @@ contract OETHVaultCore is VaultCore {
...
+    // withdrawal requests then return a total value of 0.
+    function _totalValue() internal view override returns (uint256 value) {
+        value = _totalValueInVault() + _totalValueInStrategies();
+
+        // Need to remove WETH that is reserved for the withdrawal queue.
+        WithdrawalQueueMetadata memory queue = withdrawalQueueMetadata;
+        // reserved for the withdrawal queue = cumulative queued total - total claimed
+        uint256 reservedForQueue = queue.queued - queue.claimed;
+
+        if (reservedForQueue > 0) {
```
DanielVF, line 451: If `reservedForQueue` is zero, then the rest of everything here just works fine, with no need to special case it? Can we lose this outer level if statement?

Secondly, I like having the exceptions checks return out early, and then have the success path flow through the function to the final return. This feels a little awkward here having the success return ahead of the special case 0 return.

contracts/contracts/vault/OETHVaultCore.sol
DanielVF: I like these changes.

contracts/contracts/vault/OETHVaultCore.sol
DanielVF: I'm thinking our casts in general, not just this one line. Basically unless we have a serious reason for not doing a safe cast, we should use a safe cast. Gas overhead is minimal, though I'm sad it makes the code uglier. Maybe a `using` would help?

contracts/contracts/vault/OETHVaultCore.sol
DanielVF: I'm thinking our casts in general, not just this one line. Basically unless we have a serious reason for not doing a safe cast, we should use a safe cast. Gas overhead is minimal, though I'm sad it makes the code uglier. Maybe a `using` would help with the readability?

contracts/contracts/vault/OETHVaultCore.sol
```diff
@@ -125,24 +137,338 @@ contract OETHVaultCore is VaultCore {
...
+     */
+    function requestWithdrawal(uint256 _amount)
+        external
+        whenNotCapitalPaused
+        nonReentrant
+        returns (uint256 requestId, uint256 queued)
+    {
+        // Burn the user's OETH
+        // This also checks the requester has enough OETH to burn
+        oUSD.burn(msg.sender, _amount);
```
DanielVF, line 173: Let's move this burn down to just before the post redeem. Won't actually change anything, but it's better practice since:

1. Let's us do internal calcs, storage updates first, then external calls, then checks, which is a great ordering.
2. It's consistent with the redeem method
2. Even though there are no reentrancy issues here, I like to do updates in a safe order. For OUSD/OETH, by lowering the vault value first, before reducing total supply, there's never a time when the state makes the vault look like it has money to hand out. If doing this made the code look ugly, we could skip, but it makes the code look better I think. We can leave a comment at the start of the method about the user's balance being checked by the later burn.

contracts/contracts/vault/OETHVaultCore.sol
```diff
@@ -125,24 +137,338 @@ contract OETHVaultCore is VaultCore {
...
+        emit WithdrawalClaimed(msg.sender, requestId, request.amount);
+
+        return request.amount;
+    }
+
+    /// @notice Collects harvested rewards from the Dripper as WETH then
+    /// adds WETH to the withdrawal queue if there is a funding shortfall.
+    /// @dev is called from the Native Staking strategy when validator withdrawals are processed.
+    /// It also called before any WETH is allocated to a strategy.
+    function addWithdrawalQueueLiquidity() external {
```
DanielVF, line 309: Let's add `nonReentrant` to this one, since it's updating state.

contracts/contracts/vault/OETHVaultCore.sol
```diff
@@ -125,24 +137,338 @@ contract OETHVaultCore is VaultCore {
...
+        internal
+        returns (uint256 addedClaimable)
+    {
+        WithdrawalQueueMetadata memory queue = withdrawalQueueMetadata;
+
+        // Check if the claimable WETH is less than the queued amount
+        uint256 queueShortfall = queue.queued - queue.claimable;
+
+        // No need to get WETH balance if all the withdrawal requests are claimable
+        if (queueShortfall > 0) {
```
DanielVF, line 328: ⚪ Can we change this to short-circuit if?

contracts/contracts/vault/OETHVaultCore.sol
```diff
@@ -125,24 +137,338 @@ contract OETHVaultCore is VaultCore {
...
+            uint256 wethBalance = IERC20(weth).balanceOf(address(this));
+
+            // Of the claimable withdrawal requests, how much is unclaimed?
+            uint256 unclaimed = queue.claimable - queue.claimed;
+            if (wethBalance > unclaimed) {
```
DanielVF, line 333: ⚪ Bet you know what I'm about want. :D Can we change this to short-circuit if?

contracts/contracts/vault/OETHVaultCore.sol
```diff
@@ -125,24 +137,338 @@ contract OETHVaultCore is VaultCore {
...
+
+        // Check if the claimable WETH is less than the queued amount
+        uint256 queueShortfall = queue.queued - queue.claimable;
+
+        // No need to get WETH balance if all the withdrawal requests are claimable
+        if (queueShortfall > 0) {
+            uint256 wethBalance = IERC20(weth).balanceOf(address(this));
+
+            // Of the claimable withdrawal requests, how much is unclaimed?
+            uint256 unclaimed = queue.claimable - queue.claimed;
```
DanielVF, line 332: ⚪ `unclaimed` might not be the right word for this variable. We'd use different math for that. I think it's a variable about reserved/allocated weth instead?

contracts/contracts/vault/OETHVaultCore.sol
naddison36: The changes have been pushed

contracts/contracts/vault/OETHVaultCore.sol
naddison36: I've now updated both `_wethAvailable` functions on `OETHVaultCore` and `OETHVaultAdmin`.

contracts/contracts/vault/OETHVaultCore.sol
naddison36: If `reservedForQueue` is zero then we want to return the value from `_totalValueInVault() + _totalValueInStrategies()`. If we do both your changes then the code is simpler and easier to reason.

```Solidity
function _totalValue() internal view override returns (uint256 value) {
    value = _totalValueInVault() + _totalValueInStrategies();

    // Need to remove WETH that is reserved for the withdrawal queue.
    WithdrawalQueueMetadata memory queue = withdrawalQueueMetadata;
    // reserved for the withdrawal queue = cumulative queued total - total claimed
    uint256 reservedForQueue = queue.queued - queue.claimed;

    if (value < reservedForQueue) {
        // This can happen if the vault becomes insolvent enough that the
        // total value in the vault and all strategies < outstanding withdrawals.
        // For example, there was a mass slashing event and most users request a withdrawal.
        return 0;
    }

    // Adjust the total value by the amount reserved for the withdrawal queue
    return value - reservedForQueue;
}
```

contracts/contracts/vault/OETHVaultCore.sol
naddison36: I've moved the burn down to before the post redeem.

I also moved the emit of the `WithdrawalRequested` event to the bottom. Most of the time we emit events at the end but mint and redeem does them first for some reason.

contracts/contracts/vault/OETHVaultCore.sol
naddison36: Done

contracts/contracts/vault/OETHVaultCore.sol
naddison36: I've changed `unclaimed ` to `allocatedWeth` as later we use `unallocatedWeth`

contracts/contracts/vault/OETHVaultCore.sol
naddison36: Hmm, this is not as simple as just adding `nonReentrant` to `OETHVaultCore.addWithdrawalQueueLiquidity` as `OETHVaultAdmin` does external calls to `OETHVaultCore.addWithdrawalQueueLiquidity` which will fail the reentry. These external calls are in `swapCollateral`, `withdrawFromStrategy`, `withdrawAllFromStrategy` and `withdrawAllFromStrategies`

contracts/contracts/vault/OETHVaultCore.sol
naddison36: done

contracts/contracts/vault/OETHVaultCore.sol
naddison36: done

contracts/contracts/vault/OETHVaultCore.sol
naddison36: So our options are:
1. No reentry protection on `addWithdrawalQueueLiquidity`
2. duplicate the internal `_addWithdrawalQueueLiquidity` function on `OETHVaultAdmin`
3. change `nonReentrant` to ignore reentry from itself. ie the proxy contract
4. move `_addWithdrawalQueueLiquidity` up to the common `VaultStorage` contract
4. Add a new contract for the common functions across `OETHVaultCore` and `OETHVaultAdmin`. ie `_addWithdrawalQueueLiquidity` and `_wethAvailable`

naddison36: Here's the what the inheritance of the OETH Vault currently looks like

OETHVaultHierarchy

naddison36: Adding a `OETHVaultCommon` contract with `_addWithdrawalQueueLiquidity` and `_wethAvailable` seems ok. This is what the inheritance looks like 

OETHVaultHierarchy

I've done the changes in a separate PR if we want to do it https://github.com/OriginProtocol/origin-dollar/pull/2153

contracts/contracts/vault/OETHVaultCore.sol
naddison36: For purposing of reviewing this bug, here's the problem function
```Solidity
function _totalValue() internal view override returns (uint256 value) {
    value = _totalValueInVault() + _totalValueInStrategies();

    // Need to remove WETH that is reserved for the withdrawal queue.
    // reserved for the withdrawal queue = cumulative queued total - total claimed
    WithdrawalQueueMetadata memory queue = withdrawalQueueMetadata;
    if (value + queue.claimed >= queue.queued) {
        return value + queue.claimed - queue.queued;
    }
}
```

Here's the new implementation
```Solidity
function _totalValue() internal view override returns (uint256 value) {
    value = _totalValueInVault() + _totalValueInStrategies();

    // Need to remove WETH that is reserved for the withdrawal queue.
    WithdrawalQueueMetadata memory queue = withdrawalQueueMetadata;
    // reserved for the withdrawal queue = cumulative queued total - total claimed
    uint256 reservedForQueue = queue.queued - queue.claimed;

    if (value < reservedForQueue) {
        // This can happen if the vault becomes insolvent enough that the
        // total value in the vault and all strategies is less than the outstanding withdrawals.
        // For example, there was a mass slashing event and most users request a withdrawal.
        return 0;
    }

    // Adjust the total value by the amount reserved for the withdrawal queue
    return value - reservedForQueue;
}
```

contracts/contracts/vault/OETHVaultCore.sol
naddison36: For the review of this bug, here's the relevant section in `_redeem`

```Solidity 
// If there is any WETH in the Vault available after accounting for the withdrawal queue
if (_wethAvailable() >= amountMinusFee) {
    // Use Vault funds first if sufficient
    IERC20(weth).safeTransfer(msg.sender, amountMinusFee);
} else {
    address strategyAddr = assetDefaultStrategies[weth];
    if (strategyAddr != address(0)) {
        // Nothing in Vault, but something in Strategy, send from there
        IStrategy strategy = IStrategy(strategyAddr);
        strategy.withdraw(msg.sender, weth, amountMinusFee);
    } else {
        // Cant find funds anywhere
        revert("Liquidity error");
    }
}
```

Here's the new implementation of this section
```Solidity
// Is there enough WETH in the Vault available after accounting for the withdrawal queue
require(_wethAvailable() >= amountMinusFee, "Liquidity error");

// Transfer WETH minus the fee to the redeemer
IERC20(weth).safeTransfer(msg.sender, amountMinusFee);
```

contracts/contracts/vault/OETHVaultCore.sol
naddison36: FYI, I think we should move the common internal functions between VaultAdmin and VaultCore into a common contract. I have implemented this PR https://github.com/OriginProtocol/origin-dollar/pull/2153

contracts/contracts/vault/OETHVaultCore.sol
DanielVF: What if we just did something like `require(msg.sender == addres(this))` on this method. There shouldn't be a reason someone outside the protocol actually needs to call this externally.

contracts/contracts/vault/OETHVaultAdmin.sol
```diff
@@ -1,12 +1,113 @@
...
+    }
+
+    function _swapCollateral(
+        address _fromAsset,
+        address _toAsset,
+        uint256 _fromAssetAmount,
+        uint256 _minToAssetAmount,
+        bytes calldata _data
+    ) internal override returns (uint256 toAssetAmount) {
+        require(_fromAsset != weth, "Swap from WETH not supported");
```
DanielVF, line 100: If the vault only has WETH, and no other collaterals, can we just make this revert?

contracts/contracts/vault/OETHVaultCore.sol
```diff
@@ -128,24 +140,347 @@ contract OETHVaultCore is VaultCore {
...
+
+        // The amount of WETH that is still to be claimed in the withdrawal queue
+        uint256 outstandingWithdrawals = queue.queued - queue.claimed;
+
+        // The amount of sitting in WETH in the vault
+        uint256 wethBalance = IERC20(weth).balanceOf(address(this));
+
+        // If there is more WETH in the vault than the outstanding withdrawals
+        if (wethBalance > outstandingWithdrawals) {
+            wethAvailable = wethBalance - outstandingWithdrawals;
```
DanielVF, line 380: ⚪ Cleaner with just a return here, rather than `wethAvailable =` ?

contracts/contracts/vault/OETHVaultCore.sol
```diff
@@ -128,24 +140,347 @@ contract OETHVaultCore is VaultCore {
...
+        requestId = withdrawalQueueMetadata.nextWithdrawalIndex;
+        queued = withdrawalQueueMetadata.queued + _amount;
+
+        // Store the next withdrawal request
+        withdrawalQueueMetadata.nextWithdrawalIndex = SafeCast.toUint128(
+            requestId + 1
+        );
+        // Store the updated queued amount which reserves WETH in the withdrawal queue
+        // and reduces the vault's total assets
+        withdrawalQueueMetadata.queued = uint128(queued);
```
DanielVF, line 185: ⚪ We should safe cast the other casts in this file.

contracts/contracts/vault/OETHVaultCore.sol
naddison36: All the vault actions that add WETH call `_addWithdrawalQueueLiquidity`. But we do have the Native Staking Strategy that will transfer WETH to the vault when it accounts for validator withdrawals. The strategy does not call `addWithdrawalQueueLiquidity` on the vault. A separate tx in the Defender Action has to call `addWithdrawalQueueLiquidity`.

contracts/contracts/vault/OETHVaultCore.sol
naddison36: Agree. Done

contracts/contracts/vault/OETHVaultAdmin.sol
naddison36: I've made this change

contracts/contracts/vault/OETHVaultCore.sol
naddison36: I prefer to be explicit with `SafeCast.toUint128(queued)`.
If you have a lot of using libraries it can be hard to know what is being used from which library.

contracts/contracts/vault/OETHVaultCore.sol
naddison36: @DanielVF any preference over which option we take to resolve this one?

DanielVF: I would lean then towards leaving it as it then, since we have to have this called externally by both the vault and by other code. Not worth adding complexity in the service of extra defensiveness.

contracts/contracts/vault/OETHVaultAdmin.sol
```diff
@@ -1,12 +1,102 @@
...
+    }
+
+    function _withdrawAllFromStrategies() internal override {
+        super._withdrawAllFromStrategies();
+
+        IVault(address(this)).addWithdrawalQueueLiquidity();
+    }
+
+    /// @dev Calculates the amount of WETH in the Vault that is not reserved for the withdrawal queue.
+    function _wethAvailable() internal view returns (uint256 wethAvailable) {
```
DanielVF, line 78: 🟠 Our two WETH available functions do not match.

contracts/contracts/vault/OETHVaultCore.sol
DanielVF: The other option would be to have two external functions - one nonRentrant, and callable by the world, and the other vault only without the reentrancy check.

DanielVF (changes requested):
# OETH Async Withdraws

## Requirements

We wish to allow both the public and the OETH ARM to make asynchronous, perfectly 1:1 withdraws.

We wish to avoid the theft of yield attacks that would normally happen with a rebasing token, and we wish to avoid flash loan mint/redeem attacks in imprecisions.


## Easy Checks

#### Authentication

- [x] Never use tx.origin
- [ ] Every external/public function is supposed to be externally accessible
	- [x] -`addWithdrawLiquidty` should be restricted to being called by the vault only, since it's only external for code savings reasons.- (Nope, turns out it needs to be called from outside our system)

#### Ethereum

#### Cryptographic code

_no crypto_

#### Gas problems

    - [x] A way to remove items

#### Black magic

#### Overflow

#### Proxy

#### Events

## Medium Checks

#### Rounding
	- [ ] 🟣 Redeem fee calculations round in favor of users against protocol. We can leave this for now since the rounding error is not multiplied out anywhere and will remain small.
- [ ] Casts
	- [x] 🟠 Need some safecasts on casts (done)

#### Dependencies

#### External calls

	- [ ] 🟣 _checkBalance() does not validate the address of the token used. This is acceptable because it is currently only reachable externally via checkBalance(), which does not update state. The only call outwards is via a staticcall, which precludes reentrancy. However someone could make a coin that could checkBalance cause checkBalance to return silly amounts when called by the vault. Having check balance return zero on non-valid tokens would mean that it would ignore non-valid tokens just as the vault totals do. This is not the PR for that change though.
	- [x] All external calls are to trusted addresses
	- [ ] 🟢 _Dripper calls are carefully positioned so that a compromised dripper could not reentrancy attack, even without reentrancy checks. Nice._
    - [ ] 🟣 addWithdrawalQueueLiquidity does not have reentrancy guards. However the maximum damage it could be used for, if there was a reentrancy issue elsewhere in the code, would be to allow someone to exit the queue out of order.
- [x] Oracles, one of:
  - [x] No oracles
  - [ ] Oracles can't be bent

#### Tests

#### Deploy

## Thinking

#### Logic

- [x] If's correct
- [x] For's correct

#### Internal State

**Invariant** `queued >= claimable >= claimed` on `withdrawalQueueMetadata`

Each is of these have their own single place in code that sets them. And each only goes up. 

`queued` is increased by the amount of a withdraw request.

`claimable` can only move up by the maximum of difference between it and queued, keeping the invariant.

`claimed` can only move up by withdraw requests as they are paid. This keeps the invariant with `queued >= claimed`. `claimed` can only be increased when `queue.claimable >= request.queued`, which keeps the `claimable >= claimed` invariant.

**Invariant** 



**More invariants**
- ✅ queue.claimable == max(request.queued)
- ✅ queue.nextWithdrawIndex == max(request.withdrawIndex-1)
- ✅ `(queue.queued - queue.claimed) = sum(request.amount where amount unclaimed) 


**More invariants**
- ✅ User should not be able to request a withdraw that would end with the protocol overly insolvent.
- ✅ User should not be able to claim a withdraw that would end with the protocol overly insolvent. 
- ✅ User should not be able to do a sync redeem that would end with the protocol overly insolvent. 

#### Attack

This PR affects the most of important code in the vault. This how money moves out of the vault to users and this adds two more exit paths to current redeem path. It also alters the calculation of total value, which is again as critical of code as we have. And it adds code to the mint function (again critical code).

If total value over calculated how much we have, this would lead to rebases. This should be correct because the strategy part is unchanged, the vault value part is correct, and because the reserved calculations are only on the amount that has been burned but not paid out. ✅

An attacker could attempt to withdraw to much via the redeem path. The redeem path will never return more WETH than it burned WETH. ✅

An attacker could attempt to mint too much OETH for WETH. The mint path will never mint more OETH than it takes WETH. ✅

✅ Request withdraw burns the exact amount of OETH that it stores in a request.

✅ claimWithdraw and claimWithdraws both send users the exact amount in requests, and these requests cannot be used twice.

The OETH token has occasional rounding errors. However these are only one off, and against the user as far as we know. These rounding errors do not affect the protocol during minting, redeeming, and total valuing. ✅

A concern is that the odds of the vault reading a total value of zero have just gone up from roughly impossible, to something the  code now moves values to. This should not open any attacks because rebasing will disable when the the vault total value is zero, and nothing else relies on it being non-zero. ✅


#### Flavor

_We've done many flavor improvements, and it's looking fairly good._


#### Remaining
- [x] 🟠 `OETHVaultAdmin#_wethAvailable` does not match code with the vault core one.

contracts/contracts/vault/OETHVaultAdmin.sol
naddison36: Fixed. I think this is a good argument to getting rid of the duplicate code

contracts/contracts/vault/OETHVaultCore.sol
naddison36: How about we use a library with the two duplicate functions? I can do it on a PR so you can see what it looks like

naddison36: Here's the PR with the common internal functions in a library https://github.com/OriginProtocol/origin-dollar/pull/2161

naddison36:
@DanielVF I've fixed `OETHVaultAdmin._wethAvailable` for it matches `OETHVaultCore._wethAvailable`. But I'm in favour of moving `_wethAvailable` and `_addWithdrawalQueueLiquidity` to a new compile time library `OETHVaultLibrary` that is called by both `OETHVaultAdmin` and `OETHVaultCore`. I've put this in a separate PR https://github.com/OriginProtocol/origin-dollar/pull/2161

contracts/contracts/vault/OETHVaultCore.sol
naddison36: This has been fixed in the new `_wethAvailable` implementation

DanielVF (approved):

contracts/contracts/vault/OETHVaultCore.sol
```diff
@@ -128,24 +140,349 @@ contract OETHVaultCore is VaultCore {
...
+        override
+        returns (uint256 balance)
+    {
+        balance = super._checkBalance(_asset);
+
+        if (_asset == weth) {
+            WithdrawalQueueMetadata memory queue = withdrawalQueueMetadata;
+            // Need to remove WETH that is reserved for the withdrawal queue
+            if (balance + queue.claimed >= queue.queued) {
+                return balance + queue.claimed - queue.queued;
```
shahthepro, line 406: nit: `return` inside conditional statements, Isn't this something that we just agreed not to do to make sure that the code is more readable?

shahthepro (approved):
## Requirements

Allow user to make async withdrawals. Should avoid flash loan and other attacks by incorporating a claim delay.

## Easy Checks

#### Authentication

- [x] Never use tx.origin

#### Ethereum

#### Cryptographic code

#### Gas problems

    - [x] A way to remove items

#### Black magic

#### Overflow

#### Proxy

#### Events

## Medium Checks

#### Rounding

#### Dependencies

#### External calls

- [x] No malicious behaviors
- [ ] Oracles, one of:
  - [x] No oracles
  - [ ] Oracles can't be bent

#### Tests

#### Deploy

## Strategy Specific

#### Strategy checks

#### Downstream

## Thinking

#### Logic

#### Internal State
Leaving this out since DVF already did a great job at explaining the internal state and invariants

#### Attack

- The code assumes that OETH is pegged to WETH 1:1. So, that it doesn't use any oracles that will leave the contract vulnerable. 
- Claim delay would prevent any sort of flash loan attacks of mints & redeems
- There're solvency checks wherever it's needed

#### Flavor

Code looks simple and good. There's one minor nitpick in `checkBalance` where `return` is used inside the conditional statement. This is something that we decided not to do. Would be nice to get rid of it. Also, could move the `asset == weth` check that could moved up like `if (asset == weth) return 0` that should address the `checkBalance` not checking the asset address concern that Daniel raised.

contracts/contracts/vault/OETHVaultCore.sol
naddison36: you are correct. From what we discussed at the last contract sync, this would be rewritten to

```Solidity
function _checkBalance(address _asset)
    internal
    view
    override
    returns (uint256 balance)
{
    balance = super._checkBalance(_asset);

    if (_asset != weth) {
        return 0;
    }

    WithdrawalQueueMetadata memory queue = withdrawalQueueMetadata;
    // Need to remove WETH that is reserved for the withdrawal queue
    if (balance + queue.claimed < queue.queued) {
        return 0;
    }

    return balance + queue.claimed - queue.queued;
}
```

But I'm not going to change it at this late stage

naddison36: @DanielVF @shahthepro 
In hind sight I should have done this late change as `_checkBalance` has the same bug as what `_wethAvailable` had. I'm going to change to the above

naddison36: Here's the PR to fix this issue on `master` https://github.com/OriginProtocol/origin-dollar/pull/2162

PR #2067, opened 2024-05-17
Update token names in final buyback contract

DanielVF:
We changed the variable names `ogv` to `ogn` in the BaseBuyback contract but didn't change in the contracts inherit it. Simple update just to keep things clean.


shahthepro (approved):

shahthepro:
I also renamed `BaseBuyback` to `AbstractBuyback` and also removed the unused `MockBuyback` contract. Will merge it to master and redo the deploy of implementations so that these changes are picked up

PR #2069, opened 2024-05-20
Husky pre-commit to run linter

naddison36:
# Changes

- Add Husky as a dependency at the project root
- Change the Husky pre-commit to run the contracts linter
- Upate Husky install docs

This does slow down the commit as it needs to run the linter


naddison36:
It was agreed in the contract sync to give this a try. We can change if it's not working out

PR #2070, opened 2024-05-20
correct pre-commit hook

sparrowDom:
correct pre-commit hook


naddison36:
We may fall back to this. As discussed in the contract sync, we'll change the pre-commit to `yarn lint:js` in PR https://github.com/OriginProtocol/origin-dollar/pull/2069

PR #2080, opened 2024-05-30
Lido withdraw strategy

naddison36:
# Contract changes

* Added new `LidoWithdrawalStrategy` contract to swap stETH or WETH using the Lido Withdrawa Queue

# Dependencies

* https://github.com/OriginProtocol/origin-dollar/pull/2015

# Security

Contract change checklist:
  - [ ] Unit tests pass


contracts/contracts/strategies/LidoWithdrawalStrategy.sol
```diff
@@ -0,0 +1,307 @@
...
+    }
+
+    /**
+     * @notice Takes all given stETH and creates Lido withdrawal request
+     */
+    function depositAll() external override onlyVault nonReentrant {
+        uint256 stETHStart = stETH.balanceOf(address(this));
+        require(stETHStart > 0, "No stETH to withdraw");
+
+        uint256 numWithdrawals = (stETHStart / MaxWithdrawalAmount) + 1;
```
pandadefi, line 137: In case `stETHStart mod MaxWithdrawalAmount == 0` the +1 is not necessary.

contracts/contracts/strategies/LidoWithdrawalStrategy.sol
```diff
@@ -0,0 +1,307 @@
...
+        );
+
+        emit WithdrawalRequests(requestIds, amounts);
+
+        // Is there any stETH left except 1 wei from each request?
+        require(
+            stETH.balanceOf(address(this)) <= numWithdrawals,
+            "Not all stEth in withdraw queue"
+        );
+        outstandingWithdrawals += stETHStart; // Single set for gas reasons
```
pandadefi, line 160: You could subscribe to stETHStart the very little balance of stETH.

contracts/contracts/strategies/LidoWithdrawalStrategy.sol
```diff
@@ -0,0 +1,307 @@
...
+        uint256[] memory _requestIds,
+        uint256 expectedAmount
+    ) external nonReentrant {
+        require(
+            msg.sender == IVault(vaultAddress).strategistAddr(),
+            "Caller is not the Strategist"
+        );
+        uint256 startingBalance = payable(address(this)).balance;
+        uint256 lastIndex = withdrawalQueue.getLastCheckpointIndex();
+        uint256[] memory hintIds = withdrawalQueue.findCheckpointHints(
```
pandadefi, line 209: Have you considered doing this off-chain and passing the hint to the function?

contracts/contracts/strategies/LidoWithdrawalStrategy.sol
```diff
@@ -0,0 +1,307 @@
...
+        if (payable(address(this)).balance > 0) {
+            weth.deposit{ value: payable(address(this)).balance }();
+        }
+        uint256 wethBalance = weth.balanceOf(address(this));
+        if (wethBalance > 0) {
+            // slither-disable-next-line unchecked-transfer
+            weth.transfer(vaultAddress, wethBalance);
+            emit Withdrawal(address(weth), address(0), wethBalance);
+        }
+        uint256 fraxEthBalance = stETH.balanceOf(address(this));
```
pandadefi, line 256: Here the name of the function refers to an other liquid staking protocol.

contracts/contracts/strategies/LidoWithdrawalStrategy.sol
```diff
@@ -0,0 +1,307 @@
...
+        // slither-disable-next-line unchecked-transfer
+        weth.transfer(vaultAddress, currentBalance);
+        emit Withdrawal(
+            address(weth),
+            address(withdrawalQueue),
+            currentBalance
+        );
+    }
+
+    function _abstractSetPToken(address, address) internal override {
```
DanielVF, line 238: Maybe throw this internal method to the bottom of the contract? So it's not between all the external withdraw methods.

contracts/contracts/strategies/LidoWithdrawalStrategy.sol
naddison36: `numWithdrawals` is how many withdrawal requests we are sending to the Lido withdrawal queue.
When the stETH balance is less than the `MaxWithdrawalAmount` (1000 ether), then there is one withdrawal.

contracts/contracts/strategies/LidoWithdrawalStrategy.sol
naddison36: I like the thinking as it'll save gas on the `getLastCheckpointIndex` and `findCheckpointHints` calls. Looking at our stETH ARM, it'll save around 16k in gas.

059e3103

Given this is one-off use, I think can leave the `getLastCheckpointIndex` and `findCheckpointHints` calls in the contract. gas is relatively cheap at the moment :-)

contracts/contracts/strategies/LidoWithdrawalStrategy.sol
naddison36: good pickup. I did a search and replace for frxEth but missed fraxEth

contracts/contracts/strategies/LidoWithdrawalStrategy.sol
naddison36: done. I've also made it pure

contracts/contracts/strategies/LidoWithdrawalStrategy.sol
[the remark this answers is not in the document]
naddison36: good idea. I've gone with `withdrawalLength`

contracts/contracts/strategies/LidoWithdrawalStrategy.sol
naddison36: Good point. I've changed it to
```Solidity
// Is there any stETH left except 1 wei from each request?
// This is because stETH does not transfer all the transfer amount.
uint256 stEthDust = stETH.balanceOf(address(this));
require(
    stEthDust <= withdrawalLength,
    "Not all stEth in withdraw queue"
);
outstandingWithdrawals += stETHStart - stEthDust;
```

shahthepro:
Since https://github.com/OriginProtocol/origin-dollar/pull/2097 deployed 097 already and this PR seems to be update that, we need a different deployment file

contracts/contracts/strategies/LidoWithdrawalStrategy.sol
```diff
@@ -0,0 +1,314 @@
...
+        uint256 stETHStart = stETH.balanceOf(address(this));
+        require(stETHStart > 0, "No stETH to withdraw");
+
+        uint256 withdrawalLength = (stETHStart / MaxWithdrawalAmount) + 1;
+        uint256[] memory amounts = new uint256[](withdrawalLength);
+
+        uint256 stETHRemaining = stETHStart;
+        uint256 i = 0;
+        while (stETHRemaining > MaxWithdrawalAmount) {
+            amounts[i++] = MaxWithdrawalAmount;
```
sparrowDom, line 148: Nice I haven't seen this expression in years. Had too google that the [i++] returns the initial value but after the expression still increments it.

contracts/test/strategies/lido_withdrawal_strategy.fork-test.js
```diff
@@ -0,0 +1,227 @@
...
+        .connect(strategist)
+        .withdrawAllFromStrategy(lidoWithdrawalStrategy.address);
+
+      // Check final balances
+      const finalWethBalanceVault = await weth.balanceOf(oethVault.address);
+      const finalstEthBalanceVault = await stETH.balanceOf(oethVault.address);
+      const finalNativeEthBalanceVault = await oethVault.provider.getBalance(
+        oethVault.address
+      );
+      expect(finalWethBalanceVault.sub(initialWethBalanceVault))
```
sparrowDom, line 119: very good fork tests 🙏

sparrowDom:
@shahthepro yeah the diff on 097 deployment file is pretty terrible :) there is a clean [098 deployment file](https://github.com/OriginProtocol/origin-dollar/blob/master/contracts/deploy/mainnet/098_lido_withdrawal_strategy.js) in master that is ok.

contracts/contracts/strategies/LidoWithdrawalStrategy.sol
```diff
@@ -0,0 +1,314 @@
...
+    IWETH9 private constant weth =
+        IWETH9(0xC02aaA39b223FE8D0A0e5C4F27eAD9083C756Cc2);
+    /// @notice Address of the stETH token
+    IERC20 private constant stETH =
+        IERC20(0xae7ab96520DE3A18E5e111B5EaAb095312D7fE84);
+    /// @notice Address of the Lido Withdrawal Queue contract
+    IStETHWithdrawal private constant withdrawalQueue =
+        IStETHWithdrawal(0x889edC2eDab5f40e902b864aD4d7AdE8E412F9B1);
+    /// @notice Maximum amount of stETH that can be withdrawn in a single request
+    uint256 public constant MaxWithdrawalAmount = 1000 ether;
```
shahthepro, line 94: nit: For constants, using upper snake case would be nice, not necessary to change at this point

shahthepro:
## Requirements

The strategy should interact with Lido Withdrawal Queue and should be able to redeem stETH.

## Easy Checks

#### Authentication

- [x] Never use tx.origin
- [x] Every external/public function has the correct authentication. 
- - _`claimWithdrawals` is callable by Strategist but not by Governor. Ideally I would expect Governor to be able to do everything that Strategist can do. However this is fine, since this is a one-off strategy and if needed the Governor can still call `setStrategist()` and then this method in a single proposal._


#### Ethereum

- - _It does receive ETH and wrap it into WETH during withdrawals. However, that's how it's supposed to be`_
- - _It needs to receive ETH, so needs the `receive() payable` method that it has_

#### Cryptographic code

#### Gas problems

    - [ ] A way to remove items

#### Black magic

#### Overflow

#### Proxy

- - _Nit: The contract's initialize method takes three args, however we always expect those three to be empty arrays. So, could probably change the function signature but not necessary_

#### Events

## Medium Checks

#### Rounding

#### Dependencies

#### External calls

- [x] No malicious behaviors
- [ ] Oracles, one of:
  - [x] No oracles
  - [ ] Oracles can't be bent

#### Tests

#### Deploy

## Strategy Specific

#### Strategy checks

  - _No reward tokens_

#### Downstream

## Thinking

#### Logic

  - _There's no `onERC721Receive` method on this contract. However, Lido contract doesn't call that method when minting. Those are called/checked only during transfers. This kinda works for us since the strategy won't receive any NFT that it didn't mint_

_Deployment seems to be split and moved to a different file on master, so leaving it out of this review_

#### Internal State

_Everything seems about right. The contract depends on `outstandingWithdrawals` for all withdrawal accounting._

#### Attack

_Doesn't seem to be vulnerable to common attacks._

#### Flavor

_Code is simple and elegant. Doesn't seem to have any vulnerability_

contracts/contracts/strategies/LidoWithdrawalStrategy.sol
```diff
@@ -0,0 +1,314 @@
...
+        uint256 withdrawalLength = (stETHStart / MaxWithdrawalAmount) + 1;
+        uint256[] memory amounts = new uint256[](withdrawalLength);
+
+        uint256 stETHRemaining = stETHStart;
+        uint256 i = 0;
+        while (stETHRemaining > MaxWithdrawalAmount) {
+            amounts[i++] = MaxWithdrawalAmount;
+            stETHRemaining -= MaxWithdrawalAmount;
+        }
+        amounts[i] = stETHRemaining;
```
sparrowDom, line 151: Does this revert if the `stETHRemaining` == 0? Though considering current stETH balance on the vault is ` 17,886.18` this is highly unlikely to happen

contracts/contracts/strategies/LidoWithdrawalStrategy.sol
```diff
@@ -0,0 +1,314 @@
...
+    /// @notice Total amount of stETH that has been requested to be withdrawn for ETH
+    uint256 public outstandingWithdrawals;
+
+    event WithdrawalRequests(uint256[] requestIds, uint256[] amounts);
+    event WithdrawalClaims(uint256[] requestIds, uint256 amount);
+
+    constructor(BaseStrategyConfig memory _stratConfig)
+        InitializableAbstractStrategy(_stratConfig)
+    {
+        require(MaxWithdrawalAmount < type(uint120).max);
```
sparrowDom, line 104: we can remove this line since it will always be true. To anyone interested
```
>> type(uint120).max / 1e18
>> 1329227995784915872
```

contracts/contracts/strategies/LidoWithdrawalStrategy.sol
sparrowDom: Confirmed it would revert since according to the docs (https://docs.lido.fi/contracts/withdrawal-queue-erc721/): 

_The minimal amount for a request is 100 wei, and the maximum is 1000 eth. More significant amounts should be split into several requests, which allows us to avoid clogging the queue with an extra large request._

contracts/contracts/strategies/LidoWithdrawalStrategy.sol
sparrowDom, line 94: The documentation (https://docs.lido.fi/contracts/withdrawal-queue-erc721/#claimwithdrawals) states that the withdrawal amounts must be lower than the max amount: 
_each amount in _amounts must be greater than MIN_STETH_WITHDRAWAL_AMOUNT and lower than MAX_STETH_WITHDRAWAL_AMOUNT_ 

and not lower or equal

sparrowDom: Seems the documentation is wrong. The source code https://etherscan.io/address/0xe42c659dc09109566720ea8b2de186c2be7d94d9#code is as follows: 
```
uint256 public constant MIN_STETH_WITHDRAWAL_AMOUNT = 100;
uint256 public constant MAX_STETH_WITHDRAWAL_AMOUNT = 1000 * 1e18;

function _checkWithdrawalRequestAmount(uint256 _amountOfStETH) internal pure {
        if (_amountOfStETH < MIN_STETH_WITHDRAWAL_AMOUNT) {
            revert RequestAmountTooSmall(_amountOfStETH);
        }
        if (_amountOfStETH > MAX_STETH_WITHDRAWAL_AMOUNT) {
            revert RequestAmountTooLarge(_amountOfStETH);
        }
    }
```

which obviously means withdrawal amount can be 1000 ether exactly

contracts/contracts/strategies/LidoWithdrawalStrategy.sol
```diff
@@ -0,0 +1,314 @@
...
+
+    /**
+     * @notice Claim previously requested withdrawals that have now finalized.
+     * Called by the Strategist.
+     * @param _requestIds Array of withdrawal request identifiers
+     * @param expectedAmount Total amount of ETH expect to be withdrawn
+     */
+    function claimWithdrawals(
+        uint256[] memory _requestIds,
+        uint256 expectedAmount
```
sparrowDom, line 208: super nitpicky: this could use an underscore `_expectedAmount`

sparrowDom:
## Requirements

The strategy contract is interacting with Lido withdrawal queue to natively redeem stETH for ETH.

## Easy Checks

#### Authentication

- [x] Never use tx.origin

#### Ethereum

- 🟠  Contract does not send or receive Ethereum. _It does receive it but it performs the checks for expected ETH received ignoring the starting contract balance._ 
- 🟠 Contract has no payable methods. _It has it needs it_

#### Cryptographic code

#### Gas problems

- [x] Contracts with for loops must have either: _has a while loop_
    - [ ] A way to remove items
    - [x] loop is designed in a way that it definitely exits 

#### Black magic

#### Overflow

- [] All for loops use uint256 _no for loops_

#### Proxy

#### Events

## Medium Checks

#### Rounding
- [] Code correctly multiplies before division _doesn't divide in a way this would be needed_
- [] Contract does not have bugs from zero or near zero amounts _It potentially does since stETHRemaining in `depositAll` could be 0. But it is highly unlikely and I wouldn't fix it_

#### Dependencies
- [ ] Review any new contract dependencies thoroughly (e.g. OpenZeppelin imports) when new dependencies are added or version of dependencies changes. _no OZ imports_
- [ ] If OpenZeppelin ACL roles are use review & enumerate all of them. _not used_
- [ ] Check OpenZeppelin [security vulnerabilities](https://github.com/OpenZeppelin/openzeppelin-contracts/security/advisories) and see if any apply to current PR considering the version of OpenZeppelin contract used. _not needed_

#### External calls

- [x] No malicious behaviors
- [x] Low level call() must require success. _none used_
- [x] Oracles, one of:
  - [x] No oracles
  - [ ] Oracles can't be bent

#### Tests

  - [ ] Each require() has a test _no, but not needed IMO_
  - [ ] Edge conditions are tested _this will be a 1 off called only by the strategist. Edge cases don't need covering_
  - [ ] If tests interact with AMM make sure enough edge cases (pool tilts) are tested. Ideally with fuzzing. _they don't_

#### Deploy

## Strategy Specific
#### Strategy checks

- [ ] The harvester can sell all reward tokens _harvester not utilized_

#### Downstream

- [ ] We have monitoring on all backend protocol's governances _no not needed_
- [ ] We have monitoring on a pauses in all downstream systems _no not needed_

## Thinking

#### Logic

No this is a pretty straight forward deploy. 

#### Internal State

The amount stored in `outstandingWithdrawals` always matches the amount that is being in the Lido's withdrawal queue. And only the owner of the requests can claim them, meaning no one else aside from strategy contract can claim the items in the queue and potentially manipulate the `outstandingWithdrawals`. 
`outstandingWithdrawals` makes sure that checkBalance correctly represents the funds in the strategy. Any additional native redemptions or claims accurately update the `outstandingWithdrawals`.
Same as the point a line above. 

Does this code do that?
Yes

#### Attack

If someone would be able to influence the amount returned in `checkBalance`. Or interfere with the Lido's withdrawal queue lifecycle. 

If Lido's withdrawal queue wouldn't be permissioned. If code in our contract would read the balance of tokens instead of keeping its own accounting that is immune to unexpected transfers

yes 
#### Flavor
Code is very clean and simple.

naddison36:
I mistakenly merged this PR to `master` when I merged the soETH changes.

I'll create a new branch for post merge changes `nicka/lido-withdraw-strategy2`.

sparrowDom:
Yeah I know. I've left some comments in the code, though those 2 comments that are actionable are really nitpicky. I don't think you need to do any changes.

contracts/contracts/strategies/LidoWithdrawalStrategy.sol
naddison36: Good thing you checked Lido's source code. Thanks

contracts/contracts/strategies/LidoWithdrawalStrategy.sol
naddison36: I've made this change on the `nicka/lido-withdraw-strategy2` branch in PR https://github.com/OriginProtocol/origin-dollar/pull/2116

contracts/contracts/strategies/LidoWithdrawalStrategy.sol
sparrowDom: thanks

contracts/contracts/strategies/LidoWithdrawalStrategy.sol
sparrowDom: thanks

PR #2081, opened 2024-05-31
OZ - Native Staking - M-01 All Addresses Are Registered Validators by Default

naddison36:
# Contract Changes

* Adding `NON_REGISTERED` as the default value in the `VALIDATOR_STATE` enum.

# Issue

In Solidity, an enum variable is automatically set to its first enumerated value if it is initialized without specifying a particular value. In `ValidatorRegistrator.sol` , `REGISTERED` is the default value of the `VALIDATOR_STATE` enum, meaning that it is possible to stake with any address even if `registerSsvValidator` were not called. Since the address would not be
registered as a validator in the SSV Network, it would not perform validations correctly which could result in funds being lost due to slashing.

### DIFF
--- a/contracts/contracts/strategies/NativeStaking/ValidatorRegistrator.sol
+++ b/contracts/contracts/strategies/NativeStaking/ValidatorRegistrator.sol
@@ -215,6 +215,11 @@ abstract contract ValidatorRegistrator is Governable, Pausable {
         uint256 amount,
         Cluster calldata cluster
     ) external onlyRegistrator whenNotPaused {
+        require(
+            validatorsStates[keccak256(publicKey)] ==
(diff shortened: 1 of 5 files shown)


sparrowDom (approved):
LGTM

There are some unit tests that were confirming that stake threshold amounts work correctly. I think those are missing the assertions confirming `VALIDATOR_STATE` for the pubkey has the correct value. We could also add those in

naddison36:
> LGTM
> 
> There are some unit tests that were confirming that stake threshold amounts work correctly. I think those are missing the assertions confirming `VALIDATOR_STATE` for the pubkey has the correct value. We could also add those in

I've added assertions confirming `validatorsStates` for the pubkey has the correct value in the unit and fork tests

PR #2083, opened 2024-05-31
OZ - Native Staking - N-07 Lack of Indexed Event Parameters

naddison36:
# Contract changes

* Indexed the `RegistratorChanged` and `StakingMonitorChanged` events in the `ValidatorRegistrator` contract.

# Issue

Throughout the codebase, several events do not have indexed parameters.

To improve the ability of off-chain services to search and filter for specific events, consider [indexing event parameters](https://docs.soliditylang.org/en/latest/contracts.html#events).

# Comment

Although `bytes` types can be indexed, the Waffle framework used for testing does not support it. Ideally, the `pubkey` parameters would be indexed so events for a specific validator can be filtered. But this breaks Waffle's event matchers.

### DIFF
--- a/contracts/contracts/strategies/NativeStaking/ValidatorRegistrator.sol
+++ b/contracts/contracts/strategies/NativeStaking/ValidatorRegistrator.sol
@@ -54,8 +54,8 @@ abstract contract ValidatorRegistrator is Governable, Pausable {
         EXIT_COMPLETE // validator has funds withdrawn to the EigenPod and is removed from the SSV
     }
 
-    event RegistratorChanged(address newAddress);
-    event StakingMonitorChanged(address newAddress);
+    event RegistratorChanged(address indexed newAddress);
+    event StakingMonitorChanged(address indexed newAddress);
     event ETHStaked(bytes pubkey, uint256 amount, bytes withdrawal_credentials);
     event SSVValidatorRegistered(bytes pubkey, uint64[] operatorIds);
     event SSVValidatorExitInitiated(bytes pubkey, uint64[] operatorIds);
--- a/contracts/contracts/strategies/NativeStaking/ValidatorRegistrator.sol
+++ b/contracts/contracts/strategies/NativeStaking/ValidatorRegistrator.sol
@@ -57,10 +57,27 @@ abstract contract ValidatorRegistrator is Governable, Pausable {
 
     event RegistratorChanged(address indexed newAddress);
     event StakingMonitorChanged(address indexed newAddress);
-    event ETHStaked(bytes pubkey, uint256 amount, bytes withdrawal_credentials);
-    event SSVValidatorRegistered(bytes pubkey, uint64[] operatorIds);
-    event SSVValidatorExitInitiated(bytes pubkey, uint64[] operatorIds);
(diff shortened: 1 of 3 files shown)


contracts/contracts/strategies/NativeStaking/ValidatorRegistrator.sol
```diff
@@ -54,8 +54,8 @@ abstract contract ValidatorRegistrator is Governable, Pausable {
         EXIT_COMPLETE // validator has funds withdrawn to the EigenPod and is removed from the SSV
     }
 
-    event RegistratorChanged(address newAddress);
-    event StakingMonitorChanged(address newAddress);
+    event RegistratorChanged(address indexed newAddress);
+    event StakingMonitorChanged(address indexed newAddress);
     event ETHStaked(bytes pubkey, uint256 amount, bytes withdrawal_credentials);
```
sparrowDom, line 59: Should we also index all of the event parameters that contain the `bytes pubkey`? Might be useful to be able to quickly find the relevant events of a particular validator being staked. 

The documentation states: 
```
When you attempt to index dynamic data types like `string`, `bytes`, `array`, or `struct` in Solidity, they don't get stored in their original form. Instead, the Ethereum log system stores the Keccak-256 hash of these data types.
```

I haven't tested it, but seems we would just need to keccak256(pubkey) in order to query the events of a particular `pubkey`

sparrowDom (approved):
LGTM

contracts/contracts/strategies/NativeStaking/ValidatorRegistrator.sol
naddison36: I tried but Waffle doesn't support it :-(

sparrowDom: Oh I though it just fails with tests. Fails on compile / execution? Ah that's too bad.

naddison36: compile and execution works. The Waffle `emit.withArgs` matcher doesn't work

sparrowDom: If I understand correctly, we can index the contract event, just not use the Waffle convenience functions to verify the correct behaviour. Possible approaches: 
1. we could introduce another redundant `bytes32` parameter called `pubkeyHash` which contains the keccak256 of the pubkey and index that. 
2. write an auxiliary test suite function to test for the correct value of indexed parameters. Unfortunately the test function would probably need to be aware that the `bytes` property is indexed.

I think it could be really useful to be able to quickly fetch all of the stake related events even before we have the monitoring backend setup and running.

naddison36: > When you attempt to index dynamic data types like string, bytes, array, or struct in Solidity, they don't get stored in their original form. Instead, the Ethereum log system stores the Keccak-256 hash of these data types.

from https://coinsbench.com/understanding-indexed-strings-in-solidity-events-19ba75986de6

I've gone with option 1 and added an indexed `pubKeyHash` to the events

PR #2095, opened 2024-06-07
rename fork tests

sparrowDom:
- remove the redundant '-test' in fork test names
- adds test files matching params required for fork tests of other chains


sparrowDom:
Lets merge this in next week, after we do the big SOETH deploy

PR #2106, opened 2024-06-21
WOETH lending markets adjustements

sparrowDom:
## Code Change Checklist

- [ ]  Unit tests *if needed

Internal review:

## Deploy checklist

```
```


sparrowDom:
## Requirements

## Easy Checks

#### Authentication

- [x] Never use tx.origin

#### Ethereum

#### Cryptographic code

#### Gas problems

- [ ] ~~Contracts with for loops must have either~~: _no loops_
    - [ ] ~~A way to remove items~~ _no loops_
    - [ ] ~~Can be upgraded to get unstuck~~ _no loops_
    - [ ] ~~Size can only controlled by admins~~ _no loops_
- [ ] ~~Contracts with for loops must not allow end users to add unlimited items to a loop that is used by others or admins.~~ _no loops_

#### Black magic

#### Overflow

- ~~[ ] All for loops use uint256~~ _no loops_

#### Proxy

#### Events

## Medium Checks

#### Rounding
- 🟠 Contract rounds in the protocols favor _it doesn't but it mimics what OUSD is doing using `mulTruncate` without rounding_

#### Dependencies
- ~~[ ] Review any new contract dependencies thoroughly (e.g. OpenZeppelin imports) when new dependencies are added or version of dependencies changes.~~ _no new dependancies added_

#### External calls

- [ ] No malicious behaviors
- [ ] Oracles, one of:
  - [ ] No oracles
  - [ ] Oracles can't be bent

#### Tests

#### Deploy

## Strategy Specific

#### Strategy checks

#### Downstream

## Thinking

#### Logic

#### Internal State

Does this code do that?

#### Attack

#### Flavor

_Could this code be simpler?_

contracts/contracts/token/WOETH.sol
```diff
@@ -6,17 +6,30 @@ import { ERC20 } from "@openzeppelin/contracts/token/ERC20/ERC20.sol";
...
+ * credits per token in those 4 actions. That way WOETH can keep an accurate track of the OETH balance
+ * ignoring any unexpected transfers of OETH to this contract.
  */
 
 contract WOETH is ERC4626, Governable, Initializable {
     using SafeERC20 for IERC20;
+    using StableMath for uint256;
+    // doesn't need to be public, but convenient to be able to confirm the state on the mainnet
+    uint256 public oethCreditsHighres;
+    bool _oethCreditsInitialized;
```
DanielVF, line 32: Missing an explicit visibility.

DanielVF:
The core attack we are trying to stop is someone sending the OETH to the wOETH contract, causing the value of wOETH in OETH terms to go suddenly up.

It looks like totalAssets uses the amount of OETH held by the contract as one of two multipliers. totalAssets is in turn used to calculate the exchange ratio. If someone donates to the contract, one of these two multipliers goes up, and the donation has perfectly succeeded in increasing the value of each wOETH. This attack does not appear to be blocked at all?

Or am I missing something?

DanielVF:
It also feels really scary that were are minting and burning using old ratios. That doesn't cause rektness?

contracts/contracts/token/WOETH.sol
sparrowDom: thanks nice catch, corrected

sparrowDom:
@DanielVF not completely sure what you mean. So the logic is that: 
- when the contract is initialized (on proposal execution) the internal credits of OETH are read and stored in WOETH. From that point forward the WOETH doesn't ask for internal credits OETH contract is holding but rather keeps its own accounting
- on each mint, deposit, withdraw, redeem the current credits per token are fetched from OETH (since we don't know if a rebase happened since the last time we queried it) and the accounting of credits in WOETH contracts are updated. Since only those 4 functions trigger the update of credits a normal transfer of OETH to WOETH contract has no effect. There is also [a test for this](https://github.com/OriginProtocol/origin-dollar/pull/2106/files#diff-1a55b6049ea6ba4898ff1fa84a24b20c6a6e2d20764957272f41ac0a6d01b177R109-R127) where you can see that after OETH transfer the WOETH account will still have a balance of 50 OETH and totalSupply of 0.

DanielVF:
I misread the code and got the credits per token and the credits switched. Nevermind everything I said - I'll have to stare at it more!

contracts/contracts/token/WOETH.sol
```diff
@@ -49,7 +81,161 @@ contract WOETH is ERC4626, Governable, Initializable {
...
+
+    /**
+     * @dev This function converts requested OETH token amount to its underlying OETH
+     * credits value that is stored internally in OETH.sol and is required in order to
+     * be able to rebase.
+     *
+     * @param oethAmount Amount of OETH to be converted to OETH credits
+     * @return amount of OETH credits the OETH amount corresponds to
+     */
+    function _oethToOethCreditsHighres(uint256 oethAmount)
```
DanielVF, line 98: Perhaps this would clearer if this method was named `_creditsPerAsset`, and the state variable `oethCreditsHighres` changed to `assetCredits`. 

In this way all the 4626 code gets to use clear variable names, and the OETH related stuff ends up just inside `_creditsPerAsset`.

contracts/contracts/token/WOETH.sol
```diff
@@ -49,7 +81,161 @@ contract WOETH is ERC4626, Governable, Initializable {
...
+         * of OETH.
+         */
+        return oethAmount.mulTruncate(creditsPerTokenHighres);
+    }
+
+    /** @dev See {IERC4262-totalAssets} */
+    function totalAssets() public view virtual override returns (uint256) {
+        (, uint256 creditsPerTokenHighres, ) = OETH(asset())
+            .creditsBalanceOfHighres(address(this));
+
```
DanielVF, line 122: If we used in `credits` here as well from the call from `creditsBalanceOfHighres`, we would have ever piece of data already loaded inside this function to check that our actual assets was equal or greater than our expected assets.

DanielVF:
Invariants I can think of:

- Assuming credits creditsPerTokenHighres in OETH only moves down, then woeth.convertToAssets() should only ever move up
- We should always be able to withdraw all WOETH to OETH. (This is harder to test, so another way of phrasing this is that we should always have enough OETH on hand for everyone to withdraw)
- OETH Donations to wOETH should not affect the convertToAssets() result
- WETH donations to oeth can affect the convertToAssets() result upwards after a rebase, and that's fine.

contracts/contracts/token/WOETH.sol
```diff
@@ -6,17 +6,30 @@ import { ERC20 } from "@openzeppelin/contracts/token/ERC20/ERC20.sol";
...
+ * token on mint, deposit, redeem, withdraw. The issue is that OETH is rebasing and OETH balances
+ * will change when the token rebases. For that reason we are tracking the WOETH contract credits and
+ * credits per token in those 4 actions. That way WOETH can keep an accurate track of the OETH balance
+ * ignoring any unexpected transfers of OETH to this contract.
  */
 
 contract WOETH is ERC4626, Governable, Initializable {
     using SafeERC20 for IERC20;
+    using StableMath for uint256;
+    // doesn't need to be public, but convenient to be able to confirm the state on the mainnet
```
DanielVF, line 30: Can remove this comment - we default to public on most contract variables.

contracts/contracts/token/WOETH.sol
sparrowDom: good point thanks, renamed

contracts/contracts/token/WOETH.sol
sparrowDom: Hmm, that would be a good check that would detect possible mathematical irregularities (perhaps due to rounding). Though as you point out it would be weird to revert in a view function.

Monitoring sounds like a good place for it though. Have Grafana regularly query both credits values and check that the ones in OETH.sol are greater or equal to the ones in WOETH.sol.

sparrowDom: Notion task added here: https://www.notion.so/originprotocol/WOETH-monitoring-sanity-check-695c73c0d0d04ef999a5242407d8a212

contracts/contracts/token/WOETH.sol
sparrowDom: agreed

contracts/contracts/token/WOETH.sol
```diff
@@ -31,6 +43,25 @@ contract WOETH is ERC4626, Governable, Initializable {
         OETH(address(asset())).rebaseOptIn();
     }
 
+    function initialize2() external onlyGovernor {
+        if (_oethCreditsInitialized) {
+            require(false, "Initialize2 already called");
+        }
```
shahthepro, line 49: Why not just `require(!_oethCreditsInitialized, "Initialize2 already called");`?

contracts/contracts/token/WOETH.sol
```diff
@@ -49,7 +80,161 @@ contract WOETH is ERC4626, Governable, Initializable {
...
+
+    /**
+     * @dev This function converts requested OETH token amount to its underlying OETH
+     * credits value that is stored internally in OETH.sol and is required in order to
+     * be able to rebase.
+     *
+     * @param oethAmount Amount of OETH to be converted to OETH credits
+     * @return amount of OETH credits the OETH amount corresponds to
+     */
+    function _creditsPerAsset(uint256 oethAmount)
```
shahthepro, line 97: nit: This function name confuses me a bit. When I see `per asset`, I assume it's gonna return credits for 1 OETH. But it takes in a `oethAmount` and multiplies the credits by it. So, it sounds more like `creditsForAssetAmount` than `creditsPerAsset`

contracts/contracts/token/WOETH.sol
```diff
@@ -49,7 +80,161 @@ contract WOETH is ERC4626, Governable, Initializable {
...
+            caller,
+            address(this),
+            assets
+        );
+        _mint(receiver, shares);
+        oethCreditsHighres += _creditsPerAsset(assets);
+
+        emit Deposit(caller, receiver, assets, shares);
+
+        return shares;
```
shahthepro, line 153: Could this be simplified into just (to avoid code-duplication):
```
shares = super.deposit(assets, receiver);
oethCreditsHighres += _creditsPerAsset(assets);
```

shahthepro: Same with other overridden functions as well

contracts/contracts/token/WOETH.sol
```diff
@@ -49,7 +80,161 @@ contract WOETH is ERC4626, Governable, Initializable {
...
+        uint256 shares = previewWithdraw(assets);
+
+        if (caller != owner) {
+            _spendAllowance(owner, caller, shares);
+        }
+
+        // if _asset is ERC777, transfer can call reenter AFTER the transfer happens through
+        // the tokensReceived hook, so we need to transfer after we burn to keep the invariants.
+        _burn(owner, shares);
+        oethCreditsHighres -= _creditsPerAsset(assets);
```
shahthepro, line 205: Haven't tested it yet but I wonder if we should compute the credits before burn? Let me see if it makes any difference or if it's breakable

shahthepro: I tried running some tests. Change in order didn't affect the numbers at all

contracts/contracts/token/WOETH.sol
```diff
@@ -49,7 +80,161 @@ contract WOETH is ERC4626, Governable, Initializable {
...
+        _mint(receiver, shares);
+        oethCreditsHighres += _creditsPerAsset(assets);
+
+        emit Deposit(caller, receiver, assets, shares);
+
+        return shares;
+    }
+
+    /** @dev See {IERC4262-mint} */
+    function mint(uint256 shares, address receiver)
```
pandadefi, line 157: You could save on some code complexity calling deposit from the mint function to prevent having most of the same code twice.

pandadefi:
A donation to increase the OETH price remains possible via the OETH contract but will require significant funds to increase the entire OETH vault value.

contracts/contracts/token/WOETH.sol
sparrowDom: good point: https://github.com/OriginProtocol/origin-dollar/pull/2106/commits/057469c08067ff67e31b8e1f524b61e249787499

contracts/contracts/token/WOETH.sol
sparrowDom: thanks, yes agreed. done here: https://github.com/OriginProtocol/origin-dollar/pull/2106/commits/1a6a16fc4af1f115f5c84a331df821da4b551d47

contracts/contracts/token/WOETH.sol
sparrowDom: yup agreed: https://github.com/OriginProtocol/origin-dollar/pull/2106/commits/1a6a16fc4af1f115f5c84a331df821da4b551d47

contracts/contracts/token/WOETH.sol
sparrowDom: yup good point. done here: https://github.com/OriginProtocol/origin-dollar/pull/2106/commits/1a6a16fc4af1f115f5c84a331df821da4b551d47

contracts/contracts/token/WOETH.sol
sparrowDom: good point. I have also renamed assets & shares since in the context of WOETH those can only be OETH & WOETH: https://github.com/OriginProtocol/origin-dollar/pull/2106/commits/2223600a1d8b8bf23c940a7debe124ddb1e4f44d

sparrowDom:
@pandadefi yes the donation attack on OETH with donating WETH to the Vault and calling `rebase` is ok. This is the way OETH increases its value and I can't think of a way how it could be exploited

contracts/contracts/token/WOETH.sol
```diff
@@ -6,17 +6,29 @@ import { ERC20 } from "@openzeppelin/contracts/token/ERC20/ERC20.sol";
...
+ * will change when the token rebases. For that reason we are tracking the WOETH contract credits and
+ * credits per token in those 4 actions. That way WOETH can keep an accurate track of the OETH balance
+ * ignoring any unexpected transfers of OETH to this contract.
  */
 
 contract WOETH is ERC4626, Governable, Initializable {
     using SafeERC20 for IERC20;
+    using StableMath for uint256;
+    uint256 public oethCreditsHighres;
+    bool private _oethCreditsInitialized;
```
shahthepro, line 31: `WOETHBase` inherits `WOETH`, so might as well add some gaps here now

contracts/contracts/token/WOETH.sol
```diff
@@ -49,7 +78,84 @@ contract WOETH is ERC4626, Governable, Initializable {
         external
         onlyGovernor
     {
+        //@dev TODO: we could implement a feature where if anyone sends OETH direclty to
+        // the contract, that we can let the governor transfer the excess of the token.
```
shahthepro, line 82: nit: perhaps we could just treat any donation as "yield"?

contracts/contracts/token/WOETH.sol
```diff
@@ -49,7 +78,84 @@ contract WOETH is ERC4626, Governable, Initializable {
...
+    function _oethToCredits(uint256 oethAmount) internal returns (uint256) {
+        (, uint256 creditsPerTokenHighres, ) = OETH(asset())
+            .creditsBalanceOfHighres(address(this));
+
+        /**
+         * Multiplying OETH amount with the creditsPerTokenHighres is exactly the math that
+         * is internally being done in OETH:
+         */
+        // solhint-disable-next-line max-line-length
+        /** https://github.com/OriginProtocol/origin-dollar/blob/2314cccf2933f5c1f76a6549c1f5c9cc935b6f05/contracts/contracts/token/OUSD.sol#L242-L249
```
shahthepro, line 104: nit: Points to older code (pre-yield-delegation). Would be nice to update it to point to latest commit hash on master

contracts/contracts/token/WOETH.sol
sparrowDom: good point, done

contracts/contracts/token/WOETH.sol
```diff
@@ -31,11 +43,40 @@ contract WOETH is ERC4626, Governable, Initializable {
         OETH(address(asset())).rebaseOptIn();
     }
 
-    function name() public view virtual override returns (string memory) {
+    function initialize2() external onlyGovernor {
```
DanielVF, line 46: Perhaps we should have `initialize()` call `initialize2()`. This way new contract deploys don't need to call both, and we are less likely to make a mistake.

contracts/contracts/token/WOETH.sol
DanielVF: I don't see the gaps?

contracts/contracts/token/WOETH.sol
```diff
@@ -5,24 +5,36 @@ import { ERC4626 } from "../../lib/openzeppelin/contracts/token/ERC20/extensions
...
+ * If OETH weren't rebasing this could be achieved by solely tracking the ERC20 transfers of the OETH
+ * token on mint, deposit, redeem, withdraw. The issue is that OETH is rebasing and OETH balances
+ * will change when the token rebases. For that reason we are tracking the WOETH contract credits and
+ * credits per token in those 4 actions. That way WOETH can keep an accurate track of the OETH balance
+ * ignoring any unexpected transfers of OETH to this contract.
  */
 
 contract WOETH is ERC4626, Governable, Initializable {
     using SafeERC20 for IERC20;
+    using StableMath for uint256;
```
DanielVF, line 30: This contract code is way cleaner than the last time I looked at it. Good job!

contracts/contracts/token/WOETH.sol
```diff
@@ -49,7 +90,84 @@ contract WOETH is ERC4626, Governable, Initializable {
...
+        woethAmount = super.withdraw(oethAmount, receiver, owner);
+        oethCreditsHighres -= _oethToCredits(oethAmount);
+    }
+
+    /** @dev See {IERC4262-redeem} */
+    function redeem(
+        uint256 woethAmount,
+        address receiver,
+        address owner
+    ) public virtual override returns (uint256 oethAmount) {
```
DanielVF, line 169: This contract does not currently compile. This should be fixed.

Also, although this is not the actual compile error, I'm wondering if these `virtual`s in these methods are wrong, since I think we want these functions callable without needing to be overridden by a child class.

contracts/contracts/token/WOETH.sol
DanielVF: While we shouldn't treat donations as instant yield (that's what this contract is trying to get away from), I do think we should build in a separate governor method to collect donated funds that are in excess of the backing funds.

contracts/contracts/token/WOETH.sol
```diff
@@ -49,7 +90,84 @@ contract WOETH is ERC4626, Governable, Initializable {
...
+    /**
+     * @dev This function converts requested OETH token amount to its underlying OETH
+     * credits value that is stored internally in OETH.sol and is required in order to
+     * be able to rebase.
+     *
+     * @param oethAmount Amount of OETH to be converted to OETH credits
+     * @return amount of OETH credits the OETH amount corresponds to
+     */
+    function _oethToCredits(uint256 oethAmount) internal returns (uint256) {
+        (, uint256 creditsPerTokenHighres, ) = OETH(asset())
```
DanielVF, line 108: Both `_oethToCredits()` and `totalAssets()` call `oeth.creditsBalanceOfHighres()` to get the `creditsPerToken`, discarding the other returned values.

I think it makes more sense to use the simpler `oeth.rebasingCreditsPerTokenHighres()`. 

There's two scenarios here:

1) This w token is correctly marked as rebasing. In this case, `oeth.rebasingCreditsPerTokenHighres()` will return the same value, be simpler, return only what we need, and cost less gas.

2) Governance messes up the world in a bad way and turns off yield to the contract. The current call will immediately return 1e18 instead of 1e27ish, making for a really really wrong totalAssets, and thus really really wrong conversion rate. However, if `oeth.rebasingCreditsPerTokenHighres()` is used we'll only get a gradual drift off the correct value as expected yield does not come in.

In both cases the behavior of `oeth.rebasingCreditsPerTokenHighres()` seems better.

contracts/contracts/token/WOETH.sol
```diff
@@ -49,7 +90,84 @@ contract WOETH is ERC4626, Governable, Initializable {
...
+    }
+
+    /** @dev See {IERC4262-redeem} */
+    function redeem(
+        uint256 woethAmount,
+        address receiver,
+        address owner
+    ) public virtual override returns (uint256 oethAmount) {
+        oethAmount = super.redeem(woethAmount, receiver, owner);
+        oethCreditsHighres -= _oethToCredits(oethAmount);
```
DanielVF, line 171: This is just a mathematical nit, and perhaps the code is okay without this, but in general, if you are depositing/plus-ing and withdrawing/minus-ing using the same conversion function, you are almost certainly rounding the wrong direction in one of them. It's possible we need two `_oethToCredits`, one that round up, and one that rounds down.

contracts/contracts/token/WOETH.sol
sparrowDom: Forgot to push remotely: https://github.com/OriginProtocol/origin-dollar/pull/2106/commits/be6d4734ec071df5ba9801fbc85bdfacb69c2007

contracts/contracts/token/WOETH.sol
sparrowDom: nice catch thanks: https://github.com/OriginProtocol/origin-dollar/pull/2106/commits/c38a3c659a62fd81807dadaca0151b629772deae

contracts/contracts/token/WOETH.sol
sparrowDom: Yes you are right these virtual keywords are not needed. We can add it in the future if need be: https://github.com/OriginProtocol/origin-dollar/pull/2106/commits/5ff5c5dfd0f3e81d386b6396894b954dfcf5f78b

contracts/contracts/token/WOETH.sol
sparrowDom: Great comment and great points thanks: https://github.com/OriginProtocol/origin-dollar/pull/2106/commits/3ef22194932715aac4bc2dca3d782c0769d2b420

contracts/contracts/token/WOETH.sol
```diff
@@ -62,7 +99,81 @@ contract WOETH is ERC4626, Governable, Initializable {
...
+         * up / down depending on what favours this contract
+         */
+        // solhint-disable-next-line max-line-length
+        /** https://github.com/OriginProtocol/origin-dollar/blob/c02572bd1c06eb3c2652c8692e52144be2efa741/contracts/contracts/token/OUSD.sol#L498
+         */
+        return (oethAmount * creditsPerTokenHighres + (roundUp ? 1e18 - 1 : 0)) / 1e18;
+    }
+
+    /** @dev See {IERC4262-totalAssets} */
+    function totalAssets() public view override returns (uint256) {
```
DanielVF, line 133: 🔴 `rebasingCreditsPerToken` has the property that the smaller it is, the more money a user has.

Default solidity division results in rounding down. Therefore this `totalAssets()` method is rounding up the amount of money in contract. Therefore it is very easy for the contract to be a rounding error on the wrong side of solvent.

DanielVF: That said, we should be careful to fuzz test the behavior differences. There's no perfect answer here, and so we want to make sure that erroring on the solvent side is correct.

DanielVF: This is all it takes for the contract to report insolvancy with the current code.

image

sparrowDom: a different approach has been made where the credits are read before the change and after from the OETH contract. This way the error in WOETH calculated internal credits and the one in OETH contract shouldn't diverge and cause potential conversion rate inaccuracies: https://github.com/OriginProtocol/origin-dollar/pull/2106/commits/c005824d9062cef9705fb02aaa2114ee64241362

PR #2112, opened 2024-06-27
Exit validator check

sparrowDom:
Add a check that the minimal activation period elapsed before being allowed to exit a validator


naddison36 (approved):
Looks good

PR #2114, opened 2024-06-28
Trim down defender action code build size

sparrowDom:
This PR trims down the defender action code. Unfortunately, we can not just mark the `hardhat` as an external rollup dependency and still use it in the code because the resulting compiled action code will still try to import it and it will throw an error. 

For that reason this PR separates dependency files in such a way that no defender action task imports any file that would also import hardhat. This reduces the build size to 2Mbs. 

There are other fixes and improvements in the PR as a result of testing the functionality in production.


contracts/scripts/defender-actions/rollup.config.cjs
```diff
@@ -53,9 +52,9 @@ module.exports = [
   },
   {
     ...commonConfig,
-    input: "doAccounting.js",
+    input: "registerValidators.js",
     output: {
-      file: "dist/doAccounting/index.js",
+      file: "dist/registerValidators/index.js",
```
sparrowDom, line 57: switched order since `visualize` plugin overwrites exports, and this way `registerValidator` action is the one that has dependency graphs available

naddison36 (approved):

contracts/utils/resolvers.js
```diff
@@ -16,7 +17,7 @@ const resolveAsset = async (symbol) => {
 
   // Not using helpers here as they import hardhat which won't work for Hardhat tasks
   if (process.env.FORK === "true" || hre.network.name != "hardhat") {
-    const { chainId } = await hre.ethers.provider.getNetwork();
+    const { chainId } = await ethers.getDefaultProvider().getNetwork();
```
naddison36: It turns out this will only work for mainnet :-(

https://github.com/NomicFoundation/hardhat/issues/2424#issuecomment-1066816392 

I'll see if I can get something that will work with Holesky that doesn't use Hardhat

naddison36: I've moved `registerValidators` and `stakeValidators` into `utils` without a Hardhat dependency

naddison36 (dismissed):

naddison36 (approved):

PR #2119, opened 2024-07-03
WOETH: Withdraw OETH in surplus.

clement-ux:
Allow the governor to withdraw the OETH that has been sent in surplus to the wrapper.

## Code Change Checklist

- [ ]  Unit tests *if needed

Internal review:

## Deploy checklist

```
```

### DIFF
--- a/contracts/contracts/token/WOETH.sol
+++ b/contracts/contracts/token/WOETH.sol
@@ -82,7 +82,11 @@ contract WOETH is ERC4626, Governable, Initializable {
     {
         //@dev TODO: we could implement a feature where if anyone sends OETH direclty to
         // the contract, that we can let the governor transfer the excess of the token.
-        require(asset_ != address(asset()), "Cannot collect OETH");
+        if (asset_ == address(asset())) {
+            uint256 surplus = OETH(asset()).balanceOf(address(this)) - totalAssets();
+            require(amount_ <= surplus, "Cannot collect OETH more than surplus");
+        }
+        
         IERC20(asset_).safeTransfer(governor(), amount_);
     }
 
--- a/contracts/test/token/woeth.js
+++ b/contracts/test/token/woeth.js
@@ -144,10 +144,28 @@ describe("WOETH", function () {
       await expect(woeth).to.have.a.balanceOf("0", dai);
       await expect(governor).to.have.a.balanceOf("1002", dai);
     });
-    it("should not allow a governor to collect OETH", async () => {
+    it("should allow a governor to collect less than OETH surplus", async () => {
+      await oeth.connect(josh).transfer(woeth.address, oethUnits("2"));
+      await expect(woeth).to.have.a.balanceOf("102", oeth);
+      await woeth.connect(governor).transferToken(oeth.address, oethUnits("1"));
+      await expect(woeth).to.have.a.balanceOf("101", oeth);
+      await expect(governor).to.have.a.balanceOf("1", oeth);
+      await expect(await woeth.totalAssets()).to.equal(oethUnits("100"));
+    });
+    it("should allow a governor to collect exact OETH surplus", async () => {
+      await oeth.connect(josh).transfer(woeth.address, oethUnits("2"));
+      await expect(woeth).to.have.a.balanceOf("102", oeth);
+      await woeth.connect(governor).transferToken(oeth.address, oethUnits("2"));
+      await expect(woeth).to.have.a.balanceOf("100", oeth);
+      await expect(governor).to.have.a.balanceOf("2", oeth);
+      await expect(await woeth.totalAssets()).to.equal(oethUnits("100"));
+    });
+    it("should not a allow governor to collect more than OETH surplus", async () => {
+      await oeth.connect(josh).transfer(woeth.address, oethUnits("2"));
+      await expect(woeth).to.have.a.balanceOf("102", oeth);
       await expect(
-        woeth.connect(governor).transferToken(oeth.address, oethUnits("2"))
-      ).to.be.revertedWith("Cannot collect OETH");
+        woeth.connect(governor).transferToken(oeth.address, oethUnits("3"))
+      ).to.be.revertedWith("Cannot collect OETH more than surplus");
     });
     it("should not allow a non governor to recover tokens ", async () => {
       await expect(


contracts/contracts/token/WOETH.sol
```diff
@@ -80,9 +80,11 @@ contract WOETH is ERC4626, Governable, Initializable {
         external
         onlyGovernor
     {
-        //@dev TODO: we could implement a feature where if anyone sends OETH direclty to
-        // the contract, that we can let the governor transfer the excess of the token.
-        require(asset_ != address(asset()), "Cannot collect OETH");
+        if (asset_ == address(asset())) {
+            uint256 surplus = OETH(asset()).balanceOf(address(this)) - totalAssets();
```
shahthepro, line 84: On a quick look, I'm a bit lost here. OETH is a rebasing token. wOETH's appreciation comes from the OETH yields (which are distributed during rebases with increasing balances). Won't this forgo all unrealized yields as well?

shahthepro: Looks like it has an internal log of oethCredits (updating during wrap and unwrap) but let me check

contracts/test/token/woeth.js
```diff
@@ -105,7 +106,18 @@ describe("WOETH", function () {
       await expect(woeth).to.have.approxBalanceOf("150", oeth);
       await expect(woeth).to.have.a.totalSupply("50");
     });
-
+    it("should decrease with an OETH negative rebase", async () => {
```
sparrowDom, line 109: We [only rebase upwards](https://github.com/OriginProtocol/origin-dollar/blob/master/contracts/contracts/vault/VaultCore.sol#L389-L391). If the protocol incurs a loss the rebases are "paused" until enough capital is earned for a next positive rebase

contracts/contracts/token/WOETH.sol
```diff
@@ -80,9 +80,11 @@ contract WOETH is ERC4626, Governable, Initializable {
...
+            require(amount_ <= surplus, "Cannot collect OETH more than surplus");
```
sparrowDom, line 85: This is really nitpicky but Solidity is much more effective in handling strings when [they are 31 characters long or less ](https://ethereum.stackexchange.com/questions/107282/storage-and-memory-layout-of-strings). It uses 1 slot instead of 3. 

If you don't mind can you rephrase the error message to a 31 or less char long one?

contracts/contracts/token/WOETH.sol
```diff
@@ -95,7 +97,8 @@ contract WOETH is ERC4626, Governable, Initializable {
      * @return amount of OETH credits the OETH amount corresponds to
      */
     function _creditsPerAsset(uint256 oethAmount)
-        internal
+        internal 
+        view
```
sparrowDom, line 101: nice catch 👍

contracts/contracts/token/WOETH.sol
sparrowDom: With [WOETH ignore donations PR](https://github.com/OriginProtocol/origin-dollar/pull/2106/files#diff-8bc11d78a06b7abf6bfae9f732a591fbd71cac02a935b1bc3f57ee04a22d8f41R119) WOETH keeps the balance of the internal credits (of the OETH token) that change on mint, deposit, redeem, withdraw. 

I think this approach is mathematically correct since the `surplus` figures out the difference between WOETH internal credits amount and OETH internal credits (for WOETH contract) multiplied by the credits per token - in other words operating with OETH token amounts. 

Surplus could also be calculated using another approach by finding the difference of internal credits tracked by WOETH and OETH and then multiplying it by the current credits per token ( to arrive to the OETH token surplus amount). 

I think in either of the approaches the un-realized rebase yields aren't left on the table, since as long as we do the correct math on the underlying credits per token. The rebase will take care of correct yield distribution.

sparrowDom: Still a spot on comment @shahthepro 🙏  These are the sort of checks we need to make to be sure the math works correctly.

We could add a test for this @clement-ux, where: 
- WOETH has initial supply of 100 OETH
- 2 OETH are transferred to the WOETH contract
- OETH rebases (say increases balance of OETH on WOETH contract by x2)
- governor should be able to transfer out 4 OETH and no more

sparrowDom (commented):
Left a couple of comments, some minor changes and this PR is good to go. 

Great job with it @clement-ux  🙏

contracts/contracts/token/WOETH.sol
clement-ux: Oh yes sure, nice catch!

contracts/contracts/token/WOETH.sol
shahthepro: Thanks for the detailed explanation. Will check both PRs today to get more clarity on this.

Yeah, totally agree that we should be having unit and fork tests with different scenarios involving donations and rebases.

clement-ux: I've added the test that @sparrowDom suggested in this [commit](https://github.com/OriginProtocol/origin-dollar/pull/2119/commits/e192d5175c592645ddf40e0a070f0b0d072aa877) and it passes.

clement-ux: > Surplus could also be calculated using another approach by finding the difference of internal credits tracked by WOETH and OETH and then multiplying it by the current credits per token ( to arrive to the OETH token surplus amount).

Do you mean dividing instead of multiplying maybe? 

An implementation could look something like this: 
``` solidity
(uint256 creditsBalanceHighres, uint256 creditsPerTokenHighres, ) = OETH(asset()).creditsBalanceOfHighres(address(this));
uint256 surplus = (creditsBalanceHighres - oethCreditsHighres).divPrecisely(creditsPerTokenHighres);
```

This could consume less gas but can lead to rounding issues IMO.

contracts/test/token/woeth.js
clement-ux: I was thinking of adding it so that the wrapper would continue to work even if the OETH implementation changed and allowed negative rebasing. But it's true that if the implementation changes, there will probably be stuff to modify in the wrapper anyway, so no need to create tests for a scenario that isn't at the moment ahah.

I'm removing it!

sparrowDom (approved):
LGTM

clement-ux:
Not relevant anymore due to new WOETH implementation.

PR #2140, opened 2024-07-17
Reduce delay period to 10min.

clement-ux:
Reduce the time between `requestWithdraw()` and `claimWithdraw()` from 30min to 10min.


naddison36 (approved):
LGTM

PR #2141, opened 2024-07-17
Refactor OETH Vault's `_allocate()`

naddison36:
## Contract Changes

* Changed OETH Vault's overridden `_allocate()` to
  * use the OETH total supply rather than calculating the vault's total assets when calculating the vault buffer. This fixes a previous problem where the total value was over reporting when there was more outstanding withdrawal requests than WETH in the vault.
  * Simplified the calculation of the transfer amount now we aren't looping over the vault assets


shahthepro (commented):
LGTM

PR #2150, opened 2024-07-26
OETHb: Add wOETH Strategy

shahthepro:
## Quick Overview
- `updateWOETHOraclePrice` to be scheduled on Defender (or other tools) to pull yields from mainnet
- `IStrategy.checkBalance` will underreport when price is not updated, however it should never overreport
- Deposit and Withdraw should update the price first

## Code Change Checklist

- [ ]  Unit tests *if needed

Internal review:

## Deploy checklist

```
```


contracts/contracts/strategies/BridgedWOETHStrategy.sol
```diff
@@ -0,0 +1,185 @@
...
+    // TODO: Add minAmount
+    function depositBridgedWOETH(uint256 woethAmount)
+        external
+        onlyGovernorOrStrategist
+    {
+        // Transfer in all bridged wOETH tokens
+        bridgedWOETH.transferFrom(msg.sender, address(this), woethAmount);
+
+        // Figure out how much they are worth
+        uint256 oethToMint = getBridgedWOETHValue(woethAmount);
```
DanielVF, line 85: 🟠 Whenever possible, it's important to do calculations first, before doing state updates or external state updating  calls. This way the calculations themselves are not altered by what we are doing (or by any possible reentrancy that happens). So here `oethToMint` should move up ahead of all actions.

contracts/contracts/strategies/BridgedWOETHStrategy.sol
```diff
@@ -0,0 +1,185 @@
...
+
+        // There's no pToken, however, it just uses WOETH address in the event
+        emit Deposit(address(weth), address(bridgedWOETH), oethToMint);
+
+        // Mint OETHb tokens and transfer it to the caller
+        IVault(vaultAddress).mintToForStrategy(msg.sender, oethToMint);
```
DanielVF, line 91: 🟠 It's a good habit to write reentrancy / read-only reentrancy aware code ordering, even if we have other protections. For OETH/OUSD it's okay from a rebasing point of view if the vault value is too low vs the total supply, but bad if it is too high. So we when we have a pair of commands which move the value and total supply, we should make things worse first, and then restore them to balance. 

So in this deposit function, we should mint first (making total supply vs value worse), then transfer in the funds (restoring balance).

The withdraw code also should flip. First transferring tokens out (making things worse), before burning tokens (making things better).

contracts/contracts/strategies/BridgedWOETHStrategy.sol
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    function withdrawBridgedWOETH(uint256 woethAmount, address receiver)
+        external
+        onlyGovernorOrStrategist
```
DanielVF, line 96: These should both have reentrancy protection modifiers.

contracts/contracts/strategies/BridgedWOETHStrategy.sol
```diff
@@ -0,0 +1,185 @@
...
+        uint256 oethToMint = getBridgedWOETHValue(woethAmount);
+
+        // There's no pToken, however, it just uses WOETH address in the event
+        emit Deposit(address(weth), address(bridgedWOETH), oethToMint);
+
+        // Mint OETHb tokens and transfer it to the caller
+        IVault(vaultAddress).mintToForStrategy(msg.sender, oethToMint);
+    }
+
+    function withdrawBridgedWOETH(uint256 woethAmount, address receiver)
```
DanielVF, line 94: ⚪ It's odd to have the withdraw function take an explicit receiver on where to send funds too, while the deposit function (which also sends funds) does not. I'd probably lean towards removing it from the withdraw function unless there's a need for it.

contracts/contracts/strategies/BridgedWOETHStrategy.sol
```diff
@@ -0,0 +1,185 @@
...
+        // Transfer WOETH back
+        bridgedWOETH.transfer(receiver, woethAmount);
+    }
+
+    /**
+     * @notice Returns the amount of backing WETH the strategy holds
+     * @param _asset      Address of the asset
+     * @return balance    Total value of the asset in the platform
+     */
+    function checkBalance(address _asset)
```
DanielVF, line 115: The deposit and the withdraw functions would require bad behavior by both oracle AND the strategist to have a problem. This is okay.

The checkBalance problem would have a problem if there was a problem with either mainnet woeth OETH with the oracle. This is worth some thinking about.

In the past, we not had a checkBalance function that had explicit reverts. `getBridgedWOETHValue` will revert if the updated date is wrong. (This is probably required here). I wonder if we should also revert if things are out of range in some way.

contracts/contracts/strategies/BridgedWOETHStrategy.sol
```diff
@@ -0,0 +1,303 @@
...
+     *      And stores it.
+     *
+     *      NOTE: Intentionally not caching `Vault.priceProvider` here,
+     *      since doing so would mean that we also have to update this
+     *      strategy every time there's a change in oracle router.
+     *      Besides on L2, the gas is considerably cheaper than mainnet.
+     *
+     * @return Latest price from oracle
+     */
+    function _updateWOETHOraclePrice() internal returns (uint256) {
```
DanielVF, line 103: Throughout this file, when we have an external / internal pair of functions, like here, let's run the external version first, then the internal.

contracts/contracts/strategies/BridgedWOETHStrategy.sol
```diff
@@ -0,0 +1,303 @@
...
+        // WETH price per unit of bridged wOETH
+        uint256 oraclePrice = IOracle(IVault(vaultAddress).priceProvider())
+            .price(address(bridgedWOETH));
+
+        // 1 wOETH > 1 WETH, always
+        require(oraclePrice > 1 ether, "Invalid wOETH value");
+
+        uint128 oraclePrice128 = oraclePrice.toUint128();
```
DanielVF, line 111: Can we just convert  the lastOraclePrice to uint256 here, and then we don't have to do any more casting in the rest of the function, outside of a storage set? Raw casts in the middle of math scare me.

DanielVF: We can change the events to uint256s as well.

DanielVF:
🟠 In past contracts bad oracle readings could never affect user's funds. With this strategy they can, which means our oracle code is more important. Our current oracle code does this:

```solidity
(, int256 _iprice, , uint256 updatedAt, ) = AggregatorV3Interface(_feed)
    .latestRoundData();

require(
    updatedAt + maxStaleness >= block.timestamp,
    "Oracle price too old"
);

uint8 decimals = getDecimals(_feed);
uint256 _price = uint256(_iprice).scaleBy(18, decimals);
```

The raw cast here would silently convert a negative price returned by the oracle to a very large positive value. We should change this so that it uses a safe cast which should revert.

contracts/contracts/strategies/BridgedWOETHStrategy.sol
```diff
@@ -0,0 +1,303 @@
...
+
+        // Do some checks
+        if (lastOraclePrice > 0) {
+            // Make sure the value only goes up
+            require(lastOraclePrice <= oraclePrice128, "Negative wOETH yield");
```
DanielVF, line 116: ⚪️ I think this "up only" require statement would be clearer if written as `a >= b`, since that looks more up only than `b <= a`.

contracts/contracts/strategies/BridgedWOETHStrategy.sol
```diff
@@ -0,0 +1,303 @@
...
+
+            // And that it's within the bounds.
+            require(
```
DanielVF, line 119: Maybe a local variable on maxPrice? Then this require is very clear that price <= maxPrice

contracts/contracts/strategies/BridgedWOETHStrategy.sol
```diff
@@ -0,0 +1,303 @@
...
+     *      equivalent amount of bridged wOETH.
+     * @param woethAmount Amount of bridged wOETH to get back
+     */
+    function withdrawBridgedWOETH(uint256 woethAmount)
+        external
+        onlyGovernorOrStrategist
+        nonReentrant
+    {
+        // Update wOETH price
+        uint256 oraclePrice = _updateWOETHOraclePrice();
```
DanielVF, line 197: 👍 This is very cool that it intentionally breaks the funds moving in and out of it if the oracles are doing something weird. Very good.

contracts/contracts/strategies/BridgedWOETHStrategy.sol
```diff
@@ -0,0 +1,303 @@
...
+
+        // Figure out how much they are worth
+        uint256 oethToBurn = (woethAmount * oraclePrice) / 1 ether;
```
DanielVF, line 200: 🔵 This actually rounds against the protocol, in favor of the "user", since oethToBurn is being rounded down, and this then the amount that the "user" has to pay.  This would be acceptable here with the small error size and the limited amount that this rounding could affect us.

However, I think we might (or might not) want to change this function for another reason. Let's say the strategist has 1,000 superOETHb in their wallet that they want to use to withdraw and get `bridgedWOETH` back. As it is, they will have to calculate the exchange rate to request woethAmount, then leave off some for rounding. It would be more work than just "I have a 1,000 superOETHb, let's do a withdraw". For example, OUSD mints and redeems work where the quantity in the function argument is always denominated in token that the user is holding.

Now as it is written, it's better for withdrawing the full amount held in this strategy, but it feels like this would happen less often than partial withdraws.

I'm open on this one, whichever way you like.

contracts/contracts/strategies/BridgedWOETHStrategy.sol
DanielVF: Or just do whatever you want on this one. :D It's less straightforward than I thought at first.

contracts/contracts/strategies/BridgedWOETHStrategy.sol
shahthepro: That's a good point. I thought about it while writing the code, checked other strategies, we always passed the same asset addresses in both `deposit` and `withdraw` (since Vault is the one calling it). But yeah, it makes more sense to specify amount of OETHb to burn in this strategy. Besides full withdrawal shouldn't be that complicated with OETHb amount as parameter (since you can always check the oracle for the latest price and use that to compute how much you wanna burn).

shahthepro:
> The raw cast here would silently convert a negative price returned by the oracle to a very large positive value. We should change this so that it uses a safe cast which should revert.

Thanks for catching this. I was only looking at the `AbstractOracleRouter` and didn't realize I override the price method and didn't use SafeCast there. I have fixed it now

contracts/contracts/strategies/BridgedWOETHStrategy.sol
shahthepro: Agreed, have changed it

contracts/contracts/strategies/BridgedWOETHStrategy.sol
shahthepro: This is one messy function, yeah. I removed one unnecessary cast. Let me see if I can make it less messy

contracts/contracts/strategies/BridgedWOETHStrategy.sol
shahthepro: Done

contracts/contracts/strategies/BridgedWOETHStrategy.sol
shahthepro: Have changed it now

contracts/contracts/strategies/BridgedWOETHStrategy.sol
```diff
@@ -0,0 +1,301 @@
...
+
+        uint128 oraclePrice128 = oraclePrice.toUint128();
+
+        // Do some checks
+        if (lastOraclePrice > 0) {
+            // Make sure the value only goes up
+            require(oraclePrice128 >= lastOraclePrice, "Negative wOETH yield");
+
+            // (lastOraclePrice * (1 + maxPriceDiffBps))
+            uint256 maxPrice = ((lastOraclePrice * (1e4 + maxPriceDiffBps)) /
```
DanielVF, line 127: ⚪️ Super minor - extra pair of outside parenthesis here.

contracts/contracts/strategies/BridgedWOETHStrategy.sol
```diff
@@ -0,0 +1,301 @@
...
+        uint256 _amount
+    ) external override onlyVault nonReentrant {
+        require(false, "Withdrawal disabled");
+    }
+
+    /**
+     * @notice withdrawAll() function not used for this strategy
+     */
+    function withdrawAll() external override onlyVaultOrGovernor nonReentrant {
+        require(false, "Withdrawal disabled");
```
DanielVF, line 293: I think we probably want to change this to `return false`, since we tend to want withdrawAll to succeed on the vault pulling as much funds as possible (from other strategies)

DanielVF (dismissed):
# OETHb wOETH strategy


## Requirements

We want to be able to have OETHb hold a portion of its assets as bridged wOETH. However we do not want to hard code the complexity of bridging and asynchronous balance tracking that this would involve. Instead we want a simple contract that allows a team multisig account to bear all of the risk and capital cost, while the smart contract always holds the correct number of assets.

An unavoidable part of cross chain set setup with wOETH is the need to trust the oracle for the value of wOETH. We do ensure that the oracle price can only go up, and only by a certain amount, but there is no avoiding that an anomalously high oracle reading would result in a loss of funds.




#### Authentication

- [x] Never use tx.origin

#### Ethereum

#### Cryptographic code

_no crypto_

#### Gas problems

_No for loops_

#### Black magic

_No Magic_

#### Overflow

#### Proxy

#### Events

## Medium Checks

#### Rounding and casts

#### Dependencies

#### External calls

- [x] No malicious behaviors
- [ ] Oracles, one of:
  - [ ] No oracles
  - [x] Oracles can't be bent (_to the best of our knowledge_)


#### Deploy

## Strategy Specific

#### Strategy checks

- [ ] All funds can be recovered from the strategy by some combination of depositAll, withdraw, or withdrawAll().
	- [ ] 🟣 Strategist action required to remove funds, and that is okay.
	- [ ] 🟣 Strategist withdraws will leave an equal or greater amount.
- [x] Strategist cannot steal user funds

#### Downstream

- [x] We have monitoring on all backend protocol's governances
- [x] We have monitoring on a pauses in all downstream systems

#### Logic

Only one if, no loops! 

- [x] Logic is correct.

#### Attack

🟣 Rebase too large because of bad balance data. This looks like this would only happen if the oracles are wrong. The cap on max oracle change means that damaged from a single bad oracle price is capped at a certain amount. However, if an attacker had control over the oracle prices, then they could gradually raise and update this contracts value. However that's outside the threat model of this code.

🟢 Users steal principal. Only the strategist can move funds into and out of the strategy.

🟢 Strategist steal principal. The up only nature of the price oracle prevent the strategist bouncing the oracle prices around to allow theft of principal. Future redeems will always give out less or equal amounts of WETH for the same principle. Rounding is correct and against the strategist.

🟣🟠 Public steal yield. If the amount of a rebase is higher than 0.8 bps, then yield can be stolen by minting OETHb, calling update prices, then swapping out a pool. (This threshold goes up as the percentage of this strategies assets vs rebasing OETHb goes down. At 25%, it requests a 3.2 bips change, and at 10%, it is 8 bips.) Note that since we also have asynchronous 1:1 withdraws, if someone is willing to use capital, theft of yield is possible.

image


🟣 First deposit price attack. There are no checks on the first price to ever be set on the contract other than > 1. If for some reason the oracles were bad at the moment of the first deposit, we could get bad prices locked into the contract. However this is being deployed as one of the initial strategies with very little funds around to steal, plus the oracle risk at this moment is low.

#### Flavor

_Flavor is great._

contracts/contracts/strategies/BridgedWOETHStrategy.sol
shahthepro: `withdrawAll` signature doesn't have a return value. So, I'm just removing the `require` and leaving this as a blank function

contracts/contracts/strategies/BridgedWOETHStrategy.sol
shahthepro: Fixed

clement-ux (commented):
# OETHb wOETH strategy

#### Authentication

- [x] Never use tx.origin

#### Ethereum

#### Cryptographic code

No crypto

#### Gas problems

No loops

#### Black magic

No magic

#### Overflow

#### Proxy

#### Events

## Medium Checks

#### Rounding and casts
  - To ensure, we can maybe, use [mulDivDown](https://github.com/transmissions11/solmate/blob/97bdb2003b70382996a79a406813f76417b1cf90/src/utils/FixedPointMathLib.sol#L36) or mulDivUp functions? 
  - There is one that doesn't break the contract, but doesn't act in favor of the user at all. See comment on `withdrawBridgedWOETH`
  - See comment on line 198.

#### Dependencies

#### External calls

- [x] No malicious behaviors
- [ ] Oracles, one of:
  - [ ] No oracles
  - [x] Oracles can't be bent

#### Deploy

## Strategy Specific

#### Strategy checks

  - [x] Need actions from at least Strategist.

#### Downstream

- [x] We have monitoring on all backend protocol's governances
- [x] We have monitoring on a pauses in all downstream systems

## Thinking

#### Logic

  - [x] None that I can think of

#### Attack

- The `transferToken` function seems a bit "risky for nothing" I would say, even if only the governor can call it. Because it checks that the asset to transfer is the not one of the supported asset. However, the only supported asset is the `weth`. 
This means that if the governor wants, he can transfer all bridgedOETH to himself. Not that good .

- Maybe I don't get the full picture, but it seems like too many "old" functions are still open. Like:
  - setRewardTokenAddresses
  - removePToken
  - setHarvesterAddress
  - collectRewardTokens
Even if, only the `Governor` can call it, is it necessary to keep it open? This can increase the surface of attack for "nothing"?
  
#### Flavor

_Could this code be simpler?_
I don't think so, no. Really well written!

contracts/contracts/strategies/BridgedWOETHStrategy.sol
```diff
@@ -0,0 +1,301 @@
...
+        uint256 oraclePrice = IOracle(IVault(vaultAddress).priceProvider())
+            .price(address(bridgedWOETH));
+
+        // 1 wOETH > 1 WETH, always
+        require(oraclePrice > 1 ether, "Invalid wOETH value");
+
+        uint128 oraclePrice128 = oraclePrice.toUint128();
+
+        // Do some checks
+        if (lastOraclePrice > 0) {
```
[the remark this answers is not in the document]
shahthepro: That's a really good point. 

Even if we assume that the Oracle can return a wrong price, strategist is the only one that can mint. They should always use `VaultValueChecker`. So, anything off with the first oracle price would be pretty obvious before we even put any funds to it.

I'd prefer leaving the code as-is to not complicate things further. But happy to change it, if you feel strong about it

contracts/contracts/strategies/BridgedWOETHStrategy.sol
```diff
@@ -0,0 +1,301 @@
...
+        emit Deposit(address(weth), address(bridgedWOETH), oethToMint);
+
+        // Mint OETHb tokens and transfer it to the caller
+        IVault(vaultAddress).mintForStrategy(oethToMint);
+
+        // Transfer out minted OETHb
+        oethb.transfer(msg.sender, oethToMint);
+
+        // Transfer in all bridged wOETH tokens
+        bridgedWOETH.transferFrom(msg.sender, address(this), woethAmount);
```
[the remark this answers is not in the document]
shahthepro: The first version of the code did have the code in that order. But Daniel suggested the reordering to make sure it's not reentrant. Here's the relevant discussion: https://github.com/OriginProtocol/origin-dollar/pull/2150#discussion_r1704570215

contracts/contracts/strategies/BridgedWOETHStrategy.sol
```diff
@@ -0,0 +1,301 @@
...
+    function withdrawBridgedWOETH(uint256 oethToBurn)
+        external
+        onlyGovernorOrStrategist
+        nonReentrant
+    {
+        // Update wOETH price
+        uint256 oraclePrice = _updateWOETHOraclePrice();
+
+        // Figure out how much they are worth
+        uint256 woethAmount = (oethToBurn * 1 ether) / oraclePrice;
```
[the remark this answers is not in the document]
shahthepro: That's a nice catch. I guess instead of `oethToBurn > 1`, a `woethAmount > 0` would make it a bit more cleaner. What do you think?

shahthepro:
> * The `transferToken` function seems a bit "risky for nothing" I would say, even if only the governor can call it. Because it checks that the asset to transfer is the not one of the supported asset. However, the only supported asset is the `weth`.
>   This means that if the governor wants, he can transfer all bridgedOETH to himself. Not that good .

Interesting, I actually haven't thought of it in this way. I work on the assumption that if Governor were to act maliciously, we have more things to worry about. But I really like this perspective. I will add a "supported asset" check here.

> * Maybe I don't get the full picture, but it seems like too many "old" functions are still open. Like:
>   
>   * setRewardTokenAddresses
>   * removePToken
>   * setHarvesterAddress
>   * collectRewardTokens
>     Even if, only the `Governor` can call it, is it necessary to keep it open? This can increase the surface of attack for "nothing"?

- `setRewardTokenAddresses` is something that the Governor can always change. So didn't wanna deviate much from other contracts here.

- `removePToken`, the contract hardcodes `pTokens` to `new address[](0)`, also `_abstractSetPToken` reverts if Governor tries to add it later, so this would anyway revert anyway. Can remove it though

- `setHarvesterAddress` is needed because we have to set it to the strategist address for now (will update the deployment file)

- `collectRewardTokens`, Not needed for this strategy and probably would do nothing, but will just replace it with an empty function

contracts/contracts/strategies/BridgedWOETHStrategy.sol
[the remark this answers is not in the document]
clement-ux: Ok sure, no problem, let's keep it like this 👍

contracts/contracts/strategies/BridgedWOETHStrategy.sol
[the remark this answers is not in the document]
clement-ux: Ok sorry, my bad then, I didn't see like this, but it make more sense in this order you are right 👍

contracts/contracts/strategies/BridgedWOETHStrategy.sol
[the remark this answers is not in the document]
clement-ux: You are right, 100% agree, `woethAmount > 0` check is way better!

contracts/contracts/strategies/BridgedWOETHStrategy.sol
```diff
@@ -0,0 +1,301 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import { IERC20, InitializableAbstractStrategy } from "../utils/InitializableAbstractStrategy.sol";
+import { IWETH9 } from "../interfaces/IWETH9.sol";
+import { IVault } from "../interfaces/IVault.sol";
+import { AggregatorV3Interface } from "../interfaces/chainlink/AggregatorV3Interface.sol";
```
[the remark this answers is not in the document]
shahthepro: Fixed

contracts/contracts/strategies/BridgedWOETHStrategy.sol
```diff
@@ -0,0 +1,301 @@
...
+    using SafeCast for uint256;
+
+    event MaxPriceDiffBpsUpdated(uint128 oldValue, uint128 newValue);
+    event WOETHPriceUpdated(uint128 oldValue, uint128 newValue);
+
+    IWETH9 public immutable weth;
+    IERC20 public immutable bridgedWOETH;
+    IERC20 public immutable oethb;
+
+    uint256 public constant MAX_PRICE_STALENESS = 2 days;
```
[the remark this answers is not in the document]
shahthepro: Fixed

contracts/contracts/strategies/BridgedWOETHStrategy.sol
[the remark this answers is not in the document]
shahthepro: Having a consistent round up/down behaviour here sounds good. But I wonder if it's gonna cause any issues as it is in this place. 

As of now, that division truncates and doesn't round. We have `oraclePrice128 >= lastOraclePrice` just above. So, I can't really think of any case where this would fail. But I'm not sure if I'm missing anything here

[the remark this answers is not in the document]
clement-ux: All good then 👍

clement-ux (approved):
Looks good to me, perfect 👍

PR #2151, opened 2024-07-27
OETHb: Strategy Mint Whitelist

shahthepro:
## Code Change Checklist

- [ ]  Unit tests *if needed

Internal review:

## Deploy checklist

```
```


clement-ux (commented):
LGTM 👍

contracts/contracts/interfaces/IVault.sol
```diff
@@ -210,4 +216,13 @@ interface IVault {
     function setAdminImpl(address) external;
 
     function removeAsset(address _asset) external;
+
+    function addStrategyToMintWhitelist(address strategyAddr) external;
+
+    function removeStrategyFromMintWhitelist(address strategyAddr) external;
+
+    function mintWhitelistedStrategy(address strategyAddr)
```
sparrowDom, line 224: maybe a `isMintWhitelistedStrategy` might be a bit clearer?

contracts/contracts/vault/OETHBaseVaultAdmin.sol
```diff
@@ -8,5 +8,39 @@ import { VaultAdmin } from "./VaultAdmin.sol";
...
+        mintWhitelistedStrategy[strategyAddr] = true;
+
+        emit StrategyAddedToMintWhitelist(strategistAddr);
+    }
+
+    /**
+     * @notice Removes a strategy from the mint whitelist.
+     * @param strategyAddr Strategy address
+     */
+    function removeStrategyFromMintWhitelist(address strategyAddr)
```
sparrowDom, line 33: since we don't check if strategy is valid here. Might be cool to check for a non zero address

contracts/contracts/vault/OETHBaseVaultCore.sol
```diff
@@ -1,12 +1,96 @@
...
+    }
+
+    // @inheritdoc VaultCore
+    function mintForStrategy(uint256 amount) external override {
+        _mintToForStrategy(msg.sender, amount);
+    }
+
+    // Same as mintForStrategy(uint256) but transfers
+    // minted tokens to a different address
+    function mintToForStrategy(address receiver, uint256 amount) external {
```
sparrowDom, line 55: Just curious do we need this functions where strategy contracts would mint/burn the tokens and send them directly to another contract?

I am thinking if is going to be confusing and potentially hard to track the tokens if they are not sent to strategy contracts but directly to another address

sparrowDom:
Looks good, just curious if we need those extra function that can change the target account when minting/burning tokens

contracts/contracts/interfaces/IVault.sol
shahthepro: Good point, Will rename it

contracts/contracts/vault/OETHBaseVaultAdmin.sol
shahthepro: I guess that might be unnecessary. There should be no way to add a `address(0)` to the whitelist. `addStrategyToMintWhitelist` will fail when using `address(0)` since it isn't an approved strategy on Vault

contracts/contracts/vault/OETHBaseVaultCore.sol
shahthepro: I was thinking of avoiding approve calls within the contract for OETHb. If we remove these functions, that's one additional transfer for mints. And one additional approve (from the caller) and transfer for burns.

> I am thinking if is going to be confusing and potentially hard to track the tokens if they are not sent to strategy contracts but directly to another address

That's a good point, so all printed OETHb tokens have to go through the Vault? Will make that change

contracts/contracts/vault/OETHBaseVaultAdmin.sol
sparrowDom: good point

contracts/contracts/vault/OETHBaseVaultCore.sol
sparrowDom: I was thinking that `mintForStrategy` is going to be called by the strategy contract. And similarly how the function checks if `msg.sender` is the approved strategy (meaning strategy contract is the caller) I was thinking it could also send the minted tokens (and burn excess tokens) directly from/to the strategy contract as well.

contracts/contracts/interfaces/IVault.sol
shahthepro: Have renamed it

contracts/contracts/vault/OETHBaseVaultCore.sol
sparrowDom: Ah that is what you meant by another approve and transfer, I see. 

Yup agreed, might be better to keep the Vault code cleaner.

sparrowDom (approved):
LGTM

clement-ux (approved):
Looks good to me 👍

PR #2162, opened 2024-08-02
Fix checkBalance bug in new OETHVaultCore

naddison36:
## Contract changes

* `checkBalance` has the same bug as `_wethAvailable()`. If there is more outstanding withdrawal requests than WETH in the vault and strategies, the checkBalance will not return 0. It will return the balance
* Unit test added to cover this bug

This bug is not on mainnet. It's only on `master` after the OETH Withdrawal Queue changes were merged.

## Code Change Checklist

- [ ]  Unit tests *if needed

Internal review:


shahthepro (approved):
LGTM

DanielVF (approved):

PR #2204, opened 2024-08-28
Permissioned Vault Redeem & Vault Config

shahthepro:
## Code Change Checklist

- [ ]  Unit tests *if needed

Internal review:

### DIFF
--- a/contracts/contracts/vault/OETHBaseVaultCore.sol
+++ b/contracts/contracts/vault/OETHBaseVaultCore.sol
@@ -59,4 +59,22 @@ contract OETHBaseVaultCore is OETHVaultCore {
         // Burn OTokens
         oUSD.burn(msg.sender, amount);
     }
+
+    // @inheritdoc OETHVaultCore
+    function _redeem(uint256 _amount, uint256 _minimumUnitAmount)
+        internal
+        virtual
+        override
+    {
+        // Only Strategist or Governor can redeem using the Vault for now.
+        // We don't have the onlyGovernorOrStrategist modifier on VaultCore.
+        // Since we won't be using that modifier anywhere in the VaultCore as well,
+        // the check has been added inline instead of moving it to VaultStorage.
+        require(
+            msg.sender == strategistAddr || isGovernor(),
+            "Caller is not the Strategist or Governor"
+        );
+
+        super._redeem(_amount, _minimumUnitAmount);
+    }
 }
--- a/contracts/deploy/base/009_upgrade_vault.js
+++ b/contracts/deploy/base/009_upgrade_vault.js
@@ -6,7 +6,6 @@ const { utils } = require("ethers");
 module.exports = deployOnBaseWithGuardian(
   {
     deployName: "009_upgrade_vault",
-    forceSkip: true,
   },
   async ({ ethers }) => {
     const cOETHbVaultProxy = await ethers.getContract("OETHBaseVaultProxy");
@@ -14,6 +13,7 @@ module.exports = deployOnBaseWithGuardian(
       "IVault",
       cOETHbVaultProxy.address
     );
+    const cOETHbDripperProxy = await ethers.getContract("OETHBaseDripperProxy");
 
     // Deploy new implementation
     const dOETHbVaultCore = await deployWithConfirmation("OETHBaseVaultCore", [
@@ -64,6 +64,12 @@ module.exports = deployOnBaseWithGuardian(
           contract: cOETHbVault,
           signature: "setTrusteeFeeBps(uint256)",
           args: [2000], // 20%
+        },
+        {
+          // 8. Set Dripper
+          contract: cOETHbVault,
+          signature: "setDripper(address)",
+          args: [cOETHbDripperProxy.address],
         },
       ],
     };
--- a/contracts/deploy/base/009_upgrade_vault.js
+++ b/contracts/deploy/base/009_upgrade_vault.js
@@ -1,0 +1,71 @@
+const { deployOnBaseWithGuardian } = require("../../utils/deploy-l2");
+const { deployWithConfirmation } = require("../../utils/deploy");
+const addresses = require("../../utils/addresses");
+const { utils } = require("ethers");
+
+module.exports = deployOnBaseWithGuardian(
+  {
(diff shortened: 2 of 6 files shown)


contracts/contracts/vault/OETHBaseVaultCore.sol
```diff
@@ -59,4 +59,22 @@ contract OETHBaseVaultCore is OETHVaultCore {
...
+    // @inheritdoc OETHVaultCore
+    function _redeem(uint256 _amount, uint256 _minimumUnitAmount)
+        internal
+        virtual
+        override
+    {
+        // Only Strategist or Governor can redeem using the Vault for now.
+        // We don't have the onlyGovernorOrStrategist modifier on VaultCore.
+        // Since we won't be using that modifier anywhere in the VaultCore as well,
+        // the check has been added inline instead of moving it to VaultStorage.
```
sparrowDom, line 72: 👍 

Also modifiers are extremely inefficient when it comes to contract build size, since the compiler doesn't really do good optimisations. And Vault has already had problems with build sizes in the past

contracts/contracts/vault/OETHBaseVaultCore.sol
```diff
@@ -59,4 +59,22 @@ contract OETHBaseVaultCore is OETHVaultCore {
         // Burn OTokens
         oUSD.burn(msg.sender, amount);
     }
+
+    // @inheritdoc OETHVaultCore
+    function _redeem(uint256 _amount, uint256 _minimumUnitAmount)
```
sparrowDom, line 64: I like how you've added the check in the internal function rather than 2 public facing `redeem` & `redeemAll` functions. 👍

contracts/deploy/base/009_upgrade_vault.js
```diff
@@ -0,0 +1,77 @@
...
+        {
+          // 5. Max supply diff
+          contract: cOETHbVault,
+          signature: "setMaxSupplyDiff(uint256)",
+          args: [utils.parseUnits("0.03", 18)], // 0.03 OETH
+        },
+        {
+          // 6. Set trustee address
+          contract: cOETHbVault,
+          signature: "setTrusteeAddress(address)",
```
sparrowDom, line 59: we have this set already right?

contracts/deploy/base/009_upgrade_vault.js
```diff
@@ -0,0 +1,77 @@
...
+    const dOETHbVaultCore = await deployWithConfirmation("OETHBaseVaultCore", [
+      addresses.base.WETH,
+    ]);
+    const dOETHbVaultAdmin = await deployWithConfirmation("OETHBaseVaultAdmin");
+
+    return {
+      actions: [
+        {
+          // 1. Upgrade VaultCore
+          contract: cOETHbVaultProxy,
```
sparrowDom, line 28: to upgrade the vault code we only need to execute this function. Why are other governance configuration functions present?

contracts/deploy/base/009_upgrade_vault.js
```diff
@@ -0,0 +1,77 @@
...
+          signature: "upgradeTo(address)",
+          args: [dOETHbVaultCore.address],
+        },
+        {
+          // 2. Upgrade VaultAdmin
+          contract: cOETHbVault,
+          signature: "setAdminImpl(address)",
```
sparrowDom, line 35: we don't need to call this one right?

contracts/deploy/base/009_upgrade_vault.js
```diff
@@ -0,0 +1,77 @@
...
+          args: [dOETHbVaultAdmin.address],
+        },
+        {
+          // 3. Set allocate threshold
+          contract: cOETHbVault,
+          signature: "setAutoAllocateThreshold(uint256)",
+          args: [utils.parseEther("10")], // 10 OETH
```
sparrowDom, line 42: currently on the base mainnet contract this is set to 25k. No auto allocate will be happening on base Afaik?

contracts/deploy/base/009_upgrade_vault.js
shahthepro: Strategist address is different from Trustee address. On mainnet Buyback is the trustee address that receives 20% of all yields. We don't have a Buyback here. So I'm just making it the trustee address

contracts/deploy/base/009_upgrade_vault.js
shahthepro: I actually haven't configured those other params on the Vault during deploy since we haven't made any decision on those. We need these configured before going live

contracts/deploy/base/009_upgrade_vault.js
shahthepro: We actually have to. Async withdrawal code got merged after we deployed OETHb. And that VaultAdmin adds a "setDripper" function. We need to set it so that VaultCore won't break on rebase (it's set as last step on this file).

contracts/deploy/base/009_upgrade_vault.js
shahthepro: I don't think it's 25k. Thats like more than half of OETH supply. It's 10 OETH on mainnet as well. But this is probably gonna be unused since we won't have a default strategy

clement-ux (dismissed):
Looks good to me 👍

shahthepro:
Merging this in since all open comments are about the Governance action and not the code change. Will address them in later PRs if needed

contracts/deploy/base/009_upgrade_vault.js
sparrowDom: oh got it thanks

contracts/deploy/base/009_upgrade_vault.js
sparrowDom: Screenshot 2024-08-28 at 10 55 15
Screenshot 2024-08-28 at 10 55 11

Looks like it is 25k

contracts/deploy/base/009_upgrade_vault.js
sparrowDom: Aha I see changes out of scope of this PR 👍

PR #2210, opened 2024-08-29
Add Inverted Quoter for AMO fork test. 

clement-ux:
# Description

Calculating the amount of tokens to swap for rebalancing is not easy and always changes in the fork test, as the pool situation changes too. 
Using a quoter that performs a binary search to find the amount to swap, to push the price in the correct range in the AMO,  will contribute to the stability of the fork test.

## Code Change Checklist

This PR doesn't change the code, it only changes the test.


contracts/test/strategies/aerodrome-amo.base.fork-test.js
```diff
@@ -640,19 +665,31 @@ describe("ForkTest: Aerodrome AMO Strategy (Base)", function () {
...
           oethUnits("0.009")
         )
       ).to.be.revertedWith("NotEnoughWethForSwap");
     });
 
     it("Should revert when pool rebalance is off target", async () => {
+      const currentPrice = await aerodromeAmoStrategy.getPoolX96Price();
+      const priceTick0 = await aerodromeAmoStrategy.sqrtRatioX96TickLower();
+      const priceTick1 = await aerodromeAmoStrategy.sqrtRatioX96TickHigher();
+      const targetPrice = priceTick0 * 0.2 + priceTick1 * 0.8;
```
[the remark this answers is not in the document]
sparrowDom: Also could the AMO quoter get reference to the pool and figure out the current price and the direction of the swap by itself?

contracts/contracts/utils/AerodromeAMOQuoter.sol
```diff
@@ -0,0 +1,316 @@
...
+                    revertMessage: abi.decode(reason, (string))
+                });
+        }
+    }
+}
+
+contract AerodromeAMOQuoter {
+    QuoterHelper public quoterHelper;
+
+    constructor() {
```
sparrowDom, line 263: constructor should probably also contain the `amoStrategy`  as a parameter and `clPool` (underlying aerodrome slipstream pool)

sparrowDom:
I left a couple of comments, but generally I would adjust the api between the fork tests and the quoter a little. The code that is issuing the call from the fork test doesn't really need to pass any variables to the quoter. The Quoter should be constructed using the pool & strategy contract and be able to figure out the direction and the amounts to swap on its own. 

Separately from that some of the fork tests also also try to move the price in the pool close to left tick boundary or right tick boundary. Do you think you could also write a helper functions so we don't have those swap values hardcoded and needing to adjust them frequently?

contracts/contracts/utils/AerodromeAMOQuoter.sol
```diff
@@ -0,0 +1,316 @@
...
+    ) public returns (Data memory data) {
+        QuoterHelper.QuoterParams memory params = QuoterHelper.QuoterParams({
+            strategy: strategy,
+            swapWETHForOETHB: swapWETHForOETHB,
+            minAmount: minAmount,
+            maxAmount: maxAmount,
+            maxIterations: maxIterations
+        });
+        try quoterHelper.getAmountToSwapBeforeRebalance(params) {
+            revert("Previous call should only revert, it cannot succeed");
```
sparrowDom, line 290: pretty cool that this always reverts. As to not change any state in the strategy / pool contracts 👍

clement-ux:
> I left a couple of comments, but generally I would adjust the api between the fork tests and the quoter a little. The code that is issuing the call from the fork test doesn't really need to pass any variables to the quoter. The Quoter should be constructed using the pool & strategy contract and be able to figure out the direction and the amounts to swap on its own.

Ok got it, no problem it makes sense. 


> Separately from that some of the fork tests also also try to move the price in the pool close to left tick boundary or right tick boundary. Do you think you could also write a helper functions so we don't have those swap values hardcoded and needing to adjust them frequently?

I had already built a quoter that gave the quantity to be exchanged to reach a certain price, I can integrate it without any problem, sure ;)

clement-ux:
> Separately from that some of the fork tests also also try to move the price in the pool close to left tick boundary or right tick boundary. Do you think you could also write a helper functions so we don't have those swap values hardcoded and needing to adjust them frequently?

I've updated some tests with the quoter that returns the amount to swap to reach a certain price. Let me know if you want me to apply it to other tests that I might have missed. ;)

sparrowDom (commented):
Did a pass please take a look

contracts/contracts/utils/AerodromeAMOQuoter.sol
```diff
@@ -0,0 +1,464 @@
...
+        }
+
+        revert OutOfIterations(iterations);
+    }
+
+    function isWithinAllowedVariance(
+        uint160 sqrtPriceCurrentX96,
+        uint160 sqrtPriceTargetX96
+    ) public pure returns (bool) {
+        uint256 allowedVariance = (sqrtPriceTargetX96 *
```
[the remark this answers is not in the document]
clement-ux: I'm not sure I really understand your point as I don't think we are comparing the same thing at the end.

On my side I'm calculating the "variance" around the target price, while you calculate the "variance" as the difference between two ticks. That's why there is a that huge difference I will say.

contracts/contracts/utils/AerodromeAMOQuoter.sol
```diff
@@ -0,0 +1,464 @@
...
+        strategy = _strategy;
+        quoterV2 = _quoterV2;
+        clPool = strategy.clPool();
+    }
+
+    ////////////////////////////////////////////////////////////////
+    /// --- FUNCTIONS
+    ////////////////////////////////////////////////////////////////
+    /// @notice This call can only end with a revert.
+    function getAmountToSwapBeforeRebalance() public {
```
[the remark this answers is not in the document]
clement-ux: Indeed that's a great idea! I'm doing it ;)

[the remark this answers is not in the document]
clement-ux: Done, let me know what you think of it.

contracts/contracts/utils/AerodromeAMOQuoter.sol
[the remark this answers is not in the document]
sparrowDom: I see, I guess I am interested in the logic behind how the variance is calculated. In other words why the approach to calculate variance from the total x96 format number.

For example my logic behind why to specify variance in a way where the difference between to ticks represented 100% of the variance is: 
- we deposit our whole liquidity in the [-1,0] ticker
- when we calculate the WETH to OETHb share of the liquidity the 1 tick spacing (the width of the ticker) represents 100% of the liquidity range
- meaning that 5% in the price variance calculation will also represent 5% of the WETH liquidity share in the ticker range.

contracts/contracts/utils/AerodromeAMOQuoter.sol
```diff
@@ -0,0 +1,544 @@
...
+    /// @notice Use this to get the amount to swap before rebalance and
+    ///         update allowedWethShareStart and allowedWethShareEnd on AMO.
+    /// @dev This call will only revert, check the logs to get returned values.
+    /// @dev Need to perform this call while impersonating the governor of AMO.
+    /// @param overrideBottomWethShare New value for the allowedWethShareStart on AMO.
+    ///         Use type(uint256).max to keep same value.
+    /// @param overrideTopWethShare New value for the allowedWethShareEnd on AMO.
+    ///         Use type(uint256).max to keep same value.
+    /// @return data Data struct with the amount and the number of iterations
+    function quoteAmountToSwapBeforeRebalance(
```
sparrowDom, line 475: awesome stuff!

contracts/test/strategies/aerodrome-amo.base.fork-test.js
```diff
@@ -46,6 +48,12 @@ describe("ForkTest: Aerodrome AMO Strategy empty pool setup (Base)", function ()
     await oethb
       .connect(rafael)
       .approve(aeroSwapRouter.address, oethUnits("1000"));
+
+    await deployWithConfirmation("AerodromeAMOQuoter", [
```
sparrowDom, line 52: this will fit better in here: https://github.com/OriginProtocol/origin-dollar/blob/f0d0ace54b15e64ff849c9d6b0169ddfb0fe5b19/contracts/test/_fixture-base.js#L4

contracts/contracts/utils/AerodromeAMOQuoter.sol
```diff
@@ -0,0 +1,544 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.7;
+
+import { ICLPool } from "../interfaces/aerodrome/ICLPool.sol";
+import { IQuoterV2 } from "../interfaces/aerodrome/IQuoterV2.sol";
+import { IAMOStrategy } from "../interfaces/aerodrome/IAMOStrategy.sol";
+
+contract QuoterHelper {
```
sparrowDom, line 8: can you add a small comment here why the QuoterHelper is required. It might not be immediately obvious to others that try/catch in solidity requires a separate contract in order to function

contracts/contracts/utils/AerodromeAMOQuoter.sol
clement-ux: Thanks!

contracts/test/strategies/aerodrome-amo.base.fork-test.js
```diff
@@ -322,9 +393,9 @@ describe("ForkTest: Aerodrome AMO Strategy (Base)", function () {
         balanceBefore.add(oethUnits("1"))
       );
 
-      // Little to no weth should be left on the strategy contract - 10 wei is really small
+      // Little to no weth should be left on the strategy contract - 1000 wei is really small
       expect(await weth.balanceOf(aerodromeAmoStrategy.address)).to.lte(
-        BigNumber.from("10")
+        BigNumber.from("1000")
```
sparrowDom, line 398: curious did these tests start failing?

sparrowDom: Also applies to a couple below

contracts/test/strategies/aerodrome-amo.base.fork-test.js
```diff
@@ -564,12 +635,8 @@ describe("ForkTest: Aerodrome AMO Strategy (Base)", function () {
...
       await mintAndDepositToStrategy({ amount: oethUnits("5") });
 
-      // prettier-ignore
-      const tx = await rebalance(
-        oethUnits("0.00001"),
-        true, // _swapWETHs
-        oethUnits("0.000009")
-      );
+      const { value, direction } = await quoteAmountToSwapBeforeRebalance();
+      const tx = await rebalance(value, direction, 0);
```
sparrowDom, line 639: the third value should be 1-2% below the first value.

contracts/test/strategies/aerodrome-amo.base.fork-test.js
```diff
@@ -640,70 +704,68 @@ describe("ForkTest: Aerodrome AMO Strategy (Base)", function () {
...
     it("Should revert when pool rebalance is off target", async () => {
+      const { value, direction } = await quoteAmountToSwapBeforeRebalance();
+
       await expect(
-        rebalance(
-          oethUnits("0.001"),
-          true, // _swapWETH
-          oethUnits("0.0009")
-        )
+        rebalance(value.mul("200").div("100"), direction, 0)
```
sparrowDom, line 718: we can change this by fetching the swap amount by setting the price of the say tick 1 and tick -2 from the quoter

sparrowDom:
@clement-ux did another pass

contracts/test/strategies/aerodrome-amo.base.fork-test.js
clement-ux: I'm sorry, I didn't get what do you mean. You could explain it another way please?

contracts/contracts/utils/AerodromeAMOQuoter.sol
```diff
@@ -0,0 +1,555 @@
...
+                overrideBottomWethShare,
+                overrideTopWethShare
+            );
+    }
+
+    /// @notice Internal logic for quoteAmountToSwapBeforeRebalance.
+    function _quoteAmountToSwapBeforeRebalance(
+        uint256 overrideBottomWethShare,
+        uint256 overrideTopWethShare
+    ) public returns (Data memory data) {
```
sparrowDom, line 501: nit: was this supposed to be internal?

contracts/test/strategies/aerodrome-amo.base.fork-test.js
sparrowDom: Sorry, for a bad initial comment :)

So what this test is doing is trying to trigger an error where too much funds would be swapped and the result would be a price out of tick. The way it is currently done is getting amount to swap to reach an ok target balance and multiplying the value by 2. While that might work right now I am afraid it doesn't make the test super robust, since there are probably base mainnet values where where this test won't trigger a fail. 

Now ignore my initial comment. 

I think the goal should be to get values in the test that will be always out of bounds no matter what the state of the mainnet base is.
You could probably achieve that by calling: `quoteAmountToSwapBeforeRebalance(0.98e18, 1e18)` and then multiplying that value by 2. Of if you can think of a more robust method yourself might be even better? 

Ideally we would set a price out of bounds. Say they price of the pool at ticker 1 (which is out of bounds) and get the value we need to swap for to reach that ticker when rebalancing.

contracts/contracts/utils/AerodromeAMOQuoter.sol
clement-ux: Indeed, fixed!

contracts/test/strategies/aerodrome-amo.base.fork-test.js
clement-ux: Thanks!
I've applied some changes, let me know what do you think of it ;)

> You could probably achieve that by calling: quoteAmountToSwapBeforeRebalance(0.98e18, 1e18) and then multiplying that value by 2.

I didn't multiplied it by 2, I just used the value obtained, otherwise I'll have push the price outside of bounds.
I used 0.90/0.92 for share instead of 0.98/1 due to this. 
https://github.com/OriginProtocol/origin-dollar/blob/944bb283ffa4204e41b05251173a99d11ff74ed4/contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol#L295

sparrowDom: Those changes look cool thanks, good thinking on how to bring the swap liquidity to the edge of the tick and not go over. 

My previous comment would actually apply better for testing this test `Should be reverted trying to rebalance and we are not in the correct tick` that is failing on [fork tests](https://github.com/OriginProtocol/origin-dollar/actions/runs/10885062587/job/30201758561). Which should trigger when the rebalance crosses the tick boundary. 
Would you mind adding the tick boundary tests for both directions please (crossing -1 tick going into -2 one. And crossing 0 tick going into tick 1). There are also a couple of other tests [that are failing](https://github.com/OriginProtocol/origin-dollar/actions/runs/10885062587/job/30201758561). 
But I think the approach is solid now 💪

clement-ux: I've fixed the failing test (due to the new timelock mostly). 

I've preferred to push the price at tick -2 and then try to rebalance without swapping to force reverting on `OutsideExpectedTickRange` instead of `PoolRebalanceOutOfBounds` that will occur if I were modifying WethShareStart/End. 

So there are now two tests:
- when the pool is at tick -2 and we try to rebalance with `swappingAmount == 0`.
- when the pool is at tick 1 and we try to rebalance with `swappingAmount == 0`.

Let me know if this is ok for you, or if you prefer another approach 😉

contracts/test/strategies/aerodrome-amo.base.fork-test.js
```diff
@@ -376,9 +376,9 @@ describe("ForkTest: Aerodrome AMO Strategy (Base)", function () {
         balanceBefore.add(oethUnits("1"))
       );
 
-      // Little to no weth should be left on the strategy contract - 1000 wei is really small
+      // Little to no weth should be left on the strategy contract - 1000000 wei is really small
       expect(await weth.balanceOf(aerodromeAmoStrategy.address)).to.lte(
-        BigNumber.from("1000")
+        BigNumber.from("1000000")
```
sparrowDom, line 427: @clement-ux can you add comments to these tests specifying the math/logic around why these are so high?

contracts/test/strategies/aerodrome-amo.base.fork-test.js
```diff
@@ -70,17 +70,51 @@ describe("ForkTest: Aerodrome AMO Strategy empty pool setup (Base)", function ()
...
+      priceAtTickM2
     );
-    await swap({ amount: value, swapWeth: direction });
 
     await expect(
       aerodromeAmoStrategy
         .connect(strategist)
-        .rebalance(oethUnits("0"), false, oethUnits("0"))
+        .rebalance(oethUnits("0"), direction, oethUnits("0"))
+    ).to.be.revertedWith("OutsideExpectedTickRange");
```
sparrowDom, line 95: nit: you can also check for the tick value here now that the error contains it

sparrowDom:
Couple of minor comments and a merge conflict. Then I am happy to do a final pass. Good job on this one

PR #2222, opened 2024-09-09
Add Simplified Harvester

shahthepro:
## Code Change Checklist

- [ ]  Unit tests *if needed

Internal review:

## Deploy checklist

```
```


contracts/contracts/harvest/OETHBaseHarvester.sol
```diff
@@ -0,0 +1,196 @@
...
+        weth = IERC20(_weth);
+        swapRouter = ISwapRouter(_swapRouter);
+    }
+
+    /**
+     * @notice Collects AERO from AMO strategy and
+     *      sends it to the Strategist multisig.
+     *      Anyone can call it.
+     */
+    function harvest() external {
```
DanielVF, line 62: Minor thing - anyone can call harvest.

If we are trying to operate by using harvestAndSwap(), harvest() being before called would cancel the swap or remove most of the areo from being swappable. Over any given time range we'd probably be operating either in auto harvest() mode, or in manual harvestAndSwap() mode, but overlapping both calls.

It's probably not worth adding any complications around locking out harvest though.

contracts/contracts/harvest/OETHBaseHarvester.sol
```diff
@@ -0,0 +1,196 @@
...
+        if (yield > 0) {
+            // Yields can only be sent to the Vault or the Dripper.
+            // There's no address(0) check since Vault will break if there's
+            // no Dripper address set.
+            require(
+                yieldRecipient == address(vault) ||
+                    yieldRecipient == vault.dripper(),
+                "Invalid yield recipient"
+            );
+            weth.safeTransfer(yieldRecipient, yield);
```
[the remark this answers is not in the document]
shahthepro: Thanks for flagging this. Yes, it's a known issue. However, we would start using Dripper more in the near future. Right now, harvesting and depositing yields to the vault are manual operations

clement-ux, line 141: This is really nit-picking and contains a lot of "if", but still. 

If the `strategist` make a mistake and pass `address(0)` for `yieldRecipient` and if the dripper is not enable (set to `address(0)`), then the yield can be sent to `address(0)`.

Maybe it could be worth to check that `yieldRecipient` is not `address(0)`? 

_I understand that this is a very unlikely to happen._

shahthepro: So these checks are here to ensure that yields are sent only to the Vault or the Dripper and to no other address by accident. Dripper address is read from the vault.   
    
So there shouldn't be a case where either the Vault address or Dripper is set to address(0). If Vault is set to zero address, the calls to read Dripper and strategist address would fail. If Dripper is set to zero address, our rebase won't even work and would break a lot of things. 

The only way I see this breaking is if we upgrade the Vault to not revert when Dripper is a zero address and set it. I don't feel strongly about this, but happy to accommodate the change if you see it fit

clement-ux: > If Dripper is set to zero address, our rebase won't even work and would break a lot of things.

I missed that part, so indeed it is safe 👍

contracts/contracts/harvest/OETHBaseHarvester.sol
```diff
@@ -0,0 +1,196 @@
...
+        if (aeroBalance > 0) {
+            aero.safeTransfer(strategistAddr, aeroBalance);
+        }
+
+        // Computes using all balance the contract holds,
+        // not just the WETH received from swap. Use `transferToken`
+        // if there's any WETH left that needs to be taken out
+        uint256 availableWETHBalance = weth.balanceOf(address(this));
+        // Computation rounds in favor of protocol
+        uint256 fee = (availableWETHBalance * feeBps) / 10000;
```
sparrowDom, line 128: nit can be also expressed using stableMath.sol: 
```
uint256 fee = availableWETHBalance.mulTruncateScale(feeBps, 1e4)
```

contracts/contracts/harvest/OETHBaseHarvester.sol
sparrowDom: AFAIK the Vault rebases just fine when the dripper is set to zero address. I think it is a great check to add that the yieldRecipient is not a zero address

contracts/contracts/harvest/OETHBaseHarvester.sol
sparrowDom: I would argue that it is worth adding a "harvestEnabled" boolean settable by the strategist. This way we can disable the public callable harvest function when required. I know it is additional complexity - which I also don't like.

But just 1 malicious actor is required and they can disable the convenient harvestAndSwap function for us (by calling harvest()) and we need to perform swapping and distributing the fees manually. Granted there is little incentive for anyone to be doing such a thing.

sparrowDom (commented):
Left a couple of comments

contracts/contracts/harvest/OETHBaseHarvester.sol
shahthepro: I just checked, yeah, I was wrong. It's the mint that fails not rebase. Will add that check then

contracts/contracts/harvest/OETHBaseHarvester.sol
shahthepro: That's a valid concern. However I don't think calling `harvest()` before we do a `harvestAndSwap` disables us from using that function. We can always transfer AERO to that contract in the same tx as `harvestAndSwap` to use it.

I'm trying to avoid any sort of storage variables to not mess up the storage layout when we start inheriting `AbstractHarvester` in future. One alternative I can think of is:
- in `harvestAndSwap`, when there isn't enough funds in the contracts, it can do a `transferFrom(strategist, address(this))` for the difference. This means that strategist has to approve the Harvester contract to move AERO on its behalf but that should be fine I guess. 

What do you think?

shahthepro: I have went ahead made the change to `transferFrom`, let me know if you have any further comments/suggestion on this one

contracts/contracts/harvest/OETHBaseHarvester.sol
shahthepro: Should be fixed now

contracts/contracts/harvest/OETHBaseHarvester.sol
```diff
@@ -0,0 +1,206 @@
...
+        if (aeroBalance == 0) {
+            // Do nothing if there's no AERO to transfer/swap
+            return;
+        }
+
+        if (aeroToSwap > 0) {
+            if (aeroBalance < aeroToSwap) {
+                // Transfer in balance from the multisig as needed
+                // slither-disable-next-line unchecked-transfer arbitrary-send-erc20
+                aero.safeTransferFrom(
```
clement-ux, line 114: I understand why you added this, but this seems a bit over-engineering to me. I would rather completely remove the public `harvest() `and keep only the ones that have the restricted call, to prevent front-running.
 
However, as I don't know what the project requirements are and this doesn't seem to be unsafe, so I have no problem keeping it like this.

Has this been considered to create a new role, only for simple harvest, and grant this role to a OZ Relayer?

shahthepro: > Has this been considered to create a new role, only for simple harvest, and grant this role to a OZ Relayer?

Yep, that's an option. However, I didn't wanna complicate the storage layout. Sooner or later, we want the harvester to inherit from `AbstractHarvester` like on mainnet.

Thinking of it, I guess one option is to completely discard this contract and proxy when we plan to make that transition instead of trying to upgrade the implementation at that time

clement-ux: This seems a good idea indeed!

contracts/contracts/harvest/OETHBaseHarvester.sol
shahthepro: I just made the `harvest()` method permissioned in addition to the above change. This concern should no longer be there

clement-ux (dismissed):
Permissioned `harvest()` looks way safer like this.
LGTM!

sparrowDom (approved):

contracts/contracts/harvest/OETHBaseHarvester.sol
```diff
@@ -0,0 +1,234 @@
...
+import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+
+import { IVault } from "../interfaces/IVault.sol";
+import { IStrategy } from "../interfaces/IStrategy.sol";
+import { ISwapRouter } from "../interfaces/aerodrome/ISwapRouter.sol";
+
+contract OETHBaseHarvester is Governable {
+    using SafeERC20 for IERC20;
+
+    event YieldSent(address recipient, uint256 yield, uint256 fee);
```
DanielVF, line 16: Events should be together, after storage and before modifiers.

contracts/contracts/harvest/OETHBaseHarvester.sol
```diff
@@ -0,0 +1,234 @@
...
+    function harvestAndSwap(
+        uint256 aeroToSwap,
+        uint256 minWETHExpected,
+        uint256 feeBps,
+        bool sendYieldToDripper
+    ) external onlyGovernorOrStrategist {
+        address strategistAddr = vault.strategistAddr();
+        require(strategistAddr != address(0), "Guardian address not set");
+
+        // Yields can only be sent to the Dripper.
```
DanielVF, line 128: Comment could be clearer

contracts/contracts/harvest/OETHBaseHarvester.sol
```diff
@@ -0,0 +1,234 @@
...
+        address yieldRecipient = sendYieldToDripper
+            ? vault.dripper()
+            : strategistAddr;
+        require(yieldRecipient != address(0), "Dripper address not set");
```
DanielVF, line 131: Error message might be "Yield Recipient" not set.

contracts/contracts/harvest/OETHBaseHarvester.sol
shahthepro: Fixed

contracts/contracts/harvest/OETHBaseHarvester.sol
```diff
@@ -0,0 +1,234 @@
...
+
+        // Computes using all balance the contract holds,
+        // not just the WETH received from swap. Use `transferToken`
+        // if there's any WETH left that needs to be taken out
+        uint256 availableWETHBalance = weth.balanceOf(address(this));
+        // Computation rounds in favor of protocol
+        uint256 fee = (availableWETHBalance * feeBps) / 10000;
+        uint256 yield = availableWETHBalance - fee;
+
+        // Transfer yield to Dripper if any
```
DanielVF, line 174: "Transfer yield, if any"?

contracts/contracts/harvest/OETHBaseHarvester.sol
shahthepro: Fixed it

contracts/contracts/harvest/OETHBaseHarvester.sol
shahthepro: Have updated it

clement-ux (dismissed):
LGTM 👍

sparrowDom (approved):
LGTM

PR #2248, opened 2024-09-21
Add FixedRateDripper

shahthepro:
## Changes Overview
- Rename `perBlock` to `perSecond` in the Dripper contract
- Add `FixedRateDripper` that inherits from `Dripper` but strategist sets the drip rate and it doesn't change with `collect`

## Code Change Checklist

- [ ]  Unit tests *if needed

Internal review:

### DIFF
--- a/contracts/contracts/harvest/FixedRateDripper.sol
+++ b/contracts/contracts/harvest/FixedRateDripper.sol
@@ -60,6 +60,13 @@ contract FixedRateDripper is Dripper {
     function setDripRate(uint192 _perSecond) external onlyGovernorOrStrategist {
         emit DripRateUpdated(_perSecond, drip.perSecond);
 
+        /**
+         * Note: It's important to call `_collect` before updating
+         * the drip rate especially on a new proxy contract.
+         * When `lastCollect` is not set/initialized, the elapsed
+         * time would be calculated as `block.number` seconds,
+         * resulting in a huge yield, if `collect` isn't called first.
+         */
         // Collect at existing rate
         _collect();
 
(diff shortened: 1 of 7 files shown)


DanielVF (dismissed):

clement-ux (dismissed):

contracts/contracts/harvest/FixedRateDripper.sol
```diff
@@ -0,0 +1,69 @@
...
+    /**
+     * @dev Sets the drip rate. Callable by Strategist or Governor.
+     *      Can be set to zero to stop dripper.
+     * @param _perSecond Rate of WETH to drip per second
+     */
+    function setDripRate(uint192 _perSecond) external onlyGovernorOrStrategist {
+        emit DripRateUpdated(_perSecond, drip.perSecond);
+
+        // Collect at existing rate
+        _collect();
```
sparrowDom, line 64: Maybe add a comment here as it is super important that the `_collect` be called before the `drip.perSecond` is set. And the reason is: 
- the old harvester would not have the drip.perSecond populated by an external function. In order for the dripper to be "initialised" the first call to `collect` would drip no funds since drip's `lastCollect` and `perSecond` are both zero. Meaning the available funds calculation `uint256 allowed = elapsed * _drip.perSecond` would have a very big `elapsed` number and 0 for `perSecond` number returning 0 on the first call. And initializing the drip storage value to recent `lastCollect` and some value for `perSecond`
- the new harvester directly initializes the `drip.perSecond` and if it were to do that before the `_collect()` call is made the `uint256 allowed = elapsed * _drip.perSecond` would result in a very big number. Meaning that if we were to deploy the dripper and first send funds to it, and then call the `setDripRate` all of the funds would be immediately sent to the Vault. 
- but the function correctly calls `_collect` ensuring that the `lastCollect` is not 0 but a currentBlock number. Preventing the described potential problem.

sparrowDom (approved):
Great job LGTM. Would just add one comment to the code.

contracts/contracts/harvest/FixedRateDripper.sol
shahthepro: Yep, have added the comment. Thanks

PR #2255, opened 2024-09-24
Fix issues with fork testing

sparrowDom:
This change addresses the outstanding issues with AMO fork testing: 
- When searching for the correct amount to swap `BINARY_MAX_AMOUNT_FOR_PUSH_PRICE` is increased to 5m ETH so that we can cross the [-1,0] tick liquidity in both directions
- the function with signature `getAmountToSwapToReachPrice(uint160 sqrtPriceTargetX96, uint256 maxAmount)` is removed since it serves as a false positive since it returns without reaching the required target price
- `low == high` condition is removed as a successful Quoter search price condition since it doesn't indicate that the successful price has been reached, rather serves as a false positive returning without an error when the appropriate swapAmount hasn't been calculated
- coverage tests were failing because `solidity-coverage` plugin doesn't decode error names when there are custom errors thrown. Interestingly same tests pass when not ran in coverage mode. For that reason a custom chai assertion function `revertedWithCustomError` is added that calculates the hash of the error and works in fork test mode as well as the coverage fork test mode
- Aerodrome Quoter failed to reach ticks outside the [-1, 0] default tick (e.g. tick -2 or tick 1) when those ticks didn't have any liquidity in them. As is expected with concentrated liquidity pools since if there is no liquidity in there no swap amount is going to move the pool's active price into that tick. For that reason a `depositLiquidityToPool` function is added to tests that adds liquidity to those ticks in tests that require it.
- hardhat snapshot and revert are used when tests temporarily "mess up" the state on chain and want it reverted after


contracts/contracts/utils/AerodromeAMOQuoter.sol
```diff
@@ -384,10 +381,13 @@ contract QuoterHelper {
...
 
             if (
-                low == high ||
                 isWithinAllowedVariance(sqrtPriceX96After, sqrtPriceTargetX96)
             ) {
                 return (mid, iterations, swapWETHForOETHB, sqrtPriceX96After);
+            } else if (low == high) {
+                // target swap amount not found.
+                // try increasing BINARY_MAX_AMOUNT_FOR_PUSH_PRICE
+                revert("SwapAmountNotFound");
```
clement-ux, line 390: Indeed, nice catch! I now understand why I sometimes had false positives. I thought the issue came from somewhere else, thank you for pointing that out!

contracts/contracts/utils/AerodromeAMOQuoter.sol
```diff
@@ -565,32 +565,6 @@ contract AerodromeAMOQuoter {
...
-            );
-
-        emit ValueFoundBis(
-            amount,
-            iterations,
-            swapWETHForOETHB,
-            sqrtPriceAfterX96
-        );
-    }
-
```
clement-ux, line 593: Way better without it, 100% agree.

contracts/test/helpers.js
```diff
@@ -151,6 +152,29 @@ chai.Assertion.addMethod("emittedEvent", async function (eventName, args) {
...
+          `Expected error message with signature ${errorSignature} but another was thrown.`
+        );
+    }
+
+    if (txSucceeded) {
+      chai.expect.fail(`Expected ${errorSignature} error but none was thrown`);
+    }
+  }
+);
+
```
clement-ux, line 177: 🔥🔥🔥
It is super useful and amazing to have it now for testing! 
I remember talking about this recently and that we need to upgrade to ether v6 to have it natively on chai. But you built it, that's amazing !!

clement-ux (commented):
This is truly amazing! Thanks a lot for your work 🙏

However, I've tried to run the test and some of them are failing again, I'll try to find why.

contracts/test/helpers.js
sparrowDom: Yeah I've built sort of a lite version of it :) With new chai we would probably gain support to also test the values of the error parameters. We could decode that out of the error message, but might be out of scope for this PR

contracts/test/strategies/aerodrome-amo.base.fork-test.js
```diff
@@ -50,6 +56,56 @@ describe("ForkTest: Aerodrome AMO Strategy empty pool setup (Base)", function ()
...
+      .connect(rafael)
+      .approve(aeroNftManager.address, oethUnits("1000000000"));
+
+    let blockTimestamp = (await hre.ethers.provider.getBlock("latest"))
+      .timestamp;
+
+    await weth.connect(rafael).approve(oethbVault.address, oethUnits("200"));
+    await oethbVault
+      .connect(rafael)
+      .mint(weth.address, oethUnits("200"), oethUnits("19.999"));
```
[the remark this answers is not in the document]
sparrowDom: a nice catch thanks!

sparrowDom:
@clement-ux I am merging this in for now since test pass on Base for me locally and on CI. Let me know if I can help with issues you have locally

PR #2266, opened 2024-10-03
AMO automatic deposits

sparrowDom:
- configures Vault to have Aerodrome AMO as the default WETH strategy
- `vaultBuffer` set to 100% so that automatic deposits while configured are for now paused
- when unpaused OETHb mints deposit the WETH to Aerodrome AMO strategy with behaviour:
  - if pool within allowed WETH share WETH gets deposited to underlying Aerodrome pool respecting the current WETH to OETHb split shares
  - if pool is outside allowed WETH share the deposited tokens stay on the Aerodrome AMO strategy contract
- WETH allowances are reset and only required allowances granted to `swapRouter` and `positionManager` before `add liquidity` / `swap` actions
- custom error utils detector also checks for error name aside from error hash as sometimes hardhat returns the error by name (haven't gotten to the bottom of it why that is so)

## Code Change Checklist

- [x]  The contract code is complete
- [x]  Executable deployment file
- [x]  Fork tests that test after the deployment file runs
- [ ]  Unit tests *if needed
- [x]  The owner has done a [full checklist review](https://github.com/OriginProtocol/security/blob/master/templates/Contract-Code-Review.md) of the code + tests

Internal review:

## Deploy checklist

```
```

### DIFF
--- a/contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
+++ b/contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
@@ -555,8 +555,7 @@ contract AerodromeAMOStrategy is InitializableAbstractStrategy {
 
         // approve the specific amount of WETH required
         if (_swapWeth) {
-            IERC20(WETH).safeApprove(address(swapRouter), 0);
-            IERC20(WETH).safeApprove(address(swapRouter), _amountToSwap);
+            IERC20(WETH).approve(address(swapRouter), _amountToSwap);
         }
 
         // Swap it
@@ -631,8 +630,7 @@ contract AerodromeAMOStrategy is InitializableAbstractStrategy {
         }
 
         // approve the specific amount of WETH required
-        IERC20(WETH).safeApprove(address(positionManager), 0);
-        IERC20(WETH).safeApprove(address(positionManager), _wethBalance);
+        IERC20(WETH).approve(address(positionManager), _wethBalance);
 
         uint256 _wethAmountSupplied;
         uint256 _oethbAmountSupplied;
@@ -907,8 +905,8 @@ contract AerodromeAMOStrategy is InitializableAbstractStrategy {
          * un-approving WETH to the swapRouter & positionManager and only approving
          * the required amount before a transaction
          */
-        IERC20(WETH).safeApprove(address(swapRouter), 0);
-        IERC20(WETH).safeApprove(address(positionManager), 0);
+        IERC20(WETH).approve(address(swapRouter), 0);
+        IERC20(WETH).approve(address(positionManager), 0);
     }
 
     /***************************************
--- a/contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
+++ b/contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
@@ -894,11 +894,9 @@ contract AerodromeAMOStrategy is InitializableAbstractStrategy {
         nonReentrant
     {
         // to add liquidity to the clPool
-        IERC20(OETHb).safeApprove(address(positionManager), 0);
-        IERC20(OETHb).safeApprove(address(positionManager), type(uint256).max);
+        IERC20(OETHb).approve(address(positionManager), type(uint256).max);
         // to be able to rebalance using the swapRouter
-        IERC20(OETHb).safeApprove(address(swapRouter), 0);
-        IERC20(OETHb).safeApprove(address(swapRouter), type(uint256).max);
+        IERC20(OETHb).approve(address(swapRouter), type(uint256).max);
 
         /* the behaviour of this strategy has slightly changed and WETH could be
          * present on the contract between the transactions. For that reason we are
--- a/contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
+++ b/contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
@@ -366,6 +366,13 @@ contract AerodromeAMOStrategy is InitializableAbstractStrategy {
         require(_asset == WETH, "Unsupported asset");
         require(_amount > 0, "Must deposit something");
         emit Deposit(_asset, address(0), _amount);
+
+        // if the pool price is not within the expected interval leave the WETH on the contract
+        // as to not break the mints
+        if (_checkForExpectedPoolPrice(true)) {
+            // deposit funds into the underlying pool
+            _rebalance(0, false, 0);
+        }
     }
 
     /**
@@ -400,6 +407,14 @@ contract AerodromeAMOStrategy is InitializableAbstractStrategy {
         bool _swapWeth,
         uint256 _minTokenReceived
     ) external nonReentrant onlyGovernorOrStrategist {
+        _rebalance(_amountToSwap, _swapWeth, _minTokenReceived);
+    }
+
+    function _rebalance(
+        uint256 _amountToSwap,
+        bool _swapWeth,
+        uint256 _minTokenReceived
+    ) internal {
         /**
          * Would be nice to check if there is any total liquidity in the pool before performing this swap
          * but there is no easy way to do that in UniswapV3:
@@ -422,12 +437,13 @@ contract AerodromeAMOStrategy is InitializableAbstractStrategy {
         }
         // calling check liquidity early so we don't get unexpected errors when adding liquidity
         // in the later stages of this function
-        _checkForExpectedPoolPrice();
+        _checkForExpectedPoolPrice(false);
 
         _addLiquidity();
+
         // this call shouldn't be necessary, since adding liquidity shouldn't affect the active
         // trading price. It is a defensive programming measure.
-        _checkForExpectedPoolPrice();
+        _checkForExpectedPoolPrice(false);
 
         // revert if protocol insolvent
         _solvencyAssert();
@@ -534,6 +550,12 @@ contract AerodromeAMOStrategy is InitializableAbstractStrategy {
             IVault(vaultAddress).mintForStrategy(mintForSwap);
         }
 
+        // approve the specific amount of WETH required
+        if (_swapWeth) {
+            IERC20(WETH).safeApprove(address(swapRouter), 0);
+            IERC20(WETH).safeApprove(address(swapRouter), _amountToSwap);
+        }
+
         // Swap it
         swapRouter.exactInputSingle(
             // sqrtPriceLimitX96 is just a rough sanity check that we are within 0 -> 1 tick
@@ -605,6 +627,10 @@ contract AerodromeAMOStrategy is InitializableAbstractStrategy {
             );
         }
 
+        // approve the specific amount of WETH required
+        IERC20(WETH).safeApprove(address(positionManager), 0);
+        IERC20(WETH).safeApprove(address(positionManager), _wethBalance);
+
         uint256 _wethAmountSupplied;
         uint256 _oethbAmountSupplied;
         if (tokenId == 0) {
@@ -673,8 +699,11 @@ contract AerodromeAMOStrategy is InitializableAbstractStrategy {
      *      parameters.
      *      This function works whether the strategy contract has liquidity
      *      position in the pool or not.
+     * @param dryRun  when this property is set to true the function doesn't throw
+     *        an exception but returns a bool. `True` means price within expected
+     *        range and `false` means outside that range.
      */
-    function _checkForExpectedPoolPrice() internal {
+    function _checkForExpectedPoolPrice(bool dryRun) internal returns (bool) {
         require(
             allowedWethShareStart != 0 && allowedWethShareEnd != 0,
             "Weth share interval not set"
@@ -692,6 +721,10 @@ contract AerodromeAMOStrategy is InitializableAbstractStrategy {
             _currentPrice <= sqrtRatioX96TickLower ||
             _currentPrice >= sqrtRatioX96TickHigher
         ) {
+            if (dryRun) {
+                return false;
+            }
+
             revert OutsideExpectedTickRange(getCurrentTradingTick());
         }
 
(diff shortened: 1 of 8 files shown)


contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
```diff
@@ -719,12 +752,21 @@ contract AerodromeAMOStrategy is InitializableAbstractStrategy {
...
                 allowedWethShareStart,
                 allowedWethShareEnd
             );
         }
+
+        // return before emitting the event is intentional
+        if (dryRun) {
+            return true;
+        }
         emit PoolRebalanced(_wethSharePct);
```
shahthepro, line 770: Should we rename this event? It's emitted even when we don't do a rebalance.
I'm also starting to think if we really need this event here. It looks like it serves no purpose, this function doesn't even change state. We get the `_wethSharePct` when the tx reverts from the custom error above. We are not using this at all anywhere

contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
```diff
@@ -719,12 +752,21 @@ contract AerodromeAMOStrategy is InitializableAbstractStrategy {
...
             revert PoolRebalanceOutOfBounds(
                 _wethSharePct,
                 allowedWethShareStart,
                 allowedWethShareEnd
             );
         }
+
+        // return before emitting the event is intentional
+        if (dryRun) {
+            return true;
```
shahthepro, line 768: I'd get rid of the `if` statement, and let it always return `true` when it's within the bounds

contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
```diff
@@ -605,6 +627,10 @@ contract AerodromeAMOStrategy is InitializableAbstractStrategy {
             );
         }
 
+        // approve the specific amount of WETH required
+        IERC20(WETH).safeApprove(address(positionManager), 0);
+        IERC20(WETH).safeApprove(address(positionManager), _wethBalance);
```
shahthepro, line 632: Probably better to just do an `approve(wethBalance)` here. If it gets used in the swap (which it should always do), we don't have to reset it everytime

contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
```diff
@@ -534,6 +550,12 @@ contract AerodromeAMOStrategy is InitializableAbstractStrategy {
             IVault(vaultAddress).mintForStrategy(mintForSwap);
         }
 
+        // approve the specific amount of WETH required
+        if (_swapWeth) {
+            IERC20(WETH).safeApprove(address(swapRouter), 0);
+            IERC20(WETH).safeApprove(address(swapRouter), _amountToSwap);
```
shahthepro, line 556: Same comment about resetting allowance as above

contracts/deploy/base/017_upgrade_amo.js
```diff
@@ -0,0 +1,71 @@
...
+          // 2. Reset WETH approvals to 0 on swapRouter and positionManager
+          contract: cAMOStrategy,
+          signature: "safeApproveAllTokens()",
+          args: [],
+        },
+        {
+          // 2. set auto allocate threshold to 1 - gas is cheap
+          contract: cOETHbVault,
+          signature: "setAutoAllocateThreshold(uint256)",
+          args: [utils.parseUnits("1", 18)],
```
shahthepro, line 47: nit: Might as well make it 0.1

contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
sparrowDom: I agree that approval should always be 0 after a `swap` or `mint` event I am just concerned that a malfunction in a 3rd party contract (not consuming all the WETH approved in allowance) would break the behaviour of our contract (since approves from non 0 previous approval value [will revert](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/release-v4.0/contracts/token/ERC20/utils/SafeERC20.sol#L40-L42))

contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
sparrowDom: Then it would behave the same with the `dryRun` being true or false. I didn't want the `PoolRebalanced` event to be emitted when `dryRun` equals `true`

contracts/deploy/base/017_upgrade_amo.js
sparrowDom: sure, changing to 0.1

contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
sparrowDom: It is true that this event is emitted regardless of the function changing the active trading price of the pool (WETH share percentage in other words) or not. Meaning when we "rebalance" the WETH share from 14% to 16% the event will be emitted and I think it is a useful information to have.

contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
shahthepro: I feel like we should probably start by avoiding `safeApprove` usage. It has been marked as deprecated for a while now. [`approve` of Base WETH](https://basescan.org/address/0x4200000000000000000000000000000000000006#code) doesn't have that constraint. And directly using `approve` will make our code a bit cleaner

contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
shahthepro: I don't have strong opinions about this but it still out of place since it's outside `_rebalance`. Not a big deal though

shahthepro: I'm also wondering why can't we have a `getWETHShare` public function? That'll help us get the current WETH share in the active tick and would be more apt than an event since we can't even tracking this event in monitoring (it'd only emit when there's a rebalance not when there's a swap or something)

contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
```diff
@@ -534,6 +550,12 @@ contract AerodromeAMOStrategy is InitializableAbstractStrategy {
             IVault(vaultAddress).mintForStrategy(mintForSwap);
         }
 
+        // approve the specific amount of WETH required
+        if (_swapWeth) {
+            IERC20(WETH).safeApprove(address(swapRouter), 0);
```
DanielVF, line 555: WETH is a sane, compliant ERC20 token. Here's the approval code:

```
    function approve(address guy, uint wad) public returns (bool) {
        allowance[msg.sender][guy] = wad;
        emit Approval(msg.sender, guy, wad);
        return true;
    }
```

- We don't need to do the pre approval to zero, we can just set the approval we want
- We don't need to use safeApprove, WETH will always return a boolean true.

contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
DanielVF: Yes, we shouldn't be using safeApprove. My bad on not catching it the first code review.

contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
```diff
@@ -673,8 +699,11 @@ contract AerodromeAMOStrategy is InitializableAbstractStrategy {
      *      parameters.
      *      This function works whether the strategy contract has liquidity
      *      position in the pool or not.
+     * @param dryRun  when this property is set to true the function doesn't throw
+     *        an exception but returns a bool. `True` means price within expected
+     *        range and `false` means outside that range.
      */
-    function _checkForExpectedPoolPrice() internal {
+    function _checkForExpectedPoolPrice(bool dryRun) internal returns (bool) {
```
DanielVF, line 706: ⚪ I think this boolean would be better flipped. I think it's more expected when a function flag enables features. True = bigger thing happens. In this case reverting is a much bigger and more impactful action than returning a value.

contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
DanielVF: Agree on not emitting the event.

contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
sparrowDom: Yes good point. And aside from what you've already pointed out `throwException` is clearer parameter name than `dryRun`

contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
sparrowDom: I think you have a good point. I've refactored the WETH share into a separate function and emitted the event in the `_rebalance` function which is a more appropriate place for it: https://github.com/OriginProtocol/origin-dollar/pull/2266/commits/c7346e63a48e9658dc1973327ff93e3b7f959998

contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
sparrowDom: Moved the event emission into the `_rebalance` function. The rebalance event now more appropriately only gets emitted once now: https://github.com/OriginProtocol/origin-dollar/pull/2266/commits/c7346e63a48e9658dc1973327ff93e3b7f959998

contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
sparrowDom: done here: https://github.com/OriginProtocol/origin-dollar/pull/2266/commits/c7346e63a48e9658dc1973327ff93e3b7f959998

contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
sparrowDom: thanks @shahthepro and @DanielVF! Corrected: https://github.com/OriginProtocol/origin-dollar/pull/2266/commits/1321f22abb8223a30288c5d2a6549d4352ec0a69

contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
sparrowDom: thanks addressed here: https://github.com/OriginProtocol/origin-dollar/pull/2266/commits/1321f22abb8223a30288c5d2a6549d4352ec0a69

contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
sparrowDom: thanks fixed here: https://github.com/OriginProtocol/origin-dollar/pull/2266/commits/1321f22abb8223a30288c5d2a6549d4352ec0a69

sparrowDom:
@DanielVF @shahthepro thanks for all the comments, would you guys mind another pass of this pr please? This commit is the significant one: https://github.com/OriginProtocol/origin-dollar/pull/2266/commits/c7346e63a48e9658dc1973327ff93e3b7f959998

contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
```diff
@@ -868,11 +894,19 @@ contract AerodromeAMOStrategy is InitializableAbstractStrategy {
         nonReentrant
     {
         // to add liquidity to the clPool
-        IERC20(WETH).safeApprove(address(positionManager), type(uint256).max);
+        IERC20(OETHb).safeApprove(address(positionManager), 0);
         IERC20(OETHb).safeApprove(address(positionManager), type(uint256).max);
         // to be able to rebalance using the swapRouter
-        IERC20(WETH).safeApprove(address(swapRouter), type(uint256).max);
+        IERC20(OETHb).safeApprove(address(swapRouter), 0);
```
shahthepro, line 900: nit: Any reason to use safeApprove for OETHb? It behaves as we expect it to with just `approve` as well

shahthepro (approved):
LGTM

contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
sparrowDom: nice catch, correcting

shahthepro (dismissed):

contracts/deploy/base/017_upgrade_amo.js
```diff
@@ -0,0 +1,71 @@
...
+          // 2. Reset WETH approvals to 0 on swapRouter and positionManager
+          contract: cAMOStrategy,
+          signature: "safeApproveAllTokens()",
+          args: [],
+        },
+        {
+          // 2. set auto allocate threshold to 1 - gas is cheap
+          contract: cOETHbVault,
+          signature: "setAutoAllocateThreshold(uint256)",
+          args: [utils.parseUnits("0.1", 18)],
```
clement-ux, line 47: nit: The comment says one, but the argument says 0.1 ether. I assume that 0.1 ether is the correct value, but just want to be sure about it.

clement-ux (dismissed):
LGTM 👍

contracts/deploy/base/017_upgrade_amo.js
sparrowDom: thanks! corrected

PR #2275, opened 2024-10-12
Enable async withdrawals on superOETHb

shahthepro:
## Changes Overview
- Removes `CLAIM_DELAY` constant in favour of a `withdrawalClaimDelay` variable
- `withdrawalClaimDelay` has an hardcoded limit of 10m to 7d (to avoid accidentally or intentionally setting a huge withdrawal claim period)
- `withdrawalClaimDelay` can be set to zero to disable async withdrawals. This will prevent users from making any new withdrawal requests. However, they'd be able to claim any pending withdrawals instantly after disabling it

## Code Change Checklist

- [ ]  Unit tests *if needed

Internal review:


contracts/contracts/vault/OETHVaultCore.sol
```diff
@@ -284,6 +284,14 @@ contract OETHVaultCore is VaultCore {
...
 
     function _claimWithdrawal(uint256 requestId)
+        internal
+        virtual
+        returns (uint256 amount)
+    {
+        return _claimWithdrawal(requestId, CLAIM_DELAY);
+    }
+
+    function _claimWithdrawal(uint256 requestId, uint256 claimDelay)
```
naddison36, line 294: I assume a second internal function was added so Mainnet can be backward compatible with using the `CLAIM_DELAY` constant? I'd be inclined to also change mainnet to use the new `withdrawalClaimDelay` storage variable the next time the vault is upgraded so we don't have code to handle both `CLAIM_DELAY` and `withdrawalClaimDelay`.

shahthepro: > I assume a second internal function was added so Mainnet can be backward compatible with using the CLAIM_DELAY constant?

Yes, that was the reason.

> I'd be inclined to also change mainnet to use the new withdrawalClaimDelay storage variable the next time the vault is upgraded so we don't have code to handle both CLAIM_DELAY and withdrawalClaimDelay.

I assumed that we would not make it gas expensive. But I guess it should be negligible and will keep our codebase clean, will make this change, thanks

contracts/contracts/vault/VaultAdmin.sol
```diff
@@ -163,6 +163,20 @@ contract VaultAdmin is VaultStorage {
...
     }
 
+    /**
+     * @notice Changes the async withdrawal claim period for superOETHb
+     * @param _delay Delay period (should be between 30mins to 7 days).
+     *          Set to 0 to disable async withdrawals
+     */
+    function setWithdrawalClaimDelay(uint256 _delay) external onlyGovernor {
+        require(
+            _delay == 0 || (_delay >= 30 minutes && _delay <= 7 days),
```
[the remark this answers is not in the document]
shahthepro: This is done

contracts/contracts/vault/OETHVaultCore.sol
shahthepro: Have made that change

contracts/contracts/harvest/FixedRateDripper.sol
```diff
@@ -58,7 +58,7 @@ contract FixedRateDripper is Dripper {
      * @param _perSecond Rate of WETH to drip per second
      */
     function setDripRate(uint192 _perSecond) external onlyGovernorOrStrategist {
-        emit DripRateUpdated(_perSecond, drip.perSecond);
+        emit DripRateUpdated(drip.perSecond, _perSecond);
```
sparrowDom, line 61: thanks for fixing this 👍

contracts/contracts/vault/VaultStorage.sol
```diff
@@ -236,8 +237,14 @@ contract VaultStorage is Initializable, Governable {
     /// @notice Mapping of withdrawal request indices to the user withdrawal request data
     mapping(uint256 => WithdrawalRequest) public withdrawalRequests;
 
+    /// @notice Used for OETH & superOETHb async withdrawal
```
sparrowDom, line 240: Would be nice if the comment stated the function of the variable. Smeth like: 
`@notice: Sets a minimum delay that is required to elapse between requesting async withdrawals and claiming the request. When set to 0 async withdrawals are disabled`

contracts/contracts/vault/VaultStorage.sol
```diff
@@ -236,8 +237,14 @@ contract VaultStorage is Initializable, Governable {
...
+    // slither-disable-start constable-states
+    // slither-disable-next-line uninitialized-state
```
sparrowDom, line 242: slither being annoying :)

contracts/test/vault/oethb-vault.base.fork-test.js
```diff
@@ -113,20 +113,101 @@ describe("ForkTest: OETHb Vault", function () {
...
-    it("Should be disabled", async () => {
-      const { oethbVault, nick } = fixture;
+    it("Should allow 1:1 async withdrawals", async () => {
+      const { rafael, oethbVault } = fixture;
+
+      const delayPeriod = await oethbVault.withdrawalClaimDelay();
+
+      if (delayPeriod == 0) {
+        // Skip when disabled
+        return;
```
sparrowDom, line 123: I think we should have a fixture here that sets claim delays to some constant, say 10 minutes. If we don't it could happen that: 
- we disable async withdrawals for whatever reason on the mainnet
- introduce a code change that breaks the functionality
- re-enable on the mainnet and we wouldn't detect it

contracts/test/vault/oethb-vault.base.fork-test.js
```diff
@@ -113,20 +113,101 @@ describe("ForkTest: OETHb Vault", function () {
...
+
+      // Rafael places an async withdrawal request
+      await oethbVault.connect(rafael).requestWithdrawal(oethUnits("1"));
+
+      // ... and tries to claim it after 1d
+      await advanceTime(delayPeriod);
+      await oethbVault.connect(rafael).claimWithdrawal(requestId);
+    });
+
+    it("Should not allow withdraw before claim delay", async () => {
```
sparrowDom, line 140: fixture with constant claim delay would fit in here as well

sparrowDom (approved):
Left a couple of comments inline. Looks great 👍

contracts/contracts/vault/VaultStorage.sol
shahthepro: Updated, thanks

naddison36 (approved):
LGTM

contracts/contracts/harvest/FixedRateDripper.sol
DanielVF, line 61: Did we mean to do this in this PR? Changes how the event is parsed.

If we were going to make a change, we could probably remove emitting the old rate.

contracts/contracts/vault/VaultAdmin.sol
```diff
@@ -163,6 +163,20 @@ contract VaultAdmin is VaultStorage {
...
+     * @notice Changes the async withdrawal claim period for OETH & superOETHb
+     * @param _delay Delay period (should be between 10 mins to 7 days).
+     *          Set to 0 to disable async withdrawals
+     */
+    function setWithdrawalClaimDelay(uint256 _delay) external onlyGovernor {
+        require(
+            _delay == 0 || (_delay >= 10 minutes && _delay <= 7 days),
+            "Invalid claim delay period"
+        );
+        emit WithdrawalClaimDelayUpdated(withdrawalClaimDelay, _delay);
```
DanielVF, line 176: I don't think we need to emit old values in set events.

contracts/contracts/harvest/FixedRateDripper.sol
shahthepro: Yeah, I thought I'd fix this now and deploy it later. Removing old value makes sense, will do that

contracts/contracts/vault/VaultAdmin.sol
shahthepro: Done

contracts/contracts/harvest/FixedRateDripper.sol
shahthepro: On second thought, I just got rid of this change from this branch since it doesn't really belong here. Will make that change when we update the Dripper the next time

contracts/contracts/vault/OETHVaultCore.sol
```diff
@@ -287,12 +288,16 @@ contract OETHVaultCore is VaultCore {
         internal
         returns (uint256 amount)
     {
+        // Note: When `withdrawalClaimDelay` is set to 0, users can
```
DanielVF, line 291: I think we will want to eventualy change the `request.timestamp` to become an end stamp rather than a start time stamp.

The current code has three awkward things:

- If we change the claim duration, it retroactively changes the end date of inflight requests.
- If we disable async claims by setting it to zero, it actually allows instant claims of all inflight claims.
- If someone wrapped the governance execute() of a zero set, then they could flash loan a mint and a redeem in the same block.

However, our current upgrade on OETH will be atomically setting a value for this in the upgrade, and we don't have to use the zero value for now.

Given that the current zero value is dangerous, how about we also have the require nonzero in the claim? In this way we never accidentally give someone the ability to flashloan an 1:1 mint/redeem. We could then remove it at some future date once we've changed the code to make it safe.

contracts/contracts/vault/VaultAdmin.sol
```diff
@@ -163,6 +163,20 @@ contract VaultAdmin is VaultStorage {
...
     }
 
+    /**
+     * @notice Changes the async withdrawal claim period for OETH & superOETHb
+     * @param _delay Delay period (should be between 10 mins to 7 days).
+     *          Set to 0 to disable async withdrawals
+     */
+    function setWithdrawalClaimDelay(uint256 _delay) external onlyGovernor {
+        require(
+            _delay == 0 || (_delay >= 10 minutes && _delay <= 7 days),
```
DanielVF, line 173: I think it makes sense to set the max delay to something that's greater than the max validator withdraw time. Perhaps 10 days? (At least one LST has a 15 day timed withdraw)

contracts/contracts/vault/VaultAdmin.sol
```diff
@@ -163,6 +163,20 @@ contract VaultAdmin is VaultStorage {
...
+            "Invalid claim delay period"
+        );
+        emit WithdrawalClaimDelayUpdated(_delay);
```
DanielVF, line 176: We usually like doing events last, when possible.

contracts/contracts/vault/OETHVaultCore.sol
shahthepro: I looked at changing `request.timestamp` to the end time. Looks like there's no space in the `WithdrawRequest` to fit in the same slot now. If we reuse the existing var on that struct, someone can still wrap our governance execute() and do a flash loan attack. So, yeah, leaving that aside for sometime later makes sense. When we do that change, perhaps we could consider disabling the async withdrawals first (go through governance and execute it) and then doing the upgrade would help avoid the flash loan attacks.

For now, I have changed `_claimWithdrawal` to revert when it's disabled as well.

> However, our current upgrade on OETH will be atomically setting a value for this in the upgrade, and we don't have to use the zero value for now.
This may not be true. We have a deployment file for Base. Not for mainnet. I have excluded that file intentionally since we have other changes to the Vault happening soon (rebasification). I can set a reminder to update the deployment file in those PRs to reset the withdrawalClaimDelay to 10 minutes (current value)

contracts/contracts/vault/VaultAdmin.sol
shahthepro: Since this is just an upper bound, not the actual value that we would be using, I just set it to 15 days

contracts/contracts/vault/VaultAdmin.sol
shahthepro: Done

DanielVF (approved):

contracts/contracts/vault/OETHVaultCore.sol
DanielVF: Agree on disabling withdraws right before the upgrade.

naddison36 (approved):

PR #2276, opened 2024-10-15
Enable pool fees when rebalancing Aerodrome AMO

sparrowDom:
- Leave fees on Aerodrome pool when performing rebalances
- fix Quoter contract to report correct swap values when estimating amount to swap when reaching price. Before it was overshooting
- fix Quoter contract to start the binary search at a more appropriate levels according to the strategy contract's current TVL

## Code Change Checklist

- [ ]  Unit tests *if needed

Internal review:

## Deploy checklist

```
```

### DIFF
--- a/contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
+++ b/contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
@@ -777,7 +777,7 @@ contract AerodromeAMOStrategy is InitializableAbstractStrategy {
      */
     function _burnOethbOnTheContract() internal {
         uint256 _oethbBalance = IERC20(OETHb).balanceOf(address(this));
-        if (_oethbBalance > 0) {
+        if (_oethbBalance > 1e12) {
             IVault(vaultAddress).burnForStrategy(_oethbBalance);
         }
     }
--- a/contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
+++ b/contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
@@ -486,8 +486,9 @@ contract AerodromeAMOStrategy is InitializableAbstractStrategy {
         uint256 _amountToSwap,
         bool _swapWeth
     ) internal {
+        // swapping OETHb to WETH doesn't require liquidity removal or when there 
+        // is no amount to be swapped
         if (!_swapWeth || _amountToSwap == 0) {
-            // swapping OETHb to WETH doesn't require liquidity removal
             return;
         }
 
--- a/contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
+++ b/contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
@@ -427,9 +427,13 @@ contract AerodromeAMOStrategy is InitializableAbstractStrategy {
         /**
          * When rebalance is called for the first time there is no strategy
          * liquidity in the pool yet. The liquidity removal is thus skipped.
+         * Also execute this function when WETH is required for the swap.
          */
-        if (tokenId != 0) {
-            _removeLiquidityToEnsureSwap(_amountToSwap, _swapWeth);
+        if (tokenId != 0 && _swapWeth && _amountToSwap > 0) {
+            _ensureWETHBalance(_amountToSwap);
+
+            // burn remaining OETHb -- TODO: 
+            _burnOethbOnTheContract();
         }
 
         // in some cases we will just want to add liquidity and not issue a swap to move the
@@ -471,30 +475,6 @@ contract AerodromeAMOStrategy is InitializableAbstractStrategy {
         ) {
             revert("Protocol insolvent");
         }
-    }
-
-    /**
-     * @dev Remove just enough liquidity so that the swap operation can be performed. No need to
-     * add the gaugeUnstakeAndRestake modifier since the underlying _removeLiquidity function
-     * will do that.
-     * Good to consider that remaining liquidity will remain in the pool & staked to the gauge.
-     *
-     * @param _amountToSwap The amount of the token to swap
-     * @param _swapWeth Swap using WETH when true, use OETHb when false
-     */
-    function _removeLiquidityToEnsureSwap(uint256 _amountToSwap, bool _swapWeth)
-        internal
-    {
-        // swapping OETHb to WETH doesn't require liquidity removal or when there
-        // is no amount to be swapped
-        if (!_swapWeth || _amountToSwap == 0) {
-            return;
-        }
-
-        _ensureWETHBalance(_amountToSwap);
-
-        // burn remaining OETHb
-        _burnOethbOnTheContract();
     }
 
     /**
--- a/contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
+++ b/contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
@@ -603,6 +603,9 @@ contract AerodromeAMOStrategy is InitializableAbstractStrategy {
                     : sqrtRatioX96TickHigher
             })
         );
+
+        // burn remaining OETHb
+        //_burnOethbOnTheContract();
     }
 
     /**
@@ -737,7 +740,8 @@ contract AerodromeAMOStrategy is InitializableAbstractStrategy {
      *         configured ticker.
      */
     function _checkForExpectedPoolPrice(bool throwException)
-        internal view
+        internal
+        view
         returns (bool _isExpectedRange, uint256 _wethSharePct)
     {
         require(
--- a/contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
+++ b/contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
@@ -426,10 +426,10 @@ contract AerodromeAMOStrategy is InitializableAbstractStrategy {
 
         /**
          * When rebalance is called for the first time there is no strategy
-         * liquidity in the pool yet. The full liquidity removal is thus skipped.
+         * liquidity in the pool yet. The liquidity removal is thus skipped.
          */
         if (tokenId != 0) {
-            _removeLiquidityToFacilitateSwap(_amountToSwap, _swapWeth);
+            _removeLiquidityToEnsureSwap(_amountToSwap, _swapWeth);
         }
 
         // in some cases we will just want to add liquidity and not issue a swap to move the
@@ -482,7 +482,7 @@ contract AerodromeAMOStrategy is InitializableAbstractStrategy {
      * @param _amountToSwap The amount of the token to swap
      * @param _swapWeth Swap using WETH when true, use OETHb when false
      */
-    function _removeLiquidityToFacilitateSwap(
+    function _removeLiquidityToEnsureSwap(
         uint256 _amountToSwap,
         bool _swapWeth
     ) internal {
@@ -492,7 +492,10 @@ contract AerodromeAMOStrategy is InitializableAbstractStrategy {
             return;
         }
 
-        _assureWETHBalance(_amountToSwap);
+        _ensureWETHBalance(_amountToSwap);
+
+        // burn remaining OETHb
+        _burnOethbOnTheContract();
     }
 
     /**
@@ -568,7 +571,7 @@ contract AerodromeAMOStrategy is InitializableAbstractStrategy {
         uint256 _balance = _tokenToSwap.balanceOf(address(this));
 
         if (_balance < _amountToSwap) {
-            // This should never trigger since _removeLiquidityToFacilitateSwap will already
+            // This should never trigger since _removeLiquidityToEnsureSwap will already
             // throw an error if there is not enough WETH
             if (_swapWeth) {
                 revert NotEnoughWethForSwap(_balance, _amountToSwap);
@@ -835,26 +838,28 @@ contract AerodromeAMOStrategy is InitializableAbstractStrategy {
      *
      * @param _amount  WETH balance required on the contract
      */
-    function _assureWETHBalance(uint256 _amount) internal {
+    function _ensureWETHBalance(uint256 _amount) internal {
         uint256 _wethBalance = IERC20(WETH).balanceOf(address(this));
-        if (_wethBalance < _amount) {
-            require(tokenId != 0, "No liquidity available");
-            uint256 _additionalWethRequired = _amount - _wethBalance;
-            (uint256 _wethInThePool, ) = getPositionPrincipal();
-
-            if (_wethInThePool < _additionalWethRequired) {
-                revert NotEnoughWethLiquidity(
-                    _wethInThePool,
-                    _additionalWethRequired
-                );
-            }
-
-            uint256 shareOfWethToRemove = Math.min(
-                _additionalWethRequired.divPrecisely(_wethInThePool) + 1,
-                1e18
-            );
-            _removeLiquidity(shareOfWethToRemove);
+        if (_wethBalance >= _amount) {
+            return;
         }
+
+        require(tokenId != 0, "No liquidity available");
+        uint256 _additionalWethRequired = _amount - _wethBalance;
+        (uint256 _wethInThePool, ) = getPositionPrincipal();
+
+        if (_wethInThePool < _additionalWethRequired) {
+            revert NotEnoughWethLiquidity(
+                _wethInThePool,
+                _additionalWethRequired
+            );
+        }
+
+        uint256 shareOfWethToRemove = Math.min(
+            _additionalWethRequired.divPrecisely(_wethInThePool) + 1,
+            1e18
+        );
+        _removeLiquidity(shareOfWethToRemove);
     }
 
     /**
@@ -872,7 +877,7 @@ contract AerodromeAMOStrategy is InitializableAbstractStrategy {
         require(_asset == WETH, "Unsupported asset");
         require(_recipient == vaultAddress, "Only withdraw to vault allowed");
 
-        _assureWETHBalance(_amount);
+        _ensureWETHBalance(_amount);
 
         // burn remaining OETHb
         _burnOethbOnTheContract();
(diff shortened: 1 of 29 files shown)


contracts/test/strategies/aerodrome-amo.base.fork-test.js
```diff
@@ -471,18 +475,16 @@ describe("ForkTest: Aerodrome AMO Strategy (Base)", async function () {
...
-
       await setERC20TokenBalance(
         aerodromeAmoStrategy.address,
         aero,
         "1337",
         hre
       );
-      const aeroBalanceBefore = await aero.balanceOf(strategistAddr);
-      await aerodromeAmoStrategy.connect(strategist).collectRewardTokens();
+      const aeroBalanceBefore = await aero.balanceOf(harvester.address);
```
shahthepro, line 484: Thanks for fixing this, I believe I also fixed this in another branch: #2275 
Depending on which PR goes live first, we might have a merge conflict. But I can change my branch if it comes to that

contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
```diff
@@ -429,8 +429,9 @@ contract AerodromeAMOStrategy is InitializableAbstractStrategy {
          * liquidity in the pool yet. The full liquidity removal is thus skipped.
```
shahthepro, line 429: nit: comment about "full liquidity being removed" is misleading

contracts/test/strategies/aerodrome-amo.base.fork-test.js
sparrowDom: thanks for flagging. Yeah let this be a race and the last PR to be merged needs to fix it :)

contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
```diff
@@ -803,20 +830,12 @@ contract AerodromeAMOStrategy is InitializableAbstractStrategy {
...
-        address _recipient,
-        address _asset,
-        uint256 _amount
-    ) external override onlyVault nonReentrant {
-        require(_asset == WETH, "Unsupported asset");
-        require(_recipient == vaultAddress, "Only withdraw to vault allowed");
-
+    function _assureWETHBalance(uint256 _amount) internal {
         uint256 _wethBalance = IERC20(WETH).balanceOf(address(this));
         if (_wethBalance < _amount) {
```
DanielVF, line 840: ⚪️ Can flip this, and change this to fast return, now that this if statement is the whole function?

contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
```diff
@@ -803,20 +830,12 @@ contract AerodromeAMOStrategy is InitializableAbstractStrategy {
...
      */
-    function withdraw(
-        address _recipient,
-        address _asset,
-        uint256 _amount
-    ) external override onlyVault nonReentrant {
-        require(_asset == WETH, "Unsupported asset");
-        require(_recipient == vaultAddress, "Only withdraw to vault allowed");
-
+    function _assureWETHBalance(uint256 _amount) internal {
```
DanielVF, line 838: ⚪️ `_ensureWETHBalance`.  "Assure" means "I'm trying to convince you things are good so that you feel better".  "Ensure" means "I will take the actions necessary to be sure that something is done.". We want the second.

contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
```diff
@@ -429,8 +429,9 @@ contract AerodromeAMOStrategy is InitializableAbstractStrategy {
          * liquidity in the pool yet. The full liquidity removal is thus skipped.
          */
         if (tokenId != 0) {
-            _removeLiquidity(1e18);
+            _removeLiquidityToFacilitateSwap(_amountToSwap, _swapWeth);
```
DanielVF, line 432: 🔴 `_removeLiquidity` burns the superOETHb returned from the contract. `_removeLiquidityToFacilitateSwap` does not. That could result in very fun amounts of `superOETHb` sitting on the AMO contract, increasing our TVL.

Perhaps we put a superOETHb burn  in `_ensureWETHBalance` or in `_removeLiquidity` itself?

contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
DanielVF: Given that this is called by the very active sounding `_removeLiquidityToFacilitateSwap ` and calls the active sounding `_removeLiquidity`, I might go with either more active name here. Maybe `_removeLiqudityToEnsureWETH`

contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
```diff
@@ -836,6 +855,24 @@ contract AerodromeAMOStrategy is InitializableAbstractStrategy {
             );
             _removeLiquidity(shareOfWethToRemove);
```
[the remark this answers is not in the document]
DanielVF: My feel is that it's better to keep this PR simpler and not change this here.

But what are the edge cases where this might be a problem?

contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
DanielVF: Note that `_addLiquidty` will call burn (and this happens near the end of the swap), but it won't burn if the weth balance is zero.

contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
sparrowDom: thanks fixed

contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
sparrowDom: great catch thanks, added the burn to the `_removeLiquidityToFacilitateSwap`. Will also add the functionality to tests where after the rebalance there should be no non dust WETH / OETHb left on the contract

contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
sparrowDom: hah thanks yes ensure seems much more fitting. 
You assuring me ensure is the better word has me convinced :)

contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
sparrowDom: nice short circuit spot, changed

contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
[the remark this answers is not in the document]
sparrowDom: Agree to not fix this unless it turns out to cause problems

contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
```diff
@@ -426,11 +426,12 @@ contract AerodromeAMOStrategy is InitializableAbstractStrategy {
 
         /**
          * When rebalance is called for the first time there is no strategy
-         * liquidity in the pool yet. The full liquidity removal is thus skipped.
+         * liquidity in the pool yet. The liquidity removal is thus skipped.
          */
         if (tokenId != 0) {
-            _removeLiquidity(1e18);
+            _removeLiquidityToEnsureSwap(_amountToSwap, _swapWeth);
```
shahthepro, line 432: nit: Can probably replace lines 431 to 439 with something like this. But not a big deal though since `_removeLiquidityToEnsureSwap` has a `amountToSwap > 0` check 
```
        if (_amountToSwap > 0) {
            if (tokenId != 0) {
                _removeLiquidityToEnsureSwap(_amountToSwap, _swapWeth);
            }

            _swapToDesiredPosition(_amountToSwap, _swapWeth, _minTokenReceived);
        }
        ```

File: contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol [hunk 8]
```diff
@@ -472,6 +473,30 @@ contract AerodromeAMOStrategy is InitializableAbstractStrategy {
...
+        // swapping OETHb to WETH doesn't require liquidity removal or when there
+        // is no amount to be swapped
+        if (!_swapWeth || _amountToSwap == 0) {
+            return;
+        }
+
+        _ensureWETHBalance(_amountToSwap);
+
+        // burn remaining OETHb
+        _burnOethbOnTheContract();
```
shahthepro, line 497: nit: This placement of burn is a bit off but it seems to work fine.

If `swapAmount < depositAmount`, there'd be no burn here. So that's good. If `depositAmount < swapAmount`, it pulls WETH from the pool and burns superOETHb. But there'll be nothing to add post that swap. Code seems to handle both cases fine. Will do a few checks to be sure and get back on this

shahthepro (commented):
`_checkForExpectedPoolPrice` should be `view` not `internal` I believe (though it's not part of this PR)

contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
shahthepro: @sparrowDom If I just call rebalance with swapAmount set to 10 WETH and deposit amount set to 0 WETH (i.e. no WETH from Vault), this is how the strategy behaves (assuming 20/80 splits):
- Remove 10 WETH and 40 superOETHb
- Burn 40 superOETHb
- Swap 10 WETH for 40 superOETHb
- Exit without deposit (since depositAmount is 0)

So, it looks like it still leaves some superOETHb in the contract

shahthepro: Perhaps we could consider updating `_solvencyAssert` to ensure that `oethb.balanceOf(address(this)) == 0` (or even some small amount like 1e6)

sparrowDom: this is a great point. Will add a test for this. Thinking that I'd rather burn the OETH within the swap function so that each function in the `_rebase` flow cleans up after itself. The sanity check in the end also makes sense

DanielVF: Yeah. We are on L2's with this, so gas is cheap.

shahthepro: > since someone could trigger it by donating?

I guess we can always burn any leftover superOETHb at the end of a rebalance (just before calling the solvency check method)

sparrowDom: I was curious why [the new test conditions](https://github.com/OriginProtocol/origin-dollar/pull/2276/files#diff-2fc9b69b9f62ab0307369e83cceb6e19bdf34994b331eb4f5e21a8fbf9341cefR1222-R1231) I have added before this conversation didn't trigger OETHb being left on the contract. 
Turns out that because we slightly overshoot the amount of WETH being removed to ensure the swap the `addLiquidity` function would always execute in full - adding the liquidity and burning OETHb.

It is somewhat comforting that when (in future) we would fix the slight overshooting of WETH in rebalance those test would fail. Still super awesome this was also caught in the code review 💪

contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
```diff
@@ -576,6 +603,14 @@ contract AerodromeAMOStrategy is InitializableAbstractStrategy {
...
             })
         );
+
+        /**
+         * In the interest of each function in _rebalance to leave the contract state as
+         * clean as possible the OETHb tokens here are burned. This decreases the
+         * dependence where `_swapToDesiredPosition` function relies on later functions
+         * (`addLiquidity`) to burn the OETHb. Reducing the risk of error introduction.
+         */
+        _burnOethbOnTheContract();
```
DanielVF, line 613: ⚪️ So currently, we do this to ensure that there's no leftover SuperOETHb on the contract.

image

This is ugly because two of the inner methods that can get superOETHb, check and burn locally. However the remove liquidity side is currently burning at different higher levels, two of which are in external methods, one of which is internal.

We could make this very simple by burning, and only burning, right at the places we might get superOETHb:

image

DanielVF: After some discussion with shah in discord, the current contract burns up to three times on a deposit or rebalance.

If we moved the burnOETHb calls to the following locations, we could exactly one possible burn per external call:

image

contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
```diff
@@ -472,6 +473,30 @@ contract AerodromeAMOStrategy is InitializableAbstractStrategy {
...
+    /**
+     * @dev Remove just enough liquidity so that the swap operation can be performed. No need to
+     * add the gaugeUnstakeAndRestake modifier since the underlying _removeLiquidity function
+     * will do that.
+     * Good to consider that remaining liquidity will remain in the pool & staked to the gauge.
+     *
+     * @param _amountToSwap The amount of the token to swap
+     * @param _swapWeth Swap using WETH when true, use OETHb when false
+     */
+    function _removeLiquidityToEnsureSwap(uint256 _amountToSwap, bool _swapWeth)
```
DanielVF, line 485: ⚪️ I think we should get rid of this function. The only thing it's adding over calling ` _ensureWETHBalance` is toto check two conditions, and burnOethB afterwords. But this is the wrong place to burn superOETHb, and both the conditions can easily be checked in the one if statement around the one place this is called.

sparrowDom: agreed adding another call adds more complexity over checking for these conditions in the function that is calling it. Removing this function

sparrowDom: done here: https://github.com/OriginProtocol/origin-dollar/pull/2276/commits/3bd3be1a8f5ab249153a9e90456c2126b0b18788

DanielVF: Looks good.

contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
DanielVF: Our primary objective here is to have the burning happen at one consistent level of the codebase. This makes the code understandable and hard to have bugs, because there's a simple rule, any where it's not followed is probably a bug.

So either burn only at the lowest level. (simple, hard to mess up, runs more often than needed on trade rebalances, but only once on everything else). Or run it at a higher level in the code only, and we ensure that it only runs once.

Here's current state.

image

What is a nightmare is running it at many different levels. This requires more calls (called five times vs three), is much harder to understand, and much easier to make a mistake on in the future. There's no underlying rule about where the burn code is that people can use to think about the code, or fit new code in. The current code not only keeps it called at multiple levels of the code, it doesn't simplify the number calls burns to burn, and also does the extra dust burns.

We need to either run the burns only at the low level in `_removeLiquidity`, `_swapToDesiredPosition`, and `_addLiquidity`, and accept an extra dust burn on rebalances+trades, or run it only in `withdrawAll`, `withdraw`, and in `_rebalance` (after the call to ` _addLiquidity` in `_rebalance`).

sparrowDom: I've created a separate PR for this. https://github.com/OriginProtocol/origin-dollar/pull/2290

I think it make sense to improve the accuracy of removing liquidity and add thresholds to triggers adding liquidity so the tests capture any potential issues with lingering OETHb.

sparrowDom: Just read the edited comment, yes I agree completely those are the 2 sensible options. In the PR linked in the comment above is also a change where OETHb is burned only at the lowest level

DanielVF: Alright. Now we are cooking. 
image

contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
```diff
@@ -351,7 +351,7 @@ contract AerodromeAMOStrategy is InitializableAbstractStrategy {
      */
     function depositAll() external override onlyVault nonReentrant {
         uint256 _wethBalance = IERC20(WETH).balanceOf(address(this));
-        if (_wethBalance > 0) {
+        if (_wethBalance > 1e12) {
```
DanielVF, line 354: 🟣 There's a slightly ugly bit with skipping deposits on small amounts. This means that we don't emit external events on some funds that are received from the vault. We don't have any solvency or balance tracking issues here, just emitted logs not lining up by dust amounts.

I guess this okay because any alternative might double count.

DanielVF (approved):
Approved, though we might want to return early in the burn function if we just have dust. This way we'll only do non-dust burns.

contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
DanielVF: Yes.  I think this would be cleaner.

contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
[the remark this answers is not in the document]
sparrowDom: agreed it makes sense as a defensive measure: https://github.com/OriginProtocol/origin-dollar/pull/2276/commits/cf8084c06227b7dfada58c6d85ae1e8344f2535d

DanielVF (dismissed):

PR #2281, opened 2024-10-20
Harvest Action now collects if there is more than 0.5 execution rewards

naddison36:
## Changes

* The Defender Action that harvests native staking rewards has been updated to collect if there is more than 0.5 ETH of execution rewards. It previously only collected if there was more than 1 ETH of consensus rewards.

### DIFF
--- a/contracts/scripts/defender-actions/harvest.js
+++ b/contracts/scripts/defender-actions/harvest.js
@@ -51,20 +51,33 @@ const harvest = async (
   signer,
   stratDesc
 ) => {
-  const nativeStakingStrategy1 = new ethers.Contract(
+  const nativeStakingStrategy = new ethers.Contract(
     nativeStakingProxyAddress,
     nativeStakingStrategyAbi,
     signer
   );
-  const consensusRewards = await nativeStakingStrategy1.consensusRewards();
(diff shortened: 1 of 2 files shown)


contracts/scripts/defender-actions/harvest.js
```diff
@@ -51,20 +51,33 @@ const harvest = async (
...
+  const feeAccumulatorAddress =
+    await nativeStakingStrategy.FEE_ACCUMULATOR_ADDRESS();
+  const executionRewards = await signer.provider.getBalance(
+    feeAccumulatorAddress
+  );
+  log(`Execution rewards for ${stratDesc}: ${formatUnits(executionRewards)}`);
+
+  if (
+    consensusRewards.gt(parseEther("1")) ||
+    executionRewards.gt(parseEther("0.5"))
```
shahthepro, line 71: nit: Could probably use `gte` but not a big deal

shahthepro (approved):

contracts/scripts/defender-actions/harvest.js
naddison36: yeah, the line has to be drawn somewhere. Those amounts were arbitrary

PR #2291, opened 2024-10-25
Automatic AMO rebalances

sparrowDom:
Adds a new rebalance function which a permissioned account is able to call to automatically rebalance a limited volume within a time-frame. The limited volume and rebalancer account is specified by the govenor. 

## Code Change Checklist

- [ ]  Unit tests *if needed

Internal review:

## Deploy checklist

```
```

### DIFF
--- a/contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
+++ b/contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
@@ -52,8 +52,14 @@ contract AerodromeAMOStrategy is InitializableAbstractStrategy {
     /// @notice Marks the end of the interval that defines the allowed range of WETH share in
     /// the pre-configured pool's liquidity ticker
     uint256 public allowedWethShareEnd;
+    /// @notice An account that can call rebalances respecting volume limits
+    address public rebalancer;
+    /// @notice Volume limit a rebalancer can operate within 300 blocks (~roughly an hour)
+    uint256 public rebalancerVolumeLimit;
+    /// @dev information regarding the trading volume of the rebalancer
+    RebalancerVolumeTally public rebalancerVolumeTally;
     /// @dev reserved for inheritance
-    int256[46] private __reserved;
+    int256[42] private __reserved;
 
     /***************************************
           Constants, structs and events
@@ -101,6 +107,15 @@ contract AerodromeAMOStrategy is InitializableAbstractStrategy {
     ///      protocol funds.
     uint256 public constant SOLVENCY_THRESHOLD = 0.998 ether;
 
+    /// @dev block interval within which the rebalancerVolumeLimit applies. 300 blocks equals ~1 hour
+    uint256 public constant REBALANCER_TRADE_INTERVAL_BLOCKS = 300;
+
+    /// @dev information regarding the trading volume of the rebalancer
+    struct RebalancerVolumeTally {
+        uint256 startBlock;
+        uint256 volumeRebalaced;
+    }
+
     error NotEnoughWethForSwap(uint256 wethBalance, uint256 requiredWeth); // 0x989e5ca8
     error NotEnoughWethLiquidity(uint256 wethBalance, uint256 requiredWeth); // 0xa6737d87
     error PoolRebalanceOutOfBounds(
@@ -137,6 +152,10 @@ contract AerodromeAMOStrategy is InitializableAbstractStrategy {
 
     event UnderlyingAssetsUpdated(uint256 underlyingAssets);
 
+    event RebalacerVolumeLimitUpdated(uint256 rebalancerVolumeLimit);
+
+    event RebalacerUpdated(address rebalancer);
+
     /**
      * @dev Verifies that the caller is the Governor, or Strategist.
      */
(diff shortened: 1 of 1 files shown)


contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
```diff
@@ -101,6 +107,15 @@ contract AerodromeAMOStrategy is InitializableAbstractStrategy {
     ///      protocol funds.
     uint256 public constant SOLVENCY_THRESHOLD = 0.998 ether;
 
+    /// @dev block interval within which the rebalancerVolumeLimit applies. 300 blocks equals ~1 hour
+    uint256 public constant REBALANCER_TRADE_INTERVAL_BLOCKS = 300;
+
+    /// @dev information regarding the trading volume of the rebalancer
+    struct RebalancerVolumeTally {
+        uint256 startBlock;
```
shahthepro, line 115: Nit: Could make both uint128 and fit in one slot. Block number and swap volume are not really gonna overflow

contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
```diff
@@ -149,6 +168,18 @@ contract AerodromeAMOStrategy is InitializableAbstractStrategy {
         _;
     }
 
+    /**
+     * @dev Verifies that the caller is the Rebalancer
+     */
+    modifier onlyRebalancer() {
+        require(
+            msg.sender == IVault(vaultAddress).strategistAddr() ||
```
shahthepro, line 176: There's no rebalancer check here?

contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
```diff
@@ -411,6 +459,26 @@ contract AerodromeAMOStrategy is InitializableAbstractStrategy {
...
     }
 
+    function rebalanceWithVolumeLimit(
+        uint256 _amountToSwap,
+        bool _swapWeth,
+        uint256 _minTokenReceived
+    ) external nonReentrant onlyRebalancer {
+        uint256 volumeRebalancedCurrentPeriod = _amountToSwap;
+
+        // no rebalance transaction from the rebalancer in the last hour
```
shahthepro, line 469: We can probably change it so that we are doing it as regenerative rate limit instead. Let me think and get back to this in a bit

contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
```diff
@@ -411,6 +459,26 @@ contract AerodromeAMOStrategy is InitializableAbstractStrategy {
...
+        if (rebalancerVolumeTally.startBlock + REBALANCER_TRADE_INTERVAL_BLOCKS < block.number) {
+            rebalancerVolumeTally.startBlock = block.number;
+        } else {
+            volumeRebalancedCurrentPeriod += rebalancerVolumeTally.volumeRebalaced;
+        }
+
+        require(volumeRebalancedCurrentPeriod <= rebalancerVolumeLimit, "Amount to swap exceeds limit");
+        rebalancerVolumeTally.volumeRebalaced = volumeRebalancedCurrentPeriod;
+
+        _rebalance(_amountToSwap, _swapWeth, _minTokenReceived);
```
shahthepro, line 479: Would be great to ensure that minTokenReceived isn't some crazy low amount

contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
sparrowDom: thanks fixed: https://github.com/OriginProtocol/origin-dollar/pull/2291/commits/81289d2c3f739a8ca73c42da6bfd35010c09d1e5

contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
sparrowDom: good point I guess we could enforce it to be 1bp below swapping amount. Since the automatic rebalancing shouldn't work outside the [-1,0] tick. And by work I mean start and end point being the tick

contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol
sparrowDom: I guess the implementation could be a linear decay of the tallied swap amounts. Meaning: 
- there is a certain threshold that musn't be surpassed
- the function keeps tally of the amount that has been swapped and each block decays (threshold / 900). (might make sense to make the decay a 3 hour thing instead of 1 to allow for some variance in the defender action)

Will implement this tomorrow and see if it can be done so it doesn't increase the complexity of the code too much

shahthepro: Yeah, what I was thinking was basically something like `(currentTime - lastRebalanceTime) * regenerationAmountPerSecond`. We can probably have an upper limit as well (say 500 WETH per rebalance). This will ensure that you cannot rebalance more than 500 WETH per hour. This will also limit the operator/rebalancer from doing a rebalance back to back for any reason

PR #2306, opened 2024-11-11
Use safecast for any downcasting

sparrowDom:
- Use safecast for any downcasting
 - fix one small bug that had an extra un-needed cast to `uint256`
 - fixes a bug where a negative balance would cause panic: https://github.com/OriginProtocol/origin-dollar/pull/2306/files#diff-9256f2c32e9ea588f87885ad1528ba06743860394c236fe4765adea760dd3c84L278
 - fixes unit tests

### DIFF
--- a/contracts/contracts/token/OUSD.sol
+++ b/contracts/contracts/token/OUSD.sol
@@ -18,6 +18,7 @@ import { SafeCast } from "@openzeppelin/contracts/utils/math/SafeCast.sol";
 
 contract OUSD is Governable {
     using SafeCast for int256;
+    using SafeCast for uint256;
 
     event TotalSupplyUpdatedHighres(
         uint256 totalSupply,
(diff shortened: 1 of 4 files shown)


contracts/contracts/token/OUSD.sol
```diff
@@ -277,41 +278,41 @@ contract OUSD is Governable {
         returns (int256 rebasingCreditsDiff, int256 nonRebasingSupplyDiff)
     {
         RebaseOptions state = rebaseState[account];
-        int256 currentBalance = int256(balanceOf(account));
-        uint256 newBalance = (currentBalance + balanceChange).toUint256();
```
DanielVF: Why did this change away from `uint256`? A balance is a number that can only be positive. It also looks like this changes makes us have to do many more conversions later in the code.

sparrowDom: because of the check on line 283. Though you are right this could be done in a cleaner way. Done here: https://github.com/OriginProtocol/origin-dollar/pull/2306/commits/21f7eb760ff4a50f78371a30b2ab48db133e4dc4

DanielVF (approved):

contracts/contracts/token/OUSD.sol
DanielVF: Thanks! Good catch.

PR #2307, opened 2024-11-11
Test code simplification

sparrowDom:
As per [EIP 2200](https://eips.ethereum.org/EIPS/eip-2200) ethereum EVM does some gas recycling when storage slots are being written to multiple times. This sparked the idea of updating the global states each time withing the `_adjustAccount` function. 
The code simplification is really nice, though it comes at a cost of ~3k extra gas usage. 

Gas usage on transfer from the base branch:
Screenshot 2024-11-11 at 23 48 35

Gas usage on transfer from this branch: 
Screenshot 2024-11-11 at 23 50 27

### DIFF
--- a/contracts/contracts/token/OUSD.sol
+++ b/contracts/contracts/token/OUSD.sol
@@ -254,25 +254,15 @@ contract OUSD is Governable {
             return;
         }
 
-        (
-            int256 fromRebasingCreditsDiff,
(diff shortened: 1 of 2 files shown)


sparrowDom:
not implementing this as 3k is a bit too pricey for such optimisation

PR #2308, opened 2024-11-12
Fix denomination issues with implementation 

sparrowDom:
This PR fixes a problem where `creditBalances`, `creditsPerToken`, `alternativeCreditsPerToken` are 1e27 denominated. And while using creditBalances & credits per token with 1e18 denomination will yield correct balances the issues arises when those credit balances are added to contract's global rebasingCredits


sparrowDom:
yes closing

PR #2318, opened 2024-12-01
Upgrade tests

sparrowDom:
This is only a pull request to perform the upgrade tests of token contract that check that all the token balances of OUSD, OETH and Super OETH remain the same after the contract upgrade. See `contracts/test/token/README.md` on how to execute the procedure. 

**This PR should not be merged**


contracts/test/helpers.js
```diff
@@ -752,6 +753,48 @@ async function differenceInStrategyBalance(
...
+}
+
+async function compareUpgradedContractBalances(balancesFile, upgradedTokenContract) {
+  // CSV file in format account, squidBalance, onChainBalance
+  const data = fs.readFileSync(balancesFile, "utf8").split('\n');
+  let balanceMissmatches = 0;
+  for (let i = 0; i < data.length; i++) {
+    const [account,,balanceBefore] = data[i].split(',');
+    const expectedBalance = BigNumber.from(balanceBefore);
+    const actualBalance = await upgradedTokenContract.balanceOf(account)
```
shahthepro, line 785: Would be better to do this in batches using `Promise.all`

sparrowDom: Agreed, I didn't do that in the end because all the storage slots are already cached and this function executes quite quickly

DanielVF:
Closing, since we don't plan on merging. Congrats on the successful upgrade.

PR #2325, opened 2024-12-06
N-04 Multiple Optimizable Storage Operations Addendum

sparrowDom:
**OpenZeppelin comment:**
Within the _adjustAccount function, the alternativeCreditsPerToken for non-rebasing accounts is always [set to 1e18](https://github.com/OriginProtocol/origin-dollar/blob/44951300cd2bb7eb1d387461465edbfc440f2779/contracts/contracts/token/OUSD.sol#L290), even when the mapping already has that value.

**Origin's response:** Fixed as suggested

### DIFF
--- a/contracts/contracts/token/OUSD.sol
+++ b/contracts/contracts/token/OUSD.sol
@@ -333,9 +333,14 @@ contract OUSD is Governable {
             creditBalances[_account] = newCredits;
         } else {
             _autoMigrate(_account);
-            if (alternativeCreditsPerToken[_account] > 0) {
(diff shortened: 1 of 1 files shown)


sparrowDom:
@DanielVF @clement-ux could you stamp this one please? It is a gas optimisation on a function that is called twice with each transfer.

clement-ux (approved):
Looks good to me. 
First we cache `alternativeCreditsPerToken` value and then we check if the cached value is different that the value we want to set it to avoid unnecessary operation. 

This is only a gas optimization and looks safe. All good 👍

DanielVF (approved):

PR #2327, opened 2024-12-16
Implement Curve Pool Booster.

clement-ux:
# Curve Pool Booster
This contract aims to take all funds available in the contract and create/manage bribes on VotemarketV2.
As VotemarketV2 is only available in L2s, this contract connects with VM using [the CampaignRemoteManager](https://etherscan.io/address/0xd1f0101Df22Cb7447F486Da5784237AB7a55eB4e#writeContract).

The connection with VM on L2s uses CCIP, so we need to add a bit of ETH in every transaction we do. The remaining gas is sent back to the contract. That's why I suggest ETH remain in the contract instead of sending transactions from the relayer with ethers.

The relayer will be able to perform multiple actions:
- Create a campaign
- Increase the amount for the remaining time for a campaign
- Increase campaign duration
- Adjust campaign price
- Close campaign
- Set campaign ID (as it uses cross-chain communication, a bribe is not created instantly, so we need to check logs and add it manually afterward).

Tests are pretty basic and don't test the whole process due to the complexity of testing cross-chain messaging. However, as this contract will hold a small amount of funds, I think it is ok to test it in prod. 

This contract should be deployed on BOTH mainnet and targetedChainId to completely manage the bribe. 
See this https://github.com/OriginProtocol/origin-dollar/pull/2327#issuecomment-2618146764.

## Code Change Checklist

- [x]  The contract code is complete
- [x]  Executable deployment file
- [x]  Fork tests that test after the deployment file runs
- [x]  Unit tests *if needed
- [x]  The owner has done a [full checklist review](https://github.com/OriginProtocol/security/blob/master/templates/Contract-Code-Review.md) of the code + tests

Internal review:


sparrowDom (commented):
Minor event comments. I guess prod testing is the way to go here like you suggest.

contracts/contracts/strategies/CurvePoolBooster.sol
```diff
@@ -0,0 +1,175 @@
...
+    function setCampaignId(uint256 _campaignId) external onlyOperator {
+        campaignId = _campaignId;
+    }
+
+    function setOperator(address _newOperator) external onlyGovernor {
+        operator = _newOperator;
+    }
+
+    function sendETH(address receiver) external onlyOperator {
+        payable(receiver).transfer(address(this).balance);
```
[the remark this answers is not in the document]
clement-ux: > to recover any accidentally sent tokens?
Not really. This contract will hold a bit of ETH, as we need to pay a bit when calling the CampaignRemoteManager (CCIP fees). So it is in the situation where the pool booster will not be used anymore (end of services) and the operator what to claim the remaining ETH. 

And using a payable function is not the best (imo) because a surplus of ETH is sent back to this contract, not to the sender.

What do you think of keeping `onlyOperator` modifier then?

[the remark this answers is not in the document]
sparrowDom: IMO it is fine, as there will only be small amount of ETH on the contract whose purpose is to fuel the gas for transactions and that is not part of the user funds.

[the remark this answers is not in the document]
clement-ux: Yes, exactly, only to fuel gas transactions, nothing more.

contracts/contracts/strategies/CurvePoolBooster.sol
```diff
@@ -0,0 +1,309 @@
...
+
+        // Prevent implementation contract to be governed
+        _setGovernor(address(0));
+    }
+
+    function initialize(
+        address _strategist,
+        uint16 _fee,
+        address _feeCollector
+    ) external initializer {
```
shahthepro, line 65: Should it have `onlyGovernor` modifier?

contracts/contracts/strategies/CurvePoolBooster.sol
```diff
@@ -0,0 +1,309 @@
...
+    /// @param blacklist  List of addresses to exclude from the campaign
+    /// @param bridgeFee Fee to pay for the bridge
+    /// @param additionalGasLimit Additional gas limit for the bridge
+    function createCampaign(
+        uint8 numberOfPeriods,
+        uint256 maxRewardPerVote,
+        address[] calldata blacklist,
+        uint256 bridgeFee,
+        uint256 additionalGasLimit
+    ) external onlyGovernorOrStrategist {
```
shahthepro, line 87: Can we make this function `payable`? It looks like `bridgeFee` is expected to be held by the contract. If it's `payable`, we can just do it in a call, as opposed to funding this contract and then creating a campaign

contracts/contracts/strategies/CurvePoolBooster.sol
```diff
@@ -0,0 +1,309 @@
...
+        IERC20(rewardToken).approve(campaignRemoteManager, balance);
+
+        // Create a new campaign
+        ICampaingRemoteManager(campaignRemoteManager).createCampaign{
+            value: bridgeFee
+        }(
+            ICampaingRemoteManager.CampaignCreationParams({
+                chainId: targetChainId,
+                gauge: gauge,
+                manager: address(this),
```
shahthepro, line 109: Is this the manager address on mainnet or on L2? I tried looking at the docs and couldn't find anything relevant

contracts/contracts/strategies/CurvePoolBooster.sol
```diff
@@ -0,0 +1,309 @@
...
+    event BribeCreated(
+        address gauge,
+        address rewardToken,
+        uint256 maxRewardPerVote,
+        uint256 totalRewardAmount
+    );
+    event TotalRewardAmountUpdated(uint256 extraTotalRewardAmount);
+    event NumberOfPeriodsUpdated(uint8 extraNumberOfPeriods);
+    event RewardPerVoteUpdated(uint256 newMaxRewardPerVote);
+    event RescueTokens(address token, uint256 amount, address receiver);
```
shahthepro, line 41: nit: `TokensRescued` so that it's inline with how other events are named

contracts/contracts/strategies/CurvePoolBooster.sol
```diff
@@ -0,0 +1,309 @@
...
+        emit CampaignIdUpdated(_campaignId);
+    }
+
+    /// @notice Rescue ETH from the contract
+    /// @dev Only callable by the governor or strategist
+    /// @param receiver Address to receive the ETH
+    function rescueETH(address receiver) external onlyGovernorOrStrategist {
+        //payable(receiver).transfer(amount);
+        emit RescueTokens(address(0), address(this).balance, receiver);
+        (bool success, ) = receiver.call{ value: address(this).balance }("");
```
shahthepro, line 263: Could extract `address(this).balance` to save some gas. Also, the commented line above can be removed

contracts/contracts/strategies/CurvePoolBooster.sol
```diff
@@ -0,0 +1,309 @@
...
+        onlyGovernorOrStrategist
+    {
+        campaignId = _campaignId;
+        emit CampaignIdUpdated(_campaignId);
+    }
+
+    /// @notice Rescue ETH from the contract
+    /// @dev Only callable by the governor or strategist
+    /// @param receiver Address to receive the ETH
+    function rescueETH(address receiver) external onlyGovernorOrStrategist {
```
shahthepro, line 260: Let's add `receiver != address(0)` check here

contracts/contracts/strategies/CurvePoolBooster.sol
clement-ux: The last time I discussed with the StakeDAO team was on the mainnet and L2. Let me ask again.

contracts/contracts/strategies/CurvePoolBooster.sol
clement-ux: Indeed! Done in this commit https://github.com/OriginProtocol/origin-dollar/pull/2327/commits/4ec38c8d849e8d5ba1ce1654d329e9fda1d501fd

contracts/contracts/strategies/CurvePoolBooster.sol
```diff
@@ -0,0 +1,309 @@
...
+
+    /// @notice calculate the fee amount and transfer it to the feeCollector
+    /// @param amount Amount to calculate the fee
+    /// @return Amount after fee
+    function _handleFee(uint256 amount) internal returns (uint256) {
+        uint256 feeAmount = (amount * fee) / BASE_FEE;
+
+        // If there is a fee, transfer it to the feeCollector
+        if (feeAmount > 0) {
+            IERC20(rewardToken).transfer(feeCollector, feeAmount);
```
shahthepro, line 234: nit: let's add a event for fee transfers

contracts/contracts/strategies/CurvePoolBooster.sol
clement-ux: Done in this commit: https://github.com/OriginProtocol/origin-dollar/pull/2327/commits/2b8e459de64f16e21708ce48ebbeb4b50b8afd48

contracts/contracts/strategies/CurvePoolBooster.sol
clement-ux: Done in this commit: https://github.com/OriginProtocol/origin-dollar/pull/2327/commits/345cbb92c3199560f269765a3b39dd4ca5854350

contracts/contracts/strategies/CurvePoolBooster.sol
```diff
@@ -0,0 +1,309 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import { Initializable } from "../utils/Initializable.sol";
+import { Strategizable } from "../governance/Strategizable.sol";
+import { ICampaingRemoteManager } from "../interfaces/ICampaignRemoteManager.sol";
```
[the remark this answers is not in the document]
clement-ux: Thanks for flagging it!
https://github.com/OriginProtocol/origin-dollar/pull/2327/commits/efc34e70fecafe8dab0955e08bca39454a38dc86

contracts/contracts/strategies/CurvePoolBooster.sol
clement-ux: https://github.com/OriginProtocol/origin-dollar/pull/2327/commits/8f1422470326d83057b3003093dbb155487253ee

contracts/contracts/strategies/CurvePoolBooster.sol
clement-ux: We already talked about it in the past ([here](https://github.com/OriginProtocol/origin-dollar/pull/2327#discussion_r1892898329)) and I still think that this is better to keep it not payable and have some ETH in the contract to pay the transaction. 
Because we will pay the tx, the surplus will be sent to the contract (not the sender), so the contract will have ETH either way.

contracts/contracts/strategies/CurvePoolBooster.sol
clement-ux: Confirmed by StakeDAO team, both mainnet and L2 👍

contracts/contracts/strategies/CurvePoolBooster.sol
clement-ux: Good catch! https://github.com/OriginProtocol/origin-dollar/pull/2327/commits/aeeac7c84a0f3fdb922a74daf6510fd5c9a92588

contracts/contracts/strategies/CurvePoolBooster.sol
```diff
@@ -0,0 +1,315 @@
...
+
+        // Cache current rewardToken balance
+        uint256 balance = IERC20(rewardToken).balanceOf(address(this));
+        require(balance > 0, "No reward to manage");
+
+        // Handle fee (if any)
+        balance = _handleFee(balance);
+
+        // Approve the balance to the campaign manager
+        IERC20(rewardToken).approve(campaignRemoteManager, balance);
```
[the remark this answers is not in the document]
clement-ux: Done in this commit: https://github.com/OriginProtocol/origin-dollar/pull/2327/commits/4cb5e60fb71510d7391dcde243cb21499267c53c.

shahthepro (commented):
## Requirements

## Easy Checks

#### Authentication

- [x] Never use tx.origin

#### Ethereum

    - :green_circle: Contract is expected to receive and hold ETH for the bridge fee

#### Cryptographic code

#### Gas problems

    - [ ] A way to remove items
    - [x] No `for` loops

#### Black magic

#### Overflow

#### Proxy

#### Events

## Medium Checks

#### Rounding and casts
    - :white_circle: `handleFee` doesn't round the fee in favor of the protocol. But it's neglible.

#### Dependencies

#### External calls

  - :white_circle: Not validated in constructor. But it's fine since governance proposal and deployment will be reviewed by the team.
    - :yellow_circle: Methods that are called by the governor and strategist are not protected by reentrancy guards. Shouldn't be a problem since we are trusting the 2/8 more to handle a lot more.
- [x] No malicious behaviors
- [ ] Oracles, one of:
  - [x] No oracles
  - [ ] Oracles can't be bent

#### Tests

#### Deploy

## Strategy Specific

#### Downstream

## Thinking

#### Logic

  The contract mostly acts an interface for the campaign manager with some added constraints to make sure the funds are bridged correctly and safely.

- ~~:red_circle: Deployment file makes the strategist the governor~~
- :yellow_circle: `initialize` doesn't have a `onlyGovernor` modifier. But it's probably okay since if someone initializes with malicious params, we will know it during deploy reviews.
- :yellow_circle: Since the code involves bridges and L2, it's hard to test this on fork. We should do a test with a small amount of funds once deployed.

#### Internal State

State is pretty simple and only a few variables are mutable. Only governor can change the mutable ones.

#### Attack
  - :yellow_circle: Not attack but failures on the bridge are not handled. Would be great to ensure that we can always retry when the bridge action fails

#### Flavor

Code is simple and easy to understand.

contracts/contracts/strategies/CurvePoolBooster.sol
clement-ux: From what I understand, no. The only reason we might need it could be to close a campaign, so 24 weeks after the campaign ends. However, as this implies token bridging and so on, that is something that's not available on the `ICampaignRemoteManager`. 
As it should not happen really often, we can ask the StakeDAO team to close it for us.

shahthepro (approved):

contracts/contracts/strategies/CurvePoolBooster.sol
```diff
@@ -0,0 +1,318 @@
...
+        uint256 maxRewardPerVote,
+        address[] calldata blacklist,
+        uint256 bridgeFee,
+        uint256 additionalGasLimit
+    ) external onlyGovernorOrStrategist {
+        require(campaignId == 0, "Campaign already created");
+        require(numberOfPeriods > 1, "Invalid number of periods");
+        require(maxRewardPerVote > 0, "Invalid reward per vote");
+
+        // Cache current rewardToken balance
```
sparrowDom, line 96: below 4 lines are identical to the ones in `manageTotalRewardAmount` perhaps they should be in a separate function? Also `IERC20(rewardToken).balanceOf(address(this)); ` returns a  rewardToken balance and then gets overwritten when calling `_handleFee(balance);` by deducting the fee. I think it would be better to have 1 function (can have the same name) `_handleFee` that would do: 
```
handleFee() returns (uint256 balanceWithoutFee){
  uint256 balance = IERC20(rewardToken).balanceOf(address(this));
  require(balance > 0, "No reward to manage");

   // Previously called _handleFee
  balanceWithoutFee = _transferFee(balance);

  // Approve the balance to the campaign manager
  IERC20(rewardToken).safeApprove(campaignRemoteManager, balance);
}

//and then call it in 2 places
uint256 balanceSubFee = _handleFee(balance);

```

sparrowDom (commented):
Added a few comments. Not yet done with the review

contracts/contracts/strategies/CurvePoolBooster.sol
```diff
@@ -0,0 +1,318 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import { Initializable } from "../utils/Initializable.sol";
+import { Strategizable } from "../governance/Strategizable.sol";
+import { ICampaignRemoteManager } from "../interfaces/ICampaignRemoteManager.sol";
+
+contract CurvePoolBooster is Initializable, Strategizable {
```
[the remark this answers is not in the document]
clement-ux: Fixed in this commit: https://github.com/OriginProtocol/origin-dollar/pull/2327/commits/3bc438c170f1b2f7c5e46c6b689396dccccac3a3

contracts/contracts/strategies/CurvePoolBooster.sol
```diff
@@ -0,0 +1,318 @@
...
+        _setFee(_fee);
+        _setFeeCollector(_feeCollector);
+    }
+
+    ////////////////////////////////////////////////////
+    /// --- MUTATIVE FUNCTIONS
+    ////////////////////////////////////////////////////
+    /// @notice Create a new campaign on VotemarketV2
+    /// @dev This will use all token available in this contract
+    /// @param numberOfPeriods Duration of the campaign
```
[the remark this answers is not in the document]
clement-ux: Fixed in this commit: https://github.com/OriginProtocol/origin-dollar/pull/2327/commits/3bc438c170f1b2f7c5e46c6b689396dccccac3a3

contracts/contracts/strategies/CurvePoolBooster.sol
```diff
@@ -0,0 +1,318 @@
...
+    /// @param maxRewardPerVote Maximum reward per vote
```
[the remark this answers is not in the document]
clement-ux: Fixed in this commit: https://github.com/OriginProtocol/origin-dollar/pull/2327/commits/3bc438c170f1b2f7c5e46c6b689396dccccac3a3

contracts/contracts/strategies/CurvePoolBooster.sol
clement-ux: Indeed! Fixed in this commit: https://github.com/OriginProtocol/origin-dollar/pull/2327/commits/f0b9cb40d8e6d16efcf4b4f67af76a6a509759cc.

contracts/contracts/strategies/CurvePoolBooster.sol
```diff
@@ -0,0 +1,334 @@
...
+            targetChainId,
+            additionalGasLimit
+        );
+
+        emit TotalRewardAmountUpdated(balanceSubFee);
+    }
+
+    /// @notice Manage the number of periods of the campaign
+    /// @dev This function should be called after the campaign is created
+    /// @param extraNumberOfPeriods Number of additional periods (cannot be 0)
```
sparrowDom, line 183: maybe add 
`param extraNumberOfPeriods Number of additional periods (cannot be 0) that will be added to already existing amount of periods`

contracts/contracts/strategies/CurvePoolBooster.sol
```diff
@@ -0,0 +1,334 @@
...
+                maxRewardPerVote: newMaxRewardPerVote
+            }),
+            targetChainId,
+            additionalGasLimit
+        );
+
+        emit RewardPerVoteUpdated(newMaxRewardPerVote);
+    }
+
+    /// @notice calculate the fee amount and transfer it to the feeCollector
```
sparrowDom, line 243: I would change this to:
`/// @notice Take the balance of rewards tokens owned by this contract and calculate the fee amount. Transfer the fee to the feeCollector and return the balance with the fee subtracted`

contracts/contracts/strategies/CurvePoolBooster.sol
```diff
@@ -0,0 +1,334 @@
...
+        require(campaignId != 0, "Campaign not created");
+
+        // Handle fee (if any)
+        uint256 balanceSubFee = _handleFee();
+
+        // Approve the total reward amount to the campaign manager
+        IERC20(rewardToken).safeApprove(campaignRemoteManager, balanceSubFee);
+
+        // Manage the campaign
+        ICampaignRemoteManager(campaignRemoteManager).manageCampaign{
```
sparrowDom, line 164: Maybe we can link Votemarket source code in the comments above the `manageCampaign` call as it might be useful to see how these updates are handled: 
https://github.com/stake-dao/votemarket-v2/blob/main/packages/votemarket/src/Votemarket.sol#L668

contracts/contracts/strategies/CurvePoolBooster.sol
```diff
@@ -0,0 +1,334 @@
...
+/// @author Origin Protocol
+/// @notice Contract to manage interactions with VotemarketV2 for a dedicated Curve pool/gauge.
+contract CurvePoolBooster is Initializable, Strategizable {
+    using SafeERC20 for IERC20;
+
+    ////////////////////////////////////////////////////
+    /// --- CONSTANTS && IMMUTABLES
+    ////////////////////////////////////////////////////
+    /// @notice Base fee for the contract, 100%
+    uint16 public constant BASE_FEE = 10_000;
```
sparrowDom, line 20: nit: FEE_BASE might be a more organic name for this var. BASE_FEE might indicate that this is some base fee value, where some additional fees can be applied. Where FEE_BASE sounds more like BASE digits for the FEE. Feel free to disagree though :)

contracts/contracts/strategies/CurvePoolBooster.sol
```diff
@@ -0,0 +1,334 @@
...
+    ) external onlyGovernorOrStrategist {
+        require(campaignId == 0, "Campaign already created");
+        require(numberOfPeriods > 1, "Invalid number of periods");
+        require(maxRewardPerVote > 0, "Invalid reward per vote");
+
+        // Handle fee (if any)
+        uint256 balanceSubFee = _handleFee();
+
+        // Approve the balanceSubFee to the campaign manager
+        IERC20(rewardToken).safeApprove(campaignRemoteManager, balanceSubFee);
```
sparrowDom, line 121: 🟠  safeApprove [first requires](https://github.com/OpenZeppelin/openzeppelin-contracts/blob/v4.6.0/contracts/token/ERC20/utils/SafeERC20.sol#L54) the reset of the approval to 0 before setting a new one. See examples in the code: https://github.com/search?q=repo%3AOriginProtocol%2Forigin-dollar%20SafeApprove&type=code

contracts/contracts/strategies/CurvePoolBooster.sol
```diff
@@ -0,0 +1,334 @@
...
+    /// --- MUTATIVE FUNCTIONS
+    ////////////////////////////////////////////////////
+    /// @notice Create a new campaign on VotemarketV2
+    /// @dev This will use all token available in this contract
+    /// @param numberOfPeriods Duration of the campaign in weeks
+    /// @param maxRewardPerVote Maximum reward per vote to distribute, to avoid overspending
+    /// @param blacklist  List of addresses to exclude from the campaign
+    /// @param bridgeFee Fee to pay for the bridge
+    /// @param additionalGasLimit Additional gas limit for the bridge
+    function createCampaign(
```
sparrowDom, line 106: this one should have `nonReentrant` on it

contracts/contracts/strategies/CurvePoolBooster.sol
```diff
@@ -0,0 +1,334 @@
...
+
+        emit BribeCreated(gauge, rewardToken, maxRewardPerVote, balanceSubFee);
+    }
+
+    /// @notice Manage the total reward amount of the campaign
+    /// @dev This function should be called after the campaign is created
+    /// @dev This will use all the token available in this contract
+    /// @param bridgeFee Fee to pay for the bridge
+    /// @param additionalGasLimit Additional gas limit for the bridge
+    function manageTotalRewardAmount(
```
sparrowDom, line 151: Also put `nonReentrant` on this one
(I know it is permissioned so highly unlikely to be re-entrency attacked)

contracts/contracts/strategies/CurvePoolBooster.sol
```diff
@@ -0,0 +1,334 @@
...
+
+        emit TotalRewardAmountUpdated(balanceSubFee);
+    }
+
+    /// @notice Manage the number of periods of the campaign
+    /// @dev This function should be called after the campaign is created
+    /// @param extraNumberOfPeriods Number of additional periods (cannot be 0)
+    /// @param bridgeFee Fee to pay for the bridge
+    /// @param additionalGasLimit Additional gas limit for the bridge
+    function manageNumberOfPeriods(
```
sparrowDom, line 186: `nonReentrant` on this one

contracts/contracts/strategies/CurvePoolBooster.sol
```diff
@@ -0,0 +1,334 @@
...
+
+        emit NumberOfPeriodsUpdated(extraNumberOfPeriods);
+    }
+
+    /// @notice Manage the reward per vote of the campaign
+    /// @dev This function should be called after the campaign is created
+    /// @param newMaxRewardPerVote New maximum reward per vote
+    /// @param bridgeFee Fee to pay for the bridge
+    /// @param additionalGasLimit Additional gas limit for the bridge
+    function manageRewardPerVote(
```
sparrowDom, line 217: `nonReentrant` on this one

contracts/contracts/strategies/CurvePoolBooster.sol
```diff
@@ -0,0 +1,334 @@
...
+        onlyGovernorOrStrategist
+    {
+        campaignId = _campaignId;
+        emit CampaignIdUpdated(_campaignId);
+    }
+
+    /// @notice Rescue ETH from the contract
+    /// @dev Only callable by the governor or strategist
+    /// @param receiver Address to receive the ETH
+    function rescueETH(address receiver) external onlyGovernorOrStrategist {
```
sparrowDom, line 284: Just because it is transferring ETH I would put `nonReentrant` on this one

contracts/contracts/strategies/CurvePoolBooster.sol
```diff
@@ -0,0 +1,334 @@
...
+        uint256 balance = address(this).balance;
+        (bool success, ) = receiver.call{ value: balance }("");
+        require(success, "Transfer failed");
+        emit TokensRescued(address(0), balance, receiver);
+    }
+
+    /// @notice Rescue ERC20 tokens from the contract
+    /// @dev Only callable by the governor or strategist
+    /// @param token Address of the token to rescue
+    function rescueToken(address token, address receiver)
```
sparrowDom, line 295: and maybe `nonReentrant` on this one

contracts/contracts/strategies/CurvePoolBooster.sol
```diff
@@ -0,0 +1,334 @@
...
+        uint256 feeAmount = (balance * fee) / BASE_FEE;
+
+        // If there is a fee, transfer it to the feeCollector
+        if (feeAmount > 0) {
+            // Transfer the fee to the feeCollector
+            IERC20(rewardToken).transfer(feeCollector, feeAmount);
+
+            emit FeeCollected(feeCollector, feeAmount);
+
+            // Return the balance after fee
```
sparrowDom, line 259: there are tokens (like ours) which don't have a perfect up to a WEI exact accounting when doing transfers and they might be doing some rounding. It could happen that when you transfer a `feeAmount` away, that the actual balance remaining won't be `balance - feeAmount` but rather `balance - feeAmount - 1 WEI`

For that reason I would rather return `IERC20(rewardToken).balanceOf(address(this));` here

sparrowDom: Otherwise the whole transaction amount that was about to be transferred over to the other chain might fail

contracts/contracts/strategies/CurvePoolBooster.sol
DanielVF: We should never be using safe approve. It's not needed, and just adds the extra requirement locally to first reset to zero, which we then work around by first setting to zero.

contracts/contracts/strategies/CurvePoolBooster.sol
DanielVF: Doing the IERC20(rewardToken).balanceOf(address(this)) is good.

contracts/test/strategies/curvePoolBooster.mainnet.fork-test.js
```diff
@@ -0,0 +1,427 @@
...
+    );
+  });
+
+  it("Should Create a campaign", async () => {
+    const { curvePoolBooster, ousd, wousd } = fixture;
+    const { strategistAddr } = await getNamedAccounts();
+    const woethSigner = await impersonateAndFund(wousd.address);
+    const sStrategist = await ethers.provider.getSigner(strategistAddr);
+
+    // Deal OETH and ETH to pool booster
```
sparrowDom, line 46: below 20 lines get repeated 4 times in the test file. Separating it into a function will improve the readability

contracts/test/strategies/curvePoolBooster.mainnet.fork-test.js
```diff
@@ -0,0 +1,427 @@
...
+    const { strategistAddr } = await getNamedAccounts();
+    const sStrategist = await ethers.provider.getSigner(strategistAddr);
+
+    await curvePoolBooster.connect(sStrategist).setCampaignId(12);
+
+    await expect(
+      curvePoolBooster
+        .connect(sStrategist)
+        .manageNumberOfPeriods(0, parseUnits("0.1"), 0)
+    ).to.be.revertedWith("Invalid number of periods");
```
sparrowDom, line 377: `createCampaign` could also use a `Invalid number of periods` require test

contracts/test/strategies/curvePoolBooster.mainnet.fork-test.js
```diff
@@ -0,0 +1,427 @@
...
+    const { strategistAddr } = await getNamedAccounts();
+    const sStrategist = await ethers.provider.getSigner(strategistAddr);
+
+    await curvePoolBooster.connect(sStrategist).setCampaignId(12);
+
+    await expect(
+      curvePoolBooster
+        .connect(sStrategist)
+        .manageRewardPerVote(0, parseUnits("0.1"), 0)
+    ).to.be.revertedWith("Invalid reward per vote");
```
sparrowDom, line 391: this require can be tested with `createCampaign` as well

contracts/contracts/strategies/CurvePoolBooster.sol
sparrowDom, line 295: there is no test for this function yet

sparrowDom:
## Requirements

PR is integrating with `CampaignRemoteManager` mainnet contract which communicates with the Votemarket contract that is located on L2. It is used to create and manage the incentives campaign.

## Easy Checks

#### Authentication

- [x] Never use tx.origin

#### Ethereum

  - [x] it does as it needs to pay the bridge fees
  - [x] it does as it needs to be able to receive ETH

#### Cryptographic code

#### Gas problems

- [x] no for loops

#### Black magic

#### Overflow

  - [x] no for loops

#### Proxy

#### Events

## Medium Checks

#### Rounding and casts
- [ ] Safecast is aways used when casting - no casting

#### Dependencies
- [ ] ~~If OpenZeppelin ACL roles are use review & enumerate all of them~~

#### External calls

  - [x] They will be on deployment review
- [🟠] No unsafe external calls
  - [ ] SafeApprove requires resetting the allowance to 0 before setting a non 0 allowance
- [🟡] Reentrancy guards on all state changing functions
    - most functions that interact with external contract are permissioned so they should be safe from re-entrency. Still it doesn't hurt to add re-entrency on it. 
- [x] No malicious behaviors
- [ ] Oracles, one of:
  - [x] No oracles
  - [ ] Oracles can't be bent
- [🔴] Do not call balanceOf for external contracts to determine what they will do when they use internal accounting
  - [ ] when calculating the balance without fee in `_handleFee()` we should query the contract for the balance once again. As rounding (e.g. in 4626 type of ERC20 contracts) can not assure that `balance - X` will be up to 1 WEI accurate. 

#### Tests

  - [🟡] Each publicly callable method has a test
    - some are missing
  - [🟡] Each require() has a test
    - some are missing
  - ~~If tests interact with AMM make sure enough edge cases (pool tilts) are tested. Ideally with fuzzing.~

#### Deploy

#### Downstream

## Thinking

#### Logic

 Logic is simple as it is a simple interface to a campaignRemoteManager. The one downside is that there is some manual managing of campaignId setting, but there is no other way to do this one. And we will probably need to do it only once, or very rarely.

We should be monitoring all the bridge transactions that happen as a result of transactions.

#### Internal State

- Stored state represents configurable properties.
- After campaign is created the campaignId must be set. And after that all the manage campaign functions can be called.

#### Attack

Biggest risk here is bridge transactions failing as we will need to be able to detect that off-chain.

#### Flavor

Code is nice and simple.

contracts/contracts/strategies/CurvePoolBooster.sol
clement-ux: Done in this commit: https://github.com/OriginProtocol/origin-dollar/pull/2327/commits/49a88dd301b0bf05cdd01299a6b9104d2259e64a

contracts/contracts/strategies/CurvePoolBooster.sol
clement-ux: Done in this commit: https://github.com/OriginProtocol/origin-dollar/pull/2327/commits/49a88dd301b0bf05cdd01299a6b9104d2259e64a

contracts/contracts/strategies/CurvePoolBooster.sol
clement-ux: Done in this commit: https://github.com/OriginProtocol/origin-dollar/pull/2327/commits/f9d4d9b2bd9d52b2fbfd6f6d660e5339b344bda6

contracts/contracts/strategies/CurvePoolBooster.sol
clement-ux: Done in this commit: https://github.com/OriginProtocol/origin-dollar/pull/2327/commits/8d33a72e032aa3dcc14f55caa38a9b88652b2f9f
(Done for all reentrancy comments)

contracts/contracts/strategies/CurvePoolBooster.sol
clement-ux: Done in this commit: https://github.com/OriginProtocol/origin-dollar/pull/2327/commits/82770e69ab5ff1e902f0d597dbd19ca0c21781d1

contracts/contracts/strategies/CurvePoolBooster.sol
sparrowDom: I we want to support USDT it would be best to add another approve(0): 
```
// Approve the balanceSubFee to the campaign manager
IERC20(rewardToken).safeApprove(campaignRemoteManager, 0);
IERC20(rewardToken).safeApprove(campaignRemoteManager, balanceSubFee);
```

sparrowDom: done in this commit: https://github.com/OriginProtocol/origin-dollar/pull/2327/commits/4ddbbb297901958c56d0c5de1ffc2cbd843cc6b4

contracts/contracts/strategies/CurvePoolBooster.sol
sparrowDom: I think it would be better to replace the `return balance - feeAmount` with the `IERC20(rewardToken).balanceOf(address(this));`. Since if `feeAmount == 0` then we query the `rewardsToken` contract for balance twice wasting some gas

sparrowDom (approved):

contracts/deploy/mainnet/120_pool_booster_curve_arb.js
```diff
@@ -0,0 +1,126 @@
...
+      "CurvePoolBooster",
+      addresses.zero
+    );
+    const cCurvePoolBoosterProxy = await ethers.getContractAt(
+      "CurvePoolBoosterProxy",
+      addresses.zero
+    );
+
+    console.log("\nStarting deployment using CreateX");
+    const rewardToken = addresses.mainnet.OUSDProxy;
```
naddison36, line 40: this the reward token on Arbitrum the mainnet OUSD contract? Or should it be the a token on the Arbitrum chain?
There is no bridged OUSD on Arbitrum.Only Wrapped OUSD (wOUSD) is bridged

contracts/deploy/mainnet/120_pool_booster_curve_arb.js
```diff
@@ -0,0 +1,126 @@
...
+    const gauge = addresses.mainnet.CurveOUSDUSDTGauge;
```
naddison36, line 41: Is this correct? The gauge for the Arbitrum Pool Booster is the mainnet Curve gauge?

naddison36 (commented):
I'm confused to how the Pool Boosting will work on Arbitrum.

clement-ux:
> I'm confused to how the Pool Boosting will work on Arbitrum.

Pool Boosting isn't designed to work on Arbitrum. I know it is weird, but let me explain.
This contract will be deployed at the same address as the one in mainnet for two main reasons:
- Preventing someone else to deploy it and thus manage our bribe in our stead.
- Claiming unused tokens when campaign ends.

So at the end, on Arbitrum this contract will only use the `rescueToken` function and isn't planned to do more. 
At least for the moment.

naddison36:
> > I'm confused to how the Pool Boosting will work on Arbitrum.
> 
> Pool Boosting isn't designed to work on Arbitrum. I know it is weird, but let me explain. This contract will be deployed at the same address as the one in mainnet for two main reasons:
> 
> * Preventing someone else to deploy it and thus manage our bribe in our stead.
> * Claiming unused tokens when campaign ends.
> 
> So at the end, on Arbitrum this contract will only use the `rescueToken` function and isn't planned to do more. At least for the moment.
> 
> We could have develop two different contracts, to manage bribe on mainnet and L2 at the same time. But using create2 requires to have same bytecode for both to have cross-chain replication. For the future we could use Create3 method.

ok, that explains it. Thanks

naddison36 (dismissed):

contracts/contracts/strategies/CurvePoolBooster.sol
```diff
@@ -0,0 +1,415 @@
...
+        // https://github.com/stake-dao/votemarket-v2/blob/main/packages/votemarket/src/Votemarket.sol#L668
+        ICampaignRemoteManager(campaignRemoteManager).manageCampaign{
+            value: bridgeFee
+        }(
+            ICampaignRemoteManager.CampaignManagementParams({
+                campaignId: campaignId,
+                rewardToken: rewardToken,
+                numberOfPeriods: 0,
+                totalRewardAmount: balanceSubFee,
+                maxRewardPerVote: 0
```
naddison36, line 189: Does setting the `maxRewardPerVote` to zero mean the previous `maxRewardPerVote` is left unchanged?

contracts/contracts/strategies/CurvePoolBooster.sol
```diff
@@ -0,0 +1,415 @@
...
+
+        // Manage the campaign
+        // https://github.com/stake-dao/votemarket-v2/blob/main/packages/votemarket/src/Votemarket.sol#L668
+        ICampaignRemoteManager(campaignRemoteManager).manageCampaign{
+            value: bridgeFee
+        }(
+            ICampaignRemoteManager.CampaignManagementParams({
+                campaignId: campaignId,
+                rewardToken: rewardToken,
+                numberOfPeriods: 0,
```
naddison36, line 187: Does setting the `numberOfPeriods` to zero mean the previous `numberOfPeriods` is left unchanged?

contracts/contracts/strategies/CurvePoolBooster.sol
clement-ux: Yes exactly 👍

contracts/contracts/strategies/CurvePoolBooster.sol
clement-ux: Yes exactly 👍

contracts/contracts/strategies/CurvePoolBooster.sol
```diff
@@ -0,0 +1,415 @@
...
+        // Cache current rewardToken balance
+        uint256 balance = IERC20(rewardToken).balanceOf(address(this));
+        require(balance > 0, "No reward to manage");
+
+        uint256 feeAmount = (balance * fee) / FEE_BASE;
+
+        // If there is a fee, transfer it to the feeCollector
+        if (feeAmount > 0) {
+            // Transfer the fee to the feeCollector
+            IERC20(rewardToken).transfer(feeCollector, feeAmount);
```
naddison36, line 298: Will we ever bribe with USDT? If so, we'll need to use `safeTransfer`.
If not, the Slither error can be fixed by adding this before the `transfer`
`// slither-disable-next-line unchecked-transfer unused-return`

clement-ux: Fixed in this commit: https://github.com/OriginProtocol/origin-dollar/pull/2327/commits/6997f35b6796ad46f68c91e869179ef98724f058.

PR #2332, opened 2024-12-20
Add support for Sonic chain

sparrowDom:
Add support for Sonic chain

**Current issues:**
- hardhat tracer not working
  - in console run `DEBUG=hardhat-tracer:* TRACE=true yarn run node`
  - in `contracts/node.sh` add ` --verbose` parameter to line 64 just after `hardhat node` (`hardhat node --verbose`)
   this is the error hardhat reports
Screenshot 2024-12-27 at 15 48 52
Relevant error trace: 
    - [which is failing on this line ](https://github.com/NomicFoundation/hardhat/blob/c34a43ce2027ec5196cac115e8b6494ee296e5ec/packages/hardhat-core/src/internal/hardhat-network/stack-traces/vm-trace-decoder.ts#L28)
    - and [this line](https://github.com/NomicFoundation/hardhat/blob/c34a43ce2027ec5196cac115e8b6494ee296e5ec/packages/hardhat-core/src/internal/core/providers/construction.ts#L85)
    - and the originating error [this line](https://github.com/NomicFoundation/edr/blob/main/crates/edr_solidity/src/source_map.rs#L162)

and these are the logs from the tracer: 
Screenshot 2024-12-27 at 15 51 03
it seems to fail on this line [within the tracer](https://github.com/zemse/hardhat-tracer/blob/bc75d49be050476d01f078489ba2a50823af6faa/src/extend/hre.ts#L74)

To me it looks like `hardhat` plugin has issues setting up network tracing provider and the `hardhat-tracer` can not attack to it. 
  - I've tried lowering the version of `hardhat` but then we quickly get the `eip-4895` error which prevents us from running the sonic chain
- unit/fork tests failing
- CI failing


contracts/hardhat.config.js
```diff
@@ -179,6 +187,13 @@ module.exports = {
       live: true,
       saveDeployments: true,
     },
+    sonic: {
+      url: sonicProviderUrl,
+      accounts: [
```
shahthepro, line 192: needs to have `live` and `saveDeployments` params set to true

contracts/hardhat.config.js
```diff
@@ -227,6 +243,7 @@ module.exports = {
       mainnet: MAINNET_GOVERNOR,
       holesky: HOLESKY_DEPLOYER, // on Holesky the deployer is also the governor
       base: BASE_GOVERNOR,
+      sonic: MAINNET_GOVERNOR,
```
shahthepro, line 246: nit: We don't have a governor yet, so let's set it to index 1 for now

contracts/package.json
```diff
@@ -65,6 +73,10 @@
     "@rollup/plugin-commonjs": "^25.0.7",
     "@rollup/plugin-json": "^6.1.0",
     "@rollup/plugin-node-resolve": "^15.2.3",
+    "@typechain/ethers-v6": "^0.5.0",
```
shahthepro, line 76: This is a big nope. Ethers V6 got rid of BigNumber. We use it everywhere on our repo. I know this is just types but let's not complicate our dependencies and end up breaking something

contracts/package.json
```diff
@@ -43,14 +46,19 @@
...
+    "test:coverage:sonic-fork": "REPORT_COVERAGE=true FORK_NETWORK_NAME=sonic ./fork-test.sh"
   },
   "author": "Origin Protocol Inc <support@originprotocol.com>",
   "license": "MIT",
   "devDependencies": {
     "@aws-sdk/client-kms": "^3.598.0",
     "@aws-sdk/client-s3": "^3.600.0",
+    "@nomicfoundation/hardhat-chai-matchers": "^2.0.0",
+    "@nomicfoundation/hardhat-ethers": "^3.0.0",
+    "@nomicfoundation/hardhat-ignition-ethers": "^0.15.0",
```
shahthepro, line 59: Any reason why these new dependencies were added? Would prefer to get rid of them if those are accidental

contracts/hardhat.config.js
shahthepro: Or deployer address

contracts/package.json
sparrowDom: yup thanks removing this

contracts/hardhat.config.js
sparrowDom: good catch thanks

contracts/hardhat.config.js
sparrowDom: yup thanks changed

naddison36:
closing as we are now using `nicka/sonic`

PR #2333, opened 2024-12-23
Simple harvester for OETH

clement-ux:
# Simple Harvester
This PR aims to deploy a simple harvester for the OETH AMO to replace the current. 
This new harvester is simpler, as it will only harvest reward and send it to the strategist.

Note: strategist and operator addresses are set to the deployer address but must be changed.

## Code Change Checklist

- [ ]  Unit tests *if needed

Internal review:

## Deploy checklist

```
```


contracts/contracts/harvest/OETHHarvesterSimple.sol
```diff
@@ -0,0 +1,107 @@
...
+    ////////////////////////////////////////////////////
+    /// --- CONSTRUCTOR
+    ////////////////////////////////////////////////////
+    constructor(
+        address _governor,
+        address _operator,
+        address _strategist
+    ) {
+        require(_strategist != address(0), "Invalid strategist");
+        operator = _operator;
```
DanielVF, line 44: These should emit events so we have a full even trace of all strategist and operators.

contracts/contracts/harvest/OETHHarvesterSimple.sol
```diff
@@ -0,0 +1,107 @@
...
+            }
+        }
+    }
+
+    ////////////////////////////////////////////////////
+    /// --- GOVERNANCE
+    ////////////////////////////////////////////////////
+    function setOperator(address _operator) external onlyGovernor {
+        operator = _operator;
+        emit OperatorSet(_operator);
```
sparrowDom, line 91: nit: `OperatorChanged ` might have a better ring to it

contracts/contracts/harvest/OETHHarvesterSimple.sol
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    function setStrategyStatus(address _strategy, bool _status)
+        external
+        onlyGovernor
```
DanielVF, line 96: Maybe this could be strategist only?

This does mean that the strategist could make this contract make arbitrary calls to `IERC20(address).transfer(strategist, balance);`, but that shouldn't do any damage, given that this contract is basicly just an extension of the strategist powers in the first place, all funds go to the strategist anyway, and this should not be holding money or user funds?

contracts/contracts/harvest/OETHHarvesterSimple.sol
```diff
@@ -0,0 +1,107 @@
...
+    {
+        isAuthorized[_strategy] = _status;
+        emit StrategyStatusSet(_strategy, _status);
```
sparrowDom, line 99: nit: `StrategyStatusSet` to `StrategyStatusChanged`. Same for `StrategistSet` below.

contracts/contracts/harvest/OETHHarvesterSimple.sol
```diff
@@ -0,0 +1,107 @@
...
+
+    ////////////////////////////////////////////////////
+    /// --- GOVERNANCE
+    ////////////////////////////////////////////////////
+    function setOperator(address _operator) external onlyGovernor {
+        operator = _operator;
+        emit OperatorSet(_operator);
+    }
+
+    function setStrategyStatus(address _strategy, bool _status)
```
DanielVF, line 94: `Status` is somewhat ambiguous here, both as a part of the the function name, and as variable. Perhaps it should be something more to do with authorization.

contracts/contracts/harvest/OETHHarvesterSimple.sol
```diff
@@ -0,0 +1,107 @@
...
+        external
+        onlyGovernor
+    {
```
sparrowDom, line 97: nit you can revert on zero address for `_strategy `

contracts/contracts/harvest/OETHHarvesterSimple.sol
```diff
@@ -0,0 +1,107 @@
...
+                IERC20(rewardTokens[i]).transfer(strategist, balance);
+                emit Harvested(rewardTokens[i], balance);
+            }
+        }
+    }
+
+    ////////////////////////////////////////////////////
+    /// --- GOVERNANCE
+    ////////////////////////////////////////////////////
+    function setOperator(address _operator) external onlyGovernor {
```
sparrowDom, line 89: you can also revert on zero address for operator

contracts/contracts/harvest/OETHHarvesterSimple.sol
```diff
@@ -0,0 +1,107 @@
...
+        onlyOperator
+    {
+        for (uint256 i = 0; i < _strategies.length; i++) {
+            _harvestAndTransfer(_strategies[i]);
+        }
+    }
+
+    function _harvestAndTransfer(address _strategy) internal {
+        // Ensure strategy is authorized
+        require(isAuthorized[_strategy], "Strategy not authorized");
```
sparrowDom, line 67: you can revert on 0 address _strategy here as well

sparrowDom:
@clement-ux can you refresh my memory why the harvester functions are no longer free to be called by anyone and are now locked under the `operator` account?

clement-ux:
> @clement-ux can you refresh my memory why the harvester functions are no longer free to be called by anyone and are now locked under the `operator` account?

Indeed, that is something that I've added in comparison with the current Harvester. But it doesn't make sense to restrict it, so I'll remove the operator role 👍

contracts/contracts/harvest/OETHHarvesterSimple.sol
clement-ux: I've removed the `operator` role.

contracts/contracts/harvest/OETHHarvesterSimple.sol
clement-ux: I've removed the `operator` role.

contracts/contracts/harvest/OETHHarvesterSimple.sol
clement-ux: I've removed the `operator` role.

contracts/contracts/harvest/OETHHarvesterSimple.sol
clement-ux: Agree! Changed in this commit: [23a885f](https://github.com/OriginProtocol/origin-dollar/pull/2333/commits/23a885f542f3dcc22d9a661d269ff382aa5cf19a)

contracts/contracts/harvest/OETHHarvesterSimple.sol
DanielVF: An event is very much about a change? Could probably leave that out of the name?

clement-ux: I've changed the whole wording here as Daniel suggested: [5446bde](https://github.com/OriginProtocol/origin-dollar/pull/2333/commits/5446bde0221c8fca5f6792203e9a1602fa0e7293).

contracts/contracts/harvest/OETHHarvesterSimple.sol
clement-ux: I've changed it with the same wording as the current Harvester in this commit: [5446bde](https://github.com/OriginProtocol/origin-dollar/pull/2333/commits/5446bde0221c8fca5f6792203e9a1602fa0e7293).

contracts/contracts/harvest/OETHHarvesterSimple.sol
clement-ux: Agree! Changed on this commit: [b79578b](https://github.com/OriginProtocol/origin-dollar/pull/2333/commits/b79578b4f0b02d6fcfb99beede7c18b761086c24)

contracts/contracts/harvest/OETHHarvesterSimple.sol
clement-ux: Agree! Changed on this commit: [b79578b](https://github.com/OriginProtocol/origin-dollar/pull/2333/commits/b79578b4f0b02d6fcfb99beede7c18b761086c24).

contracts/contracts/harvest/OETHHarvesterSimple.sol
DanielVF: I think fine to not have an extra require here. The isauthorized covers it, and the later function call would fail regardless.

contracts/contracts/harvest/OETHHarvesterSimple.sol
```diff
@@ -0,0 +1,90 @@
...
+    ////////////////////////////////////////////////////
+    address public strategist;
+    mapping(address => bool) public supportedStrategies;
+
+    ////////////////////////////////////////////////////
+    /// --- EVENTS
+    ////////////////////////////////////////////////////
+    event Harvested(address token, uint256 amount);
+    event StrategistChanged(address strategist);
+    event SupportedStrategyUpdate(address strategy, bool status);
```
DanielVF, line 20: I think our usual format for events that change configuration is to use "Updated" at the end of the event name.

image

DanielVF:
I'll take a pass through this with a checklist tomorrow, but otherwise looks good.

contracts/contracts/harvest/OETHHarvesterSimple.sol
clement-ux: Changed in this commit: [9e968fd](https://github.com/OriginProtocol/origin-dollar/pull/2333/commits/9e968fdd6e9173e3f22a02a9d89ccfccfb53f12f)

contracts/contracts/harvest/OETHHarvesterSimple.sol
```diff
@@ -0,0 +1,90 @@
...
+        require(msg.sender == strategist || isGovernor(), "Not strategist");
+        _;
+    }
+
+    ////////////////////////////////////////////////////
+    /// --- CONSTRUCTOR
+    ////////////////////////////////////////////////////
+    constructor(address _governor, address _strategist) {
+        require(_strategist != address(0), "Invalid strategist");
+        strategist = _strategist;
```
DanielVF, line 35: Need an event on this initial strategist set. Maybe just make it into an internal method that does the validation and the event, and have the internal used by the public method.

clement-ux: Changed in this commit: [00766a9](https://github.com/OriginProtocol/origin-dollar/pull/2333/commits/00766a9838c67a6742a7470e93fe1dd0178db50b).
However, in the `constructor`, the governor is set but doesn't emit an event. Should we emit an event here, too?

DanielVF: Good catch. We should use `_changeGovernor()` instead which does emit an event. Confusing naming.

DanielVF:
# Code review - Simple harvester

## Requirements

We want to be able to route some strategy reward token earnings directly to the strategist, for further allocation/distribution.

This contract allows anyone to harvest, does not pay incentives, and routes all to the strategist. The strategist can control which contracts this can call, while governance controls which strategies allow this contract to call it.
 
## Easy Checks

#### Authentication

- [x] Never use tx.origin

#### Ethereum

- [x] Contract does not have a vulnerablity from being sent self destruct ETH

#### Cryptographic code

_no crypto_

#### Gas problems

    - [ ] A way to remove items
    - [x] Size can be controlled by caller

#### Black magic

_no magic_

#### Overflow

#### Proxy

#### Events
	- 🟠 constructor not setting an event for strategist. Noted in code. 
	   - [x] Fixed

#### Rounding and casts

_No rounding_

#### Dependencies
_No dependancies outside our code._

#### External calls

	- [x] Tokens amounts cannot be manipulated
- [x] Oracles, one of:
  - [x] No oracles
  - [ ] Oracles can't be bent

#### Deploy

#### Logic

_Logic looks correct_

#### Internal State

_No internal state, config only_

#### Attack

This contract will move weekly/daily reward token yield. It will not touch user funds.

If an attacker could change the strategist address, they would receive the yield. However, the strategist address can only be controlled by the governor. The governor can only be changed by the governor.

If the attacker were able to whitelist a malicious strategy this would not allow taking yield, since yield only goes to the strategist, and all coins are emptied after each strategy harvest runs. This could be used as a “clever backdoor” for recovering random coins sent to this contract by accident.

A strategy that fails to be harvested can just be not called on, so that it does not block other harvests.

#### Flavor

Flavor is good.

- [x] Functions follow validate, read, compute, write, call, event, validate

contracts/contracts/harvest/OETHHarvesterSimple.sol
clement-ux: Done in this commit: [534ce82](https://github.com/OriginProtocol/origin-dollar/pull/2333/commits/534ce82b6b1adc56e019447d2ab1edd9254838b8)

DanielVF (approved):

contracts/deploy/mainnet/114_simple_harvester.js
```diff
@@ -0,0 +1,45 @@
...
+    deployerIsProposer: false,
+    // proposalId: "",
+  },
+  async ({ deployWithConfirmation }) => {
+    const { deployerAddr } = await getNamedAccounts();
+
+    // 1. Deploy contract
+    const dOETHHarvesterSimple = await deployWithConfirmation(
+      "OETHHarvesterSimple",
+      [addresses.mainnet.Timelock, deployerAddr] // Need to adjust governor and strategist
```
DanielVF, line 19: This will need to change before deploy / governance.

contracts/contracts/harvest/OETHHarvesterSimple.sol
```diff
@@ -0,0 +1,93 @@
...
+    /// --- EVENTS
+    ////////////////////////////////////////////////////
+    event Harvested(address token, uint256 amount);
+    event StrategistUpdated(address strategist);
+    event SupportedStrategyUpdated(address strategy, bool status);
+
+    ////////////////////////////////////////////////////
+    /// --- MODIFIERS
+    ////////////////////////////////////////////////////
+    modifier onlyStrategist() {
```
sparrowDom, line 25: this should be renamed to `onlyStrategistOrGovernor`

contracts/contracts/harvest/OETHHarvesterSimple.sol
```diff
@@ -0,0 +1,93 @@
...
+        require(msg.sender == strategist || isGovernor(), "Not strategist");
```
sparrowDom, line 26: nit: also this one "Not strategist" to "Not strategist or governor"

contracts/contracts/harvest/OETHHarvesterSimple.sol
sparrowDom: Actually you could further simplify this contract by using `Strategizable`. This way any logic setting strategist won't need to be part of this contract

contracts/contracts/harvest/OETHHarvesterSimple.sol
sparrowDom: in case you decide to use `Strategizable` you can omit this part.

contracts/contracts/harvest/OETHHarvesterSimple.sol
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+    function harvestAndTransfer(address[] calldata _strategies) external {
+        for (uint256 i = 0; i < _strategies.length; i++) {
+            _harvestAndTransfer(_strategies[i]);
+        }
+    }
+
+    function _harvestAndTransfer(address _strategy) internal {
+        // Ensure strategy is supported
```
sparrowDom, line 52: Nit: I think `Assure` or `Assert` will be better here instead of `Ensure`. I think the difference is that ensure is used when the code would need to make sure that something happens / holds true. 

While in this case we rather check for it and revert in case it is not true. A native english speaker will know for sure :) @DanielVF maybe?

contracts/test/harvest/simple-harvester.mainnet.fork-test.js
```diff
@@ -0,0 +1,105 @@
...
+      addresses.mainnet.Timelock
+    );
+
+    expect(
+      await simpleOETHHarvester.supportedStrategies(
+        addresses.mainnet.ConvexOETHAMOStrategy
+      )
+    ).to.be.equal(false);
+    await simpleOETHHarvester
+      .connect(timelock)
```
sparrowDom, line 41: we should also have a test for the `strategist` to set a supported strategy

sparrowDom:
## Requirements

This PR introduces a simplified Harvester which harvests the reward tokens from strategies and without having the ability to swap sends them to the strategist. The strategist has the permissions to pick a set of strategies to be harvested. The governor sets the strategist and approves the harvester on the strategy contract. 

## Easy Checks

#### Authentication

- [x] Never use tx.origin
- [X] Every external/public function is supposed to be externally accessible
- [X] Every external/public function has the correct authentication

#### Ethereum

#### Cryptographic code

- [X] This contract code does not roll it's own crypto.

#### Gas problems

- 🔴  Contracts with for loops must have either:
    - In case tokens get stuck on the contract because for-loop for some reason is not able to complete, we can not really upgrade the contract to fix the problem. I suggest we introduce a similar function we have in the non simplified harvester where a governor can recover any tokens. 
```
   function transferToken(address _asset, uint256 _amount)
        external
        onlyGovernor
    {
        IERC20(_asset).safeTransfer(governor(), _amount);
    }
```
in this case I would simplify it so that the strategist directly recovers the tokens: 
```
   function transferToken(address _asset, uint256 _amount)
        external
        onlyStrategist
    {
        IERC20(_asset).safeTransfer(strategist, _amount);
    }
```

#### Black magic

#### Overflow

#### Proxy

- [ ] ~~No storage variable initialized at definition when contract used as a proxy implementation.~~

#### Events

## Medium Checks

#### Rounding and casts
no rounding

#### Dependencies
no dependencies

#### External calls

- [ ] Reentrancy guards on all state changing functions - not needed
- [x] No malicious behaviors
- [ ] ~~Low level call() must require success.~~
- [ ] ~~No slippage attacks (we need to validate expected tokens received)~~
- [x] Oracles, one of:
  - [x] No oracles
  - [ ] Oracles can't be bent

#### Tests

  - 🟠  Each logical branch has a test
     - There should be a test for `setSupportedStrategy` to be called by the strategist
     - There should be a test for testing `setStrategists` can not be called by non govenor
     - There should be a test for testing `setSupportedStrategy` can not be called by an account that is not the governor nor the strategist
     - Currently the `timelock` in tests is calling `harvestAndTransfer`. There should be a test where a non special account can also call that function
  - [ ] ~~If tests interact with AMM make sure enough edge cases (pool tilts) are tested. Ideally with fuzzing.~~

#### Deploy

- 🔴  Deployer permissions are removed after deploy
  - currently the deployer address is still set as the strategist

## Thinking

#### Logic
Logic looks ok 

Aside from changing the strategist address deployment is ok

#### Internal State

Internal state is just config

#### Attack

This contract doesn't touch user funds, rather the strategy rewards. In worst case if the strategist is compromised then: 
  - all the existing yield up to the compromise event and the yield accrued in the governance procedure (governance proposal duration + timelock (2 day timelock)) could be stolen. 
  - Not related to the contract code, but any funds held by the strategist would also be stolen in such case.
  - strategist could also set invalid strategies as supported, or a malicious strategy which could steal all the harvester & strategist funds. Though if strategist is compromised, then there are simpler ways to steal the funds.

#### Flavor

🟠  code is simple, could be more by using `Strategizable` instead of `Governable`

contracts/contracts/harvest/OETHHarvesterSimple.sol
clement-ux: Migrated to `Strategizable` in this commit:  [639e7ae](https://github.com/OriginProtocol/origin-dollar/pull/2333/commits/639e7aeb82f62add2003cf57a3fbe437cf362de5).

clement-ux:
> 🔴 Deployer permissions are removed after deploy
> - currently the deployer address is still set as the strategist

@sparrowDom Which address should I use for the strategist at the deployment?

contracts/contracts/harvest/OETHHarvesterSimple.sol
```diff
@@ -0,0 +1,83 @@
...
+        // Cache reward tokens
+        address[] memory rewardTokens = IStrategy(_strategy)
+            .getRewardTokenAddresses();
+        for (uint256 i = 0; i < rewardTokens.length; i++) {
+            // Cache balance
+            uint256 balance = IERC20(rewardTokens[i]).balanceOf(address(this));
+            if (balance > 0) {
+                // Transfer to strategist
+                IERC20(rewardTokens[i]).safeTransfer(strategistAddr, balance);
+                emit Harvested(rewardTokens[i], balance);
```
shahthepro, line 60: nit: Let's emit the strategy address here as well?

clement-ux: Yes good idea, done in this commit: [0e8ab277](https://github.com/OriginProtocol/origin-dollar/pull/2333/commits/0e8ab277eb269cbbe30175dc2cdb0084b1d5dc83)

sparrowDom:
@clement-ux you can set it to the mainnet Strategist: https://github.com/OriginProtocol/origin-dollar/blob/master/contracts/hardhat.config.js#L47

DanielVF:
@sparrowDom I don't think we need to worry about gas on the coin loop. The number of coins is not under control, and we can upgrade the strategy if for some reason we had the hundreds of reward tokens it would take to hit gas limits.

contracts/contracts/harvest/OETHHarvesterSimple.sol
DanielVF: I think ensure is flexible enough in English that I didn't think twice about at all on this. In our programming code function names we do use "ensure" to mean that it will check something and correct it if it is not currently correct (Ideally always succeeding), vs require style reverting. 

I don't think that that changing the comment matters here though. It's pretty clear what it means. A perfectly correct comment would be "require" but then we are just duplicating the code. :D

contracts/test/harvest/simple-harvester.mainnet.fork-test.js
```diff
@@ -19,13 +19,15 @@ describe("ForkTest: CurvePoolBooster", function () {
...
   it("Should have correct parameters", async () => {
     const { simpleOETHHarvester } = fixture;
-    const { deployerAddr } = await getNamedAccounts();
+    const { strategistAddr } = await getNamedAccounts();
 
     expect(await simpleOETHHarvester.governor()).to.be.equal(
       addresses.mainnet.Timelock
     );
+
+    console.log("strategistAddr: ", strategistAddr);
```
sparrowDom, line 28: you can probably delete this one

sparrowDom (dismissed):
@DanielVF cool sounds good. 
@clement-ux all the things from my full report have been addressed. I've updated the report. Great job on this one 💪

contracts/test/harvest/simple-harvester.mainnet.fork-test.js
clement-ux: Good catch, indeed 😅

shahthepro (approved):

PR #2298, opened 2024-10-30
Yield Delegation / Rebasing Token Rewrite

sparrowDom:
We are revamping the our rebasing token contract.

The primary objective is to allow delegated yield. Delegated yield allows an account to seemlessly transfer all earned yield to another account.

PR #2168, opened 2024-08-06
OZ Audit changes for OETH Withdrawal Queue

naddison36:
## Audit changes

* https://github.com/OriginProtocol/origin-dollar/pull/2166
* https://github.com/OriginProtocol/origin-dollar/pull/2167
* https://github.com/OriginProtocol/origin-dollar/pull/2169
* https://github.com/OriginProtocol/origin-dollar/pull/2170
* https://github.com/OriginProtocol/origin-dollar/pull/2171
* https://github.com/OriginProtocol/origin-dollar/pull/2172
* https://github.com/OriginProtocol/origin-dollar/pull/2173
* https://github.com/OriginProtocol/origin-dollar/pull/2174
* https://github.com/OriginProtocol/origin-dollar/pull/2175
* https://github.com/OriginProtocol/origin-dollar/pull/2176

The first two changes have already been fixed on `master`.

## Code Change Checklist

To be completed before internal review begins:

- [ ]  The contract code is complete
- [ ]  Executable deployment file
- [ ]  Fork tests that test after the deployment file runs
- [ ]  Unit tests *if needed
- [ ]  The owner has done a [full checklist review](https://github.com/OriginProtocol/security/blob/master/templates/Contract-Code-Review.md) of the code + tests

Internal review:

- [ ] Two approvals by internal reviewers

PR #2335, opened 2024-12-30
Sonic Staking

naddison36:
## Dependencies

* https://github.com/OriginProtocol/origin-dollar/pull/2168
* https://github.com/OriginProtocol/origin-dollar/pull/2298
* https://github.com/OriginProtocol/origin-dollar/pull/2345
* https://github.com/OriginProtocol/origin-dollar/pull/2349
* https://github.com/OriginProtocol/origin-dollar/pull/2348

## Contracts

sonicContracts

### Sonic Staking Strategy

SonicStakingStrategyHierarchy

# Tests

## Sonic Unit Tests

The Sonic unit tests are in `contracts/test/vault/os-vault.sonic.js`.

```
yarn test:sonic
```

## Sonic Fork Tests

The Sonic fork tests are in:
* `contracts/test/vault/vault.sonic.fork-test.js`
* `contracts/test/strategies/sonicStaking.sonic.fork-test.js`
* `contracts/test/zapper/osonic-zapper.sonic.fork-test.js`

```
# start a local Sonic forked node
yarn run node:sonic

# In another terminal, run the Sonic fork tests
yarn test:sonic-fork
```

## Deployment

The Sonic deployment files for launch are
* `contracts/deploy/sonic/001_vault_and_token.js`
* `contracts/deploy/sonic/002_oracle_router.js`
* `contracts/deploy/sonic/003_sonic_staking_strategy.js`

To run the deployment to Sonic

```
yarn run deploy:sonic
```

When deploying to the Tenderly Testnet, change `SONIC_PROVIDER_URL` in your `.env` file to point to the Testnet.

## Code Change Checklist

- [ ]  Unit tests *if needed

Internal review:


contracts/contracts/strategies/sonic/SonicValidatorDelegator.sol
```diff
@@ -0,0 +1,216 @@
...
+        uint256 undelegatedAmount;
+    }
+    /// @notice Mapping of withdrawIds to validatorIds and undelegatedAmounts
+    mapping(uint256 => WithdrawRequest) public withdrawals;
+
+    /// @notice Address of the registrator - allowed to register, exit and remove validators
+    address public validatorRegistrator;
+
+    // For future use
+    uint256[43] private __gap;
```
sparrowDom, line 38: nit: shouldn't this be `uint256[45] private __gap;`

naddison36: yes, I had mistakenly counted the immutables. Thanks

contracts/contracts/strategies/sonic/SonicValidatorDelegator.sol
```diff
@@ -0,0 +1,318 @@
...
+                validatorIds[i]
+            );
+
+            if (rewards > 0) {
+                totalRewards += rewards;
+                ISFC(sfc).restakeRewards(validatorIds[i]);
+            }
+        }
+
+        // TODO use Deposit event or something else?
```
sparrowDom, line 242: Maybe we can add a bool flag `isRestake` to `Deposit` event indicating whether it is a normal deposit or restaking of the rewards

contracts/contracts/strategies/sonic/SonicValidatorDelegator.sol
```diff
@@ -0,0 +1,318 @@
...
+        // The SFC contract will emit Delegated and RestakedRewards events
+        emit Deposit(wrappedSonic, address(0), totalRewards);
+    }
+
+    /**
+     * @notice To receive native S from SFC and Wrapped Sonic (wS)
+     */
+    receive() external payable {
+        require(
+            msg.sender == sfc || msg.sender == wrappedSonic,
```
sparrowDom, line 252: This is a nice check, but there is [this little devil](https://github.com/OriginProtocol/origin-dollar/pull/2335/files#diff-8da7c6574d9a346b9e74b8a9b465632eb613a64e1d1b64fca9488e45f23d6c15R32) lurking around :)

sparrowDom: I don't think that donating S to the contract would throw off the math: 
- We do the `address(this).balance` only in the `withdrawFromTheSFC` and we are looking at the increment that the withdrawal has caused... ignoring previous amount
- Also the checkBalance does ignore the `S` balance on the contract. 

The only complaint I have that the check might give a false sense of security, since anyone can call `withdrawTo` and send S to the contract

contracts/contracts/strategies/sonic/SonicValidatorDelegator.sol
naddison36: I think we should drop the `Deposit` event from `restakeRewards`. I think we should leave `Deposit` for when assets are transferred from the Vault to the strategy.

contracts/contracts/vault/VaultCore.sol
```diff
@@ -23,11 +23,8 @@ import "./VaultInitializer.sol";
 contract VaultCore is VaultInitializer {
     using SafeERC20 for IERC20;
     using StableMath for uint256;
-    // max signed int
+    /// @dev max signed int
     uint256 internal constant MAX_INT = 2**255 - 1;
```
DanielVF, line 27: As long as we are cleaning up this section, we could change the `2**255 - 1` to `type(int256).max`

contracts/contracts/strategies/sonic/SonicValidatorDelegator.sol
```diff
@@ -0,0 +1,363 @@
...
+
+    /**
+     * @notice Restake any pending validator rewards for all supported validators
+     * @param validatorIds List of Sonic validator IDs to restake rewards
+     */
+    function restakeRewards(uint256[] calldata validatorIds)
+        external
+        nonReentrant
+    {
+        uint256 totalRewards = 0;
```
DanielVF, line 260: 🟠 This variable is not used and can be removed?

contracts/contracts/strategies/sonic/SonicValidatorDelegator.sol
```diff
@@ -0,0 +1,363 @@
...
+     * @notice To receive native S from SFC and Wrapped Sonic (wS)
+     *
+     * @dev This does not prevent donating S tokens to the contract
+     * as wrappedSonic has a `withdrawTo` function where a third party
+     * owner of wrappedSonic can withdraw to this contract.
+     */
+    receive() external payable {
+        require(
+            msg.sender == sfc || msg.sender == wrappedSonic,
+            "S not from allowed contracts"
```
DanielVF, line 287: ⚪️ This message does not match what the check actually does.

contracts/contracts/strategies/sonic/SonicValidatorDelegator.sol
```diff
@@ -0,0 +1,363 @@
...
+ * @title Manages delegation to Sonic validators
+ * @notice This contract implements all the required functionality to delegate to,
+   undelegate from and withdraw from validators.
+ * @author Origin Protocol Inc
+ */
+abstract contract SonicValidatorDelegator is InitializableAbstractStrategy {
+    /// @notice Address of Sonic's wrapped S token
+    address public immutable wrappedSonic;
+    /// @notice Address of Sonic's Special Fee Contract (SFC)
+    address public immutable sfc;
```
DanielVF, line 19: ⚪️ If we set the variable type here to `ISFC`, it should remove almost ten casts from the code base, at the cost of only one address() cast the other way. Would make the code read cleaner.

contracts/contracts/strategies/sonic/SonicValidatorDelegator.sol
```diff
@@ -0,0 +1,363 @@
...
+     * @dev Delegate from this strategy to a specific Sonic validator. Called
+     * automatically on asset deposit
+     * @param amount the amount of Sonic (S) to delegate.
+     */
+    function _delegate(uint256 amount) internal {
+        require(
+            isSupportedValidator(defaultValidatorId),
+            "Validator not supported"
+        );
+        require(amount > 0, "Must delegate something");
```
DanielVF, line 151: 🟠 Why must someone delegate > 0? Can we just return early? This just feels like it makes the code brittle and adds reverts (and the need for additional matching checks for them at higher levels.)

contracts/contracts/strategies/sonic/SonicValidatorDelegator.sol
```diff
@@ -0,0 +1,363 @@
...
+            IERC20(wrappedSonic).balanceOf(address(this)) +
+            pendingWithdrawals;
+
+        // For each supported validator, get the staked amount and pending rewards
+        uint256 validatorLen = supportedValidators.length;
+        for (uint256 i = 0; i < validatorLen; i++) {
+            // Get the staked amount and any pending rewards
+            balance += ISFC(sfc).getStake(
+                address(this),
+                supportedValidators[i]
```
DanielVF, line 132: ⛽️ If we do `address validator = supportedValidators[I]` at the start of the loop, we can save gas here by not hitting the same storage storage twice on loop through, (and the code might look clearer too).

contracts/utils/deploy-l2.js
```diff
@@ -112,7 +113,8 @@ function deployOnBase(opts, fn) {
 
   return main;
 }
-function deployOnBaseWithGuardian(opts, fn) {
+
+function deployOnL2WithMultisig(opts, fn, forkSignerAddress) {
```
naddison36, line 117: I think multisig is too ambiguous. Does it mean the Admin (5/8), Guardian (2/8) or the new multi-chain Admin (5/8)?

contracts/test/_fixture-sonic.js
```diff
@@ -109,6 +109,8 @@ const defaultSonicFixture = deployments.createFixture(async () => {
   const strategist = await impersonateAndFund(strategistAddr);
   strategist.address = strategistAddr;
 
+  await sonicStakingStrategy.connect(strategist).setDefaultValidatorId(18);
```
naddison36, line 112: I think this should be in the deploy script.

contracts/utils/deploy-l2.js
```diff
@@ -256,9 +265,26 @@ function deployOnSonic(opts, fn) {
   return main;
 }
 
+function deployOnSonicWithAdmin(opts, fn) {
```
naddison36, line 268: I think we can just have `deployOnSonic` that can handle both Timelock actions and no actions.

contracts/test/_fixture-sonic.js
naddison36: I see why its in the fixture. The governor can not set the default

sparrowDom: Yes we need to do it manually with the 2/8 multisig

contracts/utils/deploy-l2.js
sparrowDom: agreed

contracts/contracts/vault/VaultCore.sol
sparrowDom: agreed 👍

contracts/contracts/strategies/sonic/SonicValidatorDelegator.sol
sparrowDom: 👍

contracts/contracts/strategies/sonic/SonicValidatorDelegator.sol
sparrowDom: changed to `S not from allowed contracts`

sparrowDom: This function doesn't allow for receiving of S when not from the 2 contracts allowed: sfc & wrappedSonic. What do you think would be a better sounding message?

contracts/contracts/strategies/sonic/SonicValidatorDelegator.sol
sparrowDom: good idea 👍

contracts/contracts/strategies/sonic/SonicValidatorDelegator.sol
sparrowDom: Deleting this line, the `_delegate` gets called by the `deposit` function and the `deposit` function already performs a 0 amount check. So this one is really redundant and unreachable.

contracts/contracts/strategies/sonic/SonicValidatorDelegator.sol
sparrowDom: agreed

contracts/contracts/strategies/sonic/SonicStakingStrategy.sol
```diff
@@ -0,0 +1,154 @@
...
+        require(_asset == wrappedSonic, "Unsupported asset");
+        _withdraw(_recipient, _asset, _amount);
+    }
+
+    function _withdraw(
+        address _recipient,
+        address _asset,
+        uint256 _amount
+    ) internal {
+        require(_amount > 0, "Must withdraw something");
```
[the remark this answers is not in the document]
sparrowDom: We do this with [other](https://github.com/OriginProtocol/origin-dollar/blob/master/contracts/contracts/strategies/ConvexEthMetaStrategy.sol#L269) [strategies](https://github.com/OriginProtocol/origin-dollar/blob/master/contracts/contracts/strategies/CompoundStrategy.sol#L128) as well.

I guess it would make sense to be consistent. IMO either: 
- reverting
- not reverting and early exiting
- not reverting with no extra balance checks

will all have a very similar effect.

contracts/contracts/vault/VaultCore.sol
```diff
@@ -23,11 +23,8 @@ import "./VaultInitializer.sol";
 contract VaultCore is VaultInitializer {
     using SafeERC20 for IERC20;
     using StableMath for uint256;
-    // max signed int
-    uint256 internal constant MAX_INT = 2**255 - 1;
-    // max un-signed int
-    uint256 internal constant MAX_UINT =
-        0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff;
+    /// @dev max signed int
+    uint256 internal constant MAX_INT = type(uint256).max;
```
naddison36, line 27: This needs to be max signed int, not unsigned int as the `abs` function does
`require(x < int256(MAX_INT), "Amount too high");`

This change breaks mainnet so I'll revert. No need to redeploy Sonic as `mintForStrategy` is not used. Although it will be when an AMO is added.

MAX_UINT can still be deleted

contracts/contracts/vault/VaultCore.sol
naddison36: Solidity won't allow `type(int256).max` to be set to an unsigned int `uint256`

DanielVF: Just needs the cast so it's explicit. `uint256(type(int256).max)`

contracts/contracts/strategies/sonic/SonicValidatorDelegator.sol
DanielVF: I miss-read. Current message fine.

contracts/contracts/strategies/sonic/SonicStakingStrategy.sol
```diff
@@ -0,0 +1,154 @@
...
+    // For future use
+    uint256[50] private __gap;
+
+    constructor(
+        BaseStrategyConfig memory _baseConfig,
+        address _wrappedSonic,
+        address _sfc
+    ) SonicValidatorDelegator(_baseConfig, _wrappedSonic, _sfc) {}
+
+    /// @notice Unlike other strategies, this does not deposit assets into the underlying platform.
```
DanielVF, line 23: 🟠 Is this comment, and the later one on depositAll, accurate? It seems that these functions do delegate on deposit?

contracts/contracts/strategies/sonic/SonicStakingStrategy.sol
```diff
@@ -0,0 +1,154 @@
...
+
+        if (wSBalance > 0) {
+            _deposit(wrappedSonic, wSBalance);
+        }
+    }
+
+    /// @notice Withdraw Wrapped Sonic (wS) from this strategy contract.
+    /// Used only if some wS is lingering on the contract.
+    /// That can happen when:
+    ///   - after mints if the strategy is the default
```
DanielVF, line 65: Is this comment accurate?

contracts/contracts/strategies/sonic/SonicValidatorDelegator.sol
```diff
@@ -0,0 +1,352 @@
...
+        WithdrawRequest memory withdrawal = withdrawals[withdrawId];
+        require(withdrawal.validatorId > 0, "Invalid withdrawId");
+        return withdrawal.undelegatedAmount == 0;
+    }
+
+    /**
+     * @notice Restake any pending validator rewards for all supported validators
+     * @param validatorIds List of Sonic validator IDs to restake rewards
+     */
+    function restakeRewards(uint256[] calldata validatorIds)
```
DanielVF, line 247: ❓Do we need a way to collect rewards?

The SFC contract stores our funds in three buckets: Staked, stashed rewards, and calculated rewards. However, externally these can be treated as staked funds and reward funds. We have a mechanism to withdraw staked funds, but we do not have a mechanism to get at reward funds. This means that our only option to close out  delegator is to undelegate, wait, withdraw the staked funds, wait a while more, retake the rearwards to staked, and kick off a second undelegation cycle. This seems painful when we could just get our reward money right now.

contracts/contracts/strategies/sonic/SonicStakingStrategy.sol
sparrowDom: thanks fixed here: https://github.com/OriginProtocol/origin-dollar/pull/2335/commits/594b45adc164f9464c66836979b2bf9e1ef67724

contracts/contracts/strategies/sonic/SonicStakingStrategy.sol
sparrowDom: thanks it was not: https://github.com/OriginProtocol/origin-dollar/pull/2335/commits/11f967f74a90b828b5ae97a0a919cef3dfd2b72a

contracts/contracts/strategies/sonic/SonicValidatorDelegator.sol
```diff
@@ -0,0 +1,354 @@
...
+        withdrawals[withdrawId].undelegatedAmount = 0;
+
+        // Wrap Sonic (S) to Wrapped Sonic (wS)
+        IWrappedSonic(wrappedSonic).deposit{ value: withdrawnAmount }();
+
+        // Transfer the Wrapped Sonic (wS) to the Vault
+        // slither-disable-next-line unchecked-transfer unused-return
+        IERC20(wrappedSonic).transfer(vaultAddress, withdrawnAmount);
+
+        // withdrawal.undelegatedAmount & withdrawnAmount can differ in case of slashing
```
DanielVF, line 225: 🟠 From an events point of view, we emit `Withdraw` events when money leaves a strategy and `Deposit` events when non-reward money enters the strategy. 

However, in this code, the actual withdraw for the strategy funds is here in the `withdrawFromSFC` code, and it does not emit a `Withdraw` event. This will mean this strategy behaves differently than our other strategies for monitoring purposes. We should emit a withdraw event here. Perhaps just virtual `_withdraw()`, and then call up into the SonicStakingStrategy `_withdraw()` instead of doing a transfer.

contracts/contracts/strategies/sonic/SonicValidatorDelegator.sol
```diff
@@ -0,0 +1,352 @@
...
+        WithdrawRequest memory withdrawal = withdrawals[withdrawId];
+        require(!isWithdrawnFromSFC(withdrawId), "Already withdrawn");
+
+        uint256 sBalanceBefore = address(this).balance;
+
+        sfc.withdraw(withdrawal.validatorId, withdrawId);
+
+        // Save state to storage
+        withdrawnAmount = address(this).balance - sBalanceBefore;
+        pendingWithdrawals -= withdrawal.undelegatedAmount;
```
DanielVF, line 215: 🟠 This code is not written in the order we prefer to write in. sfc withdraw could reentrancy attack us, since it sits between the validation and the state write. I don't expect SFC to attack us (and it's not like they are going to give us our money twice), but it's just a good habit to write safe ordering, unless it's going to make a mess of things.

In this case, reordering will make things clearer.

The `withdrawals[withdrawId].undelegatedAmount` should go right after the isWithdrawnFromSFC validation, then `pendingWithdrawals` change right after that.

Then all the rest is just external contract interaction with no state writes.

contracts/contracts/strategies/sonic/SonicValidatorDelegator.sol
```diff
@@ -0,0 +1,352 @@
...
+        supportedValidators.push(validatorId);
+
+        emit SupportedValidator(validatorId);
+    }
+
+    /// @notice Removes a validator from the supported list.
+    /// Unsupported validators can still be undelegated from, withdrawn from and rewards collected.
+    function unsupportValidator(uint256 validatorId) external onlyGovernor {
+        require(isSupportedValidator(validatorId), "Validator not supported");
+        require(
```
DanielVF, line 318: ❓would it be easier to kick off a withdraw here rather than validate stake is zero. The public can move rewards to stakes, so it's going to be very hard to empty a validator particularly when there's a timelock involved.

contracts/contracts/strategies/sonic/SonicValidatorDelegator.sol
```diff
@@ -0,0 +1,354 @@
...
+        // and any undelegated amounts still pending a withdrawal
+        balance =
+            IERC20(wrappedSonic).balanceOf(address(this)) +
+            pendingWithdrawals;
+
+        // For each supported validator, get the staked amount and pending rewards
+        uint256 validatorLen = supportedValidators.length;
+        for (uint256 i = 0; i < validatorLen; i++) {
+            uint256 validator = supportedValidators[i];
+            // Get the staked amount and any pending rewards
```
DanielVF, line 130: ⚪️ This comment duplicates the comment above?

contracts/contracts/strategies/sonic/SonicValidatorDelegator.sol
```diff
@@ -0,0 +1,354 @@
...
+        // withdrawal.undelegatedAmount & withdrawnAmount can differ in case of slashing
+        emit Withdrawn(
+            withdrawId,
+            withdrawal.validatorId,
+            withdrawal.undelegatedAmount,
+            withdrawnAmount
+        );
+    }
+
+    // slither-disable-end reentrancy-no-eth
```
DanielVF, line 234: I think slither was picking up on an actual incorrectness in how the function was written. We need to be careful when we disable slither errors.

(Though disabling slither errors in here in the code is far better than disabling them in a separate json file, since this makes it easy to see them during review.)

contracts/test/behaviour/sfcStakingStrategy.js
```diff
@@ -0,0 +1,527 @@
...
+
+      const amount = oethUnits("15000");
+      await depositTokenAmount(amount);
+      await advanceSfcEpoch(1);
+
+      const stratBalanceBefore = await sonicStakingStrategy.checkBalance(
+        wS.address
+      );
+
+      await sonicStakingStrategy.restakeRewards(testValidatorIds);
```
DanielVF, line 138: ❓This is a very important test - it's critical the balance does not change. Are we sure that there are actually rewards that are being restaked here?

DanielVF:
# Code review - Sonic Strategy


## Requirements

We need safely store funds and earn in sonic validators.

## Easy Checks

#### Authentication

- [x] Never use tx.origin

#### Ethereum

	- [x] 🟣 Does only to wrapped sonic and hardcoded SFC address
	- [x] 🟣 See above
- [x] Contract does not have a vulnerability from being sent self destruct ETH
	- [x] Does not affect accounting. ETH can be recovered.

#### Cryptographic code

_No crypto_

#### Gas problems

    - [ ] A way to remove items
	    - [ ] Admin controls validator ID

#### Black magic

#### Overflow

#### Proxy

#### Events

#### Rounding and casts

_No rounding_

	- [x] 🟣 Not needed when casting an immutable value

#### Dependencies

#### External calls

- [x] No malicious behaviors
- [x] Oracles, one of:
  - [x] No oracles
  - [ ] Oracles can't be bent

#### Deploy

## Strategy Specific

#### Strategy checks

- [x] No read only reentrancy from downstream protocols during checkBalance
	- [x] At least not that I saw from their contract
	- [ ] 🔴 Rewards are left in the staking system on delegator unstakes.
	- [x] 🟣 No separate reward tokens.
	- [x] Can withdraw() or withdrawAll() wrapped sonic
	- [x] Can withdrawAll() raw sonic
	- [x] 🟣 After undelegation completes, and assuming sonic does not slash

#### Downstream

The sonic SFC contains code to upgrade it.

- [ ] We must understand the upgrade process
- [ ] We most monitor upcoming governance changes to it.
- [ ] We need some form of validator uptime monitoring?

#### Logic

Corrections noted in PR.

#### Invarients

Functions:
- no function call (except `withdrawAll()` + unaccounted gift, or `withdraw()` + recipient) should change the vault value.

State:

- `nextWithdrawId` Only goes up
- `pendingWithdraw` never goes below zero

#### Attack

This code will be holding the bulk of the protocols funds.

An attacker could try to steal funds. This contract only sends core funds to locations controlled by the vault, or immutable addresses.

The primary concern is some kind of balance change attack. Restake rewards is publicly callable. It is important that it does not change the balance. Although admin controlled, none the other methods should change the vault balance.

#### Flavor

Code split between classes is very clean.

- [ ] Functions follow validate, read, compute, write, call, validate
	- [ ] 🟠 Noted in PR where this was not true.

contracts/contracts/strategies/sonic/SonicValidatorDelegator.sol
sparrowDom: This is a great point thanks. Corrected here: https://github.com/OriginProtocol/origin-dollar/pull/2335/commits/129ce7fd78243befbaa874f62b96af9fd4b806c4

contracts/contracts/strategies/sonic/SonicValidatorDelegator.sol
sparrowDom: Thanks great point, corrected here: https://github.com/OriginProtocol/origin-dollar/pull/2335/commits/91adf813d37a810986facc8004324830d7743770

contracts/contracts/strategies/sonic/SonicValidatorDelegator.sol
```diff
@@ -0,0 +1,354 @@
...
+        uint256 undelegatedAmount,
+        uint256 withdrawnAmount
+    );
+    event RegistratorChanged(address indexed newAddress);
+    event SupportedValidator(uint256 indexed validatorId);
+    event UnsupportedValidator(uint256 indexed validatorId);
+    event DefaultValidatorIdChanged(uint256 indexed validatorId);
+
+    /// @dev Throws if called by any account other than the Registrator
+    modifier onlyRegistrator() {
```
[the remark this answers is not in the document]
sparrowDom: good point, fixed here: https://github.com/OriginProtocol/origin-dollar/pull/2335/commits/d2f3431d6cd9d663936ffb3c4565a65275b7f818

contracts/contracts/strategies/sonic/SonicValidatorDelegator.sol
sparrowDom: thanks very good point: https://github.com/OriginProtocol/origin-dollar/pull/2335/commits/5ef4bf331751a9a425cb34e89894145a1f863a47

contracts/contracts/strategies/sonic/SonicValidatorDelegator.sol
sparrowDom: good point, un-delegating everything when unsupporting a validator makes sense. Fixed here: https://github.com/OriginProtocol/origin-dollar/pull/2335/commits/51e25e719ae577017893962b2226e6bb0dce7d40

contracts/contracts/strategies/sonic/SonicValidatorDelegator.sol
sparrowDom: fixed: https://github.com/OriginProtocol/origin-dollar/pull/2335/commits/1c099468647785362478aa656c9d55ab4ad93f35

contracts/contracts/strategies/sonic/SonicValidatorDelegator.sol
sparrowDom: yes you are right slither was picking up the error for the right reasons. Manual disables removes here: https://github.com/OriginProtocol/origin-dollar/pull/2335/commits/08f6c83cb4e6c14b268ba1898a8ee20da5cbe81d

contracts/test/behaviour/sfcStakingStrategy.js
sparrowDom: good point: https://github.com/OriginProtocol/origin-dollar/pull/2335/commits/4ea53b1dcec276fb1aca2f3c5db7ce9b52e72994

contracts/contracts/strategies/sonic/SonicValidatorDelegator.sol
```diff
@@ -0,0 +1,396 @@
...
+    /**
+     * @notice Restake any pending validator rewards for all supported validators
+     * @param validatorIds List of Sonic validator IDs to restake rewards
+     */
+    function restakeRewards(uint256[] calldata validatorIds)
+        external
+        nonReentrant
+    {
+        for (uint256 i = 0; i < validatorIds.length; ++i) {
+            require(
```
DanielVF, line 246: ❓Do we need this require? It might be handy to be able to collect rewards from validators we've already closed out. In this function we are collecting money only, under non-public control.

sparrowDom: https://github.com/OriginProtocol/origin-dollar/pull/2335/commits/c76f2b92167b63172a934a39efab38d7c9d9e426

contracts/deploy/sonic/003_sonic_staking_strategy.js
```diff
@@ -62,6 +63,18 @@ module.exports = deployOnSonic(
...
     console.log("Initialized SonicStakingStrategy proxy and implementation");
 
+    // Deploy a new Wrapped Origin Sonic contract
+
+    const cOSonicProxy = await ethers.getContract("OSonicProxy");
+    const cWOSonicProxy = await ethers.getContract("WOSonicProxy");
+
+    const dWOSonic = await deployWithConfirmation("WOSonic", [
+      cOSonicProxy.address, // Base token
+      "Wrapped OS", // Token Name
```
sparrowDom, line 73: At some point we should probably override the constructor so it doesn't accept the Token Name & Symbol just to avoid confusion

contracts/deploy/sonic/003_sonic_staking_strategy.js
```diff
@@ -98,6 +111,30 @@ module.exports = deployOnSonic(
           signature: "setRegistrator(address)",
           args: [addresses.sonic.validatorRegistrator],
         },
+        // 4. Set 10% performance fee
+        {
+          contract: cOSonicVault,
+          signature: "setTrusteeFeeBps(uint256)",
```
sparrowDom, line 117: I think we need to set the Trusteee Address as well. Will add this in a separate commit

sparrowDom: done here: https://github.com/OriginProtocol/origin-dollar/pull/2335/commits/f850d728d91edd6b59cdbd2503c593e1b6fa5308

sparrowDom:
## Requirements

PR supports delegation of Wrapped Sonic to validators to earn staking rewards.

## Easy Checks

#### Authentication

- [x] Never use tx.origin

#### Sonic token

- [🟡] Contract does not send or receive S
  - It does receive S but it does not affect its `checkBalance` or behaviour of other functions. The `withdrawFromSFC` and `collectRewards` check only for the increment of S ignoring previous states. The `withdrawAll` does check for the whole balance of S and withdraws all. Donating S to contract only changes the beaviour where `withdrawAll` is able to withdraw more than was "rightfully" earned.
  - hijacking transaction execution: contracts makes sure that any storage altering operations happen before it is calling external SFC contract which can trigger transfer of S token.
- [🟡] Contract has no payable methods.
  - It does but limits the contracts that can send S to it. Though with WrappedSonic's `withdrawTo` this doesn't really prevent receiving S. Anyone can wrap S into WS and withdraw it to our staking contract via the WS contract.
- [x] Contract is not vulnerable to being sent self destruct S.
  - with exception of `withdrawAll` the contract ignores any existing S that has been received. It is able to recover all S and accounting is not affected by having received S.

#### Cryptographic code

#### Gas problems

    - [x] A way to remove items

#### Black magic

#### Overflow

#### Proxy

#### Events

## Medium Checks

#### Rounding and casts
No Rounding

#### Dependencies

#### External calls

- [x] No malicious behaviors
- [x] Oracles, one of:
  - [x] No oracles
  - [ ] Oracles can't be bent

#### Tests

  - [X] Each logical branch has a test
  - [X] Each require() has a test

#### Deploy

## Strategy Specific

#### Strategy checks

  - [x] The rewards are in S token so there is no need to swap
  - yes considering there have been no slashing of the validators. Also full withdrawal is not immediate as it takes time to undelegate and withdraw

#### Downstream
[Monitoring Ticket](https://www.notion.so/originprotocol/Monitoring-18284d46f53c8061831eef5004d6b8a8)
- we should monitor validator slashings
- we should monitor SFC contract upgrades

## Thinking

#### Logic

  The logic looks sound

Just a normal deploy governance process

#### Internal State

- `checkBalance` will return higher amount without any deposits / withdrawals each epoch (~10 minutes) as rewards are not accounted for separately
- withdrawAll is the only function picking up "unaccounted S tokens". Any other combination of Depositing / delegating / undelegating / Withdrawing / restakeRewards / CollectRewards should not change the total vault value. In case someone donates S to Sonic staking strategy and withdrawAll is called, then the vaultValue should increase
- defaultValidatorId should be one of supportedValidators
- nextWithdrawId should increase by 1

#### Attack

If a validator is slashed we could lose user funds. We will be monitoring validators and SFC contract upgrades to detect any anomalies.

A hijacked defender action relayer or strategist could undelegate all funds and prevent yield earning for 14 days. There is no way to prevent this in code.


#### Flavor

The code is a clean interface to the SFC contract

sparrowDom (approved):

PR #2379, opened 2025-02-04
Redefine Curve AMO Abstract structure.

clement-ux:
This PR aims to change how Curve AMO contract are structured.

Instead of having multiple contract, re-implementing AMO logic, the logic will be stored in a Abstract contract: "AbstractCurveAMOStrategy".
And all Curve AMO should inherit this abstract contract. 
The logic (eg. where to deposit LP or how to get LP) will be done on the main contract instead, by overriding not defined functions like:
-  `_addLiquidity`
- `_removeLiquidity`
- `_removeLiquidityOneCoin`
- `_stakeLP`
- `_unstakeLP`
- `_wrapETH`
- `_unwrapETH`
- `_claimReward`
- `_approveBase`
- `_getBalances`
- `_solvencyAssert`
-  `maxSlippage`

### DIFF
--- a/contracts/contracts/strategies/AbstractCurveAMOStrategy.sol
+++ b/contracts/contracts/strategies/AbstractCurveAMOStrategy.sol
@@ -1,0 +1,566 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+/**
+ * @title Convex Automated Market Maker (AMO) Strategy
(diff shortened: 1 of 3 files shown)


sparrowDom:
we've decided to not go for this because: 
- ConvexEthMetaStrategy has already been audited
- more inheritance makes it harder to follow the code

overall the abstraction benefits do not outweigh the downsides.

PR #2380, opened 2025-02-05
Allow guardian to redeem instantly without fees

shahthepro:
## Change overview
- Guardian (or `strategist` in the codebase) can now instantly redeem without paying the 10bps fee. Will help us a lot with operations. And this is something we already do on Base as well
- Doesn't add a separate deploy file. 108 is pending and it upgrades the Vault anyway

## Code Change Checklist

- [ ]  Unit tests *if needed

Internal review:

### DIFF
--- a/contracts/contracts/vault/OETHVaultCore.sol
+++ b/contracts/contracts/vault/OETHVaultCore.sol
@@ -130,9 +130,10 @@ contract OETHVaultCore is VaultCore {
         }
 
         // Amount excluding fees
-        uint256 amountMinusFee = _calculateRedeemOutputs(_amount)[
-            wethAssetIndex
-        ];
+        // No fee for the strategist, makes it easier to do operations
+        uint256 amountMinusFee = msg.sender == strategistAddr
+            ? _amount
+            : _calculateRedeemOutputs(_amount)[wethAssetIndex];
 
         require(
             amountMinusFee >= _minimumUnitAmount,
--- a/contracts/contracts/vault/OETHVaultCore.sol
+++ b/contracts/contracts/vault/OETHVaultCore.sol
@@ -130,7 +130,7 @@ contract OETHVaultCore is VaultCore {
         }
 
         // Amount excluding fees
-        // No fee for the strategist, makes it easier to do operations
+        // No fee for the strategist or the governor, makes it easier to do operations
         uint256 amountMinusFee = (msg.sender == strategistAddr || isGovernor())
             ? _amount
             : _calculateRedeemOutputs(_amount)[wethAssetIndex];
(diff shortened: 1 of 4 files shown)


contracts/contracts/vault/OETHVaultCore.sol
```diff
@@ -130,9 +130,10 @@ contract OETHVaultCore is VaultCore {
         }
 
         // Amount excluding fees
-        uint256 amountMinusFee = _calculateRedeemOutputs(_amount)[
-            wethAssetIndex
-        ];
+        // No fee for the strategist, makes it easier to do operations
+        uint256 amountMinusFee = msg.sender == strategistAddr
```
naddison36, line 134: On Base we allow either the Strategies or Governor to redeem. Do we want to allow the Governor, which is the Origin Timelock contract, also to redeem?

contracts/fork-test.sh
```diff
@@ -77,7 +77,11 @@ main()
 
     # Run specific files when a param is given
     if [[ ! -z "$1" ]]; then
-        params+="--testfiles $@"
+        if [[ $is_coverage == "true" ]]; then
```
naddison36, line 80: Were local fork tests not working?

shahthepro: Yes, `--testfiles` is only supported by `solcoverage` (the hardhat plugin we use for coverage tests). non-coverage tests don't support that param

contracts/contracts/vault/OETHVaultCore.sol
shahthepro: I thought of that, but it's gonna complicate the code a bit (since we cannot blindly add the modifier). I guess I can just add that to this conditional statement, will make that change

shahthepro: Have made this change

contracts/contracts/vault/OETHVaultCore.sol
```diff
@@ -130,9 +130,10 @@ contract OETHVaultCore is VaultCore {
         }
 
         // Amount excluding fees
-        uint256 amountMinusFee = _calculateRedeemOutputs(_amount)[
-            wethAssetIndex
-        ];
+        // No fee for the strategist, makes it easier to do operations
```
sparrowDom, line 133: nit: comment should also include the governor

sparrowDom (dismissed):
LGTM

clement-ux (dismissed):
LGTM 👍

contracts/contracts/vault/OETHVaultCore.sol
shahthepro: Fixed. thanks

PR #2384, opened 2025-02-06
Base curve amo - Upgrade

clement-ux:
The current `BaseCurveAMOStrategy.sol` is a copy/past/adjusted contract from `ConvexEthMetaStrategy.sol`.
This PR aims to add some improvement on `BaseCurveAMOStrategy.sol` from the original design:
- not immutable MAX_SLIPPAGE
-  using safeCast
- improveBalance modifier handles when diffBefore is 0.

### DIFF
--- a/contracts/contracts/strategies/BaseCurveAMOStrategy.sol
+++ b/contracts/contracts/strategies/BaseCurveAMOStrategy.sol
@@ -26,8 +26,6 @@ contract BaseCurveAMOStrategy is InitializableAbstractStrategy {
      */
     uint256 public constant SOLVENCY_THRESHOLD = 0.998 ether;
 
-    uint256 public constant MAX_SLIPPAGE = 1e16; // 1%, same as the Curve UI
-
     // New immutable variables that must be set in the constructor
     /**
      * @notice Address of the Wrapped ETH (WETH) contract.
@@ -62,6 +60,13 @@ contract BaseCurveAMOStrategy is InitializableAbstractStrategy {
     // Ordered list of pool assets
     uint128 public constant oethCoinIndex = 1;
     uint128 public constant ethCoinIndex = 0;
+
+    /**
+     * @notice Maximum slippage allowed for adding/removing liquidity from the Curve pool.
+     */
+    uint256 public maxSlippage;
+
+    event MaxSlippageUpdated(uint256 newMaxSlippage);
 
     /**
      * @dev Verifies that the caller is the Strategist.
@@ -133,14 +138,17 @@ contract BaseCurveAMOStrategy is InitializableAbstractStrategy {
      * well within that abstraction.
      * @param _rewardTokenAddresses Address of CRV
      * @param _assets Addresses of supported assets. eg WETH
+     * @param _maxSlippage Maximum slippage allowed for adding/removing liquidity from the Curve pool.
      */
     function initialize(
         address[] calldata _rewardTokenAddresses, // CRV
-        address[] calldata _assets // WETH
+        address[] calldata _assets, // WETH
+        uint256 _maxSlippage
     ) external onlyGovernor initializer {
         require(_assets.length == 1, "Must have exactly one asset");
         require(_assets[0] == address(weth), "Asset not WETH");
 
+        maxSlippage = _maxSlippage;
         address[] memory pTokens = new address[](1);
         pTokens[0] = address(curvePool);
 
@@ -214,7 +222,7 @@ contract BaseCurveAMOStrategy is InitializableAbstractStrategy {
             curvePool.get_virtual_price()
         );
         uint256 minMintAmount = valueInLpTokens.mulTruncate(
-            uint256(1e18) - MAX_SLIPPAGE
+            uint256(1e18) - maxSlippage
         );
 
         // Do the deposit to the Curve pool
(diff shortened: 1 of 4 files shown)


contracts/contracts/strategies/BaseCurveAMOStrategy.sol
```diff
@@ -133,14 +141,17 @@ contract BaseCurveAMOStrategy is InitializableAbstractStrategy {
...
+     * @param _maxSlippage Maximum slippage allowed for adding/removing liquidity from the Curve pool.
      */
     function initialize(
         address[] calldata _rewardTokenAddresses, // CRV
-        address[] calldata _assets // WETH
+        address[] calldata _assets, // WETH
+        uint256 _maxSlippage
     ) external onlyGovernor initializer {
         require(_assets.length == 1, "Must have exactly one asset");
         require(_assets[0] == address(weth), "Asset not WETH");
```
shahthepro, line 152: nit: why not just make it a hardcoded constant/variable instead of passing it as an argument when we know what exactly it should have?

contracts/contracts/strategies/BaseCurveAMOStrategy.sol
```diff
@@ -133,14 +141,17 @@ contract BaseCurveAMOStrategy is InitializableAbstractStrategy {
...
 
+        maxSlippage = _maxSlippage;
```
shahthepro, line 154: Would prefer making `setMaxSlippage` public and calling through that (or have an internal counterpart that's called in both places). The thing is `setMaxSlippage` has a max value check that's missing here

contracts/contracts/strategies/BaseCurveAMOStrategy.sol
clement-ux: I think simplifying deployment is safe. 
Changed in this commit: https://github.com/OriginProtocol/origin-dollar/pull/2384/commits/4e2e80c7532397747ac9a705fac0dd4279a11985.

contracts/deploy/base/025_base_curve_amo.js
```diff
@@ -48,8 +49,8 @@ module.exports = deployOnBaseWithGuardian(
 
     // Initialize Base Curve AMO implementation
     const initData = cOETHBaseCurveAMO.interface.encodeFunctionData(
-      "initialize(address[],address[])",
-      [[addresses.base.CRV], [addresses.base.WETH]]
+      "initialize(address[],uint256)",
+      [[addresses.base.CRV], oethUnits("0.2")]
```
sparrowDom, line 53: 0.2 equals 20%. we should go for 0.2% which would be: 
```
oethUnits("0.002") // 0.2%
```

sparrowDom: I think it would be cool that we also add a fork test that verifies the value of max slippage.

clement-ux: Good catch!

Fixed in this commit: https://github.com/OriginProtocol/origin-dollar/pull/2384/commits/615c275192aa52361ed6b9b793708a8825353622

contracts/contracts/strategies/BaseCurveAMOStrategy.sol
clement-ux: Fixed in this commit: https://github.com/OriginProtocol/origin-dollar/pull/2384/commits/4e2e80c7532397747ac9a705fac0dd4279a11985

contracts/contracts/strategies/BaseCurveAMOStrategy.sol
clement-ux: Fixed in this commit: https://github.com/OriginProtocol/origin-dollar/pull/2384/commits/6711b5a8027aeb69f52d50a49bc1a1dde222f4cb.

contracts/contracts/strategies/BaseCurveAMOStrategy.sol
```diff
@@ -586,6 +593,20 @@ contract BaseCurveAMOStrategy is InitializableAbstractStrategy {
...
+    /**
+     * @notice Sets the maximum slippage allowed for any swap/liquidity operation
+     * @param _maxSlippage Maximum slippage allowed, 1e18 = 100%.
+     */
+    function setMaxSlippage(uint256 _maxSlippage) external onlyGovernor {
+        _setMaxSlippage(_maxSlippage);
+    }
+
+    function _setMaxSlippage(uint256 _maxSlippage) internal {
+        require(_maxSlippage <= 1e18, "Slippage must be less than 100%");
```
naddison36, line 605: 100% slippage seems extreme. How about a max slippage of 5% (5e16)?
This should protect against the governor miss-configuring the `maxSlippage`

contracts/contracts/strategies/BaseCurveAMOStrategy.sol
```diff
@@ -132,25 +140,26 @@ contract BaseCurveAMOStrategy is InitializableAbstractStrategy {
...
-        address[] calldata _assets // WETH
+        uint256 _maxSlippage
     ) external onlyGovernor initializer {
-        require(_assets.length == 1, "Must have exactly one asset");
-        require(_assets[0] == address(weth), "Asset not WETH");
-
         address[] memory pTokens = new address[](1);
         pTokens[0] = address(curvePool);
 
+        address[] memory _assets = new address[](1);
```
naddison36, line 152: This is a good simplification. nice

PR #2393, opened 2025-02-12
Yield forwarding template

sparrowDom:
We don't need to clutter our deployments folder with extra entries every time we want to create a yield forwarding action. This PR adds a template where from & to addresses need to be added in order to schedule and execute a yield forwarding transaction.


DanielVF (approved):

sparrowDom:
Since yield forwarding is now done via the strategist this isn't needed anymore

PR #2398, opened 2025-02-14
SuperOETH Harvester

shahthepro:
## Code Changes
- Adds a new `SuperOETHHarvester` that inherits from `OETHHarvesterSimple`. 
  - Only change is the recipient of reward tokens being the guardian always
- Moving forward, all supercharged LSTs (a.k.a Super OETH/OUSD) to use `SuperOETHHarvester` and all native LST OTokens (OETH) to use `OETHHarvesterSimple`
- Rename `weth` in `OETHHarvesterSimple` to `wrappedNativeToken` to avoid confusions when using for wS on Sonic
- Move Base proxies to a separate file
- `OETHBaseHarvester` is no longer used (I'm splitting up the swap functions to a Safe module in a different branch to help with our operations better)
- **NOTE**: Storage slot changes from the existing `OETHBaseHarvester`, however the deployment file calls `initialize` and that should reset the slots

### DIFF
--- a/contracts/contracts/harvest/OETHHarvesterSimple.sol
+++ b/contracts/contracts/harvest/OETHHarvesterSimple.sol
@@ -88,6 +88,7 @@ contract OETHHarvesterSimple is Initializable, Strategizable {
 
         // Store locally for some gas savings
         address _strategist = strategistAddr;
+        address _dripper = dripper;
 
         // Harvest rewards
         IStrategy(_strategy).collectRewardTokens();
@@ -104,7 +105,7 @@ contract OETHHarvesterSimple is Initializable, Strategizable {
             if (balance > 0) {
                 // Determine receiver
                 address receiver = token == wrappedNativeToken
-                    ? dripper
+                    ? _dripper
                     : _strategist;
                 require(receiver != address(0), "Invalid receiver");
 
--- a/contracts/contracts/harvest/OETHHarvesterSimple.sol
+++ b/contracts/contracts/harvest/OETHHarvesterSimple.sol
@@ -1,4 +1,4 @@
-// SPDX-License-Identifier: MIT
+// SPDX-License-Identifier: BUSL-1.1
 pragma solidity ^0.8.0;
 
 import { Strategizable } from "../governance/Strategizable.sol";
@@ -16,8 +16,8 @@ contract OETHHarvesterSimple is Initializable, Strategizable {
     ////////////////////////////////////////////////////
     /// --- CONSTANTS & IMMUTABLES
     ////////////////////////////////////////////////////
-    /// @notice WETH address
-    address public immutable WETH;
+    /// @notice wrapped native token address (WETH or wS)
+    address public immutable wrappedNativeToken;
 
     ////////////////////////////////////////////////////
     /// --- STORAGE
@@ -46,8 +46,8 @@ contract OETHHarvesterSimple is Initializable, Strategizable {
     ////////////////////////////////////////////////////
     /// --- CONSTRUCTOR
     ////////////////////////////////////////////////////
-    constructor(address _WETH) {
-        WETH = _WETH;
+    constructor(address _wrappedNativeToken) {
+        wrappedNativeToken = _wrappedNativeToken;
     }
 
     /// @notice Initialize the contract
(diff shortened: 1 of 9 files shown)


contracts/contracts/proxies/BaseProxies.sol
```diff
@@ -0,0 +1,67 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.0;
+
+import { InitializeGovernedUpgradeabilityProxy } from "./InitializeGovernedUpgradeabilityProxy.sol";
+
+/**
+ * @notice BridgedBaseWOETHProxy delegates calls to BridgedWOETH implementation
```
clement-ux, line 7: Thanks for creating a special proxy file for Base proxies!

contracts/contracts/harvest/OETHHarvesterSimple.sol
```diff
@@ -82,10 +82,13 @@ contract OETHHarvesterSimple is Initializable, Strategizable {
     }
 
     /// @notice Internal logic to harvest rewards from a strategy
-    function _harvestAndTransfer(address _strategy) internal {
+    function _harvestAndTransfer(address _strategy) internal virtual {
         // Ensure strategy is supported
         require(supportedStrategies[_strategy], "Strategy not supported");
 
+        // Store locally for some gas savings
+        address _strategist = strategistAddr;
```
sparrowDom, line 90: There seem to be only 1 call for the strategist address in the ternary expression below. We can delete this variable?

shahthepro: That's inside a `for` loop :)

sparrowDom: ahh yes, I stand corrected :)

contracts/contracts/harvest/SuperOETHHarvester.sol
```diff
@@ -0,0 +1,40 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.0;
+
+import { OETHHarvesterSimple, IERC20, IStrategy, SafeERC20 } from "./OETHHarvesterSimple.sol";
+
+contract SuperOETHHarvester is OETHHarvesterSimple {
+    using SafeERC20 for IERC20;
+
+    constructor(address _wrappedNativeToken)
```
sparrowDom, line 9: Since it doesn't do anything extra is this constructor required?

contracts/contracts/harvest/OETHHarvesterSimple.sol
shahthepro: Now that you point it out, I should move the `dripper` as well

contracts/contracts/harvest/SuperOETHHarvester.sol
shahthepro: I think, yes, since it specifies a parameter to be set. I get a `TypeError` if I don't have it:
```
TypeError: No arguments passed to the base constructor. Specify the arguments or mark "SuperOETHHarvester" as abstract.
 --> contracts/harvest/SuperOETHHarvester.sol:6:1:
  |
6 | contract SuperOETHHarvester is OETHHarvesterSimple {
  | ^ (Relevant source part starts here and spans across multiple lines).
Note: Base constructor parameters:
  --> contracts/harvest/OETHHarvesterSimple.sol:49:16:
   |
49 |     constructor(address _wrappedNativeToken) {
   |                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

```

contracts/contracts/harvest/SuperOETHHarvester.sol
```diff
@@ -0,0 +1,40 @@
...
+        OETHHarvesterSimple(_wrappedNativeToken)
+    {}
+
+    /// @inheritdoc OETHHarvesterSimple
+    function _harvestAndTransfer(address _strategy) internal virtual override {
```
sparrowDom, line 14: you could have a `getTokenReceiver(address token) virtual` function defined in `OETHHarvestersimple`: 
```
function _getTokenReceiver(address token) internal returns(address receiver) {
    receiver = token == WETH ? dripper : strategistAddr;
}
 // and the one from SuperOETHHarveter.sol
function _getTokenReceiver(address token) internal returns(address receiver) {
    receiver = strategistAddr;
}
```

This way the `_harvestAndTransfer` function could remain the same in both files.

sparrowDom (approved):
LGTM

contracts/contracts/harvest/SuperOETHHarvester.sol
sparrowDom: Ah interesting thanks for testing it

contracts/contracts/harvest/SuperOETHHarvester.sol
shahthepro: That's a fair point and I thought about it. But we will have to call the `_getTokenReceiver` inside the loop, since we are using this on the mainnet as well, I didn't wanna be lax with gas usage

clement-ux (approved):
Looks good to me, very clean 👍

PR #2400, opened 2025-02-14
Pool booster

sparrowDom:
## Pool booster
This deploys a Pool Booster Factory on Sonic which has the ability to create individual pool booster. Pool booster is a contract that receives forwarded yield (from an AMM pool) and forwards it to corresponding Bribe contract -> which is part of the Gauge incentives. 

There is a single instance of pool booster factory for a pool booster deployment. The factories are immutable - meaning if there is a need to fix a bug in either the factory or a pool booster a new factory needs to be deployed. While the "old" factory is still live and holds references to created pool boosters.

There is a central registry that will hold references to all the approved and deployed factories.

Notice: `PoolBoosterSwapxPair` is a pool booster that supports Classic stable and classic volatile pools. Because there are no such pools on SwapX we can not fork tests its functionality. Also we can not test it on other non OS gauges since they don't accept the OS as an incentive token.  

## Code Change Checklist

- [ ]  Unit tests *if needed

Internal review:


clement-ux (commented):
I did the first pass on Factory and both PB implementation.

With more hindsight, I'm wondering if this Factory contract is really necessary? Maybe a simple UI can be enough? 

Because here,  the factory is used to keep a track of all PB deployed, in order to bribes all the PB at the same time. But: 
- `bribesAll()` have a significant amount of chance of reverting when amount of registered PB is high. 
- A simple multicall can do the same, with even more granularity (excluding some PB for example).
- as creation of PB is permissioned, factory isn't really needed imo. 

I know we already talked about this in the Spec document on Notion, but with more hindsight, I'm less convinced in the need of a Factory.
What's more, every time we will need to add a need PB type, we will need to upgrade the proxy, which is slow (governance) and unsafe! In comparison, adding it in the UI, is way faster, and doesn't imply security matter. 


I think having a Factory doesn't add value and bring a lot of hurdles. (Really sorry for changing my mind after finalizing specs).

contracts/contracts/poolBooster/PoolBoosterFactory.sol
```diff
@@ -0,0 +1,201 @@
...
+            _ammPoolAddress,
+            PoolBoosterType.SwapXClassicPool
+        );
+    }
+
+    function _deployContract(bytes memory bytecode)
+        internal
+        returns (address _address)
+    {
+        uint256 salt = 1;
```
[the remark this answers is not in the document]
sparrowDom: Thanks that is a great point. Fixed here: https://github.com/OriginProtocol/origin-dollar/pull/2400/commits/fe6005db227fdc2873309e5fa976255b4fe6d44d

contracts/contracts/poolBooster/PoolBoosterFactory.sol
```diff
@@ -0,0 +1,201 @@
...
+        );
+
+        poolBoosters.push(entry);
+        poolBoosterFromPool[_ammPoolAddress] = entry;
+
+        emit PoolBoosterDeployed(
+            poolBoosterAddress,
+            _ammPoolAddress,
+            PoolBoosterType.SwapXClassicPool
+        );
```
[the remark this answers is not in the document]
sparrowDom: Good point, I agree. Fixed here: https://github.com/OriginProtocol/origin-dollar/pull/2400/commits/c6b01f373cf3d351a495bd1b31f0709a212a7354

contracts/contracts/poolBooster/PoolBoosterFactory.sol
```diff
@@ -0,0 +1,201 @@
...
+        _setGovernor(governor);
+        _setStrategistAddr(strategist);
+    }
+
+    function bribeAll() external {
+        uint256 length = poolBoosters.length;
+        for (uint256 i = 0; i < length; i++) {
+            IPoolBooster(poolBoosters[i].boosterAddress).bribe();
+        }
+    }
```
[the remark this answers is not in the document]
pandadefi: You could keep `bribeAll`, add a try-catch mechanism to it, and add a second function to bribe a specific set of pools.

contracts/contracts/poolBooster/PoolBoosterFactory.sol
```diff
@@ -0,0 +1,201 @@
...
+            if (poolBoosters[i].boosterAddress == _poolBoosterAddress) {
+                // erase mapping
+                delete poolBoosterFromPool[poolBoosters[i].ammPoolAddress];
+                // erase array entry
+                poolBoosters[i].boosterAddress = poolBoosters[boostersLen - 1]
+                    .boosterAddress;
+                poolBoosters[i].ammPoolAddress = poolBoosters[boostersLen - 1]
+                    .ammPoolAddress;
+                poolBoosters[i].boosterType = poolBoosters[boostersLen - 1]
+                    .boosterType;
```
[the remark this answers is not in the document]
sparrowDom: For some reason I though all the values in the struct need to be overwritten, but that is clearly not the case. thanks for the comment, fixed here: https://github.com/OriginProtocol/origin-dollar/pull/2400/commits/0d7c392e06bd034ca072fba7b9f9ef3e37aeab77

contracts/contracts/poolBooster/PoolBoosterSwapxIchi.sol
```diff
@@ -0,0 +1,58 @@
...
+        bribeContractOther = IBribe(_bribeContractOther);
+        osToken = IERC20(_osToken);
+        split = _split;
+    }
+
+    function bribe() external override {
+        uint256 balance = osToken.balanceOf(address(this));
+        uint256 osBribeAmount = balance.mulTruncate(split);
+        // -1 to prevent possible rounding issues with OS token
+        uint256 otherBribeAmount = balance - osBribeAmount - 1;
```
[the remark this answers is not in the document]
sparrowDom: Great point thanks. Fix here: https://github.com/OriginProtocol/origin-dollar/pull/2400/commits/4297827b94e3f714b64f42b57dce2ad36519e18f

contracts/contracts/poolBooster/PoolBoosterSwapxIchi.sol
```diff
@@ -0,0 +1,58 @@
...
+
+        osToken.approve(address(bribeContractOS), osBribeAmount);
+        osToken.approve(address(bribeContractOther), otherBribeAmount);
+
+        bribeContractOS.notifyRewardAmount(address(osToken), osBribeAmount);
+        bribeContractOther.notifyRewardAmount(
+            address(osToken),
+            otherBribeAmount
+        );
```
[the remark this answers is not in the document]
sparrowDom: Good point, the minimum bribe amount addresses this problem as well: https://github.com/OriginProtocol/origin-dollar/pull/2400/commits/4297827b94e3f714b64f42b57dce2ad36519e18f

contracts/contracts/poolBooster/PoolBoosterFactory.sol
```diff
@@ -0,0 +1,201 @@
...
+    ) external onlyGovernor {
+        require(_bribeAddressOS != address(0), "Invalid bribeAdressOS address");
+        require(
+            _bribeAddressOther != address(0),
+            "Invalid bribeAdressOther address"
+        );
+        require(
+            _ammPoolAddress != address(0),
+            "Invalid ammPoolAddress address"
+        );
```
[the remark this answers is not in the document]
sparrowDom: Great point. Fixed here: https://github.com/OriginProtocol/origin-dollar/pull/2400/commits/5436d541e5b6b6a412b7499279bf7363726b227a

contracts/contracts/poolBooster/PoolBoosterSwapxPair.sol
```diff
@@ -0,0 +1,33 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import { IBribe } from "../interfaces/poolBooster/ISwapXAlgebraBribe.sol";
+import { IPoolBooster } from "../interfaces/poolBooster/IPoolBooster.sol";
+import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
+import { StableMath } from "../utils/StableMath.sol";
```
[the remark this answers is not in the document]
sparrowDom: thanks agreed: https://github.com/OriginProtocol/origin-dollar/pull/2400/commits/6aaa8c9bb0ea5fcf58dab4e0f694fbd996cbbf34

contracts/contracts/poolBooster/PoolBoosterFactory.sol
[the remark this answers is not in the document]
sparrowDom: agreed fixed here: https://github.com/OriginProtocol/origin-dollar/pull/2400/commits/1fb2b10e23a7ffaec65e8233fa38af07a717aaa7

contracts/contracts/poolBooster/PoolBoosterSwapxIchi.sol
```diff
@@ -0,0 +1,74 @@
...
+    function bribe() external override {
+        uint256 balance = osToken.balanceOf(address(this));
+        // balance too small, do no bribes
+        if (balance < MIN_BRIBE_AMOUNT) {
+            return;
+        }
+
+        uint256 osBribeAmount = balance.mulTruncate(split);
+        // -1 to prevent possible rounding issues with OS token
+        uint256 otherBribeAmount = balance - osBribeAmount - 1;
```
clement-ux, line 61: I have fuzz tested that `balance` could never be equal to `osBribeAmount` (due to rounding issue in `mulTruncate`). 👍

contracts/contracts/poolBooster/PoolBoostCentralRegistry.sol
```diff
@@ -0,0 +1,79 @@
...
+    /**
+     * @notice Adds a factory address to the approved factory addresses
+     * @param _factoryAddress address of the factory
+     */
+    function approveFactory(address _factoryAddress) external onlyGovernor {
+        require(_factoryAddress != address(0), "Invalid address");
+
+        factories.push(_factoryAddress);
+        emit FactoryApproved(_factoryAddress);
+    }
```
clement-ux, line 32: There is no check to ensure that the factory isn't already approved. 
But in the other hand, I don't see any issue if we have 2 times the same factory here. 
So I'm fine to keep it like this 👍

contracts/contracts/poolBooster/PoolBoosterFactorySwapxPair.sol
```diff
@@ -0,0 +1,52 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import { PoolBoosterSwapxPair } from "./PoolBoosterSwapxPair.sol";
+import { AbstractPoolBoosterFactory } from "./AbstractPoolBoosterFactory.sol";
+
+import "hardhat/console.sol";
```
clement-ux, line 7: Hehe

clement-ux (commented):
Everything looks really good!
Even though there are more contracts now, I believe it will be easier and faster to scale thanks to this architecture!

I’ve left some minor comments to check out.

contracts/contracts/poolBooster/AbstractPoolBoosterFactory.sol
```diff
@@ -0,0 +1,148 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.0;
+
+import { Governable } from "../governance/Governable.sol";
+import { PoolBoosterSwapxIchi } from "./PoolBoosterSwapxIchi.sol";
```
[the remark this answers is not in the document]
sparrowDom: thanks fixed: https://github.com/OriginProtocol/origin-dollar/pull/2400/commits/df5f5cbe7c67de98de0f41cf71e9e19baca08cbe

contracts/contracts/poolBooster/AbstractPoolBoosterFactory.sol
```diff
@@ -0,0 +1,148 @@
...
+    // @notice address of Origin Sonic
+    address public immutable oSonic;
+
+    // @notice list of all the pool boosters
+    PoolBoosterEntry[] public poolBoosters;
+    // @notice mapping of AMM pool to pool booster
+    mapping(address => PoolBoosterEntry) public poolBoosterFromPool;
+
+    /// @notice Gap for upgrade safety
+    uint256[48] private ___gap;
```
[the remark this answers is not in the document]
sparrowDom: definitely not needed. Fixed here: https://github.com/OriginProtocol/origin-dollar/pull/2400/commits/b39f08f900c9183ee4eca7987d4facc34536ac39

contracts/contracts/poolBooster/AbstractPoolBoosterFactory.sol
```diff
@@ -0,0 +1,148 @@
...
+     *         calls bribe() on them.
+     * @param _exclusionList A list of pool booster addresses to skip when
+     *        calling this function.
+     */
+    function bribeAll(address[] memory _exclusionList) external {
+        uint256 length = poolBoosters.length;
+        for (uint256 i = 0; i < length; i++) {
+            address poolBoosterAddress = poolBoosters[i].boosterAddress;
+            bool skipBribeCall = false;
+            for (uint256 j = 0; j < _exclusionList.length; j++) {
```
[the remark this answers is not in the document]
sparrowDom: good point: https://github.com/OriginProtocol/origin-dollar/pull/2400/commits/77c21d84a2c3c7ceaa7fb05aab6ebea10ec3796f

contracts/contracts/poolBooster/PoolBoosterFactorySwapxPair.sol
sparrowDom: https://github.com/OriginProtocol/origin-dollar/pull/2400/commits/713eab56634313be30ebb2f168cd7fbb9a6e813e 😬

contracts/contracts/poolBooster/PoolBoostCentralRegistry.sol
```diff
@@ -0,0 +1,79 @@
...
+
+    /**
+     * @notice Removes the factory from approved factory addresses
+     * @param _factoryAddress address of the factory
+     */
+    function removeFactory(address _factoryAddress) external onlyGovernor {
+        require(_factoryAddress != address(0), "Invalid address");
+        require(isApprovedFactory(_factoryAddress), "Not an approved factory");
```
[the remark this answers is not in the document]
sparrowDom: This is a governance action an not really expected to be called often (if at all). For that reason I wouldn't be too worried with extra gas usage?

contracts/contracts/poolBooster/PoolBoosterSwapxIchi.sol
sparrowDom: thanks for that

contracts/contracts/poolBooster/PoolBoostCentralRegistry.sol
sparrowDom: Actually, that is a great point. Adding a check and a test here: https://github.com/OriginProtocol/origin-dollar/pull/2400/commits/735f55f18fae473b309d58b438569410822fbfe5

In case we have 2 entries of the same factory we could get a false sense of security from calling remove once. Since the other factory would still be in the list and considered supported.

contracts/contracts/poolBooster/PoolBoosterSwapxIchi.sol
```diff
@@ -0,0 +1,74 @@
...
+        require(
+            _bribeContractOS != address(0),
+            "Invalid bribeContractOS address"
+        );
+        require(
+            _bribeContractOther != address(0),
+            "Invalid bribeContractOther address"
+        );
+        // expect it to be between 1% & 99%
+        require(_split > 1e16 && _split < 99e16, "Unexpected split amount");
```
pandadefi, line 44: range could be between 0 and 100%.

contracts/contracts/poolBooster/PoolBoosterSwapxIchi.sol
```diff
@@ -0,0 +1,74 @@
...
+    //         bribe contract and 60% to other bribe contract
+    uint256 public immutable split;
+
+    // @notice if balance under this amount the bribe action is skipped
+    uint256 public constant MIN_BRIBE_AMOUNT = 1e10;
+
+    constructor(
+        address _bribeContractOS,
+        address _bribeContractOther,
+        address _osToken,
```
pandadefi, line 32: The osToken should be oeth, ousd... tokens with 18 decimals. Maybe just make sure it's not a token with less decimals that is added to make sure `MIN_BRIBE_AMOUNT` is still valid

contracts/contracts/poolBooster/PoolBoosterFactorySwapxPair.sol
```diff
@@ -0,0 +1,50 @@
...
+        address _bribeAddress,
+        address _ammPoolAddress,
+        uint256 _salt
+    ) external onlyGovernor {
+        require(
+            _ammPoolAddress != address(0),
+            "Invalid ammPoolAddress address"
+        );
+        require(_salt > 0, "Invalid salt");
+
```
pandadefi, line 35: The swapX gauges can't be created by anyone, how will the deployment workflow be like?

contracts/contracts/poolBooster/PoolBoosterSwapxIchi.sol
sparrowDom: I started to implement this change but then came to this realisation: 
- the split is immutable. Whatever it gets set to that stays unless another pool booster is created and yield is re-forwarded to the new address
- if there is a need for single sided bribes (0% or 100%) then it is IMO better to use the `PoolBoosterSwapxPair` pool booster is the same as the `PoolBoosterSwapxIchi` one except that it supports only one gauge. 

This has me thinking if I should rename the `PoolBoosterSwapxIchi` & `PoolBoosterSwapxPair` to `PoolBoosterSwapxSplitBribe` & `PoolBoosterSwapxSingleBribe`. Which would better describe the functionality of the pool boosters.

Yes actually let me do the renames.

sparrowDom: Done here: https://github.com/OriginProtocol/origin-dollar/pull/2400/commits/a6d21547cfac7d80ea6922897a8c03e817ddb51d

contracts/contracts/poolBooster/PoolBoosterSwapxIchi.sol
sparrowDom: Our tokens always have 18 decimals, and if we were to break that rule, we'd have problems to address all through the codebase. Still good comment, thanks.

contracts/contracts/poolBooster/PoolBoostCentralRegistry.sol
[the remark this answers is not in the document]
clement-ux: Indeed, not a problem to keep it like this 👍

contracts/contracts/poolBooster/PoolBoostCentralRegistry.sol
clement-ux: I didn't thought about this situation but indeed! Thanks for fixing it.

clement-ux (commented):
## Requirements
This PR aims to create a Pool Booster setup for Sonic ecosystem, which will use factory to ensure that deployed PR only use selected contract bytecode.

## Easy Checks

#### Authentication

- [x] Never use tx.origin

#### Ethereum

#### Cryptographic code

- [ ] No signed data could be used in a replay attack, on our contract or others.

#### Gas problems

    - [ ] A way to remove items

#### Black magic

- [ ] Does not contain `selfdestruct`
- [ ] Does not use `delegatecall` outside of proxying. _If an implementation contract were to call delegatecall under attacker control, it could call selfdestruct the implementation contract, leading to calls through the proxy silently succeeding, even though they were failing._
- [ ] Address.isContract should be treated as if could return anything at any time, because that's reality.

#### Overflow

- [ ] Code is solidity version >= 0.8.0

#### Proxy

- [ ] No storage variable initialized at definition when contract used as a proxy implementation.

#### Events

## Medium Checks

#### Rounding and casts

#### Dependencies
- [ ] Review any new contract dependencies thoroughly (e.g. OpenZeppelin imports) when new dependencies are added or version of dependencies changes.

#### External calls

- [ ] No malicious behaviors
- [ ] Oracles, one of:
  - [ ] No oracles
  - [ ] Oracles can't be bent

#### Tests

#### Deploy

## Strategy Specific

#### Strategy checks

#### Downstream

## Thinking

#### Logic

#### Internal State

Does this code do that?

#### Attack

#### Flavor

_Could this code be simpler?_

contracts/contracts/poolBooster/PoolBoosterFactorySwapxDouble.sol
```diff
@@ -54,4 +57,43 @@ contract PoolBoosterFactorySwapxDouble is AbstractPoolBoosterFactory {
...
+        address _bribeAddressOther,
+        address _ammPoolAddress,
+        uint256 _split,
+        uint256 _salt
+    ) external view returns (address) {
+        require(
+            _ammPoolAddress != address(0),
+            "Invalid ammPoolAddress address"
+        );
+        require(_salt > 0, "Invalid salt");
```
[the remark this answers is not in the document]
sparrowDom: This is a somewhat special case I guess, and feel free to criticise the following logic: 
- `computePoolBoosterAddress` os called with the exact same parameters as `createPoolBoosterSwapxDouble`. Even though the former doesn't require the `_ammPoolAddress`
- the reason behind it is so that governance deploy actions can call both functions by passing a [stored array of arguments](https://github.com/OriginProtocol/origin-dollar/pull/2400/files#diff-072de69218436548874e61b3c8f77fd888a65d1f7b66f8ef59cb7534d551acf8R67-R73) 
- and the computed address returned and the one actually created by the create2 are the same

I wanted to make those 2 functions have the same signature (# of parameters) and the same behaviour (reverts on incorrect values) so that deploy processes are as simple (have as little complexity) as possible.

This might not be the best approach though... should we cancel the reverts in the view function? I think I like the view function reverting on deploy (and not on governance tx execution) when incorrect values are passed to the address calculation.

contracts/contracts/poolBooster/PoolBoostCentralRegistry.sol
```diff
@@ -0,0 +1,83 @@
...
```
DanielVF, line 44: This is basically doing a second scan of the list of factories. Couldn't just the remove loop handle verifying that it found it?

sparrowDom:
## Requirements

This PR creates a setup for contract (factory) creation of pool booster contract. These contracts will have yield forwarded to them and then use that yield to execute the bribe incentives on gauges. 

## Easy Checks

#### Authentication
- [x] Never use tx.origin

#### Ethereum

#### Cryptographic code

#### Gas problems

    - [X] A way to remove items

#### Black magic

#### Overflow

#### Proxy

#### Events

## Medium Checks

#### Rounding and casts
- [ ] ~~Safecast is aways used when casting~~

#### Dependencies
- [ ] ~~Check OpenZeppelin [security vulnerabilities](https://github.com/OpenZeppelin/openzeppelin-contracts/security/advisories) and see if any apply to current PR considering the version of OpenZeppelin contract used.~~

#### External calls

- [ ] ~~Reentrancy guards on all state changing functions~~
       -- no need to protect against re-entrency 
- [x] No malicious behaviors
- [ ] ~~Low level call() must require success.~~
- [ ] ~~No slippage attacks (we need to validate expected tokens received)~~
- [x] Oracles, one of:
  - [x] No oracles
  - [ ] Oracles can't be bent

#### Tests

  - [X] Edge conditions are tested
  - [] ~~If tests interact with AMM make sure enough edge cases (pool tilts) are tested. Ideally with fuzzing.~~

#### Deploy

#### Downstream

## Thinking

#### Logic

**TODO from here on forward**

#### Internal State

Does this code do that?

#### Attack

#### Flavor

_Could this code be simpler?_

contracts/contracts/poolBooster/PoolBoostCentralRegistry.sol
```diff
@@ -0,0 +1,83 @@
...
+pragma solidity ^0.8.0;
+
+import { Governable } from "../governance/Governable.sol";
+
+/**
+ * @title Contract that holds all governance approved pool booster Factory
+ *        implementation deployments
+ * @author Origin Protocol Inc
+ */
+contract PoolBoostCentralRegistry is Governable {
```
DanielVF, line 11: If we want a registry of pool boosters, would it be easier to actually store the pool booster addresses here, as well as emit the events here? Otherwise the backend has to listen to this contract, then add listeners for each factory in order to listen to new events.

If we had pool creation events here, then this would be all that would need to be tracked.

Each pool creation event could also emit the factory contract address that created it / submitted it.

contracts/contracts/poolBooster/PoolBoosterFactorySwapxDouble.sol
```diff
@@ -0,0 +1,99 @@
...
+     *        should match the one from `computePoolBoosterAddress` in order for the final deployed address
+     *        and pre-computed address to match
+     */
+    function createPoolBoosterSwapxDouble(
+        address _bribeAddressOS,
+        address _bribeAddressOther,
+        address _ammPoolAddress,
+        uint256 _split,
+        uint256 _salt
+    ) external onlyGovernor {
```
DanielVF, line 39: Would we be good with having a lower privileges address create pools, to reduce operational burden? Without yield headed in, it should be fine.

contracts/contracts/poolBooster/PoolBoosterFactorySwapxDouble.sol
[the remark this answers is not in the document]
DanielVF: I am very good with reverts here. This isn't something like balanceOf or totalSupply() that are going to be needed for the normal function of the rest of the ecosystem.

contracts/contracts/poolBooster/PoolBoosterSwapxDouble.sol
```diff
@@ -0,0 +1,75 @@
...
+    function bribe() external override {
+        uint256 balance = osToken.balanceOf(address(this));
+        // balance too small, do no bribes
+        if (balance < MIN_BRIBE_AMOUNT) {
+            return;
+        }
+
+        uint256 osBribeAmount = balance.mulTruncate(split);
+        // -1 to prevent possible rounding issues with OS token
+        uint256 otherBribeAmount = balance - osBribeAmount - 1;
```
DanielVF, line 63: Do we even have this problem any more? :D :D

sparrowDom: I know we solved this for the contract type accounts or EOA's that have opted out of rebasing. Since those had the CPT set to 1e18 and internal credits to the actual balance. 

Not sure this is solved for accounts that participate in the yield?

contracts/contracts/poolBooster/PoolBoosterFactorySwapxDouble.sol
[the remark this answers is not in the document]
clement-ux: All good then 👍

contracts/contracts/poolBooster/PoolBoosterFactorySwapxDouble.sol
sparrowDom: [For now](https://github.com/OriginProtocol/origin-dollar/pull/2400/files#diff-072de69218436548874e61b3c8f77fd888a65d1f7b66f8ef59cb7534d551acf8R80-R107) the pool boosters are going to be created via the same governance TX that will also enable yield forwarding to the boosters. Which I would argue is the lowest effort operational approach. 
We could make booster Strategizable instead of Governable, adding the extra strategist role. Though I am not sure the added complexity is necessary. Do you think we should also enable Strategists to create pool boosters?

contracts/contracts/poolBooster/PoolBoostCentralRegistry.sol
sparrowDom: I am on the fence deciding if the extra complexity justifies the monitoring convenience. The way I am imagining it the necessary complexity to add would be: 
  - Central registry needs an external function to emit pool booster events. 
  - factories would need to have a reference to central registry so it can callback when a new pool booster is created. 
  - when the mentioned function is permissioned meaning it cycles through all the factories it holds in its array

DanielVF: Just a mapping of factories is fine. Backend can have a list from events. Backend no longer needs to talk to each factory to find out what is going on.

contracts/contracts/poolBooster/PoolBoosterFactorySwapxDouble.sol
DanielVF: It's fine as you have it with the create2.

contracts/contracts/poolBooster/PoolBoosterSwapxDouble.sol
DanielVF: It's now pixel perfect for until the resolution runs out, at which point it will error towards more money in the source and more money in the destination on a transfer.

contracts/contracts/poolBooster/AbstractPoolBoosterFactory.sol
```diff
@@ -0,0 +1,173 @@
...
+        _setGovernor(_governor);
+    }
+
+    /**
+     * @notice Goes over all the pool boosters created by this factory and
+     *         calls bribe() on them.
+     * @param _exclusionList A list of pool booster addresses to skip when
+     *        calling this function.
+     */
+    function bribeAll(address[] memory _exclusionList) external {
```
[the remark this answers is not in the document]
DanielVF: Virtual means must be overridden? You can still override an external if you want to, I believe.

contracts/contracts/poolBooster/PoolBoostCentralRegistry.sol
DanielVF: Anothe way to look at the complexity here is that the factories, of which we have a lot, and will build a lot, no longer have to have any tracking code. The amount of code, and of duplicate deployed code will go down.

contracts/contracts/poolBooster/AbstractPoolBoosterFactory.sol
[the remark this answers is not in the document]
clement-ux: image

Is this what you mean?

contracts/contracts/poolBooster/PoolBoosterFactorySwapxDouble.sol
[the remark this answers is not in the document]
clement-ux: All good then 👍

contracts/contracts/poolBooster/AbstractPoolBoosterFactory.sol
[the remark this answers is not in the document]
sparrowDom: It is quite possible that we will need to override some of those functions. My thinking is why not mark them as virtual once we have a need for it?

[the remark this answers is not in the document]
clement-ux: I forget that this part will/could be deployed multiple time, so indeed, we could still change it when we will need it. All good then ;)

contracts/contracts/poolBooster/PoolBoosterSwapxDouble.sol
sparrowDom: Thanks for pointing this out. Addressed here: https://github.com/OriginProtocol/origin-dollar/pull/2400/commits/a3f537714a170276b5ca1ce3840b6a1c0a4d2bf7

contracts/contracts/poolBooster/AbstractPoolBoosterFactory.sol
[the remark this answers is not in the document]
DanielVF: I was wrong about how virtual behaves. Lesson learned.

contracts/contracts/poolBooster/PoolBoostCentralRegistry.sol
sparrowDom: You make a good point that for monitoring this will simplify things quite significantly. Change here: https://github.com/OriginProtocol/origin-dollar/pull/2400/commits/0148ddda29170e498c4d0ad97307e2ece5d6b7b2

clement-ux (dismissed):
I really like these new changes! 
Having only the CentralRegistry that emit event will simplify a lot the monitoring 👍

contracts/contracts/poolBooster/PoolBoostCentralRegistry.sol
sparrowDom: Actually re-introduced the list back from mapping. As this way it is easy to iterate over all of the supported factories and call `bribeAll` on each of them: https://github.com/OriginProtocol/origin-dollar/pull/2400/commits

contracts/contracts/poolBooster/PoolBoostCentralRegistry.sol
```diff
@@ -41,21 +47,45 @@ contract PoolBoostCentralRegistry is Governable {
...
+    /**
+     * @notice Emits a pool booster created event
+     * @dev    This has been created as a convenience method for the monitoring to have
+     *         an index of all of the created pool boosters by only listening to the
+     *         events of this contract.
+     * @param _poolBoosterAddress address of the pool booster created
+     * @param _ammPoolAddress address of the AMM pool forwarding yield to the pool booster
+     * @param _boosterType PoolBoosterType the type of the pool booster
+     */
+    function emitPoolBoosterCreated(
```
DanielVF, line 78: We really should include the factory in the emit event. This way we know what kind of pool, and exactly the version that was created.

contracts/contracts/poolBooster/AbstractPoolBoosterFactory.sol
```diff
@@ -0,0 +1,174 @@
...
+        address ammPoolAddress;
+        IPoolBoostCentralRegistry.PoolBoosterType boosterType;
+    }
+
+    event PoolBoosterCreated(
+        address poolBoosterAddress,
+        address ammPoolAddress,
+        IPoolBoostCentralRegistry.PoolBoosterType poolBoosterType
+    );
+    event PoolBoosterRemoved(address poolBoosterAddress);
```
clement-ux, line 24: Those two events seem to not be used anymore, no?

contracts/contracts/poolBooster/AbstractPoolBoosterFactory.sol
```diff
@@ -50,11 +35,21 @@ contract AbstractPoolBoosterFactory is Governable {
...
+    ) {
         require(_oSonic != address(0), "Invalid oSonic address");
         require(_governor != address(0), "Invalid governor address");
+        require(
+            _centralRegistry != address(0),
+            "Invalid central registry address"
+        );
 
         oSonic = _oSonic;
+        centralRegistry = IPoolBoostCentralRegistry(_centralRegistry);
```
DanielVF, line 52: Just noting that bribeAll() will break if the list gets too large. I think we are okay with this as long as the boosters themselves can have bribe called outside the factory.

contracts/contracts/poolBooster/AbstractPoolBoosterFactory.sol
sparrowDom: done here: https://github.com/OriginProtocol/origin-dollar/pull/2400/commits/25f0d6277b87d1e493e4b34e7e232ee2c06cecca

contracts/contracts/poolBooster/AbstractPoolBoosterFactory.sol
sparrowDom: Yeah good point. We could limit the number of poolBoosters per factory by deploying multiple factories even when the codebase doesn't change. To not hit that gas usage limit.

Or we could pass an offset & limit to the bribeAll() function, but that seems too messy. 

I guess we will need to handle the loss of convenience on the operational side.

contracts/contracts/poolBooster/PoolBoostCentralRegistry.sol
sparrowDom: Thanks great point done here: https://github.com/OriginProtocol/origin-dollar/pull/2400/commits/e45b7610f5b384c3b8dedf2bfea2f34883dec0b0

PR #2421, opened 2025-02-28
WOETH  - Fixed yield rate each day

DanielVF:
## Code Change Checklist

- [ ]  Unit tests *if needed

Internal review:

## Deploy checklist

```
```

### DIFF
--- a/contracts/contracts/token/WOETH.sol
+++ b/contracts/contracts/token/WOETH.sol
@@ -141,7 +141,7 @@ contract WOETH is ERC4626, Governable, Initializable {
         uint256 _end = yieldEnd;
         if (block.timestamp >= _end) {
             return hardAssets + yieldAssets;
-        } else if (block.timestamp < _end - YIELD_TIME) {
+        } else if (block.timestamp <= _end - YIELD_TIME) {
             return hardAssets;
         }
         return
--- a/contracts/contracts/token/WOETH.sol
+++ b/contracts/contracts/token/WOETH.sol
@@ -109,31 +109,28 @@ contract WOETH is ERC4626, Governable, Initializable {
 
     // @notice Called to start a yield period, if one is not active
     function startYield() public {
-        // If we are currently giving yield, do not adjust the yield rate
+        // If we are currently distributing yield, continue until done
         if (block.timestamp < yieldEnd) {
             return;
         }
+        // Change to next yield period
+        yieldEnd = (block.timestamp + YIELD_TIME).toUint128();
+        // Compute yield and set future yield
         uint256 _computedAssets = totalAssets();
         uint256 _actualAssets = IERC20(asset()).balanceOf(address(this));
-        if (_actualAssets == _computedAssets) {
-            // No change needed
-            return;
-        }
-        if (_actualAssets < _computedAssets) {
+        if (_actualAssets <= _computedAssets) {
             yieldAssets = 0;
+            hardAssets = _actualAssets;
         } else if (_actualAssets > _computedAssets) {
             uint256 _newYield = _actualAssets - _computedAssets;
-            // Cap yield
             uint256 _maxYield = (_actualAssets * 5) / 100; // Maximum of 5% increase in assets per day
             _newYield = _min(_min(_newYield, _maxYield), type(uint128).max);
             yieldAssets = _newYield.toUint128();
+            hardAssets = _computedAssets.toInt256();
         }
-        hardAssets = _computedAssets.toInt256();
-        yieldEnd = (block.timestamp + YIELD_TIME).toUint128();
         emit YiedPeriodStarted(hardAssets, yieldAssets, yieldEnd);
     }
 
-    /** @dev See {IERC4262-totalAssets} */
     function totalAssets() public view override returns (uint256) {
         uint256 _end = yieldEnd;
         if (block.timestamp >= _end) {
@@ -146,7 +143,6 @@ contract WOETH is ERC4626, Governable, Initializable {
         return (hardAssets + _unlockedYield.toInt256()).toUint256();
     }
 
-    /** @dev See {IERC4262-deposit} */
     function deposit(uint256 oethAmount, address receiver)
         public
         override
@@ -154,9 +150,9 @@ contract WOETH is ERC4626, Governable, Initializable {
     {
         woethAmount = super.deposit(oethAmount, receiver);
         hardAssets += oethAmount.toInt256();
+        startYield();
     }
 
-    /** @dev See {IERC4262-mint} */
     function mint(uint256 woethAmount, address receiver)
         public
         override
@@ -164,9 +160,9 @@ contract WOETH is ERC4626, Governable, Initializable {
     {
         oethAmount = super.mint(woethAmount, receiver);
         hardAssets += oethAmount.toInt256();
+        startYield();
     }
 
-    /** @dev See {IERC4262-withdraw} */
     function withdraw(
         uint256 oethAmount,
         address receiver,
@@ -174,9 +170,9 @@ contract WOETH is ERC4626, Governable, Initializable {
     ) public override returns (uint256 woethAmount) {
         woethAmount = super.withdraw(oethAmount, receiver, owner);
         hardAssets -= oethAmount.toInt256();
+        startYield();
     }
 
-    /** @dev See {IERC4262-redeem} */
     function redeem(
         uint256 woethAmount,
         address receiver,
@@ -184,6 +180,7 @@ contract WOETH is ERC4626, Governable, Initializable {
     ) public override returns (uint256 oethAmount) {
         oethAmount = super.redeem(woethAmount, receiver, owner);
         hardAssets -= oethAmount.toInt256();
+        startYield();
     }
 
     function _transfer(
--- a/contracts/contracts/token/WOETH.sol
+++ b/contracts/contracts/token/WOETH.sol
@@ -11,9 +11,10 @@ import { SafeCast } from "@openzeppelin/contracts/utils/math/SafeCast.sol";
 import { Governable } from "../governance/Governable.sol";
 import { Initializable } from "../utils/Initializable.sol";
 import { OETH } from "./OETH.sol";
+import { IVault } from "../interfaces/IVault.sol";
 
 /**
- * @title OETH Token Contract
+ * @title Wapped Token Contract
  * @author Origin Protocol
  *
  * @dev An ERC4626 contract that wraps a rebasing token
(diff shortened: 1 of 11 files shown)


sparrowDom:
I've added [a commit](https://github.com/OriginProtocol/origin-dollar/pull/2421/commits/793e37a50d47845f9b39427b7b117751e1aa57c1) to test the emission of yield start event. The 4 functions do not trigger the event - meaning that overriding `_transfer` doesn't work. Figuring out why not... 

Also with the current implementation `yieldStart` is triggered before the 4 functions alter `hardAssets`. Which I think could cause errors in `yieldStart` function.

contracts/contracts/token/WOETH.sol
```diff
@@ -109,31 +109,28 @@ contract WOETH is ERC4626, Governable, Initializable {
 
     // @notice Called to start a yield period, if one is not active
     function startYield() public {
-        // If we are currently giving yield, do not adjust the yield rate
+        // If we are currently distributing yield, continue until done
         if (block.timestamp < yieldEnd) {
             return;
         }
+        // Change to next yield period
+        yieldEnd = (block.timestamp + YIELD_TIME).toUint128();
```
sparrowDom, line 117: doesn't this extend the yield dripping even when there is no yield to be dripped out? For example in case of `_actualAssets <= _computedAssets `

contracts/contracts/token/WOETH.sol
```diff
@@ -90,103 +93,112 @@ contract WOETH is ERC4626, Governable, Initializable {
...
-        return (oethCreditsHighres).divPrecisely(creditsPerTokenHighres);
+    /* @notice Start the next yield period, if one is not active.
+     *     New yield will not start until time has moved forward.
+     */
+    function scheduleYield() public {
+        // If we are currently distributing yield, continue until end
+        if (block.timestamp < yieldEnd) {
+            return;
+        }
+        // Read current assets
```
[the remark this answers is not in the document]
DanielVF: Yes, changed.

contracts/contracts/token/WOETH.sol
```diff
@@ -6,31 +6,43 @@ import { ERC20 } from "@openzeppelin/contracts/token/ERC20/ERC20.sol";
...
 import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import { SafeCast } from "@openzeppelin/contracts/utils/math/SafeCast.sol";
 
-import { StableMath } from "../utils/StableMath.sol";
 import { Governable } from "../governance/Governable.sol";
 import { Initializable } from "../utils/Initializable.sol";
 import { OETH } from "./OETH.sol";
 
 /**
  * @title OETH Token Contract
```
rappie, line 16: OETH -> WOETH

DanielVF:
# Code Review - WOETH

## Requirements

We want to ensure that no donations can instantly affect the exchange rate used by lending platforms. This is because an instantly changing exchange rate on a redeemable asset can be fatal to a lending platform that uses it for borrows. 

The approach we chose here controls the exchange rate by slowly streaming out additional assets over time.

### Asset yield streaming

By controlling the extra asset amounts, rather than the rate, we avoids the easy critical bugs around fixed exchange rate changes in the face of changing asset amounts.

🟣 However a fixed asset per second yield does mean that the exchange rate can change over the yield period, as the total supply changes. If we assume that the standard attack requires a 2x donation, if an attacker perfectly time this around the end of a yield period, once they redeemed their half of funds, the rate of change of the exchange rate would double. However this still spreads their donation out over 23 hours, which still cancels out the attack.

### No flash loans

A core defense of this contract is that donations can only affect subsequent blocks. It is impossible to use a flash loan and see even the slightest change in exchange rate inside the same block.

🟣 However, a whale attacker can reduce the cost of their donation by coming back into woeth after the attack has landed, and collecting a portion of the streaming yield. While this reduces the cost of the attack, it does not reduce the massively increased size of the donation needed.

Sample attacks:

For each attack, we will assume that an attacker needs a 50% price instant increase for the attack to land. It might be less, but it’s what we have seen in past attacks. It must be one block, because otherwise liquidations would rekt the attacker.

Assuming 5 million wOETH supply. 10 million wOETH supply.

Required donation sizes:

Old WOETH: 5 million
No direct donation WOETH: 10 million
This contract, on 12 seconds per block: 34,500 million.
 
A flash loan cannot be used to help with this.

Having a scheduled future yield does allow others to join the contract only when this contracts yield is known to be higher than other opportunities. This does however require them to already hold OETH, or the conversion fees or redeem time would probably preclude profitability.








## Easy Checks

#### Authentication

- [x] Never use tx.origin

#### Ethereum

- [x] Contract does not have a vulnerablity from being sent self destruct ETH

#### Cryptographic code

#### Gas problems

_No for loops_

#### Black magic

_No magic_

#### Overflow

#### Proxy

#### Events

## Medium Checks

#### Rounding and casts
	- [ ] 🟣 We send funds to a dead address
	- [ ] 🟣 We send funds to a dead address
	- [ ] 🟣 Used when needed

#### Dependencies

#### External calls

    - 🟣 Not needed, since only interacting with one trusted contract.

#### Tests

	  - 🟠 Should ensure we have a test for a loss of balance case.

#### Deploy

#### Logic

Looks good.

_No_

#### Internal State

- [x] yieldAssets should be smaller than trackedAssets
- [x] totalAssets() should never be larger than trackedAssets
- [x] totalAssets() should never be smaller than trackedAssets - yieldAssets


#### Attack

This code controls 30%-50% of our our assets. It’s widely used. A critical failure could allow these assets to be stolen.

This code is also used by lending platforms or other DeFi systems, and wrong values from the exchange rate could result in losses there beyond just our token.

The two primary attacks against this kind of contract are rounding errors and donation attacks. We think we’ve blocked both of these.


#### Flavor

This code errs on the side of simplicity, using if statements to clearly define edge case behavior, as well as named intermediate variables to keep things easy to understand.

- [x] Functions follow validate, read, compute, write, call, validate

contracts/contracts/token/WOETH.sol
```diff
@@ -90,103 +91,114 @@ contract WOETH is ERC4626, Governable, Initializable {
...
+            return;
+        }
+
+        // Read current assets
+        uint256 _computedAssets = totalAssets();
+        uint256 _actualAssets = IERC20(asset()).balanceOf(address(this));
+
+        // Compute next yield period values
+        if (_actualAssets <= _computedAssets) {
+            yieldAssets = 0;
```
DanielVF, line 121: 🟠 Should ensure we have a test for a loss of balance case.

sparrowDom: done here: https://github.com/OriginProtocol/origin-dollar/pull/2421/commits/78377f86759828416daac030fe086c51f94b99c8

sparrowDom:
@DanielVF minor code change: https://github.com/OriginProtocol/origin-dollar/pull/2421/commits/72a129287cab03c0af15432e0827bde780bfd6c2

contracts/contracts/token/WOETH.sol
```diff
@@ -90,103 +91,114 @@ contract WOETH is ERC4626, Governable, Initializable {
...
+            trackedAssets = _actualAssets;
+        } else if (_actualAssets > _computedAssets) {
+            uint256 _newYield = _actualAssets - _computedAssets;
+            uint256 _maxYield = (_computedAssets * 5) / 100; // Maximum of 5% increase in assets per day
+            _newYield = _min(_min(_newYield, _maxYield), type(uint128).max);
+            yieldAssets = _newYield.toUint128();
+            trackedAssets = _computedAssets + yieldAssets;
+        }
+        // raw cast is deliberate, since this will not perma revert
+        yieldEnd = uint64(block.timestamp + YIELD_TIME);
```
sparrowDom, line 131: heppning in 580 billion years :)

sparrowDom:
## Requirements
This PR prepares the WOETH contract to be safe to use on lending platforms by preventing the possibility of rapid inflation / donation attack to either the WOETH contract or the OETH Vault. 

## Easy Checks

#### Authentication

- [x] Never use tx.origin

#### Ethereum

#### Cryptographic code

#### Gas problems

- _No loops_

#### Black magic

#### Overflow

- _no for loops_

#### Proxy

#### Events

## Medium Checks

#### Rounding and casts
- 🟡  Contract does not have bugs from loosing rounding precision
  - contract can loose precision when rounding especially when the value of `shares` is inflated. For each order of magnitude difference between underlying assets and shares the error grows. Roughly 10 wei to the pow of orders of magnitude difference. 
- 🟡 Contract does not have bugs from zero or near zero amounts
  - it does. For this reason on a fresh deploy we need to mint a sufficient amount of WOETH and send it to a dead address. A small amount of shares (e.g. 1e9) is sufficient.
- 🟡 Safecast is aways used when casting
   - it is on all but 1 place - and it is not a safecast there intentionally 

#### Dependencies
- [ ] ~~If OpenZeppelin ACL roles are use review & enumerate all of them.~~no roles

#### External calls

- [X] No unsafe external calls
- 🟡  Reentrancy guards on all state changing functions
  - not needed, since the 1 contract called is OToken trusted contract
- [x] No malicious behaviors
- [x] Oracles, one of:
  - [x] No oracles
  - [ ] Oracles can't be bent

#### Tests

#### Deploy

## Thinking

#### Logic

Looks good. 

No

#### Internal State

- `trackedAssets` should be smaller or equal to `totalAssets`
- `trackedAssets - yieldAssets` should be a positive number
- `yieldEnd` should never be more than 23 hours into the future from the current time

#### Attack

Attacks on wrapper contracts like ours will usually try either a donation attack or a rounding attack when contract balances are small. This PR addresses the possibility of a donation attack to either the WOETH contract or the OUSD Vault. 

#### Flavor

The contract needs to represent the funds entering / exiting as part of wrapping and un-wrapping the token and the other funds which are result of yield & donations. We current have `trackedAssets` representing the assets entering, exiting & yield assets. Separate `yieldAssets` representing only the yield assets. It would be more organic to have a storage variable for assets only entering / exiting without including yield assets.

Though it is easier to understand there was a lot of casting necessary making the code uglier.

## Yield limit 🔴 
There is a way to game the 5% daily yield limit the contract sets as a limitation. The steps are: 
- wait for the contract to not be emitting yield
- take a flash loan and borrow great amounts of WETH
- mint OETH and deposit it to WOETH ( this will kick off yield start)
- redeem all the WOETH
- redeem all the OETH for WETH (and pay 0.1% redemption fee)

The exchange rate between OETH and WOETH still takes 23 hours to drip. That 5% daily (23 hours rather) yield limitation can be increased to any amount by paying the 0.1% in redemption fees.

DanielVF:
"You're being let go... Your department's being downsized. You're part of an outplacement. We're going in a different direction. We're not picking up your option..."

We've decided on not using this approach, and instead controlling yield inside the vault.

PR #2422, opened 2025-02-28
make it possible to reset the yield while yield dripping is active

sparrowDom:
make it possible to reset the yield while yield dripping is active


sparrowDom:
not doing this for now

PR #2424, opened 2025-03-02
[WIP] Curve Pool Booster Factory

clement-ux:
[WIP]

- Integrate Curve Pool Booster to current PoolBooster Factoy
- Add fee mechanism on other PB implementation.

### Why?
- Pool Booster are deployed to manage a bribe campaign on Votemarket.
- As Votemarket doesn't exist on mainnet but only on L2, we use the [CampaignRemoteManager](https://etherscan.io/address/0x000000009dF57105d76B059178989E01356e4b45) contract (that exist only on mainnet) to manage bribe (which exist on L2). 
- But this is important to be sure to have the control of the manager of this bribe on L2 too! Otherwise someone can control the bribe on our behalf.
- To do so we will replicate the PoolBooster address on the L2 (where the bribe runs), by deploying a contract at this address.
- This contract will be straightforward, it will be a proxy, where the implementation only allow to rescue ERC20. At the moment, this is the only mandatory action, as when we close a campaign, remaining token will be send to the manager address (so the Pool Booster address) on **L2!!!** That's why is is mandatory to have access to this address on L2.
- If more actions are required, we will be able to update the proxy.

### How?
- We will first deploy Factories Proxies from the 2/8 cross-chain MS using CreateX. As they have the same bytecode, it is possible to achieve same address cross-chain.
- Those factories will be able to deploy PoolBooster, using local create2 opcode. The PoolBooster will be BeaconProxy. Which means they'll have the same bytecode and it will be possible to achieve same address cross-chain for BeaconProxy. 
- We will use Beacon Proxy instead of classical Proxy because we will deploy multiple instance of PB and they will all share the same implementation. It is also easier to upgrade all. 
image

## Code Change Checklist

- [ ]  Unit tests *if needed

Internal review:

### DIFF
--- a/contracts/abi/createx.json
+++ b/contracts/abi/createx.json
@@ -2,6 +2,280 @@
     {
         "inputs": [
             {
+                "internalType": "address",
+                "name": "emitter",
+                "type": "address"
+            }
+        ],
+        "name": "FailedContractCreation",
+        "type": "error"
+    },
+    {
+        "inputs": [
+            {
+                "internalType": "address",
+                "name": "emitter",
+                "type": "address"
+            },
+            {
+                "internalType": "bytes",
+                "name": "revertData",
+                "type": "bytes"
+            }
+        ],
+        "name": "FailedContractInitialisation",
+        "type": "error"
+    },
+    {
+        "inputs": [
+            {
+                "internalType": "address",
+                "name": "emitter",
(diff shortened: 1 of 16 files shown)


clement-ux:
This project has been put on hold. But the technologie created here can be useful for a futur project!

PR #2427, opened 2025-03-04
Add defender action to run bribeAll on Sonic.

clement-ux:
Simple PR that add defender script that will manage bribes on Sonic.


naddison36 (approved):
Looks good

PR #2428, opened 2025-03-04
Nobody owns the implementation contracts

shahthepro:
## Code Change Checklist

- [ ]  Unit tests *if needed

Internal review:

## Deploy checklist

```
```


contracts/contracts/utils/InitializableAbstractStrategy.sol
```diff
@@ -82,6 +82,9 @@ abstract contract InitializableAbstractStrategy is Initializable, Governable {
      * @param _config The platform and OToken vault addresses
      */
     constructor(BaseStrategyConfig memory _config) {
+        // Nobody owns the implementation
+        _setGovernor(address(0));
+
```
DanielVF, line 87: Looks great.

We need to make sure the unit tests are okay with this.

I see one error

```
  1) OETH Vault
       Config
         Fail to cacheWETHAssetIndex if WETH is not a supported asset:
     AssertionError: Expected transaction to be reverted with "Invalid WETH Asset Index", but other reason was found: "Caller is not the Governor"
```

contracts/contracts/utils/InitializableAbstractStrategy.sol
```diff
@@ -82,6 +82,9 @@ abstract contract InitializableAbstractStrategy is Initializable, Governable {
      * @param _config The platform and OToken vault addresses
      */
     constructor(BaseStrategyConfig memory _config) {
```
sparrowDom, line 84: Not sure why we don't call the `Governable()` constructor here?

sparrowDom: though all that does is set the governor to msg.sender. I guess it is ok like it is

contracts/contracts/vault/VaultStorage.sol
```diff
@@ -250,6 +250,11 @@ contract VaultStorage is Initializable, Governable {
     // For future use
     uint256[44] private __gap;
 
+    constructor() Governable() {
```
sparrowDom, line 253: We could not call `Governable()` here as well. Since it will only set the governor to `msg.sender` and then have it overwritten by the 0 governor.

sparrowDom (commented):
Good changes. We should probably be consistent in constructors calling the `Governable()` constructor or not when those constructors themselves change the govenor.
I think it is cleaner not calling the `Governable()` constructor.

contracts/contracts/utils/InitializableAbstractStrategy.sol
naddison36: The `Governable` constructor is still executed even if the contract contracts inheriting `Governable` don't explicitly call `Governable()`.

As an example, contracts `B`, `C` and `D` in the below all have `setting` equal to 1.

```Solidity
contract A {
    uint256 public setting;

    constructor() {
        setting = 1;
    }
}

contract B is A {
    constructor() {
    }
}

contract C is A {
    constructor() A() {
    }
}

contract D is A {
}
```

naddison36:
There are also the wrapped OTokens that inherit `Governable()` but but are not calling `_setGovernor(address(0));`.

That's `WOETH`, `WOSonic` and `WrappedOusd`. `WOETHBase` inherits `WOETH` so will inherit whatever the `WOETH` constructor does.

naddison36:
An alternative approach is to remove the constructor on `Governable` so governor always defaults to a zero address.
This will only work if all contracts that use `Governable` are proxied, which I think they are.

naddison36 (approved):
Looks good

sparrowDom (approved):
LGTM

DanielVF:
I would be willing to be we will forget to emit events when calling `_setGovernor()` this.

Perhaps we should move the event emit from happening that `_changeGovernor()` to happening in `_setGovernor()`. Then `_setGovernor()` would be the only thing we need to call from things that use the governor.

Otherwise, looks good.

PR #2435, opened 2025-03-10
Hardhat tasks

naddison36:
## Changes

* No contract changes
* Added `sonicStaking`, `sonicUndelegate` and `sonicWithdraw` Hardhat task
* Renamed Hardhat task `setDefaultValidator` to `sonicDefaultValidator`
* Added rebasing tokens and assets per share to `snapVault`
* Fixed failing Sonic fork test

### DIFF
--- a/contracts/tasks/vault.js
+++ b/contracts/tasks/vault.js
@@ -67,6 +67,8 @@ async function snapVault({ block }, hre) {
   const totalSupply = await oToken.totalSupply({
     blockTag,
   });
+  const nonRebasingSupply = await oToken.nonRebasingSupply({ blockTag });
+  const rebasingSupply = totalSupply.sub(nonRebasingSupply);
 
   const queue = await vault.withdrawalQueueMetadata({
     blockTag,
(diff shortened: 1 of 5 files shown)


contracts/tasks/vault.js
```diff
@@ -40,6 +40,15 @@ async function getContracts(hre, symbol, assetSymbol) {
   const oToken = await ethers.getContractAt(symbol, oTokenProxy.address);
   log(`Resolved ${network} ${symbol} OToken to address ${oToken.address}`);
 
+  // Resolve the wrapped OToken. eg wOETH, wOSonic
+  const wOTokenProxy = await ethers.getContract(
```
shahthepro, line 44: Can we instead have a mapping with hardcoded names with Network IDs as keys? Would make it more obvious during searches. Also, we can easily accommodate any contracts that don't follow this naming convention

naddison36: We can, but it means devs have to remember to add to the mapping.
I suspect the next chain we add will break the convention and we'll have to use a mapping. Let's do that change when we need to so we don't need to do mapping maintenance.

shahthepro: That sounds good, thanks

PR #2437, opened 2025-03-10
WOETH delay maximum yield restriction

sparrowDom:
Restricts WOETH contract to be manipulated to immediately increase the yield rate to high amounts. This change makes it so that a manipulation only takes effect in 23 hours.


sparrowDom:
I don't think this is a good solution to the problem

PR #2439, opened 2025-03-11
Sonic SwapX AMO Strategy

naddison36:
## Changes

* Add new `SonicSwapXAMOStrategy` contract
* Deploy `OSonicHarvester` for collecting SWPx rewards and sending them to the Guardian account
* Upgrade Base Curve AMO Strategy so `withdrawalAll` will work with no assets
* Upgrade Base Vault with complete async withdrawal functionality. eg prevent depositing WETH reserved for the withdrawal queue to a strategy.
* Upgrade Mainnet OETH Vault to support multiple AMOs
* Fixed address in `OETHVaultAdmin`'s `StrategyAddedToMintWhitelist` and `StrategyRemovedFromMintWhitelist` events

sonicContracts

The inheritance of the Vault contracts

VaultHierarchy

## Testing

Fork tests are in `contracts/test/strategies/sonic/swapx-amo.sonic.fork-test.js`

```zsh
# Start a local forked node of Sonic with
yarn run node:sonic
```

```zsh
# Run the fork tests
yarn run test:sonic-fork
```

## Deployment

As there are changes to the Vault there are three deploy scripts on Sonic, Base and Mainnet

### Sonic

The Sonic deploy script is `contracts/deploy/sonic/009_swapx_amo.js`

```zsh
yarn run deploy:sonic 
```

### Base

The Sonic deploy script is `contracts/deploy/base/027_base_curve_amo_upgrade.js`

```zsh
yarn run deploy:base 
```

### Mainnet

The Sonic deploy script is `contracts/deploy/base/126_oeth_vault_upgrade.js`

```zsh
yarn run deploy:mainnet 
```

## Code Change Checklist

- [ ]  Unit tests *if needed

Internal review:


sparrowDom (commented):
I am still reviewing, just posting the first batch of comments

contracts/contracts/strategies/sonic/SonicSwapXAMOStrategy.sol
```diff
@@ -0,0 +1,707 @@
...
+     * @dev Checks that the protocol is solvent, protecting from a rogue Strategist / Guardian that can
+     * keep rebalancing the pool in both directions making the protocol lose a tiny amount of
+     * funds each time.
+     *
+     * Protocol must be at least SOLVENCY_THRESHOLD (99,8 %) backed in order for the rebalances to
+     * function.
+     */
+    function _solvencyAssert() internal view {
+        uint256 _totalVaultValue = IVault(vaultAddress).totalValue();
+        uint256 _totalOSSupply = IERC20(os).totalSupply();
```
sparrowDom, line 663: This will probably never be 0. But if it is we divide by 0

contracts/contracts/strategies/sonic/SonicSwapXAMOStrategy.sol
```diff
@@ -0,0 +1,707 @@
...
+        // diff = wS balance - OS balance
+        int256 diffAfter = wsReservesAfter.toInt256() -
+            osReservesAfter.toInt256();
+
+        if (diffBefore == 0) {
+            require(diffAfter == 0, "Position balance is worsened");
+        } else if (diffBefore < 0) {
+            // If the pool was originally imbalanced in favor of OS, then
+            // we want to check that the pool is now more balanced
+            require(diffAfter <= 0, "Assets overshot peg");
```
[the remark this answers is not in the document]
naddison36: Yes, this is the most likely scenario. There is more OS in the pool and `swapAssetsToPool` is used to burn OS from the pool. If we burn too much OS, then the pool can end up with more wS than OS. This check prevents this.

This is the same code used for the Curve AMOs on Mainnet and Base.

contracts/deploy/sonic/009_swapx_amo.js
```diff
@@ -0,0 +1,145 @@
...
+
+    const cHarvesterProxy = await ethers.getContract("OSonicHarvesterProxy");
+    const cHarvester = await ethers.getContractAt(
+      "OETHHarvesterSimple",
+      cHarvesterProxy.address
+    );
+
+    const cDripperProxy = await ethers.getContract("OSonicDripperProxy");
+
+    const initSonicStakingStrategy = cHarvester.interface.encodeFunctionData(
```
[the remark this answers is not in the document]
naddison36: Good pick. Fixed

contracts/contracts/strategies/sonic/SonicSwapXAMOStrategy.sol
```diff
@@ -0,0 +1,707 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.0;
+
+/**
+ * @title SwapX Automated Market Maker (AMO) Strategy
```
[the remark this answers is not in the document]
naddison36: I always get that wrong. Thanks. Fixed

contracts/contracts/strategies/sonic/SonicSwapXAMOStrategy.sol
```diff
@@ -0,0 +1,707 @@
...
+    ) internal {
+        // Transfer in tokens to the pool
+        IERC20(_tokenIn).safeTransfer(pool, _amountIn);
+
+        // Calculate how much out tokens we get from the swap
+        uint256 amountOut = IPair(pool).getAmountOut(_amountIn, _tokenIn);
+
+        // Safety check that we are dealing with the correct pool tokens
+        require(
+            (_tokenIn == ws || _tokenIn == os) &&
```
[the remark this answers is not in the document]
naddison36: Nice suggestion. I've used it.

contracts/contracts/strategies/sonic/SonicSwapXAMOStrategy.sol
```diff
@@ -0,0 +1,707 @@
...
+        IPair(pool).swap(amount0, amount1, address(this), new bytes(0));
+
+        // The slippage protection against the amount out is indirectly done
+        // via the improvePoolBalance
+    }
+
+    /// @dev Calculate the value of a LP position in a SwapX stable pool
+    /// if the pool was balanced.
+    /// @param lpTokens Amount of LP tokens in the SwapX pool
+    /// @return value The wS value of the LP tokens in the pool
```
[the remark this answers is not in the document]
naddison36: That's true. I'll update the Natspec

contracts/contracts/strategies/sonic/SonicSwapXAMOStrategy.sol
```diff
@@ -0,0 +1,707 @@
...
+    {
+        _deposit(_asset, _amount);
+    }
+
+    function _deposit(address _wS, uint256 _wsAmount) internal {
+        require(_wsAmount > 0, "Must deposit something");
+        require(_wS == ws, "Unsupported asset");
+
+        // Calculate the required amount of OS to mint based on the wS amount.
+        uint256 osLiquidity = _calcTokensToMint(_wsAmount);
```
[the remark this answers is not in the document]
naddison36: Good point. I was intending to be the amount added to the pool but liquidity is misleading for concentrated pools. 
I think `osDepositAmount` is better so will go with that.

contracts/contracts/strategies/sonic/SonicSwapXAMOStrategy.sol
```diff
@@ -0,0 +1,707 @@
...
+    // solhint-disable-next-line no-unused-vars
+    function _abstractSetPToken(address _asset, address _pToken)
+        internal
+        override
+    {}
+
+    function _approveBase() internal {
+        // Approve pool for OS and wS (required for adding liquidity)
+        // slither-disable-next-line unused-return
+        IERC20(os).approve(platformAddress, type(uint256).max);
```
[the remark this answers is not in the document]
naddison36: Excellent point. I've removed these

contracts/contracts/strategies/sonic/SonicSwapXAMOStrategy.sol
```diff
@@ -0,0 +1,707 @@
...
+        );
+        _;
+    }
+
+    /**
+     * @dev Skim the SwapX pool in case any extra wS or OS tokens were added
+     */
+    modifier skimPool() {
+        IPair(pool).skim(address(this));
+        _;
```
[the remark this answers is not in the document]
naddison36: The idea is to prevent donation attacks against the pool from messing with the strategy's calculations. That is, someone front-runs a strategy's operation against the pool so there is more tokens in the pool than the strategy was expecting. The strategy is using the reserve values rather than pool balances.

Most of the time others donating to the pool will end up benefiting the strategy but there may be some cases where it throws out the strategy. At the very least, they can donate to force reverts.

There should only be dust amounts left over from deposit / withdraw but we could also skim those. Gas is cheap on Sonic.

contracts/contracts/strategies/sonic/SonicSwapXAMOStrategy.sol
```diff
@@ -0,0 +1,707 @@
...
+        // Add wS and OS liquidity to the pool and stake in gauge
+        _depositToPoolAndGauge(_wsAmount, osLiquidity);
+
+        // Ensure solvency of the vault
+        _solvencyAssert();
+
+        // Emit event for the deposited wS tokens
+        emit Deposit(_wS, pool, _wsAmount);
+        // Emit event for the minted OS tokens
+        emit Deposit(os, pool, osLiquidity);
```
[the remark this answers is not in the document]
naddison36: We have emitted `Deposit` and `Withdrawal` events for the OTokens in the Mainnet and Base Curve AMOs. It helps the analytics track the OTokens being minted and burned.

contracts/contracts/strategies/sonic/SonicSwapXAMOStrategy.sol
```diff
@@ -0,0 +1,707 @@
...
+        );
+        IERC20(ws).safeTransfer(_recipient, _wsAmount);
+
+        // Ensure solvency of the vault
+        _solvencyAssert();
+
+        // Emit event for the withdrawn wS tokens
+        emit Withdrawal(_asset, pool, _wsAmount);
+        // Emit event for the burnt OS tokens
+        emit Withdrawal(os, pool, osToBurn);
```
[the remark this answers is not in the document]
naddison36: This is the approach taken for other AMOs

contracts/contracts/strategies/sonic/SonicSwapXAMOStrategy.sol
```diff
@@ -0,0 +1,707 @@
...
+        // Get the strategy contract's wS balance.
+        // This includes all that was removed from the SwapX pool and
+        // any that was sitting in the strategy contract before the removal.
+        uint256 wsBalance = IERC20(ws).balanceOf(address(this));
+        IERC20(ws).safeTransfer(vaultAddress, wsBalance);
+
+        // Emit event for the withdrawn wS tokens
+        emit Withdrawal(ws, pool, wsBalance);
+        // Emit event for the burnt OS tokens
+        emit Withdrawal(os, pool, osToBurn);
```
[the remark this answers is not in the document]
naddison36: The same is done in the rebalancing functions of the OETH Curve AMO. eg `mintAndAddOTokens` emits a `Deposit` event for minted OETH and emits a `Withdrawal` event in `removeAndBurnOTokens` for burnt OETH.

contracts/contracts/strategies/sonic/SonicSwapXAMOStrategy.sol
```diff
@@ -0,0 +1,707 @@
...
+
+        // Work out the correct order of the amounts for the pool
+        (uint256 amount0, uint256 amount1) = _tokenIn == ws
+            ? (uint256(0), amountOut)
+            : (amountOut, 0);
+
+        // Perform the swap on the pool
+        IPair(pool).swap(amount0, amount1, address(this), new bytes(0));
+
+        // The slippage protection against the amount out is indirectly done
```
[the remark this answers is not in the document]
naddison36: The design assumes the same as the Curve AMOs in it rebalancing should always move closer to balanced.

naddison36:
Developer internal security checklist

## Requirements

This PR adds an AMO strategy to the Origin Vault on Sonic (OS) that integrates to SwapX's stable pool.

It also deploys a Harvester contract to collect SWPx rewards and transfer them to the Guardian multi-sig.

## Easy Checks

#### Authentication

- [x] Never use tx.origin

#### Ethereum

#### Cryptographic code

#### Gas problems

    - [ ] A way to remove items

#### Black magic

#### Overflow

#### Proxy

#### Events

## Medium Checks

#### Rounding and casts

#### Dependencies

#### External calls

- [x] No malicious behaviors
- [x] Oracles, one of:
  - [x] No oracles
  - [ ] Oracles can't be bent
  - `getReserves` is used on the pool rather than getting the balance of the pool

#### Tests

#### Deploy

## Strategy Specific

#### Strategy checks

  - The Harvester transfers SWPx rewards to the Guardian 
  - There can be a dust amount of wS left in the contract

#### Downstream

## Thinking

#### Logic

#### Internal State

Does this code do that?

#### Attack

#### Flavor

_Could this code be simpler?_

contracts/contracts/strategies/sonic/SonicSwapXAMOStrategy.sol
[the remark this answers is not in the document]
sparrowDom: Skim returns the token0 & token1 to the caller, and not the LP token. Meaning that this applies to `deposit`, `depositAll` and the 2 swap transactions. Which one of these do you think an attacker could manipulate by donating tokens to?

contracts/contracts/strategies/sonic/SonicSwapXAMOStrategy.sol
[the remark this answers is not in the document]
sparrowDom: I see, my mistake... thanks

contracts/contracts/strategies/sonic/SonicSwapXAMOStrategy.sol
```diff
@@ -0,0 +1,710 @@
...
+    }
+
+    /***************************************
+                    Withdraw
+    ****************************************/
+
+    /**
+     * @notice Withdraw wS and OS from the SwapX pool, burn the OS,
+     * and transfer the wS to the recipient.
+     * @param _recipient Address to receive withdrawn asset which is normally the Vault.
```
sparrowDom, line 246: nit: the vault should be the only valid recipient

contracts/test/strategies/sonic/swapx-amo.sonic.fork-test.js
```diff
@@ -0,0 +1,1234 @@
...
+        await swapXGauge.balanceOf(swapXAMOStrategy.address)
+      ).to.withinRange(
+        expectedStratGaugeBalance.sub(1),
+        expectedStratGaugeBalance.add(1),
+        "Strategy's gauge balance"
+      );
+    }
+  };
+
+  async function assertDeposit(wsDepositAmount) {
```
sparrowDom, line 924: many of these asserts could also test that after the operation the LP token balance of the strategy contract equals 0

contracts/contracts/strategies/sonic/SonicSwapXAMOStrategy.sol
```diff
@@ -0,0 +1,710 @@
...
+     * @param _wsAmount Amount of Wrapped S (wS) to withdraw.
+     */
+    function withdraw(
+        address _recipient,
+        address _asset,
+        uint256 _wsAmount
+    ) external override onlyVault nonReentrant skimPool {
+        require(_wsAmount > 0, "Must withdraw something");
+        require(_asset == ws, "Unsupported asset");
+
```
sparrowDom, line 257: We should add a require here: 
```
require(_recipient == vaultAddress, "Only withdraw to vault allowed");
```

contracts/contracts/strategies/sonic/SonicSwapXAMOStrategy.sol
```diff
@@ -0,0 +1,710 @@
...
+     * @param _asset Address of the Wrapped S (wS) contract.
+     * @param _wsAmount Amount of Wrapped S (wS) to withdraw.
+     */
+    function withdraw(
+        address _recipient,
+        address _asset,
+        uint256 _wsAmount
+    ) external override onlyVault nonReentrant skimPool {
+        require(_wsAmount > 0, "Must withdraw something");
+        require(_asset == ws, "Unsupported asset");
```
sparrowDom, line 256: These 2 requires could have unit tests

contracts/contracts/strategies/sonic/SonicSwapXAMOStrategy.sol
```diff
@@ -0,0 +1,710 @@
...
+         * Important: A downside is that the Strategist / Governor needs to be
+         * cognizant of not removing too much liquidity. And while the proposal to remove liquidity
+         * is being voted on, the pool tilt might change so much that the proposal that has been valid while
+         * created is no longer valid.
+         */
+
+        (uint256 wsReserves, , ) = IPair(pool).getReserves();
+        require(wsReserves > 0, "Empty pool");
+
+        lpTokens = (_wsAmount * IPair(pool).totalSupply()) / wsReserves;
```
sparrowDom, line 527: To express the rounding up, and not round up you could also do: 
```
lpTokens = (_wsAmount * IPair(pool).totalSupply() + 1e18 - 1) / wsReserves;
```
Pro: doesn't round up when the division is perfect  / whole numbered 
Con: affects code readability negatively

contracts/contracts/strategies/sonic/SonicSwapXAMOStrategy.sol
```diff
@@ -0,0 +1,710 @@
...
+     * @param _wsAmount Amount of Wrapped S (wS) to be removed from the pool.
+     * @return lpTokens Amount of SwapX pool LP tokens to burn.
+     */
+    function _calcTokensToBurn(uint256 _wsAmount)
+        internal
+        view
+        returns (uint256 lpTokens)
+    {
+        /* The rate between coins in the pool determines the rate at which pool returns
+         * tokens when doing balanced removal (remove_liquidity call). And by knowing how much wS
```
sparrowDom, line 512: nit: `burn` is the call instead of `remove_liquidity`

contracts/contracts/strategies/sonic/SonicSwapXAMOStrategy.sol
```diff
@@ -0,0 +1,710 @@
...
+         * we want we can determine how much of OS we receive by removing liquidity.
```
sparrowDom, line 513: not sure should this comment say: 
"... we want we can determine how much LP we need to burn in order to receive the desired amount of wS
""

contracts/contracts/strategies/sonic/SonicSwapXAMOStrategy.sol
```diff
@@ -0,0 +1,710 @@
...
+            IERC20(ws).balanceOf(address(this)) >= _wsAmount,
+            "Not enough wS removed from pool"
+        );
+        IERC20(ws).safeTransfer(_recipient, _wsAmount);
+
+        // Ensure solvency of the vault
+        _solvencyAssert();
+
+        // Emit event for the withdrawn wS tokens
+        emit Withdrawal(_asset, pool, _wsAmount);
```
sparrowDom, line 281: super nit: could be `_ws` instead of `_asset` for tiny code readability improvement

contracts/contracts/strategies/sonic/SonicSwapXAMOStrategy.sol
```diff
@@ -0,0 +1,710 @@
...
+                Pool Rebalancing
+    ****************************************/
+
+    /** @notice Used when there is more OS than wS in the pool.
+     * wS and OS is removed from the pool, the received wS is swapped for OS
+     * and the left over OS in the strategy is burnt.
+     * The OS/wS price is < 1.0 so OS is being bought at a discount.
+     * @param _wsAmount Amount of Wrapped S (wS) to swap into the pool.
+     */
+    function swapAssetsToPool(uint256 _wsAmount)
```
sparrowDom, line 335: the name could also be `swapWsToPool`

contracts/test/strategies/sonic/swapx-amo.sonic.fork-test.js
sparrowDom: and they could also assert that the strategy has a 0 OS balance after any operation

contracts/contracts/strategies/sonic/SonicSwapXAMOStrategy.sol
```diff
@@ -0,0 +1,710 @@
...
+        // Mint the required OS tokens to this strategy
+        IVault(vaultAddress).mintForStrategy(osToMint);
+
+        // 2. Swap OS for wS against the pool
+        _swapExactTokensForTokens(_osAmount, os, ws);
+
+        // 3. Add wS and OS back to the pool in proportion to the pool's reserves
+
+        // The wS is from the swap and any wS that was sitting in the strategy
+        uint256 wsDepositAmount = IERC20(ws).balanceOf(address(this));
```
sparrowDom, line 399: The code below could be replaced by calling `_deposit(wsDepositAmount)` though a new parameter would need to be added so the `_deposit()` doesn't emit events in this case

sparrowDom (commented):
Another round of comments for the PR. Overall the code is in a really good condition 💪

contracts/contracts/strategies/sonic/SonicSwapXAMOStrategy.sol
naddison36: I tried to keep the function name generic in case we reused this strategy for other assets or chains. 
The same was done for the Curve AMO which has `mintAndAddOTokens`, `removeAndBurnOTokens` and `removeOnlyAssets` rebalance functions. This worked out when the Sonic Curve AMO was done.

contracts/contracts/strategies/sonic/SonicSwapXAMOStrategy.sol
```diff
@@ -0,0 +1,710 @@
...
+
+        // This is an implementation contract. The governor is set in the proxy contract.
+        _setGovernor(address(0));
+    }
+
+    /**
+     * Initializer for setting up strategy internal state. This overrides the
+     * InitializableAbstractStrategy initializer as SwapX strategies don't fit
+     * well within that abstraction.
+     * @param _rewardTokenAddresses Address of SWPx token
```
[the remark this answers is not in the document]
naddison36: Fixed. Thanks

contracts/contracts/strategies/sonic/SonicSwapXAMOStrategy.sol
```diff
@@ -0,0 +1,710 @@
...
+        override
+        onlyVault
+        nonReentrant
+        skimPool
+    {
+        _deposit(_asset, _amount);
+    }
+
+    function _deposit(address _wS, uint256 _wsAmount) internal {
+        require(_wsAmount > 0, "Must deposit something");
```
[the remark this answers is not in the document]
naddison36: Good pickup. I've added a test

contracts/contracts/strategies/sonic/SonicSwapXAMOStrategy.sol
```diff
@@ -0,0 +1,710 @@
...
+        require(_wS == ws, "Unsupported asset");
```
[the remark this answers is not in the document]
naddison36: Good pickup. I've added a test

contracts/test/strategies/sonic/swapx-amo.sonic.fork-test.js
```diff
@@ -0,0 +1,1234 @@
...
+        .connect(oSonicVaultSigner)
+        .withdraw(oSonicVault.address, wS.address, parseUnits("10"));
+
+      await expect(tx).to.be.revertedWith("Protocol insolvent");
+    });
+    it("Should withdraw all", async () => {
+      const { oSonicVaultSigner, swapXAMOStrategy } = fixture;
+
+      // Vault withdraw alls from the strategy
+      const tx = swapXAMOStrategy.connect(oSonicVaultSigner).withdrawAll();
```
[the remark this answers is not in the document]
naddison36: There's a `not` in the assertion 
`await expect(tx).to.not.revertedWith("Protocol insolvent");`

contracts/contracts/strategies/sonic/SonicSwapXAMOStrategy.sol
naddison36: I agree. Fixed

contracts/contracts/strategies/sonic/SonicSwapXAMOStrategy.sol
naddison36: This was copied from the Curve AMO. I've corrected. Thanks

contracts/contracts/strategies/sonic/SonicSwapXAMOStrategy.sol
naddison36: we want to round up to ensure we get enough LP tokens out of the gauge and then wS tokens out off the pool.

I originally had this code but testing showed cases where not enough LP tokens were used to get the required amount of wS tokens.
```
lpTokens =
            ((_wsAmount + 1) * IPair(pool).totalSupply()) /
            ws.balanceOf(pool);
```

contracts/contracts/strategies/sonic/SonicSwapXAMOStrategy.sol
naddison36: This was copied from the Curve AMO which is a little different in Curve supports unbalanced removals.

I've changed it to this
```
  /* The SwapX pool proportionally returns the reserve tokens when removing liquidity.
   * First, calculate the proportion of required wS tokens against the pools wS reserves.
   * That same proportion is used to calculate the required amount of pool LP tokens.
   * For example, if the required wS tokens is 10% of the pool's wS reserves,
   * then 10% of the pool's LP supply needs to be burned.
```

contracts/contracts/strategies/sonic/SonicSwapXAMOStrategy.sol
naddison36: VaultCore's `redeem` code does allow withdrawing directly to the redeemer. But that's not relevant for Sonic as `redeem` has been disabled.

```Solidity
// Nothing in Vault, but something in Strategy, send from there
IStrategy strategy = IStrategy(strategyAddr);
strategy.withdraw(msg.sender, assetAddr, outputs[i]);
```

I'm inclined to keep the behaviour between the strategies consistent and allow the Vault to withdraw to a recipient.
I can be convinced the other way, though.

contracts/contracts/strategies/sonic/SonicSwapXAMOStrategy.sol
naddison36: Fixed

contracts/contracts/strategies/sonic/SonicSwapXAMOStrategy.sol
naddison36: Good idea. I had to move the `Deposit` events out of the internal `_deposit` function so it makes the external `deposit` and `depositAll` a little messier. But I think it's worth it as it moves an extra four lines of code into an exiting common function.

contracts/test/strategies/sonic/swapx-amo.sonic.fork-test.js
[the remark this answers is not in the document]
sparrowDom: Ah my mistake I'm sorry

contracts/contracts/strategies/sonic/SonicSwapXAMOStrategy.sol
sparrowDom: I guess it comes down to, can/will this ever be an asset default strategy. With AerodromAMO we knew that it won't be, so we added [the require check](https://github.com/OriginProtocol/origin-dollar/blob/master/contracts/contracts/strategies/aerodrome/AerodromeAMOStrategy.sol#L889). 

I think any strategy operating on a pool will probably never be an asset default one, and for that reason we could add a require here. Again since only vault calls this code it is a defensive programming measure that is of low importance. The require would basically only prevent the governance mistakingly setting the SwapxAMO as a default strategy. I will leave the decision up to you.

contracts/contracts/strategies/sonic/SonicSwapXAMOStrategy.sol
sparrowDom: makes sense I agree the same function names across multiple strategies is better.

contracts/contracts/strategies/sonic/SonicSwapXAMOStrategy.sol
naddison36: I'm happy to do the same as the Aerodrome AMO. I like being consistent :-)

Given deposits and withdrawals can be front-run, we definitely don't want the AMO strategy set as the default strategy in the vault.

I'll add the require

contracts/contracts/vault/OETHVaultAdmin.sol
```diff
@@ -19,6 +19,48 @@ contract OETHVaultAdmin is VaultAdmin {
...
+        isMintWhitelistedStrategy[strategyAddr] = true;
+
+        emit StrategyAddedToMintWhitelist(strategyAddr);
+    }
+
+    /**
+     * @notice Removes a strategy from the mint whitelist.
+     * @param strategyAddr Strategy address
+     */
+    function removeStrategyFromMintWhitelist(address strategyAddr)
```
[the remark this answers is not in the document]
naddison36: I've added a unit tests

contracts/contracts/strategies/sonic/SonicSwapXAMOStrategy.sol
```diff
@@ -0,0 +1,710 @@
...
+            _deposit(ws, balance);
+        }
+    }
+
+    /***************************************
+                    Withdraw
+    ****************************************/
+
+    /**
+     * @notice Withdraw wS and OS from the SwapX pool, burn the OS,
```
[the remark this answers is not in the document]
naddison36: I've updated the deposit Natspec

contracts/contracts/strategies/sonic/SonicSwapXAMOStrategy.sol
naddison36: I've updated the Natspec

contracts/test/strategies/sonic/swapx-amo.sonic.fork-test.js
naddison36: That's a good suggestion. I've added this to the deposit, withdraw and swap assert functions.

contracts/deploy/base/027_base_curve_amo_upgrade.js
```diff
@@ -0,0 +1,69 @@
...
+    const cOETHbVault = await ethers.getContractAt(
+      "IVault",
+      cOETHbVaultProxy.address
+    );
+    const cOETHBaseCurveAMOProxy = await ethers.getContract(
+      "OETHBaseCurveAMOProxy"
+    );
+
+    // Deploy Base Curve AMO implementation
+    const dOETHBaseCurveAMOImpl = await deployWithConfirmation(
```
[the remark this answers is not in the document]
naddison36: This has already been deployed

contracts/contracts/strategies/sonic/SonicSwapXAMOStrategy.sol
```diff
@@ -0,0 +1,738 @@
...
+        _swapExactTokensForTokens(_osAmount, os, ws);
+
+        // The wS is from the swap and any wS that was sitting in the strategy
+        uint256 wsDepositAmount = IERC20(ws).balanceOf(address(this));
+
+        // 3. Add wS and OS back to the pool in proportion to the pool's reserves
+        (uint256 osDepositAmount, uint256 lpTokens) = _deposit(wsDepositAmount);
+
+        // Emit event for the minted OS tokens
+        emit Deposit(os, pool, osToMint + osDepositAmount);
```
sparrowDom, line 439: I am not sure adding these 2 amounts together are correct. The `osToMint` was part of the swap transaction and there to swap OS for wS. Once we have the wS the corresponding OS part is minted and deposited. 

Hmm I think I see the thought process, so instead of emitting 2 deposit events 1 for wS amount and 1 for OS amount, we emit only 1 with OS amounts combined.

sparrowDom: I guess this way of doing it is better, since even though we deposit ws&OS into the pool, no WS assets have really moved from the Vault. Only OS has been minted, swapped and deposited. I think the way you did it is more correct.

Maybe a comment to explain the thought process?

naddison36: The thought process was to not emit three Deposit events. Instead, just emit one Deposit event for wS and the other for OS. But as we mint OS twice, the amounts are added together

contracts/deploy/sonic/009_swapx_amo.js
```diff
@@ -0,0 +1,145 @@
...
+    const dSonicSwapXAMOStrategyProxy = await deployWithConfirmation(
+      "SonicSwapXAMOStrategyProxy",
+      []
+    );
+    const cSonicSwapXAMOStrategyProxy = await ethers.getContract(
+      "SonicSwapXAMOStrategyProxy"
+    );
+
+    // Deploy Sonic SwapX AMO Strategy implementation
+    const dSonicSwapXAMOStrategy = await deployWithConfirmation(
```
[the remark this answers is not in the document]
naddison36: I've added a `deploySonicSwapXAMOStrategyImplementation` deploy action

contracts/contracts/strategies/sonic/SonicSwapXAMOStrategy.sol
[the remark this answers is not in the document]
naddison36: Deposits and withdraws are more profitable when the pool is balanced so I think it's ok to bring OS to 1:1 peg and then let the market pull it back to a fair discount given the withdrawal times.

sparrowDom:
I've added [a commit](https://github.com/OriginProtocol/origin-dollar/pull/2439/commits/2d62633bca6c964559c3a9b892fdcbdaf506cd18) where a brownie script dpeloys ~5m tokens into the AMO pool and then lets the potential attacker tilt the pool into both directions: OS heavy and WS heavy. `checkBalanace(ws)` is called on the strategy before the tilt and after the tilt. The differences in values are displayed in the below chart: 
Screenshot 2025-03-14 at 10 30 04

The amount deposited in the strategy (and returned by `checkBalance`) is roughly `5 * 1e24`. And the differences before and after the tilt are sub 10 WEI.

Which proves that `checkBalance` is resistant to pool manipulation ✅

sparrowDom:
## Requirements

This PR creates an AMO strategy for the OS/WS SwapX pool and deploys the harvester to harvest the rewards. PR also upgrades the Base Curve AMO to enable `withdrawAll()` when strategy is empty.

## Easy Checks

#### Authentication

- [X] Never use tx.origin

#### Ethereum

#### Cryptographic code

#### Gas problems

_no loops_

#### Black magic

#### Overflow

#### Proxy

#### Events

## Medium Checks

#### Rounding and casts
- 🟡  Contract does not have bugs from loosing rounding precision
  - it might have errors especially if we were to operate with tiny numbers. Normal users won't ever interact directly with this contract so it doesn't need to be fuzz tested with various edge case amounts. 
  - doesn't in `_invariant` function, though the code is copied from the SwapX pool, and it is more important that it can mimic the results SwapX pool produces
- 🟡 Contract does not have bugs from zero or near zero amounts
  - it might have, though we won't test for it, since this won't be an asset default (user interact-able) strategy

#### Dependencies

#### External calls

- [x] No malicious behaviors
- [ ] ~~Low level call() must require success.~~ no such calls
  - we deposit / withdraw in a balanced manner and using Vault value checker prevents any slippage attacks
  - when swapping only the changes towards more balanced pool are allowed, which prevents slippage attacks
- [x] Oracles, one of:
  - [x] No oracles
  - [ ] Oracles can't be bent
- 🟡  Do not call balanceOf for external contracts to determine what they will do when they use internal accounting
  - we do need to call `balanceOf` on tokens and the gauge

#### Tests

  - 🟣  Each require() has a test & Each logical branch has a test
    - "No LP tokens to burn" in `swapAssetsToPool` Maybe if `_wsAmount` is 1 WEI this will get triggered
    - "Too much OS in strategy" in `swapOTokensToPool`. Could be triggered in a test by donating OS to the strategy
    - "Unsupported asset" in `checkBalance`
    - "Position balance is worsened" in `improvePoolBalance` (though this one might be tricky to setup - lets skip it)
    - "Not enough wS removed from pool" in `withdraw` (I would skip this one as well)
    - "Empty pool" `_calcTokensToMint` (I'd skip this test)
    - "Unsupported swap" in `_swapExactTokensForTokens`. Though I think that code is unreachable
  - 🟡  If tests interact with AMM make sure enough edge cases (pool tilts) are tested. Ideally with fuzzing.
    - we don't have fuzzing, but there are some fork tests tilting the pool.

#### Deploy

## Strategy Specific
#### Strategy checks

   - this has been confirmed with brownie simulations posted one message above
  - there can be dust amounts of WS due to rounding up when removing liquidity
  - not when using Vault Value checker

#### Downstream

## Thinking

#### Logic

Logic looks ok. 

Pool and gauge are already setup. Nothing further needs to happen

#### Internal State

The strategy holds various configuration parameters in its storage state. There aren't any states that change within the strategy contract during the operation that would be important.

#### Attack

`Deposit`, `withdrawal` and `withdrawAll` can all be MEV-ed. That is why it is important that we maintain the use of VaultValueChecked when depositing / withdrawing to/from the strategy. 

#### Flavor

Code is really clean.

sparrowDom (dismissed):

sparrowDom:
I've altered the simulation to more drastically tilt the pool when comparing the `checkBalance` values. [Here is](https://gist.github.com/sparrowDom/f5f0a6b619c447f28c345e8e548968da) the raw data which includes values like pool token balances before and after the pool tilt. The differences with `checkBalance` values become more drastic with bigger tilts: 
checkBalance

contracts/contracts/strategies/sonic/SonicSwapXAMOStrategy.sol
```diff
@@ -0,0 +1,738 @@
...
+    {
+        require(_asset == ws, "Unsupported asset");
+        require(_wsAmount > 0, "Must deposit something");
+
+        (uint256 osDepositAmount, ) = _deposit(_wsAmount);
+
+        // Emit event for the deposited wS tokens
+        emit Deposit(ws, pool, _wsAmount);
+        // Emit event for the minted OS tokens
+        emit Deposit(os, pool, osDepositAmount);
```
DanielVF, line 217: The deposit and withdraw events are from the vaults perspective. What user funds is the vault investing and withdrawing. We don't emit events in our AMO's for minted protocol owned liquidity.

(Applies to all deposit/withdraw events in this contract)

contracts/contracts/strategies/sonic/SonicSwapXAMOStrategy.sol
DanielVF, line 439: In our current contract code and analytics, swaps and rebalances don't emit events.

contracts/contracts/strategies/sonic/SonicSwapXAMOStrategy.sol
```diff
@@ -0,0 +1,738 @@
...
+
+        // 2. Swap OS for wS against the pool
+        _swapExactTokensForTokens(_osAmount, os, ws);
+
+        // The wS is from the swap and any wS that was sitting in the strategy
+        uint256 wsDepositAmount = IERC20(ws).balanceOf(address(this));
+
+        // 3. Add wS and OS back to the pool in proportion to the pool's reserves
+        (uint256 osDepositAmount, uint256 lpTokens) = _deposit(wsDepositAmount);
+
```
DanielVF, line 437: Missing solvency check here.

contracts/contracts/strategies/sonic/SonicSwapXAMOStrategy.sol
```diff
@@ -0,0 +1,738 @@
...
+    function _calcTokensToMint(uint256 _wsAmount)
+        internal
+        view
+        returns (uint256 osAmount)
+    {
+        (uint256 wsReserves, uint256 osReserves, ) = IPair(pool).getReserves();
+        require(wsReserves > 0, "Empty pool");
+
+        // OS to add = (wS being added * OS in pool) / wS in pool
+        osAmount = (_wsAmount * osReserves) / wsReserves;
```
DanielVF, line 523: 🟥 🟥 rekt, rekt.

Attacker front runs a deposit transaction by depositing a huge amount on either side. We deposit matching. When the pool is imbalanced, you get less LP per funds on deposit. (You get more on a withdraw). We lose money, get back less LP tokens. Attacker rebalances pool, has more LP tokens (since they profited from us getting less).

contracts/contracts/strategies/sonic/SonicSwapXAMOStrategy.sol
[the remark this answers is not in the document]
DanielVF: image

If we skim before or after, we don't lose our funds depositing. The deposit works off reserve ratios, we work off reserves.

If we skim before or after, we get any extra funds on the contract that someone else left.

However, only if we skim after do we get any of our own funds back from us not calculating correctly on a deposit. So it seems like we should skim after.

DanielVF:
@sparrowDom How can an attacker MEV attack withdraw and WithdrawAll()?

contracts/contracts/strategies/sonic/SonicSwapXAMOStrategy.sol
naddison36: Yes, deposits can be front-run which is why they need to be wrapped with the vault value checker.

contracts/contracts/strategies/sonic/SonicSwapXAMOStrategy.sol
naddison36: The solvency check is in the internal `_deposit` function

contracts/contracts/strategies/sonic/SonicSwapXAMOStrategy.sol
naddison36: The Curve AMO does emit Deposit/Withdraw events for the OToken

contracts/contracts/strategies/sonic/SonicSwapXAMOStrategy.sol
naddison36: The Curve AMO emits Deposit and Withdraw events for protocol owned liquidity.

Here's the `Deposit` event for OETH in the OETH Curve AMO https://github.com/OriginProtocol/origin-dollar/blob/b4544dce7751298067b18084ac326142d82aef13/contracts/contracts/strategies/ConvexEthMetaStrategy.sol#L212

The `Deposit` event is also emitted in the `mintAndAddOTokens` rebalance function https://github.com/OriginProtocol/origin-dollar/blob/b4544dce7751298067b18084ac326142d82aef13/contracts/contracts/strategies/ConvexEthMetaStrategy.sol#L414

contracts/contracts/strategies/sonic/SonicSwapXAMOStrategy.sol
```diff
@@ -0,0 +1,738 @@
...
+     * To minimize loses, the pool should be rebalanced before depositing.
+     * @param _asset Address of Wrapped S (wS) token.
+     * @param _wsAmount Amount of Wrapped S (wS) tokens to deposit.
+     */
+    function deposit(address _asset, uint256 _wsAmount)
+        external
+        override
+        onlyVault
+        nonReentrant
+        skimPool
```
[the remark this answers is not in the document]
naddison36: I thought it was confusing having modifiers at different levels which is why I kept them all at the external function level.

`skimPool` can be moved from `deposit` and `depositAll` to `_deposit`.

`skimPool` needs to remain at the `withdraw` level as the pool reserves are read before the internal `_withdrawFromGaugeAndPool` is called.
`withdrawAll` doesn't read the reserves before `_withdrawFromGaugeAndPool`.
`swapAssetsToPool` need the `skimPool` modifier as the reserves are read before `_swapExactTokensForTokens` is called.
`swapOTokensToPool` is different in it doesn't read reserves before `_swapExactTokensForTokens`.

naddison36:
> @sparrowDom How can an attacker MEV attack withdraw and WithdrawAll()?

@DanielVF I've added more fork tests and shown `withdraw` and `withdrawAll` can not be front-run attacked like deposit. The strategy profits if the pool is tilted away from being balanced in either direction before a withdraw.

I've updated the `withdraw` and `withdrawAll` Natspec

contracts/contracts/strategies/sonic/SonicSwapXAMOStrategy.sol
naddison36: I've created PR https://github.com/OriginProtocol/origin-dollar/pull/2464 to add front-running protect on deposits

sparrowDom:
@DanielVF yes you are right, my mistake. Any bends of the pool before a withdrawal should only yield profit

contracts/contracts/strategies/sonic/SonicSwapXAMOStrategy.sol
naddison36: I've moved `_solvencyAssert` out of the internal `_deposit` to the external `deposit`, `depositAll` and `swapOTokensToPool` functions so it's clearer.

sparrowDom:
I've done the deposit simulation tests. There are slight -1 to -3 WEI losses when depositing, depending on the pool tilt. Since the pool allows deposits only in the 0.5% deviation from the balanced state the more tilted deposits are reverted indicated by the 0 deposit profits in the graph. 
deposit

PR #2447, opened 2025-03-17
WOETH yield buckets

sparrowDom:
This is the third attempt at the implementation of limiting the yield increase and, consequen the exchange rate between WOETH and OETH. The other 2: 
- https://github.com/OriginProtocol/origin-dollar/pull/2106
- https://github.com/OriginProtocol/origin-dollar/pull/2421

This approach: 
- does the internal accounting of credits the same way https://github.com/OriginProtocol/origin-dollar/pull/2106 does it
- the WOETH exchange rate is in sync with OETH yield as long as it doesn't surpass the 91% yearly yield rate. In such case the yield is delayed and will be distributed over the next yield period. Which lasts 23 hours 
- similarly to https://github.com/OriginProtocol/origin-dollar/pull/2421 this PR requires interaction with the contract in order to increase the exchange rate / start new yield period by calling `increaseYieldLimit()`

- [ ]  Unit tests *if needed

Internal review:


contracts/contracts/token/WOETH.sol
```diff
@@ -66,6 +86,12 @@ contract WOETH is ERC4626, Governable, Initializable {
          * environment.
          */
         oethCreditsHighres = _getOETHCredits();
+        creditsPerTokenLimit = OETH(asset())
```
DanielVF, line 89: Rounding toward higher yield limits is the best choice here, otherwise the contract would stop allowing any yield at some point.

DanielVF, line 89: Can we just remove this entire section of the initializer and call `increaseYieldLimit()`. This would deduplicate the code.

contracts/contracts/token/WOETH.sol
```diff
@@ -6,6 +6,7 @@ import { ERC20 } from "@openzeppelin/contracts/token/ERC20/ERC20.sol";
 import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 import { IERC20Metadata } from "@openzeppelin/contracts/token/ERC20/extensions/IERC20Metadata.sol";
 import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
+import { SafeCast } from "@openzeppelin/contracts/utils/math/SafeCast.sol";
```
DanielVF, line 9: In `transferToken()` the line:


```
        require(asset_ != address(asset()), "Cannot collect OETH");
```

Should probably be "core asset" or just "asset" instead of "OETH", since this code will be used for many assets.

contracts/contracts/token/WOETH.sol
```diff
@@ -28,9 +29,28 @@ import { OETH } from "./OETH.sol";
...
+     * maximum yield.
+     *
+     * If OETH has a daily yield that is larger than 91% the effective yield will spill over
+     * to WOETH in the next yield increase cadence.
+     *
+     * If the daily yield equating to 91% YOY yield isn't surpassed then the OETH rebases
+     * are in sync with WOETH. Resulting in no economical gain from wrapping OETH to WOETH
+     * and back to try to game the yield distribution.
+     */
+    uint256 public constant MAX_YIELD_INCREASE = 25e14; // 0.25%
```
DanielVF, line 53: We are the blessed summer children who have never lived hyper-inflation.

But if the underlying currency is inflating faster than this limit allows, then our users will be in deep trouble. I think we just want to head off donations that damage lending platforms, rather than bet our user funds on stable economies.

So I'd run this limit something more like 3-5%.

contracts/contracts/token/WOETH.sol
```diff
@@ -188,5 +222,44 @@ contract WOETH is ERC4626, Governable, Initializable {
...
+        increaseYieldLimit();
+    }
+
+    /**
+     * @dev This function safeguards rapid increases in WOETH -> OETH exchange ratio.
+     * Such an increase can be a result of a donation attack to the Vault contract.
+     */
+    function increaseYieldLimit() public {
+        // not yet time to increase the limit
+        if (block.timestamp < cptLimitEndTime) {
```
DanielVF, line 243: I think there should be a way where we can cleanly allow pushing forward the end time with every action, rather than requiring one limit to end before the next begins. This would be much nicer user facing behavior. We do have to do some math to see how much of the partially completed phase to carry over.

contracts/contracts/token/WOETH.sol
sparrowDom: good point, fixed here: https://github.com/OriginProtocol/origin-dollar/pull/2447/commits/9da27214906e5ca46f2b41b18708607411013db5

contracts/contracts/token/WOETH.sol
sparrowDom: Agreed thanks: https://github.com/OriginProtocol/origin-dollar/pull/2447/commits/821a75689d924644a00edbbba5f54c24eb071b65

contracts/contracts/token/WOETH.sol
DanielVF: The current approach around limits is:
- very simple
- correctly handles the normal case use of the protocol
- correctly blocks donation attacks

But it is also feels like it has a lot weirdness when the limits are hit.
- 2x the limit is possible at the end of the period
- someone can get doubled yielded at the end of a period
- ratios can change mid block, separate from OETH ratios changing

contracts/contracts/token/WOETH.sol
sparrowDom: That is a very good point.

contracts/contracts/token/WOETH.sol
sparrowDom: "someone can get doubled yielded at the end of a period"

Are you thinking that a person would be holding OETH, and knowing WOETH needs time to catch up, they would wrap their OETH just after a fat rebase?

"ratios can change mid block, separate from OETH ratios changing"

Do you think this can have negative consequences?

DanielVF: So this is talking about behavior when the limits will be hit, and the time is past the end:

- User knows/causes a BIG rebase to be about to occur.
- They have/get a large OETH position.
- The call rebase(), collecting the OETH yield
- They then deposit into WEOTH, at a privileged exchange rate.
- The end of the deposit calls increaseYield. Limits are raised again
- User withdraws, collecting OETH at the new higher rate. No one else was able to do this.

DanielVF: I'm actually concerned that is a full on attack, not just a yield steal. Minting funds at the wrong rate could be bad. This is the classic problem with the obvious path on rates based solutions.

DanielVF: I think it's just a yield steal, since the someone can't mint at a lower rate than the existing users got in at.

DanielVF: >> "ratios can change mid block, separate from OETH ratios changing"

> Do you think this can have negative consequences?

Yes, it allows someone to selectively take actions between these two. Presumably yield stealing.

contracts/contracts/token/WOETH.sol
```diff
@@ -65,7 +85,7 @@ contract WOETH is ERC4626, Governable, Initializable {
          * WOETH has already seen transactions. But it is rather annoying in unit test
          * environment.
          */
-        oethCreditsHighres = _getOETHCredits();
```
DanielVF, line 68: Still need the `oethCreditsHighres = _getOETHCredits();`

contracts/contracts/token/WOETH.sol
DanielVF: Ouch!

Confirmed with tests that this is a flashloanable enabled yield steal in the current implementation.

🟥 But it's actually much worse. Stuff does get really out of wack when things are at over the rate limit, and interacting users can actually end up with less funds than they started with depending on how they interact. It gets deeply squirrelly.

contracts/contracts/token/WOETH.sol
```diff
@@ -188,5 +222,44 @@ contract WOETH is ERC4626, Governable, Initializable {
...
+        increaseYieldLimit();
+    }
+
+    function _transfer(
+        address sender,
+        address recipient,
+        uint256 amount
+    ) internal override {
+        super._transfer(sender, recipient, amount);
+        increaseYieldLimit();
```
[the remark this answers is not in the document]
sparrowDom: remove here: https://github.com/OriginProtocol/origin-dollar/pull/2447/commits/4c4a2cb938b23748f6ebace9be341bac79854770

contracts/contracts/token/WOETH.sol
sparrowDom: uff thanks nice catch: https://github.com/OriginProtocol/origin-dollar/pull/2447/commits/b68269df644f05a6329c1ee260fd334733e972a0

sparrowDom:
closing in favour of this one: https://github.com/OriginProtocol/origin-dollar/pull/2453

PR #2452, opened 2025-03-19
[Vault] Max rebase rate control

DanielVF:
## Project

Move all yield rate limiting to the vault. This removes the need for the dripper, ensures that all yield is rate controlled, blocks donation attacks against lending platforms, and will allow us more consistent yield auto controlled yield.

## Schedule

- [ ] Code + author audit (dvf) Wednesday, March 26
- [ ] Internal code review (sparrowDom), March 28th
- [ ] Internal code review (panda), March 28th
- [ ] Audit (OpenZepelin), March 28th

## Code Change Checklist

- [ ]  Unit tests *if needed

Internal review:


naddison36:
There's a lot of tests failing

contracts/contracts/vault/VaultCore.sol
```diff
@@ -385,32 +385,102 @@ contract VaultCore is VaultInitializer {
...
+        }
+
+        // Start with the full difference available
+        uint256 yield = vaultValue - supply;
+
+        // Cap via optional automatic duration smoothing
+        uint256 _dripDuration = dripDuration;
+        if (_dripDuration > 1) {
+            // If we are able to sustain an increased drip rate for a
+            // longer duration, then increase the target drip rate
```
naddison36, line 466: nit: I think this is comment is more accurate
```
// If we are able to sustain an increased drip rate for
// double the duration, then increase the target drip rate
```

naddison36 (approved):
I like the new approach. Well done

contracts/contracts/vault/VaultCore.sol
```diff
@@ -385,32 +385,102 @@ contract VaultCore is VaultInitializer {
...
-            emit YieldDistribution(_trusteeAddress, yield, fee);
+        uint256 fee = 0;
+        if (_trusteeAddress != address(0)) {
+            fee = (yield * trusteeFeeBps) / 1e4;
+        }
+        if (fee > 0) {
+            require(fee < yield, "Fee must not be greater than yield");
+            oUSD.mint(_trusteeAddress, fee);
         }
+        emit YieldDistribution(_trusteeAddress, yield, fee);
```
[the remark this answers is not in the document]
DanielVF: Yes. Much better out here.

contracts/contracts/vault/VaultCore.sol
```diff
@@ -385,32 +385,102 @@ contract VaultCore is VaultInitializer {
...
-            require(yield > fee, "Fee must not be greater than yield");
-            if (fee > 0) {
-                oUSD.mint(_trusteeAddress, fee);
-            }
-            emit YieldDistribution(_trusteeAddress, yield, fee);
+        uint256 fee = 0;
+        if (_trusteeAddress != address(0)) {
+            fee = (yield * trusteeFeeBps) / 1e4;
+        }
+        if (fee > 0) {
```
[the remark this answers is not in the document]
DanielVF: Yeah, guess I got too de-nesting happy. While the current code works as is, better to move it back in.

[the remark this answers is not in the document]
DanielVF: Fixed in 0423f5499e26ccca68348f10c1e0341b033401a7

contracts/contracts/vault/VaultCore.sol
```diff
@@ -385,32 +385,102 @@ contract VaultCore is VaultInitializer {
...
+            elapsed == 0 || // Yield only once per block.
+            rebasing == 0 || // No yield if there are no rebasing tokens to give it to.
+            supply > vaultValue || // No yield if we do not have yield to give.
+            block.timestamp >= type(uint64).max // No yield if we are too far in the future to calculate it correctly.
+        ) {
+            return (0, targetRate);
+        }
+
+        // Start with the full difference available
+        uint256 yield = vaultValue - supply;
```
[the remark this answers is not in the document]
DanielVF: 🙏 Fixed in 6dcb859aca68e2b888b86d9cde8ea39acec46ff6

contracts/contracts/vault/VaultCore.sol
DanielVF: Changed in 864fa57b881a7ada8acf370a9eddefcc464a9378.

There's actually a bit of range that this number could be. Just ended up on 1/2, which is slightly more capital than is needed, and looks clean. 1.7 actually works out decently, if total supply is constant.

contracts/contracts/vault/VaultCore.sol
```diff
@@ -385,32 +385,102 @@ contract VaultCore is VaultInitializer {
      * @return totalUnits Total balance of Vault in units
      */
     function _rebase() internal whenNotRebasePaused returns (uint256) {
-        uint256 ousdSupply = oUSD.totalSupply();
+        uint256 supply = oUSD.totalSupply();
         uint256 vaultValue = _totalValue();
-        if (ousdSupply == 0) {
+        uint256 nonRebasing = oUSD.nonRebasingSupply();
```
sparrowDom, line 390: Fetching `nonRebasing` can be moved to into `_nextYield` function as it is only required in there

contracts/contracts/vault/VaultCore.sol
```diff
@@ -385,32 +385,102 @@ contract VaultCore is VaultInitializer {
...
+        // If no supply yet, do not rebase
+        if (supply == 0) {
             return vaultValue;
         }
 
-        // Yield fee collection
+        // Calculate yield and new supply
+        uint256 yield = 0;
+        uint256 newTarget = 0;
```
sparrowDom, line 398: nit: maybe `newTargetRate` might describe this variable better

contracts/contracts/vault/VaultCore.sol
```diff
@@ -385,32 +385,102 @@ contract VaultCore is VaultInitializer {
...
+            // drip at the new target rate
+            yield = _min(yield, targetRate * elapsed);
+        }
+
+        // Cap per second. elapsed is not 1e18 denominated
+        yield = _min(yield, (rebasing * elapsed * rebasePerSecondMax) / 1e18);
+
+        // Cap at a hard max per rebase, to avoid long durations resulting in huge rebases
+        yield = _min(yield, (rebasing * MAX_REBASE) / 1e18);
+
```
sparrowDom, line 480: The targetRate returned here could already be of type uint64 with correct casting checks: 
```
uint64(_min(newTarget, type(uint64).max));
```

contracts/contracts/vault/VaultCore.sol
```diff
@@ -385,32 +385,102 @@ contract VaultCore is VaultInitializer {
...
         }
         return vaultValue;
     }
 
+    /**
+     * @notice Calculates the amount that would rebase at at next rebase.
+     * This is before any fees.
+     * @return yield amount of expected yield
+     */
+    function previewYield() external view returns (uint256 yield) {
```
sparrowDom, line 434: very cool and useful function

contracts/contracts/vault/VaultCore.sol
```diff
@@ -385,32 +385,102 @@ contract VaultCore is VaultInitializer {
...
+        uint256 _totalSupply = oUSD.totalSupply();
+        uint256 _nonRebasing = oUSD.nonRebasingSupply();
```
sparrowDom, line 436: nit: both these don't need variables and can be directly called in `_nextYield` function call below

contracts/contracts/vault/VaultCore.sol
```diff
@@ -385,32 +385,102 @@ contract VaultCore is VaultInitializer {
...
+        ) {
+            return (0, targetRate);
+        }
+
+        // Start with the full difference available
+        yield = vaultValue - supply;
+
+        // Cap via optional automatic duration smoothing
+        uint256 _dripDuration = dripDuration;
+        if (_dripDuration > 1) {
```
sparrowDom, line 464: really elegant dripper logic implementation

sparrowDom (commented):
Left a couple of comments. Awesome work, very clean coding

contracts/contracts/vault/VaultCore.sol
```diff
@@ -385,32 +385,102 @@ contract VaultCore is VaultInitializer {
...
         }
 
-        // Yield fee collection
+        // Calculate yield and new supply
+        uint256 yield = 0;
+        uint256 newTarget = 0;
+        (yield, newTarget) = _nextYield(supply, vaultValue, nonRebasing);
+        uint256 newSupply = supply + yield;
+        // Only rebase upwards and if we have enough backing funds
+        if (newSupply <= supply || newSupply > vaultValue) {
```
sparrowDom, line 402: The `newSupply > vaultValue` is probably a sanity check? Though from the current code I don't see how this condition could ever be hit

DanielVF: Yes. Just an extra layer of sanity check, since it means that if you do something really dumb or have a bug in _nextYield, then you shouldn't be able to rekt the protocol.

contracts/contracts/vault/VaultStorage.sol
```diff
@@ -247,8 +249,30 @@ contract VaultStorage is Initializable, Governable {
...
+
+    /// @notice max rebase percentage per second
+    ///   Can be used to set maximum yield of the protocol,
+    ///   spreading out yield over time
+    // slither-disable-next-line uninitialized-state
+    uint64 public rebasePerSecondMax;
+
+    /// @notice ema over past rebases, duration is dripDuration. Only an approximation.
+    // slither-disable-next-line uninitialized-state
+    uint64 public rebasePerSecondTarget;
```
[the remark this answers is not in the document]
sparrowDom: good point

contracts/contracts/vault/VaultCore.sol
DanielVF: Awesome! `f1f5d07eb86c4af92ed30d8aa15874b77456d6d8`

contracts/contracts/vault/VaultCore.sol
```diff
@@ -385,32 +385,102 @@ contract VaultCore is VaultInitializer {
...
 
         // Only ratchet OToken supply upwards
-        ousdSupply = oUSD.totalSupply(); // Final check should use latest value
-        if (vaultValue > ousdSupply) {
-            oUSD.changeSupply(vaultValue);
+        // Final check uses latest totalSupply
+        uint256 preSupply = oUSD.totalSupply();
+        if (newSupply > preSupply) {
+            oUSD.changeSupply(newSupply);
+            rebasePerSecondTarget = uint64(_min(newTarget, type(uint64).max));
```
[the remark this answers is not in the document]
DanielVF: We no longer have an EMA. Is this mentioned in the comments anywhere?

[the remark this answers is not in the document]
DanielVF: Found it in vault storage. Updating.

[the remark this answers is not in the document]
DanielVF: Updated in 4eb85fac0ca0cc8c244d6987b7a80ec29607cf85

contracts/contracts/vault/VaultCore.sol
```diff
@@ -385,32 +385,99 @@ contract VaultCore is VaultInitializer {
...
+        // If no supply yet, do not rebase
+        if (supply == 0) {
             return vaultValue;
         }
 
-        // Yield fee collection
+        // Calculate yield and new supply
+        uint256 yield = 0;
+        uint256 targetRate = 0;
+        (yield, targetRate) = _nextYield(supply, vaultValue);
```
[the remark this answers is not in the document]
DanielVF: Done.

DanielVF:
# Code Review - Vault Slow Drip

## Requirements

We have two goals: to stop donation attacks and to provide fine grain yield control. This PR takes every source of yield that could rebase out and allows it to be throttled and limited with fine-grained control. 

The primary objective is to stop donation attacks. This is done by hard camping the maximum amount of yield that can occur in a single rebase, as well as controlling the maximum amount of yield over time. This prevents exchange rate from donations from being big enough to exploit lending platforms. 

Secondly, we want to allow for a smooth distribution of yield under normal circumstances. This PR provides two options for this. First, a hard set max cap on the maximum amount of yield can be set. If this is lower than the amount of yield coming in, you essentially have a fixed rate dripper that distributes at a fixed APY. Alternatively, we can use an auto setting dripper that automatically picks rates when we set the frequency that we expect large income events to happen at. 

The yield control is under the control of the strategist. however the anti-donation limits are hardcoded into the contract. The Strategist cannot configure a rate that would make a problem for a lending platform.

The drip duration controlled code should make for fairly consistent yield as long as the next yield arrives by the schedule that it's expecting. If it doesn't, it will begin smoothly going down at the end of this period.

#### Authentication

- [x] Never use tx.origin

#### Ethereum

#### Cryptographic code

_No crypto code_

#### Gas problems

_No for loops_

#### Black magic

_No magic._

#### Overflow

#### Proxy

#### Events

## Medium Checks

#### Rounding and casts
	- [x] Each division rounds in the direction of the protocol giving out less yield. 
	- [x] The fee calculation division rounds against the protocol receiving fees in favor of the user receiving yield. That behavior is acceptable.
- [x] Contract does not have bugs from loosing rounding precision / bugs from zero or near zero amounts
	- [x] the rebasePerSecondMax has plenty of resolution before it rounds to zero
	- [x] dripduration is not used when set to one or zero. 
	- [x] We are paranoid about what happens when you mint to the OUSD tokens, and do not assume that totalSupply changes by that exact amount.
	- [x] 🟣 At one point, we do divide yield by dripDuration times two. However, this is effectively dividing by a  constant.
	- [x] Safecast is used in all admin functions. 
	- [x] 🟣 Last rebase is deliberately not safecasted, so that a bazillion years in the future, the code will continue to allow deposits and withdrawals. However it is checked so that there will be no more yield after that point is reached.
	- [x] 🟣 Storing rebasePerSecondTarget does not use safe cast but rather minimums. This is so that it does not revert, which would again block deposits and withdrawals. 

#### Dependencies

#### External calls

	- [x] No addresses passed in. 
- [x] Reentrancy guards on all user facing state changing functions
- [x] Oracles, one of:
  - [x] No oracles
  - [ ] Oracles can't be bent

#### Tests

🔴 We currently do not have unit tests for this new behavior. 

#### Deploy

#### Downstream

This will require an analytics change to use previewYield to determine the next yield rather than subtracting totalSupply from vaultValue. The off-chain team has already been notified of this coming change. 

## Thinking

#### Logic

In the event that we distributed a fee and yet decided not to do a final rebase because of the universe being very weird, we would not set lastRebase. However, if this were to happen, you would still not be able to get extra rebases since changeSupply would not be called. Our token code that we are now using exactly adjusts totalSupply according to mint amount, which means that condition should not ever trigger.

The check on `newSupply <= supply` is not actually a duplicate. It catches the case when there is no yield.  

- [x] A max per second drip rate must be set, or the contract will distribute no yield. This is set in the current deploy file.

#### Internal State

The two most important variants are:

- ✅ The vault value must never be less than than the new total supply. (Always solvent)
- ✅ The new total supply must never be less than the old total supply. (Always up only)

✅ If there was a drop in the vault value, it's possible that both of these may not be true at the same time. If one of these is not true, we should not rebase. 

✅ MAX_REBASE should be a safe value. 

✅ rebasePerSecondMax must be equal to or less than MAX_REBASE_PER_SECOND

✅ It would be nice if preview yield only went up between as long as the vault did not lose money. This is roughly true because the only time dependent values affect this are the vault value and the time elapsed, both of which should in theory only go up. Obviously not true 69 bazillion years in the future when yield is disabled. 

✅ Preview yield should return the same yield that would actually be used by a rebase that occurred at that moment. It appears to be roughly true. 

#### Attack

An attacker causing a bad rebase is one of the primary attack vectors against OETH. In such a case, the attacker would first mint a large number of tokens, cause a bad rebase, and then either redeem or sell these tokens on AMMs. 

Because of this, the rebase function is one of the most critical functions in our code. The rebase function depends on the vault value function being correct. In theory, this set of changes massively improves the security of OETH by adding another layer of defense. In the event an attacker fools a vault strategy into returning an extremely large balance, the amount of damage an attacker can cause is limited to the instantaneous rebase limits set by this code. 

The strategist could disable the drip duration smoothing as well as max out the rebasePerSecondMax. However, in theory, this will still be a slow enough change set that lending platforms should be safe, and hack damage mitigated to a few percent of funds.  


#### Flavor

Unfortunately, I am blinded by self-bias on this code. I think it's froody.  Others will be able to see flaws that I cannot 

A primary goal of this code has been that even if there are horrible errors in the yield calculation code that these errors cannot result in unsafe rebasing. The rebase code is written so that it constrains the next supply to being between the old new supply and the vault value, regardless of the calculations

In turn, the nextYield calculations start with the vault value minus supply difference, and then each successive operation can only reduce the amount of yield from there. As a result, the nextYield function can only return a yield value which is between the vault value and the supply. 

Actual rebasing and previewing yield work off the same shared calculations. Previews should be accurate.

contracts/contracts/vault/VaultStorage.sol
```diff
@@ -72,6 +72,13 @@ contract VaultStorage is Initializable, Governable {
     event WithdrawalClaimable(uint256 _claimable, uint256 _newClaimable);
     event WithdrawalClaimDelayUpdated(uint256 _newDelay);
 
+
+    // Since we are proxy, all state should be uninitalized.
+    // Since this storage contract does not have logic directly on it
+    // we should not be checking for to see if these variables can be constant.
+    // slither-disable-start uninitialized-state
+    // slither-disable-start constable-states
+
```
sparrowDom, line 81: Yes good point 👍

shahthepro (dismissed):
## Easy Checks

#### Authentication

- [x] Never use tx.origin

#### Ethereum

#### Cryptographic code

#### Gas problems

    - [ ] A way to remove items
    - [x] No loops

#### Black magic

#### Overflow

#### Proxy

#### Events
  - :yellow_circle: `rebasePerSecondTarget` is changed inside `_rebase` but doesn't emit any event. Tracking the changes in this could potentially be useful for monitoring.

## Medium Checks

#### Rounding and casts

#### Dependencies

#### External calls

- [x] No malicious behaviors
- [ ] Oracles, one of:
  - [x] No oracles
  - [ ] Oracles can't be bent

#### Tests

#### Deploy

## Thinking

#### Logic

  - [x] Correct usage of global & local variables

There's a minor thing I noted during my tests where in an edge case the `targetRate` can go from a positive value to 0 and back to its original value. However DVF has confirmed that that's how the code is expected to behave


Almost all parameters can be set the governance or 2/8 multisig. So it should be fine even if we miss to set the values. When the values aren't set, it just does a zero-yield rebase.


#### Attack

The newly added functions are all callable by the strategist. So it only needs two bad signers. However the contract has hardcoded thresholds for the rebase rate, so it should be fine.

#### Flavor

Code looks good for the amount of stuff it gets done. I like it.

:yellow_circle: We previously agreed not to use named return variables. However this code does uses those (in `previewYield` and `_nextYield`).


P.S. I have left out things that are already mentioned by other people to be not repetitive.

contracts/contracts/vault/README - Yield Limits.md
```diff
@@ -0,0 +1,52 @@
+# Vault Yield Limits
+
+The vault does not necessarily rebase out all the value that it contains to the token on every rebase. Instead, the vault controls the maximum amount it rebases at one time. 
+
+This limiting behavior protects against donation attacks against other protocols. Some lending platforms don't handle sudden value change well. 
+
+ Secondly, it allows us to smoothly distribute yield over time. One example is that if we do a big sale of airdropped reward tokens, it would be much better if that yield was spread out over a longer period than just one block. Because this rate is controlled by the strategist, it can be set to a high rate to only block donation attacks, but not smooth regular yield, to a middle rate to only smooth extraordinary yield over a some blocks, or a low rate to set the protocols APR to end users, and provide a predictable APY.
```
sparrowDom, line 7: should 
`only smooth extraordinary yield over a some blocks `
be 
`only smooth extraordinary yield over a sum blocks `
or 
`only smooth extraordinary yield over some blocks `

contracts/contracts/vault/VaultCore.sol
```diff
@@ -372,32 +372,99 @@ contract VaultCore is VaultInitializer {
...
+
+        // Cap via optional automatic duration smoothing
+        uint256 _dripDuration = dripDuration;
+        if (_dripDuration > 1) {
+            // If we are able to sustain an increased drip rate for
+            // double the duration, then increase the target drip rate
+            targetRate = _max(targetRate, yield / (_dripDuration * 2));
+            // If we cannot sustain the target rate any more,
+            // then rebase what we can, and reduce the target
+            targetRate = _min(targetRate, yield / _dripDuration);
```
sparrowDom, line 454: The way I understand how 2 lines of code. (_min & _max) above function:
- the rate at which the yield is paid out is limited between a larger rate (yield / _dripDuration) and a smaller rate (yield / _dripDuration * 2).
- the target rate will default to a smaller `yield / (_dripDuration * 2)` if there were no pre-existing yield rates stored in the `rebasePerSecondTarget`. Is this expected? - I would expect the that the yield would be by default distributed over a `_dripDuration` and not half that rate. 
- the `targetRate` is set by the previous rebase's rate (`rebasePerSecondTarget`). Target rate is probably used here so that pre-existing yield rate is respected when subsequent rebases are called. 

I find the comments confusing. The 
```
// If we are able to sustain an increased drip rate for
// double the duration, then increase the target drip rate
```
might be better as:
```
// If the previous rebase's yield rate is higher than half the rate
// of yield spread over default drip duration keep the pre-existing higher rate
```

sparrowDom (commented):
Left some comments.... not done with the review yet

sparrowDom:
## Requirements
This PR makes some useful additions to the Vault where vault is able to: 
- control the maximum rebase amount possible - to make the token safe for the landing markets preventing any donation attacks. 
  - the rebase is limited by the maximum one time rebase
  - and the rebase per second calculated from the seconds since last rebase
- implement the Dripper logic right into the Vault being able to drip yield at a constant rate if the limits allow for it
- even if strategist is compromised the rebase can not exceed 5% / day

## Easy Checks

#### Authentication

- [x] Never use tx.origin

#### Ethereum

#### Cryptographic code

#### Gas problems

- [x] no loops

#### Black magic

#### Overflow

#### Proxy

#### Events
  - ⚪ no event emitted when the `rebasePerSecondTarget` changes

## Medium Checks

#### Rounding and casts
  - ⚪ a really low `dripDuration` (say 2 seconds) would cause rounding precision. But the effects would be just a more aggressive rounding down in protocol's favour. 
  - ⚪ we don't when calling `yield / (_dripDuration * 2)` but that potentially just acts as slight loss of resolution that isn't impactful
  - there are a couple of places where the cast is direct. But both of those are safe

#### Dependencies

#### External calls

- [x] Reentrancy guards on all state changing functions.
- [x] No malicious behaviors
- [x] Oracles, one of:
  - [x] No oracles
  - [ ] Oracles can't be bent

#### Tests

#### Deploy

- [X] Deployer permissions are removed after deploy

#### Downstream

- [ ] no changes to downstream required. Ideally the code could emit the altered `rebasePerSecondTarget` in an event and that could be useful to monitoring

## Thinking

#### Logic

#### Internal State

- Protocol always need to rebase only upwards. This code keeps that behaviour
- Protocol shouldn't rebase more than the backing assets support making it insolvent. This is also ok
- The protocol shouldn't send so much fees to the trustee to cause insolvency. Check for this is also made

With this change the value accrual of the token is controlled using multiple mechanisms: maximum single rebase limit & per second limit. These are all controlled by the governor and also strategist. Even in case of the strategist going rouge there is still a hardcoded limit of 5% / day. Which should be a sufficient safeguard for the lending platforms

#### Attack

The attack could somehow do a bad rebase. E.g. manipulating some of the underlying strategies to report too big of a value and do a positive rebase. This code imposes extra limits on such actions limiting the profitability of similar type of attacks

#### Flavor

The code is very clean and readable. It also touches one of the "most core" parts of our protocol, requiring in-depth knowledge of the protocol and its functions. Just reading the code without good understanding of the protocol might leave the reader confused. 
There are Readme-s that help to understand some of the principles. If we wanted to make our code better understandable, we'd probably need to invest more time into documentation. I am not sure that is high on the priority list.

sparrowDom (approved):

PR #2453, opened 2025-03-19
Simplified donation attack prevention

sparrowDom:
We are planning to prevent big yield jumps by modifying code directly in VaultCore in this PR:  https://github.com/OriginProtocol/origin-dollar/pull/2452 
This allows for simplification in WOETH to completely rely on rebasingCreditsPerToken to deduce its own exchange rate. 

## Code Change Checklist

- [ ]  Unit tests *if needed

Internal review:


contracts/test/helpers.js
```diff
@@ -36,17 +36,17 @@ chai.Assertion.addMethod(
     if (expected.gte(BigNumber.from(0))) {
       chai
         .expect(actual, message)
-        .gte(expected.mul(10000 - maxTolerancePct * 100).div(10000));
+        .gte(expected.mul(1000000 - maxTolerancePct * 10000).div(1000000));
```
sparrowDom, line 39: this just adds a sub basis point precision to the `approxEqualTolerance `

contracts/test/token/woeth.mainnet.fork-test.js
```diff
@@ -262,11 +265,42 @@ describe("ForkTest: wOETH", function () {
...
+        .sub(initialOethBalance)
+        .mul(BigNumber.from("1000000000000000000"))
+        .div(initialOethBalance);
+      const woethRateIncrease = assetTransfered
+        .sub(initialDeposit)
+        .mul(BigNumber.from("1000000000000000000"))
+        .div(initialDeposit);
+
+
+      await expect(oethRateIncrease).to.equal(woethRateIncrease);
```
sparrowDom, line 279: I wonder if we should allow for 1-2 WEI rounding errors here

sparrowDom: we will add these in if they start failing

contracts/contracts/token/WOETH.sol
```diff
@@ -6,6 +6,7 @@ import { ERC20 } from "@openzeppelin/contracts/token/ERC20/ERC20.sol";
 import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
 import { IERC20Metadata } from "@openzeppelin/contracts/token/ERC20/extensions/IERC20Metadata.sol";
 import { SafeERC20 } from "@openzeppelin/contracts/token/ERC20/utils/SafeERC20.sol";
```
DanielVF, line 8: ⚪ I don't think we use `StableMath` or `SafeCast` in this code?

contracts/contracts/token/WOETH.sol
```diff
@@ -106,87 +113,8 @@ contract WOETH is ERC4626, Governable, Initializable {
 
     /** @dev See {IERC4262-totalAssets} */
     function totalAssets() public view override returns (uint256) {
```
DanielVF, line 115: We could remove the todo comment in transferToken?

contracts/contracts/token/WOETH.sol
```diff
@@ -6,6 +6,7 @@ import { ERC20 } from "@openzeppelin/contracts/token/ERC20/ERC20.sol";
 import { IERC20 } from "@openzeppelin/contracts/token/ERC20/IERC20.sol";
```
DanielVF, line 6: Switch to BSL?

contracts/contracts/token/WOETH.sol
```diff
@@ -28,8 +29,15 @@ import { OETH } from "./OETH.sol";
 contract WOETH is ERC4626, Governable, Initializable {
```
DanielVF, line 29: Some of the comment above this is no longer accurate.

contracts/contracts/token/WOETH.sol
```diff
@@ -28,8 +29,15 @@ import { OETH } from "./OETH.sol";
...
+     * OETH's rebase since inception (expressed with rebasingCreditsPerToken) and OWETH to OETH
+     * conversion.
+     *
+     * If WOETH and OETH are deployed at the same time, the value of _adjuster is a neutral 1e27
+     */
+    uint256 public _adjuster;
+    bool private _oethExchangeRateInitialized;
     uint256[48] private __gap;
 
     // no need to set ERC20 name and symbol since they are overridden in WOETH & WOETHBase
```
DanielVF, line 43: From shah, we don't need Governable anymore.

DanielVF:
🟠 We need to make sure that each implementation inherits from WOETH. WOSonic and WOETH currently do not.

contracts/contracts/token/WOETH.sol
```diff
@@ -106,87 +113,8 @@ contract WOETH is ERC4626, Governable, Initializable {
```
DanielVF, line 113: transferToken should use the generic "core asset" in both the require and the comment, so that it can be used unchanged in other code bases.

DanielVF:
🟥 There was a critical bug with rounding, allowing an attacker to steal from, and I think drain the contract.

totalAssets() rounds down. When it rounds down, it favors deposits. It's possible for an attacker to make a tiny mint to place the contract such that it has max down rounding error, then deposit a massive amount at an advantage. The attacker sizes this big amount, such that the tiny end then causes the ending rounding error to be minimal, and withdraws, collecting a profit.

DanielVF:
🟥 A mint at a zero totalSupply can mint at a privileged rate and thus make the contract insolvent. This does requires either a rebase to have happened between the initialize and the first mint, or the wrapped token to have gone back down to zero total supply and then a mint happen there.

The first deposit would mint at a 1 privileged rate. 

Also fixed by the above changes. 

image

contracts/contracts/token/WOETH.sol
sparrowDom: thanks fixed: https://github.com/OriginProtocol/origin-dollar/pull/2453/commits/c97a327e60a1c333e73ab1e179a21dc6736d6c06

contracts/contracts/token/WOETH.sol
sparrowDom: Agreed: https://github.com/OriginProtocol/origin-dollar/pull/2453/commits/ca1a9984fb2481f3267bf507dd57527aa1f1825e

We don't need to think about doing this. Lets implement it once someone actually donates a substantial amount

contracts/contracts/token/WOETH.sol
sparrowDom: Fixed: https://github.com/OriginProtocol/origin-dollar/pull/2453/commits/599c41524ca3698056f12515b2fd1ba1c448f376

contracts/contracts/token/WOETH.sol
```diff
@@ -111,10 +111,25 @@ contract WOETH is ERC4626, Governable, Initializable {
         IERC20(asset_).safeTransfer(governor(), amount_);
     }
 
+    /**
+     * @dev See {IERC4262-convertToShares}
```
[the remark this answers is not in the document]
sparrowDom: changed it to the natspec compliant `@inheritdoc`: https://github.com/OriginProtocol/origin-dollar/pull/2453/commits/a2a69c9e2ff8b5de1d2189de48ba6d467ade0492

contracts/contracts/token/WOETH.sol
```diff
@@ -54,18 +59,16 @@ contract WOETH is ERC4626, Governable, Initializable {
...
-         * finicky since it reports Highres version of credits and creditsPerToken
-         * when the account is a fresh one. That doesn't have an effect on mainnet since
-         * WOETH has already seen transactions. But it is rather annoying in unit test
-         * environment.
-         */
-        oethCreditsHighres = _getOETHCredits();
+        require(_adjuster == 0, "Initialize2 already called");
+
+        if (totalSupply() == 0) {
+            _adjuster = 1e27;
```
sparrowDom, line 65: Another valid approach here would be that in case of 0 `totalSupply` we could set the adjuster to rebasingCredits: 
```
_adjuster = rebasingCreditsPerTokenHighres();
```

The difference in behaviour occurs if we deposit a fresh wrapped contract after rebases on the underlying token have already occurred. In case we set the adjuster to `rebasingCreditsPerTokenHighres()` the WOETH -> OETH exchange rate would start at 1:1 and follow the underlying token's rebase rate.
If we keep it at `1e27` the exchange rate will follow the value accrual of OETH from inception.

DanielVF: I think I prefer the 1e27 approach - bigger number is more accuracy? Plus we'll be deploying this soon after launch on new tokens, so the end number should be very similar.

DanielVF (dismissed):
# WOETH Audit

## Requirements

We want to prevent donation attacks against lending protocols using our system. We want to prevent someone donating to this contract to instantly increase the exchange rate. 

In this code, the exchange rates are directly tied to the OETH exchange rates without any consideration for the shares or assets on WOETH. Once initialized, there is no way to affect the exchange rate on this contract without going through the vault (which now has limits). This should meet our needs. 

## Easy Checks

#### Authentication

- [x] Never use tx.origin

#### Ethereum

#### Cryptographic code

_No crypto_

#### Gas problems

_No for loops_

#### Black magic

_No magic_

#### Overflow

#### Proxy

#### Events

## Medium Checks

#### Rounding and casts
	- [x] 🟢 TotalAssets rounds down, which underestimates actual assets.
	- [x] 🟢 Have fork tested stupidly large OETH rebase ratios with precision loss, and WOETH errors in the correct direction.
	- [x] 🟢 When adjuster is zero, mints/deposits are disabled because of a divide by zero error. On new contracts there can be no supply until after adjuster is set.
	- [x] 🟢 Adjuster cannot be changed back to zero once it has a positive number, because of the require on the init.
	- [x] 🟣 If you were to migrate (initialize2) a contract with a very small totalSupply (say 1), and have donated some assets to it, you could force the contract into having big rounding errors. Ajuster errors favor deposits. Something like the classic first donation attack on 4626 contracts. However, we only have a few old wrapped tokens, and they already have lots of funds in them and will have sane adjustor values. New deploys are immune since they will go to a hardcoded 1e27. 
	- [x] zero totalSupply() or zero totalAssets() does not affect exchange rates in or out.
#### Dependencies
- [x] Check OpenZeppelin [security vulnerabilities](https://github.com/OpenZeppelin/openzeppelin-contracts/security/advisories) and if any apply to current PR considering the version of OpenZeppelin contract used.

#### External calls

	- [x] 🟢 No user controlled addresses
- [x] Reentrancy safe
- [x] Oracles, one of:
  - [ ] No oracles
  - [x] Oracles can't be bent

#### Deploy

## Thinking

#### Logic

  Good, see notes under flavor.

#### Internal State

We want all shares to be redeemable. The worst case for this is a single withdraw of all shares, which is represented by `totalAssets()`. Therefor `totalAssets()` should return a number that is less than the actual assets held by the contract. This has been the case in all my testing. Rounding errors on deposits and withdraws add to the solvency, it is not given to users, and rebasing also rebases the solvency dust.

#### Attack

These wrapped token contracts usually contain a large amount of funds - up to half the tokens for a given Origin product may be in the wrapped contract. If an attacker found a way to drain these funds, they would be very happy. 

🟢 The easiest way to drain the assets would be to have a time when convert shares and convert assets cross each other. Because  `convertAssets()` and `convertShares()` are computed from the same numbers, just with different rounding directions, they should never cross.

🟢 Alternately, you can find a way to get convert shares and convert assets to cross each other over time. Obviously they cross each other in a way as yield goes up, but that’s just earning yield, and does not negatively affect other depositors. In theory the rounding error in deposit changes from equal to unfavorable. However this should not let an attacker beat the ratio. 

😜 Once rounding errors became a thing, an attacker could front run deposits with tiny deposits of their own, maximizing depositor rounding error. The attacker wouldn’t even profit if holding OETH, since all rounding error is by design stored by the contract and not shared with users.

🟢 If the exchange rate could be manipulated rapidly either up or down, it could cause problems for AMMs or lending platforms using this. In theory, the exchange rate cannot be manipulated down because OE's exchange rate cannot go down. Upwards should be controlled by the vault, and rate limited.

🟢 This code is sensitive to the exact nuances of how OUSD rebases and may have tiny insolvency issues if there is some mismatch in behavior. I’ve fork tested that with ludicrous totalSupply growth, the contract gets more solvent, not less.

🟢 Trappped dust will continue to stay in the contract and grow. Should be fine though. 

🟢 There is an instant update of the exchange rate when the exchange rate of OUSD changes. However, this is actually a more restrictive situation than it used to be - before the exchange rate would instantly update on OUSD exchange rate changes as well as on donations or changes in totalAssets()/totalSupply(). At a minimum we are no worse than before.


#### Flavor

Probably not worth making the below changes after so much security review has happened, but I think it would be more beautiful to:

- use `convertToAssets()` in `totalAssets()`, rather than duping the calculation
- and change `rebasingCreditsPerTokenHighres()` to `_rebasingCreditsPerTokenHighres()`

shahthepro (approved):
## Requirements

Fix the logic to prevent donation attacks on wOETH contract.

## Easy Checks

#### Authentication

- [x] Never use tx.origin

#### Ethereum

#### Cryptographic code

#### Gas problems

    - [ ] A way to remove items
    - [x] No loops

#### Black magic

#### Overflow

#### Proxy

#### Events

## Medium Checks

#### Rounding and casts

#### Dependencies

#### External calls

- [x] No malicious behaviors
- [ ] Oracles, one of:
  - [x] No oracles
  - [ ] Oracles can't be bent

#### Tests

  - [x] If tests interact with AMM make sure enough edge cases (pool tilts) are tested. Ideally with fuzzing.

#### Deploy

#### Downstream

## Thinking

#### Logic

  - [x] Correct usage of global & local variables

N/A

#### Internal State

`adjuster` is the only new variable added to the contract. Ideally it could have been a immutable set in the constructor. However, it relies on OETH's state and a few local functions (`totalSupply`), which might not be available in the implementation contract.

#### Attack

Couldn't think of anything. Simulated a few edge case scenarios and they all worked as expected.

#### Flavor

The code is simple and easy to understand.

:red_circle: The code inherits from another older PR, but almost all of the changes from that PR are removed in this branch. Would be great if we can point this to `master` instead of the older PR. The code is not synced with `master` for over a month, it __**may**__ lead to merge conflicts and unnecessary behaviour of contracts/tests as it is.


:yellow_circle: The code only updates the wOETH contract. Doesn't change wOUSD and wOETHb tokens.

sparrowDom:
thanks @shahthepro all issues addressed: 
- PR branched off of mainnet
- WrappedOUSD now extends WOETH 
- added deploy file and test files for WOUSD

DanielVF (approved):

shahthepro (approved):

PR #2462, opened 2025-03-25
Remove DAI migration strategy contract

shahthepro:
The proposal is executed. We are not gonna reuse this contract anymore. So this just deletes it


sparrowDom (approved):
LGTM

PR #2474, opened 2025-04-07
Merkl Pool Booster

clement-ux:
## Description 
Create a Pool booster and the relative factory to create campaign on [Merkl Distributor](https://app.merkl.xyz/). 
Tutorial to generate correct payload: [Notion](https://www.notion.so/originprotocol/Merkl-Pool-Booster-1ce84d46f53c80cb94b4f468439226af?pvs=4#1ce84d46f53c80538cdfe5947b411ae7).

## Code Change Checklist

- [ ]  Unit tests *if needed

Internal review:

## Deploy checklist

```
```


contracts/contracts/poolBooster/PoolBoosterFactoryMerkl.sol
```diff
@@ -0,0 +1,163 @@
+// SPDX-License-Identifier: MIT
```
naddison36, line 1: We are now using `BUSL-1.1` by default

clement-ux: Thanks for flagging it! Changed in this commit [1adf5dd](https://github.com/OriginProtocol/origin-dollar/pull/2474/commits/1adf5ddff3c7612a4c29d27ac4cd07e35c73f4e1)

contracts/contracts/poolBooster/PoolBoosterMerkl.sol
```diff
@@ -0,0 +1,125 @@
...
+                campaignId: bytes32(0),
+                creator: creator,
+                rewardToken: address(rewardToken),
+                amount: balance,
+                campaignType: campaignType,
+                startTimestamp: getTomorrowRoundedTimestamp(),
+                duration: duration,
+                campaignData: campaignData
+            }),
+            bytes("")
```
naddison36, line 102: How does `signAndCreateCampaign` work with an empty signature?

contracts/contracts/poolBooster/PoolBoosterMerkl.sol
```diff
@@ -0,0 +1,125 @@
...
+        campaignData = _campaignData;
+    }
+
+    /// @notice Create a campaign on the Merkl distributor
+    function bribe() external override {
+        // Ensure token is approved for the Merkl distributor
+        uint256 minAmount = merklDistributor.rewardTokenMinAmounts(
+            address(rewardToken)
+        );
+        require(minAmount > 0, "Invalid minAmount for the reward token");
```
naddison36, line 76: nit: I like to keep error strings to 31 chars or less so they fit in one word (256 bytes)

contracts/contracts/poolBooster/PoolBoosterMerkl.sol
```diff
@@ -0,0 +1,125 @@
...
+        hashToSign = _hashToSign;
+        creator = _creator;
+
+        merklDistributor = IMerklDistributor(_merklDistributor);
+        rewardToken = IERC20(_rewardToken);
+        campaignData = _campaignData;
+    }
+
+    /// @notice Create a campaign on the Merkl distributor
+    function bribe() external override {
```
naddison36, line 71: Since anyone can call `bribe`, do we need to check where the last campaign has finished?
I'm concerned about having overlapping campaigns trying to spend the same reward tokens in the pool booster contract. Or does Merkl prevent overlapping campaigns?

contracts/deploy/sonic/018_merkl_pool_booster.js
```diff
@@ -0,0 +1,84 @@
...
+      cPoolBoostCentralRegistryProxy.address
+    );
+
+    // ---------------------------------------------------------------------------------------------------------
+    // ---
+    // --- Deploy PoolBoosterFactoryMerkl
+    // ---
+    // ---------------------------------------------------------------------------------------------------------
+    const hashToSign =
+      "0x97bd015d4e48fc7d8e7db116ba2d83567597ea6eb64e0694c6fccccd5a4b1841";
```
naddison36, line 48: Was this randomly generated? Or is it a hash of something?

naddison36 (commented):
I've had an initial pass

clement-ux:
> what's the expected length of a campaign?

@naddison36 This is an example of generated campaignData:

```
0x00000000000000000000000069dfa0235f5df814fb15e1d7c774dc1bc52ab338000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000012000000000000000000000000000000000000000000000000000000000000001400000000000000000000000000000000000000000000000000000000000000160000000000000000000000000000000000000000000000000000000000000018000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
```

clement-ux:
> Have I understand this correctly? We create a new PoolBoosterMerkl contract via the PoolBoosterFactoryMerkl contract for each campaign and update the yield forwarding address on the OS token for the Euler pool to the new PoolBoosterMerkl contract.

@naddison36 Yes exactly.

contracts/contracts/poolBooster/PoolBoosterFactoryMerkl.sol
```diff
@@ -0,0 +1,163 @@
...
+    address public merklDistributor;
+    /// @notice address of the Merkl hash to sign
+    bytes32 public merklHashToSign;
+
+    /**
+     * @param _oSonic address of the OSonic token
+     * @param _governor address governor
+     * @param _centralRegistry address of the central registry
+     * @param _merklDistributor address of the Merkl distributor
+     */
```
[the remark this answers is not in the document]
clement-ux: Changed in this commit: [7ddfc4e](https://github.com/OriginProtocol/origin-dollar/pull/2474/commits/7ddfc4e8ec48471cc0cc99a9934f7a374e787450)

contracts/contracts/poolBooster/PoolBoosterMerkl.sol
clement-ux: Here is the process for signing:
- On Merkle Distributor:
image
- Then on the SignatureChecker library, it will go to `isValidERC1271SignatureNow` as `ECDSA.tryRecover` fails as this is not an EOA. 
image
- It come back on the MerklePoolBooster contract and call `isValidSignature(signature)`.
- Capture d’écran 2025-04-09 à 08 56 33
-  And this should return the selector of `isValidSignature`. So the `signature` will be only used in the MerklePoolBooster. And because it is called by the PoolBooster itself, I think that the signature isn't necessary.

Let me know if this makes sense for you?

contracts/contracts/poolBooster/PoolBoosterMerkl.sol
clement-ux: Thanks for catching it! Changed in this commit: [a9bd169](https://github.com/OriginProtocol/origin-dollar/pull/2474/commits/a9bd16909e4ce660f2e7508ac4352f619e70e086)

contracts/deploy/sonic/018_merkl_pool_booster.js
clement-ux: I copied it from this [`messageHash()`](https://sonicscan.org/address/0x8BB4C975Ff3c250e0ceEA271728547f3802B36Fd#readProxyContract) view function 28.

contracts/contracts/poolBooster/PoolBoosterMerkl.sol
clement-ux: Merkl prevent creating campaign with the same Id to prevent overlapping campaigns.
This is how campaignId are generated: 
image

Only startTimestamp and duration will have an impact on the two different call of `MerklePoolBooster::bribe()`. 

So there is two call on the same day, it will revert yes, as Merkl prevent overlapping campaigns .

naddison36:
> > what's the expected length of a campaign?
> 
> @naddison36 This is an example of generated campaignData:
> 
> ```
> 0x00000000000000000000000069dfa0235f5df814fb15e1d7c774dc1bc52ab338000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000012000000000000000000000000000000000000000000000000000000000000001400000000000000000000000000000000000000000000000000000000000000160000000000000000000000000000000000000000000000000000000000000018000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
> ```

Sorry, I meant duration of a campaign. That is, how many days?

clement-ux:
> > > what's the expected length of a campaign?
> > 
> > 
> > @naddison36 This is an example of generated campaignData:
> > ```
> > 0x00000000000000000000000069dfa0235f5df814fb15e1d7c774dc1bc52ab338000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000010000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000012000000000000000000000000000000000000000000000000000000000000001400000000000000000000000000000000000000000000000000000000000000160000000000000000000000000000000000000000000000000000000000000018000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
> > ```
> 
> Sorry, I meant duration of a campaign. That is, how many days?

Oh yes, it should be 1 day by default. This will be immutable on each PoolBooster, duration decided when deploying it with the factory.

naddison36:
#### Proxy

- `PoolBoosterFactoryMerkl` and `PoolBoosterMerkl` are not proxied

#### Black magic

#### Dependencies

#### Deploy

#### Authentication

- [x] Never use tx.origin

#### Cryptographic code

- [ ] Beware of signed data being used in a replay attack to other contracts.

#### Gas problems

    - [ ] A way to remove items

#### External calls

- [x] Unsafe external calls
- [ ] Malicious behaviors
- [x] Oracles, one of:
  - [x] No oracles
  - [ ] Oracles can't be bent

#### Ethereum

#### Flavor

Could this code be simpler?

It feels like `merklDistributor` and `merklHashToSign` would never change in `PoolBoosterFactoryMerkl`.  If that's the case, the setter functions could be removed. But given the factory is not proxied, I'm happy to leave `merklDistributor` and `merklHashToSign` as storage variables.

Could this code be less vulnerable to other code behaving weirdly?

I was worried about the `signAndCreateCampaign` call to the merklDistributor not including a signature. I think it's ok if we tighten up the `isValidSignature` callback by checking the caller is the merklDistributor contract.

contracts/contracts/poolBooster/PoolBoosterMerkl.sol
naddison36: I see. As we calculate the `startTimestamp` from `getTomorrowRoundedTimestamp`, there can only be one bribe per day

contracts/contracts/poolBooster/PoolBoosterMerkl.sol
```diff
@@ -0,0 +1,125 @@
...
+    {
+        // Check if the signature is valid for the given hash
+        // bytes4(keccak256("isValidSignature(bytes32,bytes)")) == 0x1626ba7e
+        return (hash == hashToSign ? bytes4(0x1626ba7e) : bytes4(0x00000000));
+    }
+
+    /// @notice Returns the current timestamp rounded to the next day
+    function getTomorrowRoundedTimestamp() public view returns (uint32) {
+        // Calculate the timestamp for the start of tomorrow (next midnight)
+        return uint32((block.timestamp / 1 days + 1) * 1 days);
```
naddison36, line 123: I think this needs to use the `duration` and not `1 days`. What if the `duration` was 7 days? Wouldn't that mean we can get overlapping campaigns?

The alternative is to fix the `duration` to 1 day.

contracts/contracts/poolBooster/PoolBoosterMerkl.sol
```diff
@@ -0,0 +1,125 @@
...
+    address public immutable creator;
+    /// @notice Campaign data
+    bytes public campaignData;
+
+    constructor(
+        address _rewardToken,
+        address _merklDistributor,
+        uint32 _duration,
+        uint32 _campaignType,
+        bytes32 _hashToSign,
```
naddison36, line 47: Instead of passing in the hash from Merkl's `DistributionCreator` contract, lets just call `messageHash` on the `merklDistributor` contract to get it.

This would also mean `setMerklHashToSign` could be replaced with `updateMerklHashToSign`

contracts/contracts/poolBooster/PoolBoosterFactoryMerkl.sol
```diff
@@ -0,0 +1,164 @@
...
+ * @title Pool booster factory for creating Merkl pool boosters.
+ * @author Origin Protocol Inc
+ */
+contract PoolBoosterFactoryMerkl is AbstractPoolBoosterFactory {
+    uint256 public constant version = 1;
+
+    /// @notice address of the Merkl distributor
+    address public merklDistributor;
+    /// @notice address of the Merkl hash to sign
+    bytes32 public merklHashToSign;
```
naddison36, line 17: Since we don't sign this hash, I think we can rename it to just `merklHash`. That would remove some of the signing confusion.

contracts/contracts/poolBooster/PoolBoosterFactoryMerkl.sol
```diff
@@ -0,0 +1,164 @@
...
+        _setMerklDistributor(_merklDistributor);
+    }
+
+    function _setMerklDistributor(address _merklDistributor) internal {
+        require(
+            _merklDistributor != address(0),
+            "Invalid merklDistributor address"
+        );
+        merklDistributor = _merklDistributor;
+    }
```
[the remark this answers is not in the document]
clement-ux: [bc3b8e0](https://github.com/OriginProtocol/origin-dollar/pull/2474/commits/bc3b8e0ccd760077fe97788046c10253de4a278e)

contracts/contracts/poolBooster/PoolBoosterFactoryMerkl.sol
```diff
@@ -0,0 +1,164 @@
...
+        external
+        onlyGovernor
+    {
+        _setMerklHashToSign(_merklHashToSign);
+    }
+
+    function _setMerklHashToSign(bytes32 _merklHashToSign) internal {
+        require(_merklHashToSign != bytes32(0), "Invalid merklHashToSign");
+        merklHashToSign = _merklHashToSign;
+    }
```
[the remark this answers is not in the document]
clement-ux: [bc3b8e0](https://github.com/OriginProtocol/origin-dollar/pull/2474/commits/bc3b8e0ccd760077fe97788046c10253de4a278e)

contracts/contracts/poolBooster/PoolBoosterMerkl.sol
naddison36: Thinking about this some more, I think we can just drop the `hashToSign` storage variable. Validating the `hash` param in `isValidSignature` doesn't give us much. It's much better to validate that the caller of `isValidSignature` is from the `merklDistributor` contract.

contracts/contracts/poolBooster/PoolBoosterMerkl.sol
clement-ux: Nice catch thanks! 
Indeed I should have use duration and not days and I prefer to keep duration flexible.
Fixed in this commit: [7b20bfe](https://github.com/OriginProtocol/origin-dollar/pull/2474/commits/7b20bfeb1cb5788047f686b840e6c8bccccb7bdf)
I changed the function name to have something more meaningful.

contracts/contracts/poolBooster/PoolBoosterMerkl.sol
clement-ux: I thought about it at the beginning, we could even remove this and fetch it will signing. 
I was worried about Merkl going crazy and make us sign bad stuff. However as this is not really signing, I'm not sure this is a bad thing. 

Should I go for the first option and fetch the message to sign when creating a campaign?

clement-ux: > Thinking about this some more, I think we can just drop the `hashToSign` storage variable. Validating the `hash` param in `isValidSignature` doesn't give us much. It's much better to validate that the caller of `isValidSignature` is from the `merklDistributor` contract.

I wrote the previous comment before seeing this answer. I agree that we don't really need to validate the `hash`. I'll remove it.

clement-ux: I completely removed `hashToSign` from the contract: [c39d0c1](https://github.com/OriginProtocol/origin-dollar/pull/2474/commits/c39d0c1a224be71f960903008dedd551aea7d6c0)

PR #2480, opened 2025-04-11
Add managePassThrough script and update rollup configuration

clement-ux:
## Description
- Add script to manage pass through for OZ Defender actions.


contracts/scripts/defender-actions/managePassThrough.js
```diff
@@ -0,0 +1,59 @@
...
+const handler = async (event) => {
+  console.log(
+    `DEBUG env var in handler before being set: "${process.env.DEBUG}"`
+  );
+
+  // Initialize defender relayer provider and signer
+  const provider = new DefenderRelayProvider(event);
+  const signer = new DefenderRelaySigner(event, provider, { speed: "fastest" });
+
+  const OUSD = "0x2A8e1E676Ec238d8A992307B495b45B3fEAa5e86";
```
naddison36, line 20: It's ok to import the JS addresses. ie 
```js
const addresses = require("../../utils/addresses");
```

naddison36 (commented):
I like to have the logic in a Defender Action also callable by a Hardhat task. That way it can be tested against a local node impersonating the Relayer. Or the node can be started from a past block to test.

An example is `contracts/scripts/defender-actions/sonicRequestWithdrawal.js` which uses the `undelegateValidator` function in  `contracts/utils/sonicActions.js`. 
The same `undelegateValidator` function is also called from the `sonicUndelegate` task in `contracts/tasks/tasks.js`.

It's not essential. We do have other Actions that don't have Hardhat tasks. We even have simple Actions with just the code living in the Action and not this repo.

contracts/scripts/defender-actions/managePassThrough.js
clement-ux: Fixed in this commit: [a7705a7](https://github.com/OriginProtocol/origin-dollar/pull/2480/commits/a7705a7fff1ad9dd191effd7674d28dbdcd4f0e9)

clement-ux:
> I like to have the logic in a Defender Action also callable by a Hardhat task. That way it can be tested against a local node impersonating the Relayer. Or the node can be started from a past block to test.

Yes sure, no problem, I'll create a task for it.

naddison36 (approved):
Looks good

naddison36 (changes requested):
Just need to run prettier and I'll merge

naddison36 (approved):
I've ran prettier

PR #2499, opened 2025-04-29
Plume Rooster AMO

sparrowDom:
## Code Change Checklist

- [ ]  Unit tests *if needed

Internal review:

## Deploy checklist

```
```


contracts/contracts/strategies/plume/RoosterAMOStrategy.sol
```diff
@@ -0,0 +1,1301 @@
...
+        );
+        require(
+            _poolDistributor != address(0),
+            "Pool distributor zero address not allowed"
+        );
+
+        uint256 _tickSpacing = IMaverickV2Pool(_mPool).tickSpacing();
+        require(_tickSpacing == 1, "Unsupported tickSpacing");
+        // -1 is the tick where the contract deposits liquidity to
+        tickNumber = -1;
```
clement-ux, line 239: Just by curiosity, why not use a constant here instead of immutable? For the sake of harmony?

contracts/contracts/strategies/plume/RoosterAMOStrategy.sol
```diff
@@ -0,0 +1,1301 @@
...
+        liquidityManager = IMaverickV2LiquidityManager(_liquidityManager);
+        poolLens = IMaverickV2PoolLens(_poolLens);
+        maverickPosition = IMaverickV2Position(_maverickPosition);
+        quoter = IMaverickV2Quoter(_maverickQuoter);
+        mPool = IMaverickV2Pool(_mPool);
+        votingDistributor = IVotingDistributor(_votingDistributor);
+        poolDistributor = IPoolDistributor(_poolDistributor);
+
+        require(address(mPool.tokenA()) == WETH, "WETH not TokanA");
+        require(address(mPool.tokenB()) == OETHp, "OETHp not TokanB");
```
clement-ux, line 261: This seems a duplicate with the previous requires, at the beginning of the constructor.
> require(address(IMaverickV2Pool(_mPool).tokenA()) == _wethAddress, "Only WETH supported as tokenA");
> require(address(IMaverickV2Pool(_mPool).tokenB()) == _oethpAddress,"Only OETHp supported as tokenB");

contracts/contracts/strategies/plume/RoosterAMOStrategy.sol
```diff
@@ -0,0 +1,1301 @@
...
+        uint256 _allowedWethShareEnd
+    ) external onlyGovernor {
+        require(
+            _allowedWethShareStart < _allowedWethShareEnd,
+            "Invalid interval"
+        );
+        // can not go below 1% weth share
+        require(_allowedWethShareStart > 0.01 ether, "Invalid interval start");
+        // can not go above 95% weth share
+        require(_allowedWethShareEnd < 0.95 ether, "Invalid interval end");
```
clement-ux, line 307: nit: Maybe we could replace magic numbers (`0.01 ether` and `0.95 ether`) by a constant/immutable with an intuitive name?

contracts/contracts/strategies/plume/RoosterAMOStrategy.sol
```diff
@@ -0,0 +1,1301 @@
...
+        _deposit(_asset, _amount);
+    }
+
+    /**
+     * @notice Deposits all the funds to the strategy which deposits them to the
+     * underlying Rooster pool if the pool price is within the expected interval.
+     */
+    function depositAll() external override onlyVault nonReentrant {
+        uint256 _wethBalance = IERC20(WETH).balanceOf(address(this));
+        if (_wethBalance > 1e12) {
```
clement-ux, line 342: nit: Same here for magic number (`1e12`).

clement-ux (commented):
Just a quick first pass, nothing very important, but I'd like your opinion.

contracts/contracts/strategies/plume/RoosterAMOStrategy.sol
DanielVF: If a constant number is only used once and is used in a numeric sense, I often prefer it to live where it is used. This makes reviewing and understanding the code easier.

If however, it actually represents something that's not a number, then it's good to constant it up.

For example, if a "2" is actually "MODE_EXACT_INPUT_IN", it really helps to make it a constant. But in this, case, "don't allow the range to be out of 95%" is best have it clear in the code.

contracts/contracts/strategies/plume/RoosterAMOStrategy.sol
sparrowDom: Agreed constant would be more correct: https://github.com/OriginProtocol/origin-dollar/pull/2499/commits/7a970ed694b618b77937cd63283e1ef2d23ec860

contracts/contracts/strategies/plume/RoosterAMOStrategy.sol
sparrowDom: nice catch: https://github.com/OriginProtocol/origin-dollar/pull/2499/commits/14f8ba4cb4cc12e60dec3f5422c3ceaa80faeb03

contracts/contracts/strategies/plume/RoosterAMOStrategy.sol
```diff
@@ -0,0 +1,1301 @@
...
+    error InsufficientTokenBalance(
+        uint256 tokenOffered,
+        uint256 tokenRequired,
+        address token
+    ); // 0xe23d5ff7
+
+    /**
+     * @dev Verifies that the caller is the Governor, or Strategist.
+     */
+    modifier onlyGovernorOrStrategist() {
```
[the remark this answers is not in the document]
sparrowDom: agreed: https://github.com/OriginProtocol/origin-dollar/pull/2499/commits/4204c2af9430c4a264b55a0a5e247ae268397b0b

clement-ux (commented):
More comments.
A more global question, what do you think of having all view as public instead of internal? I think that it can only help off-chain work and there is no downside of doing so.  Happy to have your thought on it.

contracts/contracts/strategies/plume/RoosterAMOStrategy.sol
```diff
@@ -0,0 +1,1274 @@
...
+     * @notice Tick dominance denominated in 1e18
+     *
+     */
+    function tickDominance() public view returns (uint256 _tickDominance) {
+        uint256 _currentPrice = getPoolSqrtPrice();
+        (
+            IMaverickV2Pool.TickState memory tickState,
+            ,
+
+        ) = _reservesInTickForGivenPrice(tickNumber, _currentPrice);
```
clement-ux, line 1134: This can be replaced by something like this as we only used the tickState, and `_currentPrice` is not used: 
```solidity
IMaverickV2Pool.TickState memory tickState = mPool.getTick(tickNumber).
```

contracts/contracts/strategies/plume/RoosterAMOStrategy.sol
sparrowDom: great point fixed here: https://github.com/OriginProtocol/origin-dollar/pull/2499/commits/9ad942f2df69f6d2ccd66e15305a51c319ec3237

contracts/contracts/strategies/plume/RoosterAMOStrategy.sol
```diff
@@ -0,0 +1,1301 @@
...
+     *      position in the pool or not. The function returns _wethSharePct
+     *      as a gas optimization measure.
+     * @param throwException  when set to true the function throws an exception
+     *        when pool's price is not within expected range.
+     * @return _isExpectedRange  Bool expressing price is within expected range
+     * @return _wethSharePct  Share of WETH owned by this strategy contract in the
+     *         configured ticker.
+     */
+    function _checkForExpectedPoolPrice(bool throwException)
+        internal
```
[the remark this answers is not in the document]
sparrowDom: I think the off-chain can use more appropriate: `getWETHShare`, `allowedWethShareStart`, `allowedWethShareEnd` to get more insight to the current price and the strategy imposed limits.
IMO the `_checkForExpectedPoolPrice` is in light of the other functions somewhat redundant

contracts/contracts/strategies/plume/RoosterAMOStrategy.sol
```diff
@@ -0,0 +1,1301 @@
...
+        returns (uint256)
+    {
+        (
+            uint256 wethAmount,
+            uint256 oethpAmount
+        ) = _reservesInTickForGivenPriceAndLiquidity(
+                sqrtPriceTickLower,
+                sqrtPriceTickHigher,
+                _currentPrice,
+                1e24
```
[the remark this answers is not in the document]
sparrowDom: The liquidity amount is not important here. It could be any number or even the current liquidity.

contracts/contracts/strategies/plume/RoosterAMOStrategy.sol
```diff
@@ -0,0 +1,1290 @@
...
+            );
+
+        return wethAmount.divPrecisely(wethAmount + oethpAmount);
```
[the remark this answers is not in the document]
sparrowDom: Yeah the Rooster and Aerodrome (Uniswap V3 fork) provide different liquidity utility functions even though the underlying concentrated liquidity math basics are almost the same. The different approach is due to that

contracts/contracts/strategies/plume/RoosterAMOStrategy.sol
```diff
@@ -0,0 +1,1290 @@
...
+    {
+        int32[] memory ticks = new int32[](1);
+        uint128[] memory amounts = new uint128[](1);
+        ticks[0] = tickNumber;
+        amounts[0] = 1e24;
+
+        // construct value for Quoter with arbitrary LP amount
+        IMaverickV2Pool.AddLiquidityParams memory addParam = IMaverickV2Pool
+            .AddLiquidityParams({
+                kind: 0, // static kind
```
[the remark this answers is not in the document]
sparrowDom: Agreed great point: https://github.com/OriginProtocol/origin-dollar/pull/2499/commits/3c55be1f12d8ab9a81105e6bf9a0738ba287b687

contracts/contracts/strategies/plume/RoosterAMOStrategy.sol
```diff
@@ -0,0 +1,1278 @@
...
+     *
+     * @dev refactored from here:
+     * https://github.com/rooster-protocol/rooster-contracts/blob/main/v2-supplemental/contracts/libraries/LiquidityUtilities.sol#L665-L695
+     */
+    function _reservesInTickForGivenPriceAndLiquidity(
+        uint256 lowerSqrtPrice,
+        uint256 upperSqrtPrice,
+        uint256 newSqrtPrice,
+        uint256 liquidity
+    ) internal view returns (uint128 reserveA, uint128 reserveB) {
```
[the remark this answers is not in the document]
sparrowDom: great catch: https://github.com/OriginProtocol/origin-dollar/pull/2499/commits/769acd8ad263e822b66f1bad94bf8cab7b7d95b5

contracts/contracts/strategies/plume/RoosterAMOStrategy.sol
```diff
@@ -0,0 +1,1278 @@
...
+     */
+    function _removeLiquidity(uint256 _liquidityToDecrease) internal {
+        require(_liquidityToDecrease > 0, "Must remove some liquidity");
+        require(
+            _liquidityToDecrease <= 1e18,
+            "Can not remove more than 100% of liquidity"
+        );
+
+        IMaverickV2Pool.RemoveLiquidityParams memory params = maverickPosition
+            .getRemoveParams(tokenId, 0, _liquidityToDecrease);
```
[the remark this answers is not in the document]
sparrowDom: yes :) This is a nice treat Maverick gave us :)

contracts/contracts/strategies/plume/RoosterAMOStrategy.sol
```diff
@@ -0,0 +1,1278 @@
...
+        ) = _reservesInTickForGivenPrice(tickNumber, _currentPrice);
+
+        uint256 wethReserve = tickState.reserveA;
+        uint256 oethpReserve = tickState.reserveB;
+
+        (
+            uint256 _amountWeth,
+            uint256 _amountOethp,
+
+        ) = _getPositionInformation();
```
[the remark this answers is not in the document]
sparrowDom: As mentioned somewhere previously, the passing of governance proposal will [mint the](https://github.com/OriginProtocol/origin-dollar/pull/2499/files#diff-f775042f777aabc78a490a61c6f13b9727ab8c2d6365b7b0cf57d1339f4d5febR130) initial position

contracts/contracts/strategies/plume/RoosterAMOStrategy.sol
```diff
@@ -0,0 +1,1278 @@
...
+     * @param _asset      WETH address
+     * @param _amount     Amount of WETH to withdraw
+     */
+    function withdraw(
+        address _recipient,
+        address _asset,
+        uint256 _amount
+    ) external override onlyVault nonReentrant {
+        require(_asset == WETH, "Unsupported asset");
+        require(_amount > 0, "Must withdraw something");
```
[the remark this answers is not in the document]
sparrowDom: Yes good spotting this!

The `withdrawAll` in both strategies already checks for balance before calling `_withdrawal` for that reason I made more sense to me to move the check to the only function (`withdrawal`) that can trigger it.

I would argue, this is easier to read, since the check is verified is 1 less code path. And changing it in Aerodrome would probably make sense. Though I am not sure if that to some tiny extent invalidates our existing OZ audit of Aerodrome.

sparrowDom:
@clement-ux regarding the change of internal functions to public. I guess this could be applied to the following 3: 
- _balanceInPosition
- _reservesInTickForGivenPriceAndLiquidity
- _reservesInTickForGivenPrice

- The `_balanceInPosition` returns the `underlyingAssets` which is a public var the off chain script can read. 
- The `_reservesInTickForGivenPriceAndLiquidity` seems to me like off-chain wouldn't have much use of it 
- We could expose the `_reservesInTickForGivenPrice` one though. Could be useful in some cases. moved to public: https://github.com/OriginProtocol/origin-dollar/pull/2499/commits/1b09a9aade965e0f97ea4112a562da248fa1c835

contracts/contracts/strategies/plume/RoosterAMOStrategy.sol
```diff
@@ -0,0 +1,1274 @@
...
+            "More OETHp required than specified"
+        );
+
+        // organize values to be used by manager
+        addParams = new IMaverickV2Pool.AddLiquidityParams[](1);
+        addParams[0] = addParam;
+        packedArgs = liquidityManager.packAddLiquidityArgsArray(addParams);
+        uint88[] memory prices = new uint88[](1);
+        // price can stay 0 if array only has one element
+        packedSqrtPriceBreaks = liquidityManager.packUint88Array(prices);
```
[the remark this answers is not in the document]
sparrowDom: nice, agreed: https://github.com/OriginProtocol/origin-dollar/pull/2499/commits/a50e3b6772d1df756217120a4bdb3e5a3042ac85

contracts/contracts/strategies/plume/RoosterAMOStrategy.sol
```diff
@@ -0,0 +1,1274 @@
...
+     */
+    // slither-disable-end reentrancy-no-eth
+    function _getAddLiquidityParams(uint256 _maxWETH, uint256 _maxOETHp)
+        internal
+        returns (
+            bytes memory packedSqrtPriceBreaks,
+            bytes[] memory packedArgs,
+            IMaverickV2Pool.AddLiquidityParams[] memory addParams,
+            uint256 WETHRequired,
+            uint256 OETHpRequired
```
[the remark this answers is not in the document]
sparrowDom: That is a good point. I've been on the fence about this one, since it is just a modified function of the one offered by the LiquidityManager. I've removed the `addParams` as a return value, since it doesn't really offer much value outside of the function. It still needs to be present to create the `packedArgs`: 
```
packedArgs = liquidityManager.packAddLiquidityArgsArray(addParams);
```

For now I'd keep the `WETHRequired` in, I've used it to validate the accuracy of Quoter calculations. Though it could be argued that we might want to remove that one as well from the return arguments.

https://github.com/OriginProtocol/origin-dollar/pull/2499/commits/437d3f916395a231af23ca918dccbcb12588cf4c

contracts/contracts/strategies/plume/RoosterAMOStrategy.sol
```diff
@@ -0,0 +1,1274 @@
...
+        view
+        returns (
+            IMaverickV2Pool.TickState memory tickState,
+            bool tickLtActive,
+            bool tickGtActive
+        )
+    {
+        tickState = mPool.getTick(_tick);
+        (uint256 lowerSqrtPrice, uint256 upperSqrtPrice) = TickMath
+            .tickSqrtPrices(mPool.tickSpacing(), _tick);
```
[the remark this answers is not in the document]
sparrowDom: Yes this is more of a sanity check than anything else. I wouldn't change it to an immutable or a constant due to not being used outside of the constructor.

contracts/contracts/strategies/plume/RoosterAMOStrategy.sol
```diff
@@ -0,0 +1,1274 @@
...
+
+    /**
+     * @dev The function creates liquidity parameters required to be able to add liquidity to the pool.
+     * The function needs to handle the 3 different cases of the way liquidity is added:
+     *  - only WETH present in the tick
+     *  - only OETHp present in the tick
+     *  - both tokens present in the tick
+     *
+     */
+    // slither-disable-end reentrancy-no-eth
```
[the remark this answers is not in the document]
sparrowDom: Thanks, have addressed all slither issues here: https://github.com/OriginProtocol/origin-dollar/pull/2499/commits/ddb22dcff1bcbe1b8258d5d59d42b242519938e2

contracts/contracts/strategies/plume/RoosterAMOStrategy.sol
```diff
@@ -0,0 +1,1290 @@
...
+        // Remove the required amount of liquidity
+        if (_liquidityToRemovePct > 0) {
+            _removeLiquidity(_liquidityToRemovePct);
+        }
+
+        // In case liquidity has been removed and there is still not enough WETH owned by the
+        // strategy contract remove additional required amount of WETH.
+        if (_swapWeth && _amountToSwap > 0) {
+            _ensureWETHBalance(_amountToSwap);
+        }
```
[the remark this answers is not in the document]
clement-ux: Indeed all good 👍

contracts/contracts/strategies/plume/RoosterAMOStrategy.sol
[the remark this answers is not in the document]
clement-ux: Change looks good 👍

contracts/contracts/strategies/plume/RoosterAMOStrategy.sol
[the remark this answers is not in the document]
clement-ux: Ok, I'm fine to keep it like this too 👍

contracts/contracts/strategies/plume/RoosterAMOStrategy.sol
```diff
@@ -0,0 +1,1228 @@
...
+            _votingDistributor != address(0),
+            "Voting distributor zero address not allowed"
+        );
+        require(
+            _poolDistributor != address(0),
+            "Pool distributor zero address not allowed"
+        );
+
+        uint256 _tickSpacing = IMaverickV2Pool(_mPool).tickSpacing();
+        require(_tickSpacing == 1, "Unsupported tickSpacing");
```
clement-ux, line 228: `_tickSpacing` is not used anymore, so we can refactor this into:
```solidity 
require(IMaverickV2Pool(_mPool).tickSpacing() == 1, "Unsupported tickSpacing");
```

And remove the immutable defined at line 65:
>``` solidity 
>/// @notice tick spacing of the pool (set to 1)
>int24 public immutable tickSpacing;
>```

contracts/contracts/strategies/plume/RoosterAMOStrategy.sol
```diff
@@ -0,0 +1,1226 @@
...
+    function tickDominance() public view returns (uint256 _tickDominance) {
+        IMaverickV2Pool.TickState memory tickState = mPool.getTick(TICK_NUMBER);
+
+        uint256 wethReserve = tickState.reserveA;
+        uint256 oethpReserve = tickState.reserveB;
+
+        (
+            uint256 _amountWeth,
+            uint256 _amountOethp,
+
```
DanielVF, line 1143: ⚪ We probably don't need this blank line.

contracts/contracts/strategies/plume/RoosterAMOStrategy.sol
```diff
@@ -0,0 +1,1226 @@
...
+        ) = _getPositionInformation();
+
+        if ((wethReserve + oethpReserve) > 0) {
```
DanielVF, line 1146: ⚪ Clearer code flow if we early return for the zero case.

contracts/contracts/strategies/plume/RoosterAMOStrategy.sol
```diff
@@ -0,0 +1,1239 @@
...
+
+        _approveTokenAmounts(1e16, OETHpRequired);
+
+        (, , , uint256 _tokenId) = liquidityManager.mintPositionNftToSender(
+            mPool,
+            packedSqrtPriceBreaks,
+            packedArgs
+        );
+        // burn remaining OETHp
+        _burnOethOnTheContract();
```
DanielVF, line 985: ⚪ Sometimes we burnOethOnTheContract() before we update the _updateUnderlyingAssets(), and sometimes after. Would be cleaner to be consistent. I think it would be cleanest to burn before, but either way should work.

contracts/deploy/plume/005_rooster_amo.js
```diff
@@ -0,0 +1,148 @@
...
+          strategyImplInitData
+        )
+    );
+
+    const cAMOStrategy = await ethers.getContractAt(
+      "RoosterAMOStrategy",
+      cAMOStrategyProxy.address
+    );
+
+    if (isFork) {
```
DanielVF, line 100: Is there a reason for doing things differently in the fork vs the real deploy? This means the real deploy settings won't have been actually tested in the fork.

contracts/contracts/strategies/plume/RoosterAMOStrategy.sol
sparrowDom: Agreed: https://github.com/OriginProtocol/origin-dollar/pull/2499/commits/58f849f1e0338d821a375a2fa16e102ad8626a6c

contracts/deploy/plume/005_rooster_amo.js
sparrowDom: agreed, thanks: https://github.com/OriginProtocol/origin-dollar/pull/2499/commits/3640538075eb907172b07b124089a0c8cb3292a1

contracts/contracts/strategies/plume/RoosterAMOStrategy.sol
```diff
@@ -0,0 +1,1181 @@
...
+        // strategy contract remove additional required amount of WETH.
+        if (_swapWeth && _amountToSwap > 0) {
+            _ensureWETHBalance(_amountToSwap);
+        }
+
+        // in some cases (e.g. deposits) we will just want to add liquidity and not
+        // issue a swap to move the active trading position within the pool
+        if (_amountToSwap > 0) {
+            _swapToDesiredPosition(_amountToSwap, _swapWeth, _minTokenReceived);
+        }
```
clement-ux, line 668: What do you think of combining this two sections in a single one? Something like this:
```solidity
// in some cases (e.g. deposits) we will just want to add liquidity and not
// issue a swap to move the active trading position within the pool
if (_amountToSwap > 0) {
    // In case liquidity has been removed and there is still not enough WETH owned by the
    // strategy contract remove additional required amount of WETH.
    if (_swapWeth) _ensureWETHBalance(_amountToSwap);
    
    _swapToDesiredPosition(_amountToSwap, _swapWeth, _minTokenReceived);
}
```
It reduces one check, but more important for me, it groups actions.

sparrowDom: I like the suggestion, thanks: https://github.com/OriginProtocol/origin-dollar/pull/2499/commits/97af511f7a980410133e66a11485565d5ebb0643

contracts/contracts/strategies/plume/RoosterAMOStrategy.sol
```diff
@@ -0,0 +1,1181 @@
...
+        emit LiquidityAdded(
+            _wethBalance, // wethAmountDesired
+            OETHpRequired, // oethpAmountDesired
+            _wethAmount, // wethAmountSupplied
+            _oethAmount, // oethbAmountSupplied
+            tokenId, // tokenId
+            underlyingAssets
+        );
+
+        // burn remaining OETHp and skip check because liquidityManager takes
```
DanielVF, line 463: What check is being skipped here?

contracts/contracts/strategies/plume/RoosterAMOStrategy.sol
```diff
@@ -0,0 +1,1181 @@
...
+    ///      set to a zero address.
+    /// @param _stratConfig the basic strategy configuration
+    /// @param _wethAddress Address of the Erc20 WETH Token contract
+    /// @param _oethpAddress Address of the Erc20 OETHp Token contract
+    /// @param _liquidityManager Address of liquidity manager to add
+    ///         the liquidity
+    /// @param _poolLens Address of the pool lens contract
+    /// @param _maverickPosition Address of the Maverick's position contract
+    /// @param _maverickQuoter Address of the Maverick's Quoter contract
+    /// @param _mPool Address of the Aerodrome concentrated liquidity pool
```
DanielVF, line 172: "Aerodrome" :D

contracts/contracts/strategies/plume/RoosterAMOStrategy.sol
```diff
@@ -0,0 +1,1181 @@
...
+    /// @notice sqrtPriceTickHigher
+    /// @dev tick higher represents 1:1 price parity of WETH to OETHp
+    uint256 public immutable sqrtPriceTickHigher;
+    /// @dev price at parity (in OETHp this is equal to sqrtPriceTickHigher)
+    uint256 public immutable sqrtPriceAtParity;
+    /// @notice The tick where the strategy deploys the liquidity to
+    int32 public constant TICK_NUMBER = -1;
+    /// @notice Minimum liquidity required to continue with the action
+    /// e.g. deposit, add liquidity, burn OETHp
+    uint256 public constant MIN_BALANCE_THRESHOLD = 1e12;
```
DanielVF, line 112: A "Min balance threshold" sounds like we want to keep this much money in the contract. Perhaps it should be `ACTION_THRESHOLD` to show it controls if something happens or not?

contracts/contracts/strategies/plume/RoosterAMOStrategy.sol
```diff
@@ -0,0 +1,1181 @@
...
+        }
+
+        (
+            bytes memory packedSqrtPriceBreaks,
+            bytes[] memory packedArgs,
+            ,
+            uint256 OETHpRequired
+        ) = _getAddLiquidityParams(_wethBalance, 1e30);
+
+        if (OETHpRequired > _oethBalance) {
```
DanielVF, line 435: Thought this code we use "oeth", "oethb", and "oethp" interchangeably. This gets to be a little confusing.

I think we shouldn't add the "oethp"'s. That's at least one less thing.

Event names should probably match the previous AMOs, so the backend team doesn't have to do extra work, and support multiple naming schemes.

Beyond that, my slight preference would be to shift the "oethb"'s back to "oeth", though up for second opinions there.

contracts/contracts/strategies/plume/RoosterAMOStrategy.sol
sparrowDom: `BurnOethOnTheContract` used to take a parameter to skip the minimal threshold check. That has been removed some time ago. Updated the comment, thanks: https://github.com/OriginProtocol/origin-dollar/pull/2499/commits/51cd18bdf9ad23e66fbb500e405e70174dfc6ecd

contracts/contracts/strategies/plume/RoosterAMOStrategy.sol
sparrowDom: oops :): https://github.com/OriginProtocol/origin-dollar/pull/2499/commits/9b1ba1a7e5261386e2fda875681133639ef0ecc1

contracts/contracts/strategies/plume/RoosterAMOStrategy.sol
sparrowDom: better thanks: https://github.com/OriginProtocol/origin-dollar/pull/2499/commits/12350c8f57307f7949edc90cae0479ddccb9964c

contracts/contracts/strategies/plume/RoosterAMOStrategy.sol
```diff
@@ -0,0 +1,1181 @@
...
+        _deposit(_asset, _amount);
+    }
+
+    /**
+     * @notice Deposits all the funds to the strategy which deposits them to the
+     * underlying Rooster pool if the pool price is within the expected interval.
+     */
+    function depositAll() external override onlyVault nonReentrant {
+        uint256 _wethBalance = IERC20(WETH).balanceOf(address(this));
+        if (_wethBalance > MIN_BALANCE_THRESHOLD) {
```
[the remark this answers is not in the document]
sparrowDom: Agreed, the event should still be emitted even if the funds don't get deposited in the underlying pool. Fixed here: https://github.com/OriginProtocol/origin-dollar/pull/2499/commits/cb60e3471637a3be3db4ffa059188200ee8ff325

contracts/contracts/strategies/plume/RoosterAMOStrategy.sol
```diff
@@ -0,0 +1,1181 @@
...
+            mPool,
+            packedSqrtPriceBreaks,
+            packedArgs
+        );
+        // burn remaining OETHp
+        _burnOethOnTheContract();
+
+        // Store the tokenId before calling updateUnderlyingAssets as it relies on the tokenId
+        // not being 0
+        tokenId = _tokenId;
```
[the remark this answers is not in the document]
sparrowDom: good point: https://github.com/OriginProtocol/origin-dollar/pull/2499/commits/de417562faa9b15ba23fd09699570cab8641afed

sparrowDom:
## Requirements
This is a Vault Strategy contract that is designed to deposit funds into the underlying Rooster (Maverick) pool. The strategy is capable of depositing / withdrawing and rebalancing (moving the current trading price) the pool. It can harvest the incentive rewards (presented in Rooster token) and is reporting a `checkBalance` which is not affected by the price action in the underlying pool. 

#### Authentication

- [x] Never use tx.origin

#### Ethereum

#### Cryptographic code

#### Gas problems
- no loops

#### Black magic

#### Overflow

- [ ] ~~All for loops use uint256~~no loops

#### License
- [x] The contract uses the appropriate limited BUSL-1.1 (Business) or the open MIT license
- [x] If the contract license changes from MIT to BUSL-1.1 any contracts importing it need to also have their license set to BUSL-1.1

#### Proxy

#### Events

## Medium Checks

#### Rounding and casts

#### Dependencies
- [ ] ~~If OpenZeppelin ACL roles are use review & enumerate all of them.~~ no acl roles

#### External calls

- [x] No malicious behaviors
- [x] Oracles, one of:
  - [x] No oracles

#### Tests

#### Deploy

## Strategy Specific
#### Strategy checks

- [x] WithdrawAll() does not revert when strategy has 0 assets
- [ ] 🟡  Strategist cannot steal funds
  - Strategist *can* steal funds, but is limited to 0.2% due to `_solvencyAssert`

#### Downstream

- [ ] 🟡  We have monitoring on all backend protocol's governances -> being setup by Chris
- [ ] 🟡 We have monitoring on a pauses in all downstream systems -> not yet

## Thinking

#### Logic

The logic correctly deposits and withdraws the capital from the underlying concentrated liquidity pool

#### Internal State

- after the deployment governance is executed `tokenId` should never be 0
- `underlyingAssets` should always be greater than the cumulative amount of WETH deposited to the strategy
- unless donated the strategy shouldn't hold more than 1e12 of WETH
- on withdrawal the strategy will receive either WETH, OETH, or a MIX of them. WETH is retained, OETH is burned, reducing the debt. The balance of WETH and debt reduction of burning OETH shouldn't result in more than 1bp difference to what checkBalance reported while the funds have still been deployed. 

#### Attack

- the strategist could using continuous disadvantaged rebalancing cause loss of funds. The total loss is limited to 0.2% since solvency checks are performed on any liquidity action.
- strategy can have WETH & OETH donated which would immediately count towards inflated `checkBalance`. Recent Vault changes which limit maximum rebase & rebase rate increase mitigates this problem. 

#### Flavor
The strategy code seems quite complex due to the way it integrates with the Maverick pool. It doesn't look clean to me, though I don't think further simplifications are possible.

Also see [liquidity simulation tests](https://www.notion.so/originprotocol/Internal-Code-Review-20884d46f53c8039b8d8cce928176672).

contracts/contracts/strategies/plume/RoosterAMOStrategy.sol
```diff
@@ -0,0 +1,1177 @@
...
+        address _maverickPosition,
+        address _maverickQuoter,
+        address _mPool,
+        bool _upperTickAtParity,
+        address _votingDistributor,
+        address _poolDistributor
+    ) initializer InitializableAbstractStrategy(_stratConfig) {
+        require(
+            address(IMaverickV2Pool(_mPool).tokenA()) == _wethAddress,
+            "WETH not TokanA"
```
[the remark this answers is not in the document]
sparrowDom: thanks: https://github.com/OriginProtocol/origin-dollar/pull/2499/commits/dc6a0b0aa6b7e08b250aac729d0ff6b89f734008

DanielVF (approved):
## Requirements

This strategy is intended to bring Origin’s AMO magic to the Rooster/Maverick AMM on Plume. It should behave like our aerodrome AMO strategy.

#### Authentication

#### Ethereum

#### Cryptographic code

_No Crypto_

#### Gas problems

_No Loops_

#### Black magic

_No Magic_

#### Overflow

#### License
- [x] The contract uses the appropriate limited BUSL-1.1 (Business) or the open MIT license

#### Proxy

#### Events

#### Rounding and casts
Rounding is less important in this code because we just need to make the AMM happy and to give users exactly what they are owed.

But beyond that, any rounding errors just end up as funds sitting on the contract, correctly accounted for and eventually cleaned up and swept back.

Because we check expected WETH share both before and after adding liquidity you can't put the pool into a position where it looks safe to add to because of rounding error, but would then end up outside bounds.

#### Dependencies

#### External calls

#### Deploy

## Strategy Specific

#### Strategy checks

- [ ] WithdrawAll() does not revert when strategy has 0 assets
	- [x] Limited

#### Downstream

Downstream has no pauses

#### Logic

_Looks correct_

Pool already deployed.

#### Internal State

When at rest, underlyingAssets should always be less than or equal to the amount of WETH+OETH withdrawable out of the pool by the strategy.

#### Attack

This code has high impact if there were bugs.

- Holds ETH on behalf of the users.
- Can print unlimited OTokens
- Controls a balance used to drive token rebasing.

To prevent the theft of WETH, WETH can only leave the contract as directed by the Vault WETH, or deposited to the Pool.

OTokens are only sent to the pool or burned.

The primary risk, therefore, is some form of incorrect accounting, or incorrect assumptions, which would allow either value to be taken from the pool, or bad accounting resulting in incorrect rebases.

We have fuzz tested our accounting to a limited degree, as well as range simulated it with python simulations.

clement-ux (commented):
Some small gas comments, feel free to ignore them.

clement-ux (commented):
## Requirements

## Easy Checks

#### Authentication

- [ ] Never use tx.origin
- [ ] Every external/public function is supposed to be externally accessible
- [ ] Every external/public function has the correct authentication

#### Ethereum

- [ ] Contract is not vulnerable to being sent self destruct ETH

#### Cryptographic code

- [ ] This contract code does not roll it's own crypto.
- [ ] No signature checks without reverting on a 0x00 result.
- [ ] No signed data could be used in a replay attack, on our contract or others.

#### Gas problems

    - [ ] A way to remove items

#### Black magic

- [ ] Does not contain `selfdestruct`
- [ ] Does not use `delegatecall` outside of proxying. _If an implementation contract were to call delegatecall under attacker control, it could call selfdestruct the implementation contract, leading to calls through the proxy silently succeeding, even though they were failing._
- [ ] Address.isContract should be treated as if could return anything at any time, because that's reality.

#### Overflow

- [ ] Code is solidity version >= 0.8.0

#### License
- [ ] The contract uses the appropriate limited BUSL-1.1 (Business) or the open MIT license
- [ ] If the contract license changes from MIT to BUSL-1.1 any contracts importing it need to also have their license set to BUSL-1.1

#### Proxy

- [ ] No storage variable initialized at definition when contract used as a proxy implementation.

#### Events

## Medium Checks

#### Rounding and casts

#### Dependencies
- [ ] Review any new contract dependencies thoroughly (e.g. OpenZeppelin imports) when new dependencies are added or version of dependencies changes.

#### External calls

- [ ] No malicious behaviors
- [ ] Oracles, one of:
  - [ ] No oracles
  - [ ] Oracles can't be bent

#### Tests

#### Deploy

## Strategy Specific

#### Strategy checks

- [ ] WithdrawAll() does not revert when strategy has 0 assets

#### Downstream

## Thinking

#### Logic

#### Internal State

Does this code do that?

#### Attack

#### Flavor

_Could this code be simpler?_

clement-ux (dismissed):

PR #2533, opened 2025-06-16
Safe Module for bridging

shahthepro:
## Code Change

- Adds Safe modules for Ethereum and Plume that help in bridging WETH and wOETH

### TODO
- Add tests for deposit and redeem methods (it's hard to test bridge methods, so it's okay as long as it doesn't revert and we can check on Tenderly simulations when we use them first)


contracts/contracts/automation/PlumeBridgeHelperModule.sol
```diff
@@ -0,0 +1,291 @@
...
+     * @param slippageBps Slippage in 10^4 basis points.
+     * @return Amount of WETH received.
+     */
+    function depositWOETHAndBridgeWETH(uint256 woethAmount, uint256 slippageBps)
+        external
+        onlyOperator
+        returns (uint256)
+    {
+        uint256 wethAmount = _depositWOETH(woethAmount, true);
+        _bridgeTokenWithLz(
```
sparrowDom, line 113: you could change the `bridgeWETHToEthereum` to a public function and replace this bit by just calling: 
`bridgeWETHToEthereum(wethAmount, slippageBps)`

contracts/contracts/automation/EthereumBridgeHelperModule.sol
```diff
@@ -0,0 +1,257 @@
...
+     * @param wethAmount Amount of WETH to mint.
+     * @param slippageBps Bridge slippage in 10^4 basis points.
+     */
+    function mintWrapAndBridgeToPlume(uint256 wethAmount, uint256 slippageBps)
+        external
+        payable
+        onlyOperator
+    {
+        uint256 woethAmount = _mintAndWrap(wethAmount);
+        _bridgeTokenWithLz(
```
sparrowDom, line 173: similarly to other comments you could make `bridgeWOETHToPlume` public

contracts/contracts/automation/EthereumBridgeHelperModule.sol
```diff
@@ -0,0 +1,257 @@
...
+     * @param slippageBps Bridge slippage in 10^4 basis points.
+     */
+    function unwrapRedeemAndBridgeToPlume(
+        uint256 woethAmount,
+        uint256 slippageBps
+    ) external onlyOperator {
+        uint256 wethAmount = _unwrapAndRedeem(woethAmount);
+        // Unwrap into ETH
+        weth.withdraw(wethAmount);
+        _bridgeTokenWithLz(
```
sparrowDom, line 248: and here use a public `bridgeWETHToPlume`

contracts/contracts/automation/AbstractLZBridgeHelperModule.sol
```diff
@@ -0,0 +1,138 @@
...
+            to: bytes32(uint256(uint160(address(safeContract)))),
+            amountLD: amount,
+            minAmountLD: minAmount,
+            extraOptions: options,
+            composeMsg: bytes(""),
+            oftCmd: bytes("")
+        });
+
+        // Compute fees
+        MessagingFee memory msgFee = lzAdapter.quoteSend(sendParam, false);
```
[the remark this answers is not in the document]
shahthepro: We will still be having a verifier on the calls and we are using nativeFee (which should be ETH or PLUME), so any anomalies or huge transfers should raise some eyebrows. I'd suggest just letting it be. But happy to add checks if you think we need it

contracts/contracts/automation/AbstractLZBridgeHelperModule.sol
```diff
@@ -0,0 +1,138 @@
...
+    /**
+     * @dev Bridges token using LayerZero.
+     * @param lzEndpointId LayerZero endpoint id.
+     * @param token Token to bridge.
+     * @param lzAdapter LZ Adapter to use.
+     * @param amount Amount of token to bridge.
+     * @param slippageBps Slippage in 10^4 basis points.
+     * @param isNativeToken Whether the token is native token.
+     */
+    function _bridgeTokenWithLz(
```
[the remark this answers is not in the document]
shahthepro: LayerZero already emits event for these. And we also emit events from all the underlying actions on the OTokens and Vaults. Is there any other benefit of doing that?

contracts/contracts/automation/PlumeBridgeHelperModule.sol
shahthepro: That's a really good suggestion, have cleaned up all instances

sparrowDom (approved):
LGTM

PR #2540, opened 2025-06-20
remove vault calls to the Dripper when doing rebases

sparrowDom:
This PR removes VaultCore calls to the dripper when doing rebases. Also Dripper is no longer set on the Vault on deploys of mocks. 

Current state of Vault, Dripper and Harvester configuration: 
**OUSD:** 
- the harvester is not used anymore: https://etherscan.io/address/0x21Fb5812D70B3396880D30e90D9e5C1202266c89
- the Dripper's `collectAndRebase` is being called by Defender Action
  - no USDT is being dripped
  - on rebase Vault sends OUSD to the multichainBuybackOperator

**OETH:**
- the harvester is not used anymore: https://etherscan.io/address/0x0D017aFA83EAce9F10A8EC5B6E13941664A6785C
- the dripper is not used anymore: https://etherscan.io/address/0xc0F42F73b8f01849a2DD99753524d4ba14317EB3

**OETHb:** 
- the Dripper is configured on the Vault and regularly called by a gnosis safe account: https://basescan.org/address/0x02f2C609950E90934ce99e58b4d7326aD0d7f8d6

**OETHp:**
- the Dripper is configured on the Vault and regularly called by a gnosis safe account: https://explorer.plume.org/address/0x3b56c09543D3068f8488ED34e6F383c3854d2bC1


contracts/contracts/vault/OETHVaultCore.sol
```diff
@@ -269,10 +265,7 @@ contract OETHVaultCore is VaultCore {
             withdrawalRequests[_requestId].queued >
             withdrawalQueueMetadata.claimable
         ) {
-            // Stream any harvested rewards (WETH) that are available to the Vault
-            IDripper(dripper).collect();
-
-            // Add any WETH from the Dripper to the withdrawal queue
+            // Add any WETH to the withdrawal queue
             _addWithdrawalQueueLiquidity();
```
naddison36, line 269: `_addWithdrawalQueueLiquidity` can also be deleted since we aren't getting any more WETH. 
I think the entire if statement to check for more liquidity can be removed.

contracts/contracts/vault/OETHVaultCore.sol
```diff
@@ -305,13 +298,7 @@ contract OETHVaultCore is VaultCore {
...
     {
-        // Just call the Dripper instead of looping through _requestIds to find the highest id
-        // and checking it's queued amount is > the queue's claimable amount.
-
-        // Stream any harvested rewards (WETH) that are available to the Vault
-        IDripper(dripper).collect();
-
-        // Add any WETH from the Dripper to the withdrawal queue
+        // Add any WETH to the withdrawal queue
         _addWithdrawalQueueLiquidity();
```
naddison36, line 302: `_addWithdrawalQueueLiquidity` can also be deleted

contracts/contracts/vault/OETHVaultCore.sol
shahthepro: We aren't getting any WETH from external Dripper but the Vault might still have unrealized yields in the form of WETH, so that can be accounted towards withdrawal queue shortfall. So, I think we can leave this line (`_addWithdrawalQueueLiquidity`) as it is.

Let's say we have 500 WETH shortfall. And we fund the Vault with 500 WETH. Unless someone explicitly calls the external `addWithdrawalQueueLiquidity()` method, the `_claimWithdrawal` internal method will revert since the metadata isn't updated. This isn't a big deal since we can always call the public method whenever we fund the Vault, but it'd be a minor inconvenience if we remove this

shahthepro (approved):
LGTM

contracts/contracts/vault/OETHVaultCore.sol
shahthepro: > Automatically calling _addWithdrawalQueueLiquidity might prevent us from depositing the desired amount into the strategy B

That makes sense. I don't feel strongly about any of our options. So, I'm fine with whatever you guys decide

naddison36 (dismissed):
LGTM

PR #2571, opened 2025-07-04
Add support for general requests from execution to consensus layer

sparrowDom:
- replaces the Execution layer contracts that are responsible for issuing EL -> CL communication
- adds support to verify amounts these contracts are called with 
- the `enableExecutionLayerGeneralPurposeRequests` can be removed and the requests will still succeed. Though the tests won't be able to verify the values
- the original problem why tests were failing is that mock contracts (e.g. `MockBeaconConsolidation` didn't have 1 wei of ETH to pay for the fee)


naddison36 (approved):
This is very good. Thanks

PR #2602, opened 2025-07-28
add initial files to separate consolidation into a separate strategy

sparrowDom:
This PR: 
- separates consolidations into a separate intermediate strategy
- verifies the target validator balance instead for balances of the source validators
- process: 
  - create SSV validator on the Consolidation strategy, stake 32 ETH to it, verify validator, verify deposit
  - start consolidation on the old Native Staking strategy 
  - verify the consolidation on the Consolidation strategy
  - Consolidation strategy hands off the validator to the Compounding SSV strategy

### DIFF
--- a/contracts/contracts/strategies/NativeStaking/ValidatorRegistrator.sol
+++ b/contracts/contracts/strategies/NativeStaking/ValidatorRegistrator.sol
@@ -3,7 +3,7 @@ pragma solidity ^0.8.0;
 
 import { Pausable } from "@openzeppelin/contracts/security/Pausable.sol";
 import { Governable } from "../../governance/Governable.sol";
-import { IConsolidationTarget } from "../../interfaces/IConsolidation.sol";
+import { IConsolidationStrategy, IConsolidationSource } from "../../interfaces/IConsolidation.sol";
 import { IDepositContract } from "../../interfaces/IDepositContract.sol";
 import { IVault } from "../../interfaces/IVault.sol";
 import { IWETH9 } from "../../interfaces/IWETH9.sol";
@@ -21,7 +21,7 @@ struct ValidatorStakeData {
  * @notice This contract implements all the required functionality to register, exit and remove validators.
  * @author Origin Protocol Inc
  */
-abstract contract ValidatorRegistrator is Governable, Pausable {
+abstract contract ValidatorRegistrator is Governable, Pausable, IConsolidationSource {
     /// @notice The maximum amount of ETH that can be staked by a validator
     /// @dev this can change in the future with EIP-7251, Increase the MAX_EFFECTIVE_BALANCE
     uint256 public constant FULL_STAKE = 32 ether;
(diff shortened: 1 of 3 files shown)


contracts/contracts/strategies/NativeStaking/ValidatorRegistrator.sol
```diff
@@ -56,9 +56,7 @@ abstract contract ValidatorRegistrator is Governable, Pausable {
 
     /// @notice Number of validators currently being consolidated
     uint256 public consolidationCount;
-    address public consolidationTargetStrategy;
-    /// @notice Mapping of support target staking strategies that can be used for consolidation
-    mapping(address => bool) public consolidationTargetStrategies;
+    address public intermediateConsolidationStrategy;
 
     // For future use
     uint256[44] private __gap;
```
naddison36, line 62: The gap can be changed to 45 since we are removing one of the slots added

contracts/contracts/strategies/NativeStaking/CompoundingValidatorManager.sol
```diff
@@ -125,14 +124,18 @@ abstract contract CompoundingValidatorManager is Governable, Pausable {
         EXITED, // The validator has been verified to have a zero balance
         REMOVED // validator has funds withdrawn to the EigenPod and is removed from the SSV
     }
+
+    // For future use
+    uint256[49] private __gap;
```
naddison36, line 129: this is a new contract so I've started the gap at 50

contracts/contracts/strategies/NativeStaking/CompoundingValidatorManager.sol
```diff
@@ -109,13 +109,12 @@ abstract contract CompoundingValidatorManager is Governable, Pausable {
...
-    bytes32 public consolidationLastPubKeyHash;
-    address public consolidationSourceStrategy;
-    mapping(address => bool) public consolidationSourceStrategies;
-
-    // For future use
-    uint256[50] private __gap;
+    /// @notice Mapping of the validator hashed keys to balances at last verification.
+    /// Once all the Validators have been consolidated from the old strategy contract this field can be
+    /// deprecated
+    mapping(bytes32 => uint256) public validatorBalances;
```
naddison36, line 117: Could this also live in the ConsolidateValidator contract?

contracts/contracts/strategies/NativeStaking/CompoundingValidatorManager.sol
```diff
@@ -796,6 +763,9 @@ abstract contract CompoundingValidatorManager is Governable, Pausable {
                         IBeaconProofs.BalanceProofLevel.Container
                     );
 
+                // deprecate once we have consolidated all of the validators
```
naddison36, line 766: I've prefer this in the `ConsolidateValidator` contract

contracts/contracts/strategies/NativeStaking/CompoundingValidatorManager.sol
sparrowDom: In some contracts we did it in a way where even at contract creation the slots + the gap would make up the number 50. This way each element in inheritance would always take up multiples of 50  or a 100. I personally prefer it that way. What do you think?

contracts/contracts/strategies/NativeStaking/CompoundingValidatorManager.sol
sparrowDom: indeed

PR #2610, opened 2025-08-04
Accounting explanation

sparrowDom:
add an in depth explanation of the accounting behaviour


naddison36 (approved):
Excellent work. Thanks for this

PR #2613, opened 2025-08-05
Add manual gasLimit for Morpho harvest script

shahthepro:
- gasEstimate seems to fail for the module at times. This sets a hardcoded gasLimit that should be higher than usual


clement-ux (approved):
LGTM 👍

PR #2635, opened 2025-08-26
various minor changes to SSV staking

sparrowDom:
- adds/corrects comments in various place
- adds CEI pattern to `removeSsvValidator`
- reduces contract size by 0.07 kB with [this commit](https://github.com/OriginProtocol/origin-dollar/commit/336477e1c85340c3d15aec504e82260a89efe046)


contracts/contracts/strategies/NativeStaking/CompoundingValidatorManager.sol
```diff
@@ -553,12 +575,16 @@ abstract contract CompoundingValidatorManager is Governable {
             withdrawalAddress
         );
 
+        // Store the validator state as verified
+        validator[pubKeyHash] = ValidatorData({
```
naddison36, line 579: why move this to before verifying the validator? It will be invalid if the withdrawal address does not match the strategy's address. Or is this to set the validatorIndex?

PR #2641, opened 2025-08-29
[SP - 2] multiple full validator exits

sparrowDom:
**Sigma Prime report:** 
validatorWithdrawal() should not mark the validator as EXITING, as the withdrawal request could be rejected on the beacon chain. Full withdrawal requests are rejected when:
 - the validator has not been activated (activation_epoch > current epoch)
 - the validator has not been active long enough (current_epoch < activation_epoch + SHARD_COMMITTEE_PERIOD)
 - the validator has pending balance to withdraw from a previous partial withdrawal request

**Solution:** allow for full/partial withdrawals even once the validator has already been marked as fully exited. This allows for subsequent exit requests even if previous full exits have failed


naddison36 (approved):
LGTM

PR #2642, opened 2025-08-29
[SP - 1] change validator index from uint64 to uint40

sparrowDom:
**Sigma prime report:**
As mentioned in the kickoff call, validatorIndex should be uint40 instead of uint64. Providing a validatorIndex that is out-of-bounds can contaminate the generalizedIndex and hence the leaf that is being proven. We looked deep into this and found that a 64-bit validatorIndex allows an attacker to completely change the branch selection of the entire proof (from any Validator container field or validator balance leaf all the way to the beacon block root) due to how concatGenIndices() works. The attack surface is quite limited though since you can only flip 0 bits to 1. A potential attack was to divert the proof through historical summaries and try to reach BeaconBlockBody.graffiti, but that proof would require 55 witness hashes, and verifyValidator() limited us to 53. In the end, we couldn't find a valid exploit path for this vulnerability, though we still recommend fixing it as it affects the index that's stored onchain and emitted in events. A general way to prevent contaminating the generalized index in concatGenIndices() is to do input validation by checking that indexBits <= height

**Solution:** done as suggested

### DIFF
--- a/contracts/contracts/beacon/BeaconProofs.sol
+++ b/contracts/contracts/beacon/BeaconProofs.sol
@@ -22,7 +22,7 @@ contract BeaconProofs is IBeaconProofs {
         bytes32 beaconBlockRoot,
         bytes32 pubKeyHash,
         bytes calldata proof,
-        uint64 validatorIndex,
+        uint40 validatorIndex,
         address withdrawalAddress
     ) external view {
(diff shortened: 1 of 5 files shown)


naddison36 (approved):
LGTM

PR #2644, opened 2025-08-29
[SP - 4] Verify validator type

sparrowDom:
**Sigma Prime report:**
BeaconProofsLib.verifyValidator() does not check that the withdrawal credentials are 0x01 or 0x02. Through a validator frontrunning attack, the registrator can change the withdrawal credentials to any other prefix (e.g. 0x00 or 0x03) and be able to successfully verify the validator. The withdrawal credentials is invalid, so the ETH cannot be withdrawn to the strategy contract. Since the validator can be verified on the strategy contract, firstDeposit does not protect against this attack. Furthermore, it's impossible to remove this validator from the strategy, so it uses up the MAX_VERIFIED_VALIDATORS limit and cannot be removed with removeSsvValidator(), since VERIFIED validators cannot be removed

**Solution:**
Fixed

### DIFF
--- a/contracts/contracts/beacon/BeaconProofsLib.sol
+++ b/contracts/contracts/beacon/BeaconProofsLib.sol
@@ -123,6 +123,7 @@ library BeaconProofsLib {
             withdrawalAddressFromProof == withdrawalAddress,
             "Invalid withdrawal address"
         );
+        require(proof[0] == 0x02, "Invalid validator type");
 
         require(
             // 53 * 32 bytes = 1696 bytes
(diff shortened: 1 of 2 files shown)


contracts/contracts/beacon/BeaconProofsLib.sol
```diff
@@ -123,6 +123,7 @@ library BeaconProofsLib {
             withdrawalAddressFromProof == withdrawalAddress,
             "Invalid withdrawal address"
         );
+        require(proof[0] == 0x02, "Invalid validator type");
```
naddison36, line 126: That ended up being an easy check. Well done

naddison36 (approved):
excellent

contracts/contracts/beacon/BeaconProofsLib.sol
sparrowDom: Do you think we should verify that bytes 1 - 11 are of `0x0` value as well?

PR #2645, opened 2025-09-01
add check for zero padding in withdrawal credentials

sparrowDom:
add check for zero padding in withdrawal credentials


contracts/contracts/beacon/BeaconProofsLib.sol
```diff
@@ -125,6 +125,10 @@ library BeaconProofsLib {
         );
         require(proof[0] == 0x02, "Invalid validator type");
 
+        for (uint256 i = 1; i < 12; i++) {
```
naddison36, line 128: iterating over 12 bytes doesn't seem very efficient. What if we constructed the full bytes32 of the withdrawal credential and compared that?

naddison36: Thanks. We discussed casting to bytes32 rather than taking the hash but casting bytes to bytes32 is not easy. The keccak 256 is a precompiled function so is reasonably efficient.

naddison36 (approved):
Thanks for this

PR #2651, opened 2025-09-03
Fix for: Deposit Balance Undercount Due To Missing Timestamp Validation In verifyDeposit()

sparrowDom:
**Sigma Prime report:**
The CompoundingValidatorManager contract uses a balance verification system where snapBalances() records the
The verifyDeposit() function allows systematic ETH balance undercounting due to missing timestamp validation,
enabling removal of deposits that were processed after the current balance snapshot.
contract’s ETH balance and timestamp, and verifyBalances() calculates the total balance at that timestamp by sum-
ming the contract balance with validator balances and pending deposits. However, the verifyDeposit() function
lacks crucial timestamp validation that prevents deposits processed after from being verified and
removed from the pending deposits list.
This creates a systematic accounting error where:
1. snapBalances() records ETH balance at time T and sets lastSnapTimestamp = T
2. Deposits are processed on the beacon chain after time T but before verifyBalances() is called
3. verifyDeposit() allows these deposits to be verified and removed from totalDepositsWei calculation
4. The removed deposits were not included in validator balances at snapshot time T , causing an undercount
The current verifyDeposit() function only validates that the deposit was processed before the validator creation slot,
but fails to check if the deposit was processed after the balance snapshot:

**Response:** fixed as suggested

### DIFF
--- a/contracts/contracts/strategies/NativeStaking/CompoundingValidatorManager.sol
+++ b/contracts/contracts/strategies/NativeStaking/CompoundingValidatorManager.sol
@@ -705,6 +705,19 @@ abstract contract CompoundingValidatorManager is Governable {
         // The verification slot must be after the deposit's slot.
         // This is needed for when the deposit queue is empty.
         require(deposit.slot < depositProcessedSlot, "Slot not after deposit");
+        uint64 snapTimestamp = snappedBalance.timestamp;
+
+        // This check prevents an accounting error that can happen if:
+        //  - snapBalances are snapped at the time of T
+        //  - deposit is processed on the beacon chain after time T and before verifyBalances()
+        //  - verifyDeposit is called before verifyBalances which removes a deposit from depositList
+        //    and deposit balance from totalDepositsWei
+        //  - verifyBalances is called under-reporting the strategy's balance
+        require(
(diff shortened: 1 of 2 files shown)


contracts/contracts/strategies/NativeStaking/CompoundingValidatorManager.sol
```diff
@@ -705,6 +705,19 @@ abstract contract CompoundingValidatorManager is Governable {
...
+        // This check prevents an accounting error that can happen if:
+        //  - snapBalances are snapped at the time of T
+        //  - deposit is processed on the beacon chain after time T and before verifyBalances()
+        //  - verifyDeposit is called before verifyBalances which removes a deposit from depositList
+        //    and deposit balance from totalDepositsWei
+        //  - verifyBalances is called under-reporting the strategy's balance
+        require(
+            (_calcNextBlockTimestamp(depositProcessedSlot) <= snapTimestamp) ||
+                snapTimestamp == 0,
+            "Deposit processed after last balance snapshot"
```
naddison36, line 719: nit: I like error strings to be 31 characters or less so they fit in one EVM work (32 bytes). 1 byte is used for the string length.

PR #2662, opened 2025-09-10
Restrict full withdrawals for non activateable validators

sparrowDom:
There are at least 2 approaches on how to solve the problem where the strategy contract needs to prevent a malicious or negligent registrator from putting internal validators in a non recoverable state. This can be caused by:
- stake 1 ETH
- verifyValidator
- verifyDeposit
- so full validator withdrawal
That puts the validator into EXITING state meaning no further deposits to it are allowed. And the initial 1 ETH deposit is stuck to that not activated validator.

**Possible approaches**
**Option 1 (Sigma Prime recommended)** Only set the validator state to EXITING after receiving confirmation that the beacon chain has accepted the withdrawal request.
We could implement this by: 
 - Issue a full validator withdrawal on the execution layer to the beacon chain and not change validator states
 - wait for the beacon chain to process the withdrawal
 - call a newly implemented function `verifyFullExit` which would include Merkle Tree proofs that the `exit_epoch` on the validator is not set to a FAR_FUTURE_EPOCH.
 
**Pros:**
- get certain confirmation that the exit has been processed

**Cons:**
- while waiting for the exit to be processed deposits are still allowed to the validator
- more Merkle Tree proofs slightly increases maintenance requirements

**Option 2 (Implemented in this PR)** Add additional field to validator data: `bool canActivate` which is set to true when validator balance once surpasses 32 ETH minimal activation balance. Once that threshold has been surpassed the validator is allowed to be fully withdrawable. If for some other reason the full withdrawal fails, it can be issued again.
**Pros:**
- lesser increase in complexity
- deposits to fully withdrawn validator are immediately blocked

**Cons:**
- not as accurate as approach 1, since the strategy contract makes an optimistic assumption and marks the validator as fully exited even when that is not the case


contracts/contracts/strategies/NativeStaking/CompoundingValidatorManager.sol
```diff
@@ -129,6 +131,7 @@ abstract contract CompoundingValidatorManager is Governable, Pausable {
     struct ValidatorData {
         ValidatorState state; // The state of the validator known to this contract
         uint40 index; // The index of the validator on the beacon chain
+        bool canActivate; // Has validator ever had minimal activation balance
```
naddison36, line 134: Should we add a `ACTIVE` validator state rather than introduce a new state flag flag?

PR #2672, opened 2025-09-17
pin dependancies and replace yarn with pnpm

sparrowDom:
This pins all the repo dependancies and replaces the yarn with pnpm


naddison36 (dismissed):
This works on my pc

PR #2773, opened 2026-01-29
Script to trigger withdraw all from strategies

sparrowDom:
In case of the incident this is a script that triggers withdraw all from the underlying strategies. The current problem is that if one strategy isn't able to withdrawAll (like Morpho strategy) because of liquidity issues, the script will fail.


naddison36 (approved):
I was looking for this on Sunday :-)

PR #2824, opened 2026-03-03
Add script to generate pool booster safe transactions

clement-ux:
This pull request introduces a new script, `generate-pool-booster-safe.sh`, designed to automate the generation of Safe Transaction Builder JSON files for deploying Curve Pool Booster contracts and delegating yield. The script validates user input, verifies gauge registration, computes necessary parameters, and outputs a structured JSON transaction bundle suitable for Safe multisig execution.

The most important changes are:

**Addition of deployment automation script:**

* Added `generate-pool-booster-safe.sh`, a Bash script for generating Safe Transaction Builder JSON to deploy `CurvePoolBoosterPlain` contracts and delegate yield for specified gauges. The script accepts multiple command-line options and validates all required arguments.

**Validation and verification logic:**

* Implements logic to fetch all registered gauges from the on-chain gauge controller and verifies that all provided gauge addresses are registered, ensuring only valid gauges are processed.

**Transaction building and output:**

* Constructs transaction objects for both the pool booster creation and the yield delegation, assembling them into a single JSON output compatible with the Safe Transaction Builder.
* Dynamically computes unique salts for each deployment using the factory address and a timestamp-based counter, and calculates the expected booster contract address for each gauge.

**User experience and safety:**

* Provides clear usage documentation, error messages, and cleans up temporary files on exit, making the script robust and

### DIFF
--- a/contracts/scripts/deploy/generate-pool-booster-safe.sh
+++ b/contracts/scripts/deploy/generate-pool-booster-safe.sh
@@ -1,0 +1,216 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# ── Defaults ──────────────────────────────────────────────────────────────────
+REWARD_TOKEN=""
+FEE_COLLECTOR=""
+FEE=""
+CAMPAIGN_REMOTE_MANAGER=""
+VOTEMARKET=""
+FACTORY=""
+RPC_URL=""
+SAFE=""
+GAUGES=()
+
+# ── Usage ─────────────────────────────────────────────────────────────────────
+usage() {
+  cat <<'EOF'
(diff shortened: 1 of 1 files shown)


naddison36 (approved):
I would have done this using a Hardhat task rather than a bash script for these reasons:
Hard coded addresses rather than reusing addresses in a common file.
Hardhat has CLI infrastructure that helps a lot of the user input validation.
I recently moved Safe file generation out into a js utils file in contracts/utils/safe.js

No need to redo it.  But we should have a discussion about what we want to build CLIs in moving forward. I'm not fixed on Hardhat. A js cli package like commander would work just as well. Foundry scripts are also a good option if there's not much off-chain logic.

PR #2848, opened 2026-03-20
Migrate test suite from Hardhat to Foundry

clement-ux:
## Summary

Migrates the smart contract test and deployment validation workflow from Hardhat to Foundry.

Also adds a fresh OUSD core deployment and hardens Foundry governance simulation and transaction broadcasting for Prague.

## Changes

- Adds Foundry unit, fork, and smoke tests.
- Adds `DeployManager` and `Resolver` to test deployed contracts with pending governance applied.
- Adds a fresh OUSD deployment script for the token, Vault, Wrapped OUSD, and `VaultValueChecker`.
- Uses the Guardian multisig to propose and vote during mainnet governance simulations, since the Timelock has no delegated xOGN voting power.
- Avoids Foundry's EIP-2935 block-history backfill during artificial governance block jumps by temporarily using Cancun execution semantics around `vm.roll`, then restoring the active EVM version.
- Disables call isolation for real deployments with `--no-isolate`, ensuring nonces remain sequential when contract creations and calls are interleaved.
- Ignores local Foundry broadcast artifacts.
- Adds Foundry CI, Makefile targets, environment helpers, and test-generation skills.
- Removes obsolete Hardhat, Echidna, oracle, and documentation files.

## Foundry deployment notes

- On Prague, the current Foundry nightly backfills the EIP-2935 history contract for every block skipped by `vm.roll`. Large governance jumps therefore trigger thousands of RPC block-hash and storage reads. Governance simulation temporarily switches only the roll cheatcode to Cancun and immediately restores Prague; the skipped blocks' synthetic EIP-2935 history is intentionally not populated.
- Foundry isolation produced duplicate broadcast nonces when deployment calls and contract creations were interleaved. Real deployment commands use `--no-isolate` so each captured transaction increments the deployer nonce before the next address is calculated.

## Test plan

- [x] `make test`
- [x] Fork tests
- [x] `make test-smoke`
- [x] Foundry CI
- [x] `forge build scripts/deploy/`
- [x] `make simulate`
- [x] Mainnet test deployment with governance calldata generation


.github/workflows/foundry.yml
```diff
@@ -0,0 +1,247 @@
...
+          submodules: recursive
+      - uses: ./.github/actions/foundry-setup
+      - name: Run unit tests
+        working-directory: contracts
+        run: make test-unit
+
+  # ── Fork Tests ──────────────────────────────────────────────
+  fork-tests-mainnet:
+    name: Fork Tests (Mainnet ${{ matrix.chunk_id }})
+    runs-on: ubuntu-latest
```
shahthepro, line 77: We used to split fork tests into 4 because with the hardhat fixtures were too slow and it took us about ~3 to 5 mins for each of those batches. Looks like each batch now only takes around ~1 min, so we could get rid of the splitting of tests in CI

.gitignore
```diff
@@ -85,11 +87,15 @@ coverage
...
 # Certora #
 .certora_internal
 
+# Foundry / Soldeer
+contracts/dependencies/
+contracts/out/
 # Possible Agent.md file
-AGENTS.md
\ No newline at end of file
+AGENTS.md
```
shahthepro: why is agents.md in .gitignore when it's already committed to the repo?

contracts/package.json
```diff
@@ -4,6 +4,7 @@
   "description": "Origin DeFi Contracts",
   "main": "index.js",
   "scripts": {
+    "check:storage": "node scripts/check-storage-layout.js",
     "deploy": "rm -rf deployments/hardhat && npx hardhat deploy",
     "deploy:mainnet": "VERIFY_CONTRACTS=true npx hardhat deploy --network mainnet --verbose",
     "deploy:arbitrum": "FORK_NETWORK_NAME=arbitrumOne npx hardhat deploy --network arbitrumOne --tags arbitrumOne --verbose",
```
shahthepro, line 10: Should we also remove all these hardhat related functions from package.json?

.github/workflows/foundry.yml
clement-ux: Yes, why not, let me try it.

contracts/package.json
clement-ux: I was planning to remove all the hardhat (test and packages etc) in a second PR. Would you prefer to have it done in the same PR?

.github/workflows/foundry.yml
clement-ux: Applied in this [bdb9dc5](https://github.com/OriginProtocol/origin-dollar/pull/2848/commits/bdb9dc55f84c8fdf073be37b41efa206eb2965c1), this is almost the same time, thanks for the advice!

.gitignore
clement-ux: My bad on this, fixed in this [af8fc19](https://github.com/OriginProtocol/origin-dollar/pull/2848/commits/af8fc1954c05bb182c999c9f03ece426d5348a06).

contracts/package.json
shahthepro: A followup PR sounds good

contracts/build/deployments-1.json
```diff
@@ -0,0 +1,118 @@
+{
```
sparrowDom, line 1: Should we also support deploying on Hoodie? Last deploy was February 2026, so maybe not so important

contracts/scripts/deploy/helpers/GovHelper.sol
```diff
@@ -0,0 +1,362 @@
...
+    ///      2. Active: Voting period open
+    ///      3. Succeeded: Voting passed, ready for queue
+    ///      4. Queued: In timelock, waiting for execution delay
+    ///      5. Executed: Proposal actions have been executed
+    ///
+    /// @param log Whether logging is enabled
+    /// @param prop The proposal to simulate
+    function simulate(bool log, GovProposal memory prop) internal {
+        // ===== Setup: Label addresses for trace readability =====
+        address govMultisig = Mainnet.Timelock;
```
sparrowDom: This should probably be 5/8 multisig and not Timelock

sparrowDom (dismissed):
A couple of comments inline otherwise LGTM

contracts/scripts/deploy/helpers/GovHelper.sol
clement-ux: Yep, indeed, I faced the issue while testing deployment, good catch. It should be fixed with this commit: [4f18675](https://github.com/OriginProtocol/origin-dollar/pull/2848/commits/4f1867554d1b42e34d58d75cb5d1e43fe8d2439c)

contracts/build/deployments-1.json
clement-ux: It is easy to add a new chain, let's keep it like this and add the new chain when we need it.

shahthepro (approved):

PR #2871, opened 2026-04-05
Added consol option to withdrawValidator, exitValidators and stakeValidator Hardhat tasks

naddison36:
## Changes

* Added `consol` option to the `withdrawValidator`, `exitValidators` and `stakeValidator` Hardhat tasks so it can call via the `ConsolidationController`
* Changed `getValidators` Hardhat task to use GET validator state API


sparrowDom (approved):
LGTM

PR #2876, opened 2026-04-13
Add shared Postgres nonce queue for cross-repo transaction serialization

sparrowDom:
Both `origin-dollar` and `arm-oeth` cron containers use the same KMS signer, risking nonce collisions when transactions are sent concurrently.

Functionality:
- This adds postgres backed nonce lock, so different processes don't have nonce race conditions
- Timeout mechanism added and configured via `NONCE_QUEUE_LOCK_TIMEOUT_MS`. If process is waiting for longer than this configuration the process exits
- add per chain gas price limit
- functionality to repeat failed brodcasts is added
- functionality to re-broadcast a transaction with a higher gas price is added 

- nonceQueue.ts: withNonceLock + wrapWithNonceQueue signer wrapper
- action.ts: conditionally wraps signer when DATABASE_URL is set
- docker-compose.yml: adds postgres service for local dev
- cron-supervisor.ts: pool cleanup on shutdown
- Includes integration tests verifying lock serialization


contracts/tasks/lib/nonceQueue.test.ts
```diff
@@ -0,0 +1,230 @@
...
+    provider: {
+      getNetwork: async () => ({ chainId: 1 }),
+    },
+    sendTransaction: async (tx: any) => {
+      await onSend(tx.nonce);
+      return {
+        hash: "0x" + Math.random().toString(16).slice(2),
+        from: address,
+        nonce: tx.nonce,
+        wait: async () => ({ status: 1 }),
```
sparrowDom, line 47: nit: might be cool to wait a random 200-300ms to somewhat simulate waiting for the transaction to be mined

contracts/tasks/lib/nonceQueue.ts
```diff
@@ -0,0 +1,192 @@
...
+export function wrapWithNonceQueue(
+  signer: ethers.Signer,
+  chainId: number
+): ethers.Signer {
+  const p = getNoncePool();
+  if (!p) return signer;
+
+  const originalSendTransaction = signer.sendTransaction.bind(signer);
+  const addressPromise = signer.getAddress().then((a) => a.toLowerCase());
+  const getOnChainNonce = () => signer.getTransactionCount("pending");
```
sparrowDom, line 199: I am not sure this is correct. If I am not mistaking the nonce is here fetched from the provider? I think it would be better if the database would be the one supplying the nonce to the transaction being signed.

sparrowDom: Nevermind my comment. What you've done here is correct. The nonce is fetched from the provider only if one with the same chainId & signerAddress doesn't already exist in the db

PR #2877, opened 2026-04-13
Simulate hypothetical balance on the rebalancer

shahthepro:
## Code Change 
- Add support to simulate hypothetical balance changes to see how the rebalancer would respond
- Try out step-wise filling algorithm to maximize APY
- Add support to write to JSON file

```bash
# What if the vault had $500K more idle USDC?
npx hardhat planRebalance --sim-vault 500000

# What if Base Morpho had $200K more?
npx hardhat planRebalance --sim-base 200000

# Combine: vault +$500K, Ethereum Morpho -$100K
npx hardhat planRebalance --sim-vault 500000 --sim-eth -100000

# Write state to a JSON file
npx hardhat planRebalance --json rebalancer-targets.json
```

### DIFF
--- a/contracts/utils/rebalancer.js
+++ b/contracts/utils/rebalancer.js
@@ -308,6 +308,223 @@ function _enforceMinimums(
 }
 
 /**
+ * Compute APY for a strategy at a given delta from its current on-chain balance.
+ * Positive delta = deposit, negative = withdrawal, zero = current APY.
+ *
+ * @param {object} strategy - strategy config with morphoChainId, metaMorphoVaultAddress
+ * @param {BigNumber} delta - cumulative change from current balance
+ * @param {object} currentApys - { metaMorphoVaultAddress: apy }
+ * @returns {Promise<number>} post-action APY
+ */
+async function _computeApyAtDelta(strategy, delta, currentApys) {
+  if (delta.isZero()) {
+    return currentApys[strategy.metaMorphoVaultAddress] || 0;
+  }
+  const rpcUrl = getRpcUrl(strategy.morphoChainId);
+  if (!rpcUrl) {
+    return currentApys[strategy.metaMorphoVaultAddress] || 0;
+  }
+  try {
+    const result = delta.gt(0)
+      ? await computeDepositImpactRpc(
+          rpcUrl,
+          strategy.morphoChainId,
+          strategy.metaMorphoVaultAddress,
+          delta.toBigInt()
+        )
+      : await computeWithdrawalImpactRpc(
+          rpcUrl,
+          strategy.morphoChainId,
+          strategy.metaMorphoVaultAddress,
+          delta.abs().toBigInt()
+        );
+    return result.newApy;
+  } catch (err) {
+    log(`APY computation failed for ${strategy.name}: ${err.message}`);
+    return currentApys[strategy.metaMorphoVaultAddress] || 0;
+  }
+}
+
+/**
+ * Step-wise marginal APY allocation: allocate capital in chunks, always to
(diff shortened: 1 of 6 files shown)


contracts/utils/rebalancer.js
```diff
@@ -1628,7 +1890,7 @@ async function buildRebalancePlan() {
     state.shortfall,
     state.vaultBalance,
     {},
```
clement-ux, line 1723: Could this be dropping data that `_resolveDeposit()` still needs? `buildRebalancePlan()` now passes an empty `depositCapacities` object into `buildExecutableActions()`, but `_resolveDeposit()` still seems to use `capacity?.postDepositApy` for the post-impact spread check on withdrawal-funded deposits. If `capacity` is missing here, are we effectively skipping that guard in production?

contracts/utils/rebalancer.js
```diff
@@ -1538,10 +1768,39 @@ async function fetchMaxWithdrawals(strategies) {
...
 
+  // Apply simulation overrides (balance adjustments in whole-dollar USDC)
+  if (simulation) {
+    const parts = [];
+    if (simulation.vault != null) {
+      const delta = parseUnits(
+        Math.abs(simulation.vault).toString(),
+        USDC_DECIMALS
+      );
+      state.vaultBalance =
```
clement-ux, line 1614: Should the simulation be adjusting `shortfall` before `vaultBalance` here? I think `state.vaultBalance` already represents vault cash available after reserving for the withdrawal queue. If `state.shortfall` is nonzero, could this end up overstating available cash on positive simulations, or mis-modeling removals on negative ones?

contracts/utils/rebalancer.js
shahthepro: This is a good catch. Yes, it kinda overrides two of the checks we do in the _resolveDeposit function:
- Capping deposit amount to capacity.maxDeposit to stay under a desired APY impact
- postDepositApy check

The new step-wise algorithm kinda handles both in `computeImpactAwareAllocation`. That method wouldn't recommend depositing into a strategy if the APY impact is too high.

The primary reason I am trying this new algorithm is that I saw HyperEVM over 8% but a blind 0.5% max apy impact check kinda caps the max deposit to a smaller amount so that the APY can stay at 7.5%. However, even if we earn 5% with a higher deposit, that'll be much higher than what we would earn on Ethereum. So, this new alg will be okay even if the APY impact is greater than 0.5% as long as it still is more profitable than other strategies. Since this new alg does try the deposits in steps (of 50k), it figures out the maxDeposit amount by itself (i.e. it stops any deposits if withdrawal APY impact will get pushed too high or above other strategies or if OETH/wstETH market constraints aren't met for Ethereum Morpho Vault).

So, both the checks that use the `capacity` as effectively replaced witht his one. I'll remove the dead code

shahthepro: Removed those dead checks and code in: https://github.com/OriginProtocol/origin-dollar/pull/2877/commits/16d8b46a6debe688b7a198376813c67a944a24f7

contracts/utils/rebalancer.js
shahthepro: > Should the simulation be adjusting shortfall before vaultBalance here? 

It already does in `readOnChainState`: https://github.com/OriginProtocol/origin-dollar/blob/shah/rebalancer-tweaks/contracts/utils/rebalancer.js#L144-L149

Also, the `simulation` exists only on CLI for us to test the "what if" cases. So we might wanna test with a Vault shortfall or surplus

clement-ux (dismissed):

clement-ux (approved):

PR #2878, opened 2026-04-14
Include domain in the key for crosschain relay

shahthepro:
Include CCTP Destination domain ID in the key stored so that there's no overlap between txs across chains


sparrowDom (approved):
LGTM

PR #2881, opened 2026-04-22
fix: decode BalancesSnapped event via strategy interface in snapBalances task

sparrowDom:
## Summary
- When the `snapBalances` hardhat task runs with `--consol true`, the tx is routed through `ConsolidationController`, but the `BalancesSnapped` event is emitted by the target `CompoundingStakingSSVStrategy`. Ethers only decodes `receipt.events[*].event` against the ABI of the contract used to send the tx, so the event lookup failed and the task threw even though the on-chain call succeeded.
- Fix locates the raw log by topic on the strategy's address and parses it through the strategy's interface. No contract changes required.


naddison36 (approved):
LGTM

PR #2885, opened 2026-05-04
Allow initial deposits to activate validators

naddison36:
## Changes

This updates the Compounding Staking SSV Strategy so new validators can be activated with a larger initial deposit while the Ethereum validator deposit queue is long.

The governance proposal deploys a new `CompoundingStakingSSVStrategy` implementation, upgrades the existing proxy, and sets `initialDepositAmountWei` to 32.25 ETH. This allows a validator to become active from a single deposit instead of requiring an initial 1 ETH deposit followed by a second top-up after the first deposit is processed.

The strategy still caps the first deposit amount, but now allows the first deposit to be less than or equal to the stored `initialDepositAmountWei`. This keeps the 32.25 ETH cap while preserving flexibility for smaller initial deposits.

Once the validator deposit queue returns to a more reasonable length, for example under 10 days, governance can set the initial deposit amount back to 1 ETH.

## Code Change Checklist

- [ ]  Unit tests *if needed

Internal review:


contracts/test/strategies/compoundingSSVStaking.js
```diff
@@ -768,11 +822,11 @@ describe("Unit test: Compounding SSV Staking Strategy", function () {
       ).to.equal(2, "Validator state not 2 (STAKED)");
     };
 
-    it("Should stake to a validator: 1 ETH", async () => {
-      await stakeValidators(0, 1);
+    it("Should stake to a validator: 32.25 ETH", async () => {
+      await stakeValidators(0, INITIAL_DEPOSIT_AMOUNT);
```
sparrowDom, line 823: Multiple places in the test files comments state that 32.25 ETH has been deposited while using a constant `INITIAL_DEPOSIT_AMOUNT` which is set to 1 ETH.

sparrowDom (commented):
Looks good. There is one thing in the tests where comments don't align with code. Also I think it would be cool, that both compounding staking strategy contract setups are tested:  when initial deposit is configured to 1 ETH / 32.24 ETH.

contracts/test/strategies/compoundingSSVStaking.js
naddison36: Good spot. I've fixed with commit https://github.com/OriginProtocol/origin-dollar/pull/2885/changes/e2fbec612fde09a22da51ac3cabedd831b40f506

sparrowDom:
I think we should have a couple of tests where we try to deposit `32.24` ETH to confirm this new functionality

sparrowDom (approved):
LGTM

PR #2889, opened 2026-05-06
Add permissioned rebase

shahthepro:
## Code Changes
- Add permissioned rebase
- Remove rebaseThreshold and auto rebase on large mints and redeems


sparrowDom (approved):
LGTM

clement-ux (commented):
Changes look good.
We need to ensure that no integrator call `rebaseThreshold`, otherwise we can just create view function that return 0? 
I need to spend more time thinking about the implication of removing `_rebase()` for large mint.

sparrowDom:
Here are the OUSD and OETH Dune queries looking at which addresses are calling `rebase`: 
- https://dune.com/queries/7444815
- https://dune.com/queries/7445293

We are ok making rebase non public

naddison36 (commented):
One of the reasons `mint` did a `rebase` was to prevent large mints getting yield from when they were not a oToken holder. For example, if `rebase` is executed once a day and a new mint doubles the oTokens just before `rebase` is called, the new oToken holder will get half the yield of the rebase even though they have only just minted. The impact of this is pretty limited as the minter can not use a flash loan due to async withdrawals. They have to have the capital to mint, hold for the rebase and then redeem post rebase.

Assumptions: 27.61k OETH existing supply, 2.42% net OETH APY, one rebase per day, no gas/slippage/opportunity cost.

| Mint size | Captured yield / day | One-shot ROI | one opportunity/day APY |
|---:|---:|---:|---:|
| 1 WETH | 0.000066 OETH | 0.0066% | 2.42% |
| 10 WETH | 0.000662 OETH | 0.0066% | 2.42% |
| 100 WETH | 0.006598 OETH | 0.0066% | 2.41% |
| 1,000 WETH | 0.0639 OETH | 0.0064% | 2.36% |
| 10,000 WETH | 0.4874 OETH | 0.0049% | 1.79% |
| 27,610 WETH | 0.9149 OETH | 0.0033% | 1.22% |

For small mints, the attack APY is basically the normal OETH APY because the attacker barely dilutes the pool. The absolute profit is tiny though: a 1 WETH attacker captures only about **0.000066 OETH/day** before gas.
Larger mints capture more absolute yield, but that yield is taken from existing rebasing oToken holders.

Note the capital is only used for the above for just over 10 minutes. The capital can be redeployed for the rest of the day earning more yield. eg deposit into a lending market.

Something that mitigate the above is `allocate` being called on larger mints. For example, mints over 10 OETH will allocate the WETH to the default Compounding Staking Strategy. There is currently no way for users to move that liquidity back to the Vault so they can claim a redeem. That means the capital will be locked up for a lot longer than just over 10 minutes. This effectively kills the incentive to mint just before rebase to capture yield.

naddison36 (approved):
I approve the changes

clement-ux:
- I couldn't find any situation where someone can profit from minting large amount with rebase disable.
- With a Dune query, I checked that nobody call `rebaseThreshold()`, so it doesn't break integration to remove it.  
LGTM 👍

clement-ux (approved):

pandadefi:
## VaultStorage.sol

- `rebaseThreshold` deprecated as `__deprecatedRebaseThreshold`
- `operatorAddr` added
- `__gap` reduced by one to `41`
- ⚠️ `rebaseThreshold` is removed and might break contracts integrating with it.

## VaultAdmin.sol

- `setRebaseThreshold(uint256)` removed
- `setOperatorAddr(address)` added as governor-only admin action

## VaultCore.sol

- Mints no longer auto-call `_rebase()`
- Withdrawal no longer auto-call `_rebase()`
- Unauthorised `rebase()` calls now revert with `Caller not authorized`

## VaultInitializer.sol

- `rebaseThreshold = 1000e18` removed
- `operatorAddr` is not initialized

pandadefi (approved):

PR #2927, opened 2026-06-16
Add Credit Market AMO Strategy

shahthepro:
## Code Changes
  
-  Adds CreditMarketAMOStrategy, an AMO that mints an OToken and supplies it as the sole lender into a deposit-gated Morpho Vault V2 ("credit vault"). Borrowers post their own collateral and borrow the
  OToken; the interest they pay grows the strategy's position and reaches OToken holders as rebase yield. The strategy only ever mints OToken in and burns OToken out — it never touches the Vault's backing
  asset.

-  The contract is generic: the OToken, hard asset, and credit vault are set at deploy time, so one implementation serves OUSD or OETH. Wired up for OUSD first.

###  Key design decisions

  - Phantom backing. checkBalance reports the full live position value (keeps total value and rebase correct), but the strategy provides zero redemption capacity — it can only burn OToken. mintCap is the
  only on-chain bound and must be sized below the protocol's real redeemable liquidity.
  - mintCap-gated minting. Minting is off until governance raises the cap; mintAndSupply / redeemAndBurn are Governor-or-Strategist; withdrawAll tolerates illiquidity so cleanup never reverts.
  - Yield-first accounting. netMinted tracks principal still deployed, drawing accrued interest down before principal, so netMinted ≤ positionValue() always.
  - Shared merkle claim. Extracted merkleClaim into AbstractMerkleClaimStrategy, now shared with Generalized4626Strategy.

### DIFF
--- a/contracts/contracts/strategies/AbstractMerkleClaimStrategy.sol
+++ b/contracts/contracts/strategies/AbstractMerkleClaimStrategy.sol
@@ -1,0 +1,53 @@
+// SPDX-License-Identifier: BUSL-1.1
+pragma solidity ^0.8.0;
+
+/**
+ * @title Abstract strategy with Merkl reward claiming
+ * @notice Adds the ability to claim rewards from the Merkl Distributor to a strategy.
+ * @dev Holds no storage so it is safe to insert into the inheritance chain of an
+ *      already deployed upgradeable strategy without shifting any storage slots.
+ * @author Origin Protocol Inc
+ */
+import { InitializableAbstractStrategy } from "../utils/InitializableAbstractStrategy.sol";
+import { IDistributor } from "../interfaces/IMerkl.sol";
+
+abstract contract AbstractMerkleClaimStrategy is InitializableAbstractStrategy {
+    /// @notice The address of the Merkle Distributor contract.
+    IDistributor public constant merkleDistributor =
+        IDistributor(0x3Ef3D8bA38EBe18DB133cEc108f4D14CE00Dd9Ae);
+
+    event ClaimedRewards(address indexed token, uint256 amount);
+
+    /**
+     * @param _baseConfig The platform and OToken vault addresses
+     */
+    constructor(BaseStrategyConfig memory _baseConfig)
+        InitializableAbstractStrategy(_baseConfig)
+    {}
+
+    /// @notice Claim tokens from the Merkle Distributor
+    /// @param token The address of the token to claim.
+    /// @param amount The amount of tokens to claim.
+    /// @param proof The Merkle proof to validate the claim.
+    function merkleClaim(
+        address token,
+        uint256 amount,
+        bytes32[] calldata proof
+    ) external {
+        address[] memory users = new address[](1);
+        users[0] = address(this);
+
+        address[] memory tokens = new address[](1);
+        tokens[0] = token;
+
+        uint256[] memory amounts = new uint256[](1);
+        amounts[0] = amount;
+
+        bytes32[][] memory proofs = new bytes32[][](1);
+        proofs[0] = proof;
+
+        merkleDistributor.claim(users, tokens, amounts, proofs);
+
+        emit ClaimedRewards(token, amount);
+    }
+}
--- a/contracts/contracts/strategies/CreditMarketAMOStrategy.sol
+++ b/contracts/contracts/strategies/CreditMarketAMOStrategy.sol
@@ -63,14 +63,6 @@ contract CreditMarketAMOStrategy is AbstractMerkleClaimStrategy {
     event Redeemed(uint256 oTokenAmount, uint256 sharesBurned);
     event MintCapUpdated(uint256 oldCap, uint256 newCap);
 
-    error MintCapExceeded(
-        uint256 requested,
-        uint256 currentNetMinted,
-        uint256 cap
-    );
-    error NothingToWithdraw();
-    error UnsupportedFunction();
-
     /**
      * @param _baseConfig platformAddress = the Morpho V2 credit vault,
      *        vaultAddress = the OToken Vault (eg VaultProxy or OETHVaultProxy).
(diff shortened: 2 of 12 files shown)


contracts/contracts/strategies/AbstractMerkleClaimStrategy.sol
```diff
@@ -0,0 +1,53 @@
...
+ * @title Abstract strategy with Merkl reward claiming
+ * @notice Adds the ability to claim rewards from the Merkl Distributor to a strategy.
+ * @dev Holds no storage so it is safe to insert into the inheritance chain of an
+ *      already deployed upgradeable strategy without shifting any storage slots.
+ * @author Origin Protocol Inc
+ */
+import { InitializableAbstractStrategy } from "../utils/InitializableAbstractStrategy.sol";
+import { IDistributor } from "../interfaces/IMerkl.sol";
+
+abstract contract AbstractMerkleClaimStrategy is InitializableAbstractStrategy {
```
naddison36, line 14: I like how the merkle abstraction has been moved out of `Generalized4626Strategy`

contracts/contracts/strategies/CreditMarketAMOStrategy.sol
```diff
@@ -0,0 +1,301 @@
...
+    }
+
+    /// @notice Same as maxWithdrawable(); the liquid portion of the position.
+    function liquidValue() external view returns (uint256) {
+        return maxWithdrawable();
+    }
+
+    /// @notice Full position value (principal plus accrued interest), in OToken units.
+    function positionValue() public view returns (uint256) {
+        return creditVault.previewRedeem(creditVault.balanceOf(address(this)));
```
naddison36, line 214: this assumes Morpho v2 vault's previewRedeem is share-accounting math, not a liquidity check.

In Morpho V2 it calls accrueInterestView() and returns:
`shares * (newTotalAssets + 1) / (newTotalSupply + virtualShares)`

So it reflects the vault’s total asset value per share, including assets deployed through adapters. High utilization can make the position illiquid, but it does not by itself lower the accounting value.

For Aave vaults, `previewRedeem` is liquidity-aware and can return less than `convertToAssets` when liquidity is constrained.

So the code is fine when using Morpho v2 vaults. `convertToAssets` would be a better option if non-morpho vaults could be used in the future.

Given CreditMarketAMOStrategy is Morpho v2 specific, I think it's fine to use `previewRedeem`

contracts/contracts/strategies/CreditMarketAMOStrategy.sol
```diff
@@ -0,0 +1,301 @@
...
+    }
+
+    /// @notice The principal, ie net OToken minted into the position.
+    function principal() external view returns (uint256) {
```
naddison36: nit: is this view function adding value over the already public `netMinted` storage variable?

Also, `netMinted` may drift below the true principal once accrued interest is burned (redeemAndBurn lowers it by the amount withdrawn, floored at 0).

contracts/contracts/strategies/CreditMarketAMOStrategy.sol
```diff
@@ -0,0 +1,301 @@
...
+        return netMinted;
+    }
+
+    /// @notice Interest accrued on the position above the principal, in OToken units.
+    function accruedYield() external view returns (uint256) {
```
naddison36, line 220: this won't be the true accrued yield after yielded oTokens are burned leaving `netMinted` not reflecting the true principal amount.

naddison36 (commented):
A simple and clean implementation. Well done.

I think we need to decide if `principal` and `accruedYield` should be the true values or we just document that these aren't accurate as the `netMinted` amount will be less than the true principal amount once a burn has been done.

contracts/contracts/strategies/CreditMarketAMOStrategy.sol
shahthepro: This is good point, we would be only using it for Morpho V2 in the near future. I'll add a comment regardless

shahthepro: Done in: https://github.com/OriginProtocol/origin-dollar/pull/2927/commits/f5d87df47088da20fba53ca6957f52042f245894

contracts/contracts/strategies/CreditMarketAMOStrategy.sol
shahthepro: I removed `principal` and another duplicate method. Re: the drift, I changed how the strategy behaves: it now burns through the yield/interest tokens first before it eats into the printed tokens. 

So, if there's 100 OUSD minted and an interest of 10 OUSD is earned on top of it, when the strategist burns 50 OUSD, netMinted will be 60 OUSD (instead of 50 OUSD with the original implementation): https://github.com/OriginProtocol/origin-dollar/pull/2927/commits/f5d87df47088da20fba53ca6957f52042f245894

contracts/contracts/strategies/CreditMarketAMOStrategy.sol
shahthepro: Fixed by changing the behaviour to burn through yield first: https://github.com/OriginProtocol/origin-dollar/pull/2927/commits/f5d87df47088da20fba53ca6957f52042f245894

contracts/contracts/strategies/CreditMarketAMOStrategy.sol
naddison36: Does this work when the burn about is less than the interest amount. Using the above scenario, 100 OUSD minted and an interest of 10 OUSD is earned on top of it, when the strategist burns 5 OUSD, netMinted will be 105 OUSD and accrued yield will be 0 OUSD. The `redeemAndBurn` effectively moves any left over yield to principal. 

What if the `netMinted` can only be decreased and not increased in `redeemAndBurn`? The scenario then becomes, 100 OUSD minted and an interest of 10 OUSD is earned on top of it, when the strategist burns 5 OUSD, netMinted will remain at the lower 100 OUSD and accrued yield will be 5 OUSD.

shahthepro: That's how it works right now. If there's an interest of 10 OUSD and 5 of that is burned, netminted will still be 100 OUSD. Position value would be 105 OUSD (so accruedYield would be reported as 5)

shahthepro: Here're the specific lines: https://github.com/OriginProtocol/origin-dollar/commit/f5d87df47088da20fba53ca6957f52042f245894#diff-7e729b9f7930dfc3c5241433dd882a09e5388c0cb1996df54ff5437e0cadf3d8R173-R177

They update `netMinted` only if position value is lower than `netMinted` (i.e. we burnt some from printed capital)

naddison36: Awesome work

PR #2934, opened 2026-07-09
OToken Vault Loss Socialization (Mint Protection)

shahthepro:
## Summary

This PR introduces proportional loss socialization across live OToken holders and queued withdrawal requests.

### Mint protection

User mints now revert whenever the Vault is under-backed, based on the pre-mint state. Mints remain 1:1 when permitted. `mintForStrategy()` is intentionally exempt so strategist operations remain available.

## Deployment scope

The upgrade and configuration are applied to:

- OUSD Vault on Ethereum
- OETH Vault on Ethereum
- superOETHb Vault on Base

## Code Change Checklist

- [x] The contract code is complete
- [x] Executable deployment file
- [x] Fork tests that test after the deployment file runs
- [x] Unit tests, if needed
- [ ] The owner has done a [full checklist review](https://github.com/OriginProtocol/security/blob/master/templates/Contract-Code-Review.md) of the code + tests

Internal review:


contracts/contracts/vault/VaultCore.sol
```diff
@@ -74,6 +74,13 @@ abstract contract VaultCore is VaultInitializer {
     function _mint(uint256 _amount) internal virtual {
         require(_amount > 0, "Amount must be greater than 0");
 
+        // Block mints when the vault is under-backed by more than mintTolerance.
+        // Checked on the pre-mint state (the deposit is transferred in below).
+        // Stops new minters from buying OTokens above their real value and from
+        // subsidising the withdrawal queue at par during a depeg. mintForStrategy
+        // is intentionally not gated here; that stays a strategist decision.
+        require(_backingRatio() + mintTolerance >= 1e18, "Vault under-backed");
```
sparrowDom: I assume this check deliberately isn't set on `mintForStrategy`

contracts/contracts/vault/VaultCore.sol
```diff
@@ -605,6 +655,63 @@ abstract contract VaultCore is VaultInitializer {
...
+            IStrategy strategy = IStrategy(allStrategies[i]);
+            if (strategy.supportsAsset(asset)) {
+                balance = balance + strategy.checkBalance(asset);
+            }
+        }
+    }
+
+    /**
+     * @dev Gross vault + strategy asset value, scaled to 18 decimals. Unlike
+     * `_checkBalance` this is not netted against the withdrawal queue and is
```
sparrowDom: nit: Could also say: ... Unlike `_checkBalance` this ignores the withdrawal queue.

contracts/contracts/vault/VaultCore.sol
```diff
@@ -324,39 +342,45 @@ abstract contract VaultCore is VaultInitializer {
...
+        // what walks the fixed FIFO gate forward under impairment.
         withdrawalQueueMetadata.claimed =
             queue.claimed +
             SafeCast.toUint128(
                 StableMath.scaleBy(request.amount, assetDecimals, 18)
             );
 
+        // Nominal amount is emitted (the full size of the settled request); the
+        // asset actually transferred is the haircut `amount` return value.
         emit WithdrawalClaimed(msg.sender, requestId, request.amount);
```
sparrowDom, line 340: Might make sense to add another field to `WithdrawalClaimed` that indicates the actual paid out amount - which in case of socialized loss would be different from the `request.amount`

contracts/contracts/vault/VaultCore.sol
sparrowDom, line 327: 🔴  `request.queued` is in nominal units while the `queue.claimable` is in actual balances -> which in case of socialization losses is smaller than queued. In case users would decide to empty the whole protocol, this would trigger preventing the full withdrawals. 

Would probably also be cool to have a test for this: 
- 3 users with 10 WETH in the vault
- Vault experiences a loss cutting the ratio to 0.9
- see if all 3 can claim the withdrawal

contracts/contracts/vault/VaultAdmin.sol
```diff
@@ -365,6 +365,15 @@ abstract contract VaultAdmin is VaultCore {
         emit MaxSupplyDiffChanged(_maxSupplyDiff);
     }
 
+    /**
+     * @notice Sets the maximum shortfall in the backing ratio below 1.0 that
+     *      still permits user mints. 18 decimals. eg 0.01e18 = 1%.
+     */
+    function setMintTolerance(uint256 _mintTolerance) external onlyGovernor {
+        mintTolerance = _mintTolerance;
```
sparrowDom: There should probably be some sensible bounding here: 
- `require(_mintTolerance <= 1e18)` is probably a must 
- `require(_mintTolerance >= 0.95e18)` might also make sense

contracts/contracts/vault/VaultAdmin.sol
```diff
@@ -365,6 +365,15 @@ abstract contract VaultAdmin is VaultCore {
         emit MaxSupplyDiffChanged(_maxSupplyDiff);
     }
 
+    /**
+     * @notice Sets the maximum shortfall in the backing ratio below 1.0 that
+     *      still permits user mints. 18 decimals. eg 0.01e18 = 1%.
+     */
+    function setMintTolerance(uint256 _mintTolerance) external onlyGovernor {
```
sparrowDom: Lets also add a fork test for this confirming that the deploy has set the expected setting

sparrowDom (commented):
Let some comments inline

contracts/contracts/vault/VaultCore.sol
shahthepro: Yep, that's intentionally left out because the Guardian controls that. Also, if we add those checks to mintForStrategy and burnForStrategy methods, we might be prevented from exiting or rebalancing AMO pools at times

sparrowDom: yeah that makes sense

contracts/contracts/vault/VaultAdmin.sol
shahthepro: Could add bounds here. But I'm not sure if that'd end up causing issues for us in the future. If we set it to 5% max and if for any reason we need to allow a higher tolerance, that'll start causing issues for us. I can keep a higher upper bound though (something like 50%, either way this is onlyGovernor action, so it already goes through a proposal). Lower bound seems unnecessary since 0% tolerance is the best case

contracts/contracts/vault/VaultCore.sol
shahthepro: This is a good find. This is actually a byproduct of one of the decision decisions: https://app.notion.com/p/originprotocol/OToken-Vault-Loss-Socialization-39784d46f53c8020bea8ef588ffde2b9?source=copy_link#39784d46f53c80028347c6aa8b95722b

Let me brainstorm with Claude to see if it can come up with a better solution without complicating much

contracts/contracts/vault/VaultAdmin.sol
sparrowDom: Agreed, the upper bound might be the only one sensible

contracts/test/vault/redeem.js
```diff
@@ -1426,33 +1426,40 @@ describe("OUSD Vault Withdrawals", function () {
...
+          userBefore.add(payout6)
+        );
+        expect(await usdc.balanceOf(vault.address)).to.equal(
+          vaultBefore.sub(payout6)
         );
+        // claimed bumps by the full nominal even though only the haircut is paid
+        const queueAfter = await vault.withdrawalQueueMetadata();
+        expect(queueAfter.claimed).to.equal(queueBefore.claimed.add(nominal6));
+        // ratio is ~invariant; a tiny upward drift from 6-dp rounding favours the vault
+        expect(await vault.backingRatio()).to.be.gte(ratio);
```
sparrowDom: is the `gte` check required? Can the ratio really only drift in 1 direction

sparrowDom (commented):
1 minor comment

contracts/contracts/vault/VaultCore.sol
```diff
@@ -661,21 +665,22 @@ abstract contract VaultCore is VaultInitializer {
      * withdrawal queue and never clamped to zero. Shared by `_checkBalance`
      * (which nets the queue) and `_grossAssets` (which does not).
      */
-    function _rawAssetBalance() internal view returns (uint256 balance) {
-        balance = IERC20(asset).balanceOf(address(this));
```
sparrowDom: This change is probably unnecessary?

sparrowDom: One of the things I have in my global Claude.md file is to not change the adjecent code.

shahthepro: I explicitly asked Claude to do this change (and in other functions that this PR added) because as a security precaution we decided not to use named return variables

contracts/contracts/vault/VaultCore.sol
```diff
@@ -323,7 +326,7 @@ abstract contract VaultCore is VaultInitializer {
     ///        The caller computes this once so a batch is treated identically.
     function _claimWithdrawal(uint256 requestId, uint256 ratio)
         internal
-        returns (uint256 amount)
+        returns (uint256)
```
sparrowDom: nit: this could remain a named return param

contracts/contracts/vault/VaultStorage.sol
```diff
@@ -50,10 +50,17 @@ abstract contract VaultStorage is Initializable, Governable {
...
+    /// @param _paidAmount Asset actually transferred out after the loss-socialisation
+    ///     haircut, in ASSET decimals (eg 6 for USDC, 18 for WETH). Equals `_amount`
+    ///     scaled to asset decimals while the vault is fully backed, and is strictly
+    ///     less once a loss has been socialised.
     event WithdrawalClaimed(
         address indexed _withdrawer,
         uint256 indexed _requestId,
-        uint256 _amount
+        uint256 _amount,
+        uint256 _paidAmount
```
sparrowDom: nit: we will need to notify @apexearth once this is deployed that the event signatures have changed

sparrowDom (dismissed):
LGTM

contracts/test/vault/redeem.js
shahthepro: Interestingly, it's always one directional for 18 decimal. But for OUSD which has USDC w/ 6 decimals, there seems to be an edge where it can lower by 1 wei or so. Will update it

naddison36 (commented):
The `mintTolerance` feels like an over complication to me. New minters can still have significant losses with a 1% mint tolerance.

Here's a contrived example with $10 million in assets and a `0.991` backing ratio:

```text
Effective supply: $10.09m
Existing deficit: ~$90.8k
```

A user can deposit $10 million because the Vault is within the 1% `mintTolerance`. They receive 10 million OTokens at 1:1, but after minting those tokens are worth only about $9.955 million.

```text
Immediate minter loss: ~$45.2k
```

That value is transferred to existing holders and queued withdrawals. A simpler rule—blocking all mints whenever `backingRatio < 1`—avoids this value transfer and removes the need for `mintTolerance`.

I'd prefer to go with a simpler mint gate of 
```Solidity
require(
    _grossAssets() >= _effectiveSupply(),
    "Vault under-backed"
);
```

contracts/contracts/vault/VaultCore.sol
```diff
@@ -74,12 +74,12 @@ abstract contract VaultCore is VaultInitializer {
...
-        // Stops new minters from buying OTokens above their real value and from
-        // subsidising the withdrawal queue at par during a depeg. mintForStrategy
-        // is intentionally not gated here; that stays a strategist decision.
-        require(_backingRatio() + mintTolerance >= 1e18, "Vault under-backed");
+        // Block mints whenever the vault is under-backed. Checked on the pre-mint
+        // state (the deposit is transferred in below). Stops new minters from
+        // buying OTokens above their real value and from subsidising the
+        // withdrawal queue at par during a depeg. mintForStrategy is intentionally
+        // not gated here; that stays a strategist decision.
+        require(_grossAssets() >= _effectiveSupply(), "Vault under-backed");
```
sparrowDom: Looks good. We should put a note that we need a critical alerts whenever this error is triggered.

naddison36 (dismissed):
I like the simpler design.

LGTM

clement-ux (dismissed):
LGTM

sparrowDom (dismissed):
LGTM

naddison36 (dismissed):
LGTM

PR #2938, opened 2026-07-13
Check nonce before relaying message

shahthepro:
## Code Change Checklist

- Check the message nonce before relaying a CCTP message
- Add a txHash optional arg that'll relay messages from that specific tx


sparrowDom (approved):
Looks like a nice solution lets test it

PR #2943, opened 2026-07-17
Sparrow dom/foundry migration/talos actions

sparrowDom:
This migrates the Talos actions to Foudry: 
- actions no longer use hardhat rather the hardcoded addresses
- Claude is directed to check all deploys agains the talos actions to fix paths to any new deployments 

**Important** -> In Talos all actions will need to be edited to support new way of invocation


contracts/hardhat.config.js
clement-ux, line 1: Why do we still need an `hardhat.config.js`?

clement-ux: Nevermind, we will remove all related to Hardhat in a second PR.

clement-ux (approved):

PR #2946, opened 2026-07-21
chore: update references from @talos/client to @oplabs/talos-client, …

toniocodo:
## What

Migrates the origin-dollar runner off git-installing `@talos/client` from the private `oplabs/talos` repo (SSH deploy key) to the **published `@oplabs/talos-client` GitHub Package**. Companion to `oplabs/talos#19` (which publishes it) and `OriginProtocol/arm-oeth#305`.

## Changes

- **`@talos/client` → `@oplabs/talos-client`**, declared as an **optional peer dependency**; the runner Dockerfile installs the pinned version with a `read:packages` PAT (`TALOS_PACKAGE_TOKEN` secret). No more SSH deploy key / git install.
- **`autoInstallPeers: false`** (`pnpm-workspace.yaml` + lockfile) so pnpm never tries to resolve the optional peer without a token — token-less `pnpm install` passes locally, in CI, and in the Docker build.
- **Lazy `require`** of the client in `contracts/utils/signers.js`, behind the `DATABASE_URL` gate, so hardhat boots when the optional package is absent.
- **Dockerfile**: `# syntax=docker/dockerfile:1.10` (for the `env=` secret mount) + `pnpm add --save-prod`.
- **`@lodestar` `^1.41.0 → ^1.44.0`** *(the one non-obvious change — worth a look)*: 1.44 drops `@lodestar/config`'s spurious `@lodestar/spec-test-util` dependency, which shipped a broken `vitest: "catalog:"` that broke any workspace re-resolution (i.e. the runner image's `pnpm add`). Net lockfile shrink (~600 lines); retired the now-dead `spec-test-util` `packageExtension`.
- Pin `@oplabs/talos-client@0.0.28`; remove an accidental root `pnpm-workspace.yaml` / `pnpm-lock.yaml`.

## Validation

- **defi.yml green.** The failing **Base Fork Tests** job is pre-existing and non-blocking — it fails on `master` too and is unrelated to `@lodestar` (Base is L2, no consensus-layer usage).
- Runner image builds locally end-to-end (frozen install + token-authed `pnpm add @oplabs/talos-client@0.0.28`).

> `TALOS_PACKAGE_TOKEN` (a `read:packages` PAT) must stay set on the repo. The image build runs on `master`; the actual runner roll is a separate signed release.

PR #2950, opened 2026-07-27
Talos actions fix

sparrowDom:
## Problem

Every `origin-dollar` scheduled action in Talos prod has been failing with:

```
ERR_PNPM_RECURSIVE_EXEC_FIRST_FAIL  Command "tsx" not found
```

#2848 moved the action runtime off hardhat to `tsx tasks/run.ts` and rewrote all 46 schedule commands in `migrations/seed_schedules.sql` accordingly — but **`tsx` was never declared**. It is absent from `contracts/package.json` and from `pnpm-lock.yaml`, and the image installs with `pnpm install --frozen-lockfile`, so `pnpm exec tsx` fails in the runner container.

The same change added `tasks/lib/network.ts` and `tasks/lib/signer.ts` importing **`@talos/client`**. The package is published as `@oplabs/talos-client` (renamed in #2946, which landed the day before), and `dockerfile-actions` installs that name — so both modules fail to resolve in the image. No `tsconfig` path alias covers the old name.

## Why CI stayed green

`dockerfile-actions` wraps the catalog dump in a swallow-all fallback:

```
pnpm exec tsx dump-actions-catalog.ts > /app/actions-catalog.json \
  || (echo "[build] action catalog dump failed; shipping empty catalog" \
      && echo "{}" > /app/actions-catalog.json)
```

`tsx` was already missing at build time, the dump failed, and the `||` branch shipped `{}` and let the build succeed. The runner logged `loaded actions catalog: 0 tasks` on boot. The build had the exact signal needed to catch this and discarded it.

## Changes

- Declare `tsx@^4.23.1` in `dependencies` — it is a genuine runtime dependency of the container now, not tooling, so it belongs there rather than `devDependencies` (survives any future `--prod` install).
- Point `tasks/lib/{network,signer}.ts` at `@oplabs/talos-client`.

Lockfile movement is `tsx` + its esbuild platform binaries only; no version changes to existing packages.

## Verification

Reproduced the exact prod failure locally, then confirmed the fix:

| | actions in catalog |
|---|---|
| before (prod image) | 0 |
| `tsx` added, stale import | 15 (27 modules skipped: `Cannot find module '@talos/client'`) |
| both fixes | **42, zero skips** |

Cross-checked the catalog against the schedule table: all 46 scheduled rows reference 38 distinct actions, and **every one resolves** in the catalog.

`pnpm run typecheck` fails identically on `master` with and without this change (`@types/node` is undeclared) — pre-existing, left alone.

## Follow-up (not in this PR)

The `|| echo "{}"` fallback should fail the build, or at minimum fail when the dump yields `{}`. A fallback that turns a fundamentally broken image into a green build is why this reached production rather than CI.

### DIFF
--- a/CODEOWNERS
+++ b/CODEOWNERS
@@ -1,2 +1,2 @@
-/contracts @sparrowDom @DanielVF @naddison36 @shahthepro @clement-ux
-/brownie @sparrowDom @DanielVF @naddison36 @shahthepro @clement-ux
+/contracts @sparrowDom @naddison36 @shahthepro @clement-ux
+/brownie @sparrowDom @naddison36 @shahthepro @clement-ux
--- a/contracts/package.json
+++ b/contracts/package.json
@@ -145,7 +145,8 @@
     "@lodestar/utils": "^1.44.0",
     "@types/pg": "^8.20.0",
     "origin-morpho-utils": "^0.1.1",
-    "pg": "^8.20.0"
+    "pg": "^8.20.0",
+    "tsx": "^4.23.1"
   },
   "peerDependencies": {
     "@oplabs/talos-client": "0.0.28"
--- a/contracts/pnpm-lock.yaml
+++ b/contracts/pnpm-lock.yaml
@@ -46,6 +46,9 @@ importers:
       pg:
         specifier: ^8.20.0
         version: 8.20.0
+      tsx:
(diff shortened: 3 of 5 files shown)


naddison36 (approved):
LGTM

PR #2954, opened 2026-07-30
Fix snap Hardhat tasks

naddison36:
## Summary

Fixes the following Hardhat snapshot tasks:

- `snapStakingStrat`
  - Removes the unconditional `@oplabs/talos-client` dependency from the shared network helper.
  - Restores access to the Hardhat provider and network context.
  - Uses the existing `IERC20Metadata` ABI for `IERC20` contract resolution.
  - Decodes beacon data using the fork returned by the beacon API.

- `snapStaking`
  - Removes the unnecessary signer lookup from this read-only task.

- `snapVault`
  - Normalizes the token symbol so lowercase values such as `oeth` resolve the correct deployment.

## Test plan

```bash
pnpm hardhat snapStakingStrat --network mainnet --ssv false
pnpm hardhat snapStaking --network mainnet --index 2
pnpm hardhat snapVault --network mainnet --symbol oeth
```

### DIFF
--- a/contracts/tasks/lib/contracts.ts
+++ b/contracts/tasks/lib/contracts.ts
@@ -58,7 +58,10 @@ function readDeployment(
 // Resolve an ABI by contract/interface name: curated abi/<name>.json first
 // (interfaces like IVault), then the deployment artifact's abi.
 function readAbiByName(chainId: number, name: string): unknown[] {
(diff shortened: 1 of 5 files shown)


sparrowDom (approved):
NIce, thank you for this

PR #2959, opened 2026-08-04
Remove cache of beacon data files

naddison36:
## Summary

- Store downloaded beacon state SSZ files in temporary storage instead of the project cache directory.
- Create the temporary directory on demand before writing beacon state data.
- Delete all downloaded beacon state files after `verifyBalances` and `verifyDeposits` complete.
- Run cleanup in a `finally` block so files are removed even when verification fails.

## Background

Talos containers no longer create `/app/cache` during their build. Beacon verification actions consequently failed with `ENOENT` when attempting to write `./cache/state_<slot>.ssz`.

Beacon state files are large and only needed while an action is running, so they do not need to persist between actions or container deployments.

## Testing

- Ran JavaScript and TypeScript Prettier.
- Verified `utils/beacon.js` loads successfully.
- Ran `git diff --check`.

### DIFF
--- a/contracts/tasks/actions/verifyBalances.ts
+++ b/contracts/tasks/actions/verifyBalances.ts
@@ -58,7 +58,10 @@ action({
     );
   },
   run: async ({ signer, args }) => {
-    await verifyBalances({ ...args, signer });
-    cleanStateCache();
+    try {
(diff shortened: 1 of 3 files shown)


sparrowDom (approved):
thanks looks more resiliant

PR #2960, opened 2026-08-04
Better timeout handling of beacon chain data

naddison36:
## Summary

- Extend the beacon-state fetch timeout from 5 to 15 minutes.
- Keep the timeout active while downloading the complete SSZ response body.
- Add progress logs when response headers arrive and when the body download completes.
- Log when a beacon-state request is aborted after reaching the timeout.

## Background

Beacon-state responses are large and can take approximately seven minutes to download. The existing five-minute timeout was cleared as soon as the response headers arrived, leaving the subsequent `response.arrayBuffer()` operation without any timeout.

This caused Talos actions to remain in the `running` state indefinitely if the response body stalled.

The updated timeout covers both the initial request and the full body download while allowing enough time for the observed response duration.

This applies to all actions using the shared beacon-state loader, including `daily_verify_deposits` and `daily_verify_balances`.

## Testing

- Ran `pnpm prettier:js`.
- Verified `utils/beacon.js` loads successfully.
- Ran `git diff --check`.

### DIFF
--- a/contracts/utils/beacon.js
+++ b/contracts/utils/beacon.js
@@ -19,6 +19,7 @@ const fetchImpl =
         import("node-fetch").then(({ default: fetch }) => fetch(...args));
 
 const SLOTS_PER_EPOCH = 32;
+const BEACON_STATE_FETCH_TIMEOUT_MS = 15 * 60 * 1000;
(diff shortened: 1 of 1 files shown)


sparrowDom (approved):
LGTM

PR #2967, opened 2026-08-10
Remove SSV validators from old staking strategy

naddison36:
## Summary

Updates the Talos `removeValidator` action to remove validators from the legacy Native Staking Strategy 2 instead of the retired Compounding Staking Strategy.

- Hardcodes Native Staking Strategy 2 and its SSV operator IDs (`752,753,754,755`)
- Allows `--pubkey` to contain one or more comma-separated validator public keys
- Validates all supplied public keys before submitting transactions
- Removes validators sequentially with a 30-second delay so the SSV API can refresh cluster state
- Reuses the signer supplied by the standalone Talos runtime
- Updates the Talos action catalog and documentation
- Avoids loading the consolidation signer when no consolidation confirmation transaction is required

## Example

```bash
cd /app && pnpm exec tsx tasks/run.ts removeValidator --network mainnet --pubkey 0xKEY1,0xKEY2 --consol true
```

## Testing

- Generated and verified the Talos action catalog
- Ran Prettier on the changed TypeScript, JavaScript, and Markdown files
- Verified the diff with `git diff --check`

### DIFF
--- a/contracts/docs/ACTIONS.md
+++ b/contracts/docs/ACTIONS.md
@@ -98,7 +98,7 @@ each run (see notes in `seed_schedules.sql`).
 | Action                       | Description                                                                          |
 | ---------------------------- | ------------------------------------------------------------------------------------ |
 | `stakeValidator`             | Convert WETH to ETH and deposit to a validator from the Compounding Staking Strategy |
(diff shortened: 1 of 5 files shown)


sparrowDom (approved):
LGTM

PR #2968, opened 2026-08-10
Fix remove validator Talos action

naddison36:
## Summary

Fixes Talos actions that referenced Hardhat-provided global `ethers` and `hre` objects, which are unavailable in the standalone Talos runtime.

- Use the action signer’s provider to resolve the chain ID in `removeValidator`.
- Remove the unused `hre.network` argument from the SSV cluster lookup.
- Use the signer’s provider when checking strategy ETH balances for automatic validator deposits and withdrawals.

## Testing

- Ran Prettier on the changed JavaScript files.
- Ran focused ESLint checks.
- Ran `git diff --check`.
- Verified all registered Talos action modules load successfully.


sparrowDom (approved):
LGTM

PR #2982, opened 2026-08-25
Remove legacy Hardhat contract tests

clement-ux:
## Why

Foundry is already the tool we use to test smart contracts. The repository still contained the old Hardhat contract test suite, so we had two test systems even though only one was meant to be the source of truth.

This PR removes that retired Hardhat test layer. It does not change the contracts or replace the tests again; the contract coverage already lives in Forge.

## What changes

- remove the legacy Hardhat smart-contract tests and their coverage/fork commands
- remove dependencies that were only needed by those tests
- keep the remaining JavaScript task tests and run them separately in CI
- move a few live files out of `test/` to where they actually belong:
  - the Talos task test moves to `tasks/test/`
  - Forge Beacon and validator fixtures move next to their Forge consumers
  - ABIs still used by Hardhat tooling move to `abi/hardhat/`

## What stays

Hardhat is still present for operational tasks and, at this point in the stack, the historical deployment and ABI tooling. Those are handled separately in the follow-up PRs.

## Stack

This is the first PR in the cleanup:

1. **#2982 — remove legacy Hardhat contract tests**
2. #2983 — remove legacy Hardhat deployment scripts
3. #2985 — generate the ABI package with Foundry

## Validation

- `forge build contracts/ --force -j 1`
- affected Forge fixture subtrees compile successfully
- CompoundingStakingStrategy Forge suite: 116 passing
- `pnpm run test:tasks`: 20 passing
- `pnpm run test:scripts`: 61 passing
- `pnpm run test:layouts`: 11 passing
- JavaScript lint, typecheck, Prettier and Hardhat task loading pass

The full all-at-once Forge build exceeded the 8 GB local runner limit; the production contracts and affected test subtrees were therefore validated separately.

PR #2983, opened 2026-08-25
Remove legacy Hardhat deployment scripts

clement-ux:
## Why

Deployments have moved to Foundry, but the repository still contained 129 old `hardhat-deploy` scripts and the helpers built around them. They were no longer the deployment system, yet generating `@origin/defi` could still load and execute that old pipeline.

This PR removes the executable Hardhat deployment path while keeping the operational tools that still need historical deployment information.

## What changes

- remove the 129 legacy scripts under `contracts/deploy/**`
- remove deploy-only helpers and npm commands
- stop ABI generation from running `hardhat deploy`
- keep only the small network, unit, funding and governance helpers still used by operational tasks
- point the deployment-verification skill at the Foundry deployment scripts

At this point in the stack, ABI generation uses a clean Hardhat compilation only; follow-up #2985 replaces that final compilation with Foundry.

## What stays

`deployments/**` is intentionally kept. These files are not executable deployment scripts: they are the address and ABI records used by Talos and other operational tasks to find existing contracts.

`hardhat-deploy` also remains as a read-only adapter for those records. It is no longer used to deploy anything.

## Stack

This PR is intentionally based on #2982, not `master`:

1. #2982 — remove legacy Hardhat contract tests
2. **#2983 — remove legacy Hardhat deployment scripts**
3. #2985 — generate the ABI package with Foundry

## Validation

- `@origin/defi` generation succeeds without executing deployments: 292 ABI files
- Hardhat governance tasks `execute` and `executeOnFork` load successfully
- `pnpm run test:tasks`: 22 passing
- `pnpm run test:scripts`: 61 passing
- `pnpm run test:layouts`: 11 passing
- `forge build contracts/ --force -j 1`
- JavaScript/TypeScript lint, typecheck, Prettier and `git diff --check` pass


sparrowDom (dismissed):
LGTM

PR #2985, opened 2026-08-25
Generate ABI package with Foundry

clement-ux:
## Why

After #2983, Hardhat was no longer testing or deploying contracts. One dependency remained in the ABI release path: `@origin/defi` was still built from Hardhat compilation artifacts.

This PR moves that last compilation step to Foundry without changing the package consumed by applications.

## Before and after

Before:

```text
Hardhat compile → Hardhat artifacts → @origin/defi
```

After:

```text
Forge build → compatibility exporter → @origin/defi
```

The compatibility exporter reads Foundry output and writes the same public artifact shape used by the existing package. Consumers keep the same filenames, contract/source identities and ABI entries.

## What changes

- compile the 297 production Solidity files with `forge build contracts/ -j 1`
- export the Foundry output into the existing `@origin/defi` package format
- keep the same 292 published ABI filenames and semantic ABIs
- handle the two duplicate contract names deterministically and fail rather than silently choosing a different source
- update the ABI release workflow to install Foundry dependencies instead of the full npm/Talos dependency tree

## Talos and Hardhat

Talos is unchanged. It continues to use `deployments/**`, the task ABI files and its standalone action runner. Its action registry loads without skipped actions, and all task tests pass.

Hardhat also remains available as the operational task CLI. It is simply no longer part of contract testing, deployment or ABI compilation.

## Stack

This PR is intentionally based on #2983:

1. #2982 — remove legacy Hardhat contract tests
2. #2983 — remove legacy Hardhat deployment scripts
3. **#2985 — generate the ABI package with Foundry**

## Validation

- Foundry clean build: 297 Solidity files
- ABI export: 292 production artifacts, with 31 mocks excluded
- comparison with the previous Hardhat output: identical filenames, contract/source identities and semantic ABIs
- `npm pack --dry-run --json`: `@origin/defi` with 292 ABI files
- Talos action registry loads without skipped actions
- `pnpm run test:tasks`: 22 passing
- `pnpm run test:scripts`: 66 passing
- `pnpm run test:layouts`: 11 passing
- JavaScript/TypeScript lint, typecheck, Prettier, Hardhat CLI loading and `git diff --check` pass


sparrowDom (approved):
LGTM

PR #2987, opened 2026-08-26
Added lastVerifiedBalanceTimestamp to CompoundingStakingStrategy

naddison36:
## Summary

Adds `lastVerifiedBalanceTimestamp` to `CompoundingStakingStrategy` so integrations can determine when beacon-chain balance data was last successfully verified.

The value records the timestamp captured by `snapBalances()`, not the later transaction timestamp of `verifyBalances()`.

## Changes

- Add the public `uint64 lastVerifiedBalanceTimestamp` storage variable.
- Update it only after all validator and pending-deposit proofs pass in `verifyBalances()`.
- Leave it unchanged when:
  - `snapBalances()` is called without a successful verification.
  - `verifyBalances()` reverts.
- Expose the getter through `ICompoundingStakingStrategy`.
- Document the new accounting behavior.

Existing deployments will return `0` until their first successful balance verification after upgrading. No migration or backfill is required.

## Storage layout

The new variable consumes one reserved storage slot:

- `lastVerifiedBalanceTimestamp` is appended at slot 61.
- `__gap` is reduced from 40 to 39 slots.
- The gap still ends at slot 101, so downstream inherited storage is unchanged.
- The pinned storage-layout test has been updated accordingly.

## Tests

Added unit coverage confirming:

- The initial value is zero.
- `snapBalances()` alone does not update it.
- A successful verification stores the snapshot timestamp rather than the verification transaction timestamp.
- Clearing `snappedBalance.timestamp` does not clear the last verified timestamp.
- A failed verification preserves the previous value.
- A later successful verification advances the value.

Verification performed:

- `make build-contracts`
- Focused Compounding Staking Strategy tests: 28 passed
- `pnpm test:layouts`: 11 passed
- `git diff --check`

## Code Change Checklist

- [x] Contract code is complete
- [ ] Executable deployment file — not included in this PR
- [ ] Fork tests after deployment — not applicable without a deployment file
- [x] Unit tests
- [ ] Owner completed the full contract code review checklist


sparrowDom (approved):
Great thanks

PR #2997, opened 2026-09-01
Remove obsolete historical scripts and artifacts

clement-ux:
## Plain-English summary

This repository still contains many files from workflows that are no longer used: old staking and liquidity scripts, generated diagrams, unused ABIs, copied libraries, and empty placeholders.

This PR removes only that historical baggage. It does **not** replace Hardhat or change how active commands and local forks run. Keeping this cleanup separate makes the three functional migration steps much easier to review.

## Context

Foundry already handles contract compilation, testing, deployment, and ABI generation. Hardhat remains temporarily for two active jobs:

1. running operational commands;
2. running the local/forked node.

The next three PRs replace those jobs one at a time, then remove Hardhat.

## What changes in this PR

- remove retired Resolution, staking, compensation, and liquidity scripts;
- remove obsolete Hardhat-only ABIs and the historical `localhost.json` snapshot;
- remove orphaned diagrams and their generator entries;
- remove unused Rooster library copies and stale analysis configuration;
- remove empty test/deployment placeholders and dead utility modules;
- remove documentation and package-script references to those deleted files.

## What does not change yet

- active operational commands still run through Hardhat;
- Hardhat still provides the local/forked node;
- no runtime replacement happens in this layer.

## Stack and merge order

Base: `master`

1. **#2997 — remove obsolete historical scripts and artifacts (this PR)**
2. #2998 — introduce standalone ops CLI
3. #2999 — migrate local forks to Anvil
4. #3000 — remove the Hardhat runtime

Merge in cascade: **#3000 → #2999 → #2998 → #2997 → `master`**.

## What to review

The important question is not the implementation of a new runtime; there is none here. Please check that every deleted category is genuinely retired and that no surviving file still points to a deleted path.

This layer changes 80 files: 24,782 deletions and 8 small documentation/configuration additions.

## Validation

- `git diff --check`
- audited tracked references to all 77 deleted paths
- full GitHub CI passes
- verified that the complete four-layer stack produces exactly the same final tree as the previous #2993–#2995 stack on current `master`

PR #2998, opened 2026-09-01
Introduce standalone ops CLI

clement-ux:
## Plain-English summary

Hardhat is currently acting as a command launcher for operational tasks. The useful logic already lives in ordinary TypeScript handlers, but operators must still enter through `hardhat` to reach it.

This PR adds a small standalone CLI that calls those same handlers directly. The commands, parameters, signers, roles, and deployment data stay the same; only the entry point changes.

The local/forked node still uses Hardhat in this layer. That is replaced separately in #2999.

## Before and after

```text
Before: pnpm hardhat <command> → Hardhat task wrapper → existing handler
After:  pnpm ops <command>     → standalone catalogue → existing handler
```

This is a runtime CLI migration, **not** a contract compilation or deployment migration. Foundry already handles those jobs.

## What changes in this PR

- add `pnpm ops` with the complete 74-command public catalogue, derived from the `task()` declarations in `tasks/tasks.js` and pinned as a snapshot in `tasks/test/fixtures/ops-command-catalog.json` (the `accounts` task is dropped: `eth_accounts` is empty on a live RPC and the Hardhat task printed private keys);
- preserve command parameters and type coercion; `pnpm ops <command> --help` lists a command's options, `--key=value` is accepted, a repeated option is rejected as under Hardhat;
- preserve signer modes, role resolution, deployment lookups, and Foundry artifact loading;
- connect every registry-backed command to its existing handler, with `isMainnet` and `replaceContractAt` derived from the standalone network instead of `hre`;
- add catalogue, parser, signer, governance, and Talos action tests;
- remove active tooling that was only needed by the retired contract-test/deployment workflow;
- standardize the remaining JavaScript workflow on pnpm and the standalone CLI.

## What does not change yet

- Hardhat still runs the local/forked node;
- Hardhat packages and configuration remain installed;
- Anvil migration happens in #2999;
- final Hardhat removal happens in #3000.

## Stack and merge order

Base: #2997 (`chore/remove-obsolete-historical-artifacts`)

1. #2997 — remove obsolete historical scripts and artifacts
2. **#2998 — introduce standalone ops CLI (this PR)**
3. #2999 — migrate local forks to Anvil
4. #3000 — remove the Hardhat runtime

Merge in cascade: **#3000 → #2999 → #2998 → #2997 → `master`**.

## What to review

Please focus on behavioral parity:

- are all public commands still present?
- do parameters keep the same names, required/optional status, and types?
- do signer, role, network, and deployment lookups behave as before?
- does each command still reach the intended existing handler?

## Validation

- `pnpm install --frozen-lockfile`
- `pnpm test:tasks` — 44 passing
- `pnpm test:scripts` — 66 passing
- `pnpm test:layouts` — 11 passing
- `pnpm typecheck`
- `pnpm lint:js`, `pnpm lint:ts`, `pnpm prettier:check`
- `forge build contracts/ -j 1`
- full GitHub CI passes


contracts/utils/hardhat-task-helpers.js
```diff
@@ -1,38 +1,9 @@
 const hre = require("hardhat");
-const { BigNumber } = require("ethers");
-const { parseUnits } = require("ethers").utils;
 
 const isFork = process.env.FORK === "true";
 const isMainnet = hre.network.name === "mainnet";
```
sparrowDom: `pnpm ops` never hands `--network` to Hardhat, so the `require("hardhat")` above boots the default in-process network and `hre.network.name` is `hardhat`. `isMainnet` is therefore always `false` under the new CLI (`node -e 'require("./utils/hardhat-task-helpers")'` in this tree prints `{isFork:false,isMainnet:false}`). Two consumers depend on it: `pnpm ops capital --symbol OUSD --pause true --network mainnet` skips the print-a-governance-proposal branch (`tasks/vault.js:241-248`) and calls `vault.connect(signer).pauseCapital()` directly, and `pnpm ops execute --network mainnet` loses its "can not be used on mainnet" guard (`tasks/governance.js:31-33`). `tasks/test/governance.js:5` mocks this module, so the unit test cannot see it. Talos schedules are unaffected, `run.ts` never loads `tasks.js`.

This is fixed in #3000, which replaces this file with `utils/runtime-helpers.js` (derives `isMainnet` from the standalone `getNetworkName()`) and pins both states in `tasks/test/runtime-helpers.js`. Since each layer is meant to be reviewable on its own, could you hoist that file and its test into this PR? It has no Hardhat dependency, so nothing else needs to move. Otherwise a line in the PR body saying `pnpm ops` is not mainnet-safe until #3000 would do.

contracts/tasks/lib/command.ts
```diff
@@ -0,0 +1,242 @@
...
+export type CommandContext = {
+  chainId: number;
+  networkName: string;
+  ethers: typeof ethers & {
+    provider: ethers.providers.JsonRpcProvider;
+    getContract: typeof contracts.getContract;
+    getContractAt: typeof contracts.getContractAt;
+    getContractFactory: typeof contracts.getContractFactory;
+  };
+  network: {
```
sparrowDom, line 26: `mockBeaconRoot` cannot run through this context: `utils/hardhat.js#replaceContractAt` (reached via `tasks/beaconTesting.js:112`) calls `hre.ethers.getSigners()` and `hre.network.provider.request({ method: "hardhat_setCode" })`, and neither `ethers.getSigners` nor `network.provider.request` exists here (`provider` is an ethers `JsonRpcProvider`, which has `.send`). It throws `TypeError` at first use, so the catalogue test's "every registry-backed command has a live handler" only holds in the `typeof handler === "function"` sense. Dev-only command, but the stated invariant does not hold on this branch.

Fixed in #3000, which replaces `utils/hardhat.js` with `utils/anvil.js` (`getProvider().send("anvil_setCode", …)`). Same ask as for `isMainnet`: hoisting that file into this PR keeps the layer self-consistent.

contracts/tasks/commands.ts
```diff
@@ -0,0 +1,41 @@
...
+import {
+  legacyHandler,
+  type CommandDefinition,
+  type CommandParam,
+} from "./lib/command";
+
+type CatalogueEntry = Omit<CommandDefinition, "handler">;
+const catalogue = JSON.parse(
+  readFileSync(
+    join(__dirname, "test", "fixtures", "ops-command-catalog.json"),
```
sparrowDom: This makes the JSON under `test/fixtures/` the only source of truth for every command's params: the `makeDefinition` shim in `command.ts` (~156-185) turns `addParam` / `addOptionalParam` / `addFlag` into no-ops, so the `.addParam(...)` chains that still decorate every `task()` in `tasks.js` are dead text. It is correct today because the JSON was generated from those declarations (diffed all 75 against master's `hre.tasks`: zero drift), but the first edit in the obvious place fails silently: an optional param added in `tasks.js` reaches the handler as `undefined` instead of its default (`coerceParams` only iterates the JSON specs), a changed default in `tasks.js` is ignored and the JSON one wins (e.g. `depositToStrategy` `symbol`), a removed param is still accepted. Only a new required param is loud (`Unknown option`). `command-catalog.js` will not catch any of it: it compares names from `tasks.js` against the fixture, then the runtime commands against the fixture, which is tautological since the runtime *is* the fixture. Talos is unaffected (actions use `lib/action.ts`'s `ParamBuilder`), so this is about `pnpm ops`.

Proposal: have the registry record the declarations instead of dropping them, in `task-registry.ts` once #3000 lands (the same shim lives here in `command.ts`). `addParam(name, description, defaultValue, type)`, `addOptionalParam(...)` and `addFlag(name, description)` push `{ name, description, type: type?.name ?? "string", optional, flag, default }` onto the task entry (~15 lines; `types.*` already carry `.name`), `commands.ts` builds the catalogue from `registeredTasks()` plus the inline `accounts`, and the JSON becomes a pinned snapshot that `command-catalog.js` asserts deep-equality against. A param edit in `tasks.js` then either updates the snapshot or fails the test, as it did under Hardhat; `tasks.js` stays the single source of truth and the JSON can move out of `test/fixtures/`.

sparrowDom (commented):
Some comments inline

contracts/tasks/lib/command.ts
```diff
@@ -0,0 +1,242 @@
...
+  for (let index = 0; index < rest.length; index++) {
+    const token = rest[index];
+    if (!token.startsWith("--")) {
+      throw new Error(`Unexpected positional argument '${token}'`);
+    }
+    const key = kebabToCamel(token.slice(2));
+    const next = rest[index + 1];
+    if (next === undefined || next.startsWith("--")) flags[key] = true;
+    else {
+      flags[key] = next;
```
sparrowDom, line 77: A repeated option silently last-wins here: `pnpm ops rebase --symbol OUSD --network mainnet --symbol OETH` ran with OETH. Hardhat rejected that (HH308). Cheap to keep: throw if `key` is already in `flags`.

contracts/tasks/ops.ts
```diff
@@ -0,0 +1,41 @@
...
+    "usage: pnpm ops <command> --network <network> [options]\n\n" +
+    commands
+      .map(({ name, description }) => `  ${name.padEnd(31)} ${description}`)
+      .join("\n")
+  );
+}
+
+export async function main(argv = process.argv.slice(2)): Promise<void> {
+  const { name, network, flags } = parseCli(argv);
+  if (!name || name === "help" || flags.help === true) {
```
sparrowDom: Two small parity gaps with `npx hardhat <task> --help`:

- `--help` after a command name prints the global command list, not that command's params; there is no per-command help at all, which matters more now that the param definitions are only readable in the JSON catalogue (see the `commands.ts` comment). Printing `command.params` (name, type, optional/default, description) when `name` is set and `flags.help` is true would cover it.
- `--network=mainnet` fails with "--network is required" (line 25) because `parseCli` only understands the space-separated form. Hardhat also rejected the `=` form, so this is parity, but the message misleads; either accept `--key=value` in `parseCli` or say "use --network <name>".

contracts/tasks/commands.ts
```diff
@@ -0,0 +1,41 @@
...
+  params: CommandParam[];
+  destination: string;
+}>;
+
+export const commands: CommandDefinition[] = catalogue.map((entry) => ({
+  ...entry,
+  handler:
+    entry.name === "accounts"
+      ? async (_args, context) => {
+          const accounts = await context.ethers.provider.listAccounts();
```
sparrowDom: `provider.listAccounts()` is `eth_accounts`, which a live RPC answers with `[]`, so `pnpm ops accounts --network mainnet` now prints nothing. On master the task printed the configured signer addresses (`DEPLOYER_PK` / `GOVERNOR_PK`, `tasks/account.js:5-19`, together with the keys, which we can happily stop printing). It is only useful against a local node now. Either print the address the standalone signer would use (`getOptionalSigner()` → `getAddress()`), or drop the command; `tasks/account.js` is left without an importer once #3000 deletes `hardhat.config.js`, so it can go together with it. This is also the only description drift in the catalogue: "Prints the list of accounts" → "Prints the list of configured accounts".

sparrowDom (commented):
some more comments inline

contracts/utils/hardhat-task-helpers.js
clement-ux: Hoisted `utils/runtime-helpers.js` and `tasks/test/runtime-helpers.js` here, `utils/hardhat-task-helpers.js` is gone and the two consumers plus the governance test mock point at the new module. Fixed in [6ea230994](https://github.com/OriginProtocol/origin-dollar/commit/6ea230994045d80ceba50cbb4b536cfabe0a4ad7).

contracts/tasks/lib/command.ts
clement-ux: Hoisted the standalone version here as a rewrite of `utils/hardhat.js` (`getProvider().send(...)`), with a test. I kept `hardhat_setCode` rather than `anvil_setCode` because this layer still ships the Hardhat node, which does not know the `anvil_*` names; Anvil honours the `hardhat_*` aliases so #2999 only has to rename. Fixed in [c4b6f4fdb](https://github.com/OriginProtocol/origin-dollar/commit/c4b6f4fdb88b2ee94fd111219f05f4a52cb2dcac).

contracts/tasks/commands.ts
clement-ux: Done as proposed, in `command.ts` since the shim lives here for now: `addParam` / `addOptionalParam` / `addFlag` / the variadic variants record `{ name, description, type, optional, flag, variadic, default? }` with Hardhat's semantics (override inherits description and params, `addParam` with a default is optional, `default` only when declared, duplicate declaration throws), `commands.ts` builds the catalogue from `registeredCommands()`, and the JSON is now a pinned snapshot that `command-catalog.js` deep-equals against (regeneration one-liner in the test header). Regenerating from the declarations gave zero drift against the previous fixture. I left the JSON under `test/fixtures/` since nothing but the test reads it any more. Fixed in [da5031aa7](https://github.com/OriginProtocol/origin-dollar/commit/da5031aa7576e20631c430915a5f361621ae6858).

contracts/tasks/lib/command.ts
clement-ux: Now throws `Option --symbol is repeated`, covered in `tasks/test/command.js`. Fixed in [f70ad2ccc](https://github.com/OriginProtocol/origin-dollar/commit/f70ad2ccc44226a706a13dccfeca95e925dd9fd4).

contracts/tasks/ops.ts
clement-ux: Both: `pnpm ops <command> --help` prints the command's options with type, default and required/optional status, and `parseCli` accepts `--key=value` (first `=` splits, so values containing `=` survive). README mentions the per-command help. Fixed in [f70ad2ccc](https://github.com/OriginProtocol/origin-dollar/commit/f70ad2ccc44226a706a13dccfeca95e925dd9fd4).

contracts/tasks/commands.ts
clement-ux: Dropped the command: with the private keys gone there was nothing left worth printing on a live RPC. `tasks/account.js` goes with `hardhat.config.js` in #3000. The catalogue is 74 entries and the description drift is gone with it. Fixed in [da5031aa7](https://github.com/OriginProtocol/origin-dollar/commit/da5031aa7576e20631c430915a5f361621ae6858).

sparrowDom (approved):
Verified at 31556eb5e: runtime-helpers hoisted (isMainnet correct under pnpm ops), replaceContractAt works through the standalone context, catalogue is now derived from the tasks.js declarations with Hardhat's param semantics and pinned as a snapshot with zero drift against the previous Hardhat catalogue (74 commands), per-command --help / --key=value / repeated-option rejection all behave as requested. test:tasks, test:ops, typecheck, lint and prettier clean; Talos catalogue dump byte-identical to master. Approving for the cascade merge.

sparrowDom (approved):
Re-approving on the auto-rebased head: the tree is byte-identical to the verified head and CI is green on it.

PR #2999, opened 2026-09-01
Migrate local forks to Anvil

clement-ux:
## Plain-English summary

After #2998, operational commands no longer need Hardhat. One active Hardhat job remains: starting a local blockchain or a fork of a live network for development and simulation.

This PR replaces that node with Anvil. The previous Hardhat node is kept temporarily as `hardhat-node.sh` so the two implementations could be compared while migrating; it is deleted in #3000.

## Before and after

```text
Before: scripts/tests → Hardhat node → forked network
After:  scripts/tests → Anvil       → forked network
```

Safe batches produced by both nodes at the same block were compared by hand once, on `chainId`, `to`, `value` and `data` (gas estimates and metadata are node-specific and were ignored). No comparison tooling is shipped: the legacy node has no future, so there is nothing left to compare against once it is gone.

## What changes in this PR

- make Anvil the canonical local/forked node for mainnet, Arbitrum, Base, Sonic, Plume, Hoodi, and HyperEVM;
- connect Brownie to external Anvil networks;
- migrate impersonation, balance, and code-mutation calls to `anvil_*` RPC methods (strategist runlogs are left untouched: they are a record of what ran, and Anvil accepts the `hardhat_*` aliases anyway);
- preserve the previous node wrapper as `hardhat-node.sh` until #3000;
- add Anvil and Brownie bridge tests.

## What does not change yet

- the standalone ops CLI introduced in #2998 is unchanged;
- Hardhat packages and configuration remain only to run the temporary legacy node;
- Hardhat is deleted in #3000 after this replacement layer is reviewable on its own.

## Stack and merge order

Base: #2998 (`chore/introduce-standalone-ops-cli`)

1. #2997 — remove obsolete historical scripts and artifacts
2. #2998 — introduce standalone ops CLI
3. **#2999 — migrate local forks to Anvil (this PR)**
4. #3000 — remove the Hardhat runtime

Merge in cascade: **#3000 → #2999 → #2998 → #2997 → `master`**.

## What to review

Please focus on node and simulation parity:

- does each supported network use the correct chain ID and fork configuration?
- are impersonation, balance changes, and code changes translated to the correct Anvil RPC methods?
- does Brownie connect to the external node correctly?

## Validation

- `pnpm test:tasks` — 44 passing
- `pnpm test:anvil` — 3 passing
- `pnpm test:brownie` — 2 passing
- `pnpm test:scripts` — 66 passing
- `pnpm test:layouts` — 11 passing
- `pnpm typecheck`
- `pnpm lint:js`, `pnpm lint:ts`, `pnpm prettier:check`
- full GitHub CI passes


brownie/scripts/safe_payload_smoke.py
```diff
@@ -0,0 +1,38 @@
+import json
+import os
+from pathlib import Path
+
+from brownie import network, web3
+
+
+def main():
+    if web3.eth.chain_id != 1337:
```
sparrowDom: This guard means the smoke cannot run against either node this PR ships: `anvil.sh` passes the real chain id via `--chain-id` for every network, including `ANVIL_NO_FORK=true`, and `node:legacy` (`./hardhat-node.sh fork`) runs with `FORK=true`, so `getHardhatNetworkChainId()` returns 1 there too. The only 1337 node left is a bare `./hardhat-node.sh` without the `fork` argument, which no script invokes, so the "generate the same Safe transaction with both implementations and compare" flow from the PR body raises this `RuntimeError` as shipped. The payload built below is also a synthetic one from a raw `eth_sendTransaction`; it never goes through `to_gnosis_json` (`world_abstract.py:111`), which is the batch-generation path the comparator exists to validate.

And once #3000 deletes `hardhat-node.sh` and `node:legacy`, this script, `compare_safe_batches.py`, `tests/test_safe_payload_compare.py` and the README paragraph ("compare a batch generated by the legacy node…", line 94) are left without a legacy side to compare against.

Two ways out: (a) make it real here: drop the 1337 guard (record the chain id, or assert against an `EXPECTED_CHAIN_ID` env), have the smoke wrap a real `TemporaryForkForReallocations` run so both engines emit a `to_gnosis_json` batch, and document how `legacy-batch.json` / `anvil-batch.json` are produced (today the batch only goes to stdout); or (b) treat the comparison as a one-off already done, and delete the three files plus the README paragraph in #3000 together with `hardhat-node.sh`.

brownie/brownie-config.yml
```diff
@@ -1 +1 @@
-dotenv: .env
+# Environment variables are inherited from the shell. A local .env is optional.
```
sparrowDom: Unrelated to Anvil: this stops Brownie auto-loading `brownie/.env`. The README (line 139) still tells strategists to keep `CMC_API_KEY` in the environment, so anyone holding it in `brownie/.env` silently loses it. Revert, or say in the PR body why it goes.

brownie/runlogs/2025_02_strategist.py
```diff
@@ -110,7 +110,7 @@ def main():
     requestId = 187
 
     # Hack to make weth.withdraw work
-    brownie.network.web3.provider.make_request('hardhat_setCode', [MULTICHAIN_STRATEGIST, '0x'])
+    brownie.network.web3.provider.make_request('anvil_setCode', [MULTICHAIN_STRATEGIST, '0x'])
```
sparrowDom: These eleven `hardhat_setCode` → `anvil_setCode` edits (here, `2025_05:539`, `2025_06:47,250,411,700`, `2025_07:625,655`, `2026_01:61,148,170`) are the only change to the runlogs. Anvil accepts the `hardhat_*` aliases unchanged (checked against anvil 1.7.1: `hardhat_impersonateAccount` and `hardhat_setCode` both succeed), so the edit is not needed for Anvil, and it makes these runlogs fail if replayed on the legacy node this PR keeps as the A/B oracle. Runlogs are records of what was executed; I would leave them untouched, and the same goes for the three files #3000 edits.

sparrowDom (commented):
Some comments inline

brownie/scripts/safe_payload_smoke.py
clement-ux: Went with (b). The A/B comparison was a one-off done by hand during the migration and the legacy node is deleted in #3000 anyway, so `compare_safe_batches.py`, `safe_payload_smoke.py`, `test_safe_payload_compare.py` and the README paragraph are removed here rather than shipped dead. PR body updated accordingly. Fixed in [42024ec](https://github.com/OriginProtocol/origin-dollar/commit/42024ecf2417c6f554d728c40494063a2ceffd6d).

brownie/brownie-config.yml
clement-ux: Reverted, `dotenv: .env` is back. Fixed in [7f9a1d5](https://github.com/OriginProtocol/origin-dollar/commit/7f9a1d531eb35bf45300a16876067df757777cf5).

brownie/runlogs/2025_02_strategist.py
clement-ux: Reverted all eleven edits, runlogs are back to what was executed (confirmed `hardhat_setCode` / `hardhat_impersonateAccount` work on anvil 1.8.1 too). The three runlogs touched in #3000 will get the same treatment there. Fixed in [1d7fc25](https://github.com/OriginProtocol/origin-dollar/commit/1d7fc25735a00ca2e8bbb54aad3223cec99f5f18).

sparrowDom (approved):
Verified at 372b16fee: Safe-payload harness and README paragraph removed, dotenv restored, runlogs back to what was executed. anvil.sh probed on mainnet/base/hyperevm forks with the real chain ids; test:anvil and test:brownie pass. Approving for the cascade merge.

sparrowDom (dismissed):
Re-approving on the auto-rebased head: the tree is byte-identical to the verified head and CI is green on it.

sparrowDom (approved):
Re-approving on the auto-rebased head after merging #2998: tree byte-identical to the verified head, CI green.

